PR #2, opened 2022-01-14
Init CICD setup

### LINKED ISSUE #1: [FEA] Initiate CICD setup
The CICD should include:

1. signoff check
2. pre-merge check
3. nightly build and deploy
4. periodical submodule sync
5. auto-merge from pre-release branch to dev branch

pxLi:
fix #1 

this is the initial PR for CICD setup, we will refine most of components in the future

**signoff check**
mostly the same as spark-rapids one

**nightly build and deploy**
build and deploy scripts, internal jenkinsfile is ready

**periodical submodule sync**
script. internal jenkinsfile is ready

periodically
1. check if cudf submodule need to be updated (compare to current BASE ref)
2. if yes, try run compile and test internally
3. force push new change to intermediate branch
4. comment test result (passed: True or False) in PR thread
5. If test passed, auto-merge the PR. Otherwise keep the PR open

Why not use pre-merge CI?
security team is going to add DUO push for triggering pre-merge, this is not promising for our automation account.

**auto-merge from pre-release branch to dev branch**
script, and github action

Why force push an intermediate branch to act as the merge HEAD?
since we cannot set merge strategy for specific file in github PR, `thirdparty/cudf` in our case, and the periodical submodule sync could happen frequently (new commits merged to cudf). To avoid frequent merge conflict, we need to make an intermediate branch to check if any change to `thirdparty/cudf`, if yes we will checkout cudf commit from dev branch and then create the PR.

**pre-merge check**
test script. internal blossom pipeline is WIP

this PR also update thirdparty/cudf to https://github.com/rapidsai/cudf/commit/1eceaed26d8242401f2be12b50eb635872fe1bf6 since recent RMM change could break current(https://github.com/rapidsai/cudf/commit/582cc6e466c7d941e1b34893fd56fbd42fe90d68) cudf build 

This PR will also require careful verification after merge. I will following this~


ci/premerge-build.sh
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+nvidia-smi
+
+git submodule update --init --recursive
+bash $WORKSPACE/build/submodule-check
```
jlowe, line 23: Same submodule-check and ninja comments as in the nightly build file.

ci/submodule-sync.sh
```diff
@@ -0,0 +1,82 @@
...
+
+test_pass="False"
+if [[ "${ret}" == "0" ]]; then
+  echo "Test failed, will update the result"
+  test_pass="True"
+else
+  echo "Test passed, will try merge the change"
+fi
+
+# force push the intermediate branch and create PR against REF
```
jlowe, line 70: I'm still a little unclear why we need to force push instead of just push.  Is it because we're sharing the same branch across multiple sync PRs?  Even if we share the branch, I'm not sure why subsequent pushes wouldn't always be essentially a fast-forward of the previous version on that branch.

Couple of questions:
- If there is the potential of sharing, should we ensure each upmerge attempt uses a separate branch to avoid the potential of accidentally colliding?  Seems like we would never need to force-push in that scenario.
- How are these intermediate branches cleaned up?

.github/workflows/action-helper/python/submodule-sync
```diff
@@ -0,0 +1,79 @@
...
+            sha = exist[0].get('head').get('sha')
+        else:
+            params = {
+                'title': f"[submodule-sync] {pr.head} to {pr.base} [skip ci] [bot]",
+                'head': f"{pr.head_owner}:{pr.head}",
+                'base': pr.base,
+                'body': "submodule-sync to create a PR keeping thirdparty/cudf up-to-date.\n"
+                        f"HEAD commit SHA: {args.sha}, "
+                        f"cudf commit SHA: https://github.com/rapidsai/cudf/commit/{args.cudf_sha}\n\n"
+                        "The scheduled sync pipeline gets triggered every 6 hours.\n"
```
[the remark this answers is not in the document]
pxLi: yes, I will remove this one or at lease taking value from ENV

ci/deploy.sh
```diff
@@ -0,0 +1,76 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+###
+# Script to deploy cudf jar files along with other classifiers,
```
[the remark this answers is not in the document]
pxLi: good catch! I will refine the comment part

ci/deploy.sh
```diff
@@ -0,0 +1,76 @@
...
+CLASS_TYPES=${CLASS_TYPES#*,}
+CLASS_FILES=${CLASS_FILES#*,}
+IFS="$ORI_IFS"
+
+###### Copy jar so we strip off classifier  #######
+# Use the first classifier(aka jar file) as the default jar
+FIRST_FILE=${CLASS_FILES%%,*}
+cp -f "$FIRST_FILE" "$FPATH.jar"
+
+###### Deploy cudf jar with all its additions ######
```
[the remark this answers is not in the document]
pxLi: will fix

ci/nightly-build.sh
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+nvidia-smi
+
+git submodule update --init --recursive
+bash $WORKSPACE/build/submodule-check
```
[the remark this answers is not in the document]
pxLi: ah no, this is actually copy-pasted between different scripts. Let me clean a few for mvn build ones

ci/nightly-build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+PARALLEL_LEVEL=${PARALLEL_LEVEL:-4}
+mvn clean package \
```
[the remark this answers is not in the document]
pxLi: sure, created https://github.com/NVIDIA/spark-rapids-jni/issues/3 as a follow-up of this one

ci/premerge-build.sh
pxLi: will cleanup this line

ci/submodule-sync.sh
pxLi: > If there is the potential of sharing, should we ensure each upmerge attempt uses a separate branch to avoid the potential of accidentally colliding? Seems like we would never need to force-push in that scenario.

For current setup, there should be exactly one branch (bot-auto-merge-<BASE>) for BASE branch's submodule sync which also means there should be zero or only one opened PR for BASE branch's submodule sync, all updates will go into the same branch to avoid opening multiple sync PRs. And also the single sync branch will be manually removed after we disable the corresponding sync job
 
The reason I decide to force push here is
1. If someone (not sync job) accidentally modified the sync branch, we may resolve the potential conflict
2. the sync job would need to do an extra upmerge for current sync branch instead of always simply fetching from remote each time, then try test and update submodule commit. And we may need to use the `squash merge` for this PR, so auto-merge conflict fix for this kind of PR may require another steps (conflict fix for submodule sync should not happen frequently, though)
3. The comment history should help leave the test results in thread, we do not have to keep the commit history here

Yes, to use the force push is not a must but for simplifying the workflow. I can make it to just push, and some potential changes could be,
1. every job fetch both remote **sync branch** and BASE branch
2. merge BASE into sync branch to make sure local test is up to date
3. use `squash` merge strategy for the PR to avoid merge independent failed-test cudf commit

@jlowe Do you think this one looks better? Thanks~

jlowe: > If someone (not sync job) accidentally modified the sync branch, we may resolve the potential conflict

This is one of the main reasons I'm _not_ a fan of the force-push approach.  The force-push can silently destroy changes on the branch.  I would prefer if this created a new, temporary branch for the PR, then there's no need to force-push (or if the branch already exists we _want it to fail_ to avoid clobbering any other changes).  When the PR merges, the action merging would then also clean up the temporary branch.  We would need a way to resolve cleaning up the temporary branches when the PR is closed in other ways (e.g.: by resolving the merge conflict manually, etc.).  Maybe always delete the temporary branch via a github trigger when the corresponding auto-update PR is closed (even if not merged)?

> And we may need to use the squash merge for this PR

As for squashing instead of merging the PR, that seems appropriate here since this should be treated just like a regular development PR which we would also squash.

pxLi: > Maybe always delete the temporary branch via a github trigger when the corresponding auto-update PR is closed (even if not merged)?

Yeah, the github could add the strategy to automatically delete the HEAD branch after PR gets merged, but it cannot tell which one is from auto bot. Let me try simply add some code to clean up the temp branch if `auto-merged` succeed.

I will also refactor the process here to not force push

jlowe (approved):

PR #164, opened 2022-04-07
Statically link the CUDA runtime into a single shared library

### LINKED ISSUE #146: [FEA] Statically link the CUDA runtime
This requests something similar to what was implemented in rapidsai/cudf#9873 for cudf.  Ideally spark-rapids-jni should provide a single shared library that statically links the CUDA runtime.

jlowe:
Closes #146.

This is a major update to the way this project builds and is designed to build all of the libcudf, libcudfjni, and libspark_rapids_jni code into a single shared library that is statically linked with the CUDA runtime.  Because we are no longer leveraging the contents of a cudf-like jar, the project has been collapsed into a single top-level Maven module that pulls in the cudf Java sources.

During the build, CMake is invoked to build libcudf and libcudfjni as static libraries.  Those static libraries are then whole-archive linked into a libcudf.so library which is statically linked with the CUDA runtime.  In addition a stub libcudfjni.so library is created as well that depends on libcudf.so.  This somewhat contorted build is done for the following reasons:
- only have one static link of the CUDA runtime to minimize the number of separate CUDA contexts
- preserve backwards compatibility with the cudf Java source that expects to find a libcudf.so and libcudfjni.so
- preserve backwards compatibility with RAPIDS accelerated UDFs which may directly depend on libcudf.so and/or libcudfjni.so

This depends on rapidsai/cudf#10619.


jlowe:
build

jlowe:
build

jlowe:
build

pxLi (approved):
LGTM. I will monitoring other nightly pipeline's status to see if we need to do an update

PR #176, opened 2022-04-11
Add spark-rapids-jni-version-info.properties

jlowe:
This adds a spark-rapids-jni-version-info.properties file to the jar to capture the git information of the repository separate from the cudf version info, which is kept for backwards-compatibility.

In addition the build-info script was updated to handle older git versions that do not support the `-C` flag.


jlowe:
build

pxLi (approved):

PR #182, opened 2022-04-13
Add a build script to build in a Docker container

jlowe:
Adds a convenience script to build within a Docker container as the user and produce artifacts in the same places as if it were built outside of the Docker container.


jlowe:
build

pxLi (approved):

PR #199, opened 2022-04-25
Use C++ to parse and filter parquet footers.

revans2:
We have seen on some parquet files where there are a lot of columns that reading the footer can be the bottleneck. This will help to fix it, but it is just a first step. It does not include predicate push down yet.  Just range filtering for row groups that fall into a given split and column pruning. It still needs a lot of documentation and tests, but I wanted to get the code up sooner than later.


revans2:
build

src/main/cpp/CMakeLists.txt
```diff
@@ -102,9 +102,27 @@ find_library(CUDFJNI_LIB "libcudfjni.a" REQUIRED NO_DEFAULT_PATH
...
+# thrift
+find_library(THRIFT_LIB "libthrift.a" REQUIRED NO_DEFAULT_PATH
+    HINTS "${CUDF_DIR}/cpp/build/_deps/arrow-build/thrift_ep-install/lib/"
+)
+
 set(CUDFJNI_INCLUDE_DIRS
   "${CUDF_DIR}/java/src/main/native/include"
   "${CUDF_DIR}/java/src/main/native/src"
+  "${GENERATED_PARQUET_INCLUDE}"
+  "${CUDF_DIR}/cpp/build/_deps/arrow-build/thrift_ep-install/include/"
```
jlowe, line 125: I wasn't able to find this path after a build?

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+      int64_t mid_point = start_index + total_size / 2;
+      if (mid_point >= part_offset && mid_point < (part_offset + part_length)) {
+        filtered_groups.push_back(row_group);
+      }
+    }
+    return filtered_groups;
+}
+
+void deserialize_parquet_footer(uint8_t* buffer, uint32_t len, parquet::format::FileMetaData * meta) {
```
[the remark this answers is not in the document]
jlowe: Filed #202

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+    }
+
+    /**
+     * Given a schema from a parquet file create a set of pruning maps to prune columns from the rest of the footer
+     */
+    column_pruning_maps filter_schema(std::vector<parquet::format::SchemaElement> & schema, bool ignore_case) {
+      CUDF_FUNC_RANGE();
+      // The maps are sorted so we can compress the tree...
+      // These are the outputs of the computation
+      std::map<int, int, std::less<int>> chunk_map;
```
abellina, line 108: mostly curious, why do you need to specify `std::less<int>` as the comparator? that looks to be the default for `std::map`.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+      std::map<int, int, std::less<int>> num_children_map;
+      // Start off with 0 children in the root, will add more as we go
+      schema_map[0] = 0;
+      num_children_map[0] = 0;
+
+      // num_children_stack and tree_stack hold the current state as we walk though schema
+      std::vector<int> num_children_stack;
+      std::vector<column_pruner*> tree_stack;
+      tree_stack.push_back(this);
+      num_children_stack.push_back(schema[0].num_children);
```
abellina, line 119: are there any odd empty schemas in parquet we should guard against?

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+      uint64_t chunk_index = 0;
+      // We are skipping over the first entry in the schema because it is always the root entry, and
+      //  we already processed it
+      for (uint64_t schema_index = 1; schema_index < schema.size(); schema_index++) {
+        // num_children is optional, but is supposed to be set for non-leaf nodes. That said leaf nodes
+        // will have 0 children so we can just default to that.
+        int num_children = 0;
+        if (schema[schema_index].__isset.num_children) {
```
abellina, line 128: we pick out `schema[schema_index]` several times in this function, perhaps we can take out a reference?

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+          num_children = schema[schema_index].num_children;
+        }
+        std::string name;
+        if (ignore_case) {
+          name = unicode_to_lower(schema[schema_index].name);
+        } else {
+          name = schema[schema_index].name;
+        }
+        column_pruner * found = nullptr;
+        if (tree_stack.back() != nullptr) {
```
abellina, line 138: can we push nulls to `tree_stack`? if not could we replace this with `!tree_stack.empty()`?

abellina: also maybe consider out a reference to `tree_stack.back()`

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+      std::vector<int> final_num_children_map;
+      for (auto it = num_children_map.begin(); it != num_children_map.end(); it++) {
+        final_num_children_map.push_back(it->second);
+      }
+
+      std::vector<int> final_chunk_map;
+      for (auto it = chunk_map.begin(); it != chunk_map.end(); it++) {
+        final_chunk_map.push_back(it->second);
+      }
```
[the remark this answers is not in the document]
abellina: We could also construct the vectors using the iterators:

```
std::vector<int> final_chunk_map { chunk_map.begin(), chunk_map.end() };
```

build/Dockerfile.centos7
```diff
@@ -0,0 +1,36 @@
...
+###
+# Build the image for cudf development environment.
+#
+# Arguments: CUDA_VERSION=11.5.0
+#
+###
+ARG CUDA_VERSION=11.5.0
+# use rapids gpuci/cuda images until nvidia/cuda cuda 11.5+ images are available in docker hub
+# FROM nvidia/cuda:$CUDA_VERSION-devel-centos7
+FROM cudf-build:$CUDA_VERSION-devel-centos7
```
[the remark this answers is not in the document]
jlowe: Agree we can address it in that issuel

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+#include "cudf_jni_apis.hpp"
+#include "jni_utils.hpp"
+
+namespace rapids {
+namespace jni {
+
+/**
+ * Convert a string to lower case. It uses std::tolower per character which has limitations
+ * and may not produce the exact same result as the JVM does. This is probably good enough
+ * for now.
```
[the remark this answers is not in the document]
revans2: I filed https://github.com/rapidsai/cudf/issues/10747 for this.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+    throw std::invalid_argument("invalid character sequence...");
+  }
+  std::vector<wchar_t> wide(wide_size + 1);
+  std::mbstowcs(wide.data(), input.data(), wide_size);
+  for (auto wit = wide.begin(); wit != wide.end(); wit++) {
+    *wit = std::towlower(*wit);
+  }
+  int mb_size = std::wcstombs(nullptr, wide.data(), 0);
+  std::string ret(mb_size, '\0');
+  std::wcstombs(ret.data(), wide.data(), mb_size);
```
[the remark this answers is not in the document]
revans2: What error would it return the second time when we actually do the translation that it wouldn't have returned the first time just to get the length?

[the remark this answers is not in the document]
jlowe: Agree it shouldn't be possible, but if it somehow did that would be particularly bad.

src/main/cpp/src/NativeParquetJni.cpp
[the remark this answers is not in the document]
revans2: No we cannot do it using the iterators directly because the chunk_map is an iterator of std::pair<int, Int> not an iterator of ints.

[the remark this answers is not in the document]
abellina: oh sorry, that makes sense

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,500 @@
...
+                done = true;
+              } else {
+                tree_stack.pop_back();
+                num_children_stack.pop_back();
+              }
+
+              if (tree_stack.size() <= 0) {
+                done = true;
+              }
+          }
```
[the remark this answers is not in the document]
revans2: Yes there is, because we are walking two flattened depth first trees. I really don't have the desire to try and commonize that though.

src/main/cpp/src/NativeParquetJni.cpp
revans2: It is I just wanted to be explicit that this is a sorted map, and that less gave us the desired order.

src/main/cpp/src/NativeParquetJni.cpp
revans2: Yes we can and do push a nullptr to treestack when the item we are looking for is not found. We still need to keep track of the number of children to finish walking the tree so we push in a nullptr.

revans2:
build

revans2:
@pxLi could you take a look at the CI failures. I am in over my head with the errors at this point.

src/main/cpp/src/NativeParquetJni.cpp
abellina: It would be nice just because there doesn't seem to be bounds checking here.

revans2:
I am also running into some issues with the java parquet parser if there are no columns in the footer. I am going to have to make a special case bypass for the empty read schema case, where we just get out the number of rows from the matching row groups. But I still need to figure out how to fit it all together.

pxLi:
build

pxLi:
> @pxLi could you take a look at the CI failures. I am in over my head with the errors at this point.

just realize this change includes some dockerfile change, I will adjust some internal setup to meet the requirements

build/build-in-docker
```diff
@@ -43,9 +44,14 @@ fi
 
 $DOCKER_CMD build -f $REPODIR/thirdparty/cudf/java/ci/Dockerfile.centos7 \
   --build-arg CUDA_VERSION=$CUDA_VERSION \
-  -t $IMAGE_NAME \
+  -t $CUDF_IMAGE_NAME \
   $REPODIR/thirdparty/cudf/java/ci
 
+$DOCKER_CMD build -f $REPODIR/build/Dockerfile.centos7 \
```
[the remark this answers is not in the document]
pxLi: I am not sure who are maintaining some nv official accounts in dockerhub or maybe other public docker registries.

cc @GaryShen2008 @NvTimLiu do you have any idea?

ci/Dockerfile
```diff
@@ -27,7 +27,6 @@ FROM gpuci/cuda:$CUDA_VERSION-devel-centos7
 RUN yum install -y centos-release-scl
 RUN yum install -y devtoolset-9 rh-python38 epel-release
 RUN yum install -y zlib-devel maven tar wget patch ninja-build
-RUN yum -y install https://packages.endpoint.com/rhel/7/os/x86_64/endpoint-repo-1.7-1.x86_64.rpm && yum install -y git
```
pxLi, line 30: the CI also required git to do some work, otherwise it will fail 
```bash
ci/premerge-build.sh: line 22: git: command not found
```

pxLi: seeing the original repo is invalid.
Our CI requires some newer git version instead of default yum pkg. Let me try find available one

pxLi: merged yum repo fix https://github.com/NVIDIA/spark-rapids-jni/pull/208. and also fixed internal pipeline to cover dockerfile change cases in this PR.

Please help upmerge your branch and try re-trigger the build, thanks~

revans2:
Moving back to draft. I have found a number of incorrect assumptions I made about column pruning and the structure of the data. I am going to have to re-write a lot of that code to make it generic enough to match things the way we want them.

revans2:
build

revans2:
After talking this over with others I am going to put this in as is and then we can improve it later on.

revans2:
build

revans2:
build

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,612 @@
...
+          // has an entry that does not match the tree
+          auto found_it = tree_stack.back()->children.find(name);
+          if (found_it != tree_stack.back()->children.end()) {
+            found = &(found_it->second);
+            int parent_mapped_schema_index = tree_stack.back()->s_id;
+            num_children_map[parent_mapped_schema_index]++;
+          }
+        }
+
+        if (schema_item.__isset.type) {
```
abellina, line 229: it seems like code inside of this if statement (other than `chunk_index++`) executes if `found != nullptr`. Could it be moved over to line 222 where we set `found` to be not null?

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,612 @@
...
+      int local_s_id = 0; // There is always a root on the schema
+      int local_c_id = -1; // for columns it is just the leaf nodes
+      auto num = names.size();
+      std::vector<column_pruner*> tree_stack;
+      std::vector<int> num_children_stack;
+      tree_stack.push_back(this);
+      num_children_stack.push_back(parent_num_children);
+      for(uint64_t i = 0; i < num; i++) {
+        auto name = names[i];
+        auto num_c = num_children[i];
```
abellina, line 312: could we expand `num_c` to something like `node_num_children`

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,612 @@
...
+        local_s_id++;
+        int tmp_c_id = -1;
+        if (num_c == 0) {
+          // leaf node...
+          local_c_id++;
+          tmp_c_id = local_c_id;
+        }
+        tree_stack.back()->children.try_emplace(name, local_s_id, tmp_c_id);
```
abellina, line 320: curious, why `try_emplace`, instead of `emplace`?

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,612 @@
...
+      int local_c_id = -1; // for columns it is just the leaf nodes
+      auto num = names.size();
+      std::vector<column_pruner*> tree_stack;
+      std::vector<int> num_children_stack;
+      tree_stack.push_back(this);
+      num_children_stack.push_back(parent_num_children);
+      for(uint64_t i = 0; i < num; i++) {
+        auto name = names[i];
+        auto num_c = num_children[i];
+        local_s_id++;
```
abellina, line 313: same thing with `local_s_id`. I think this is a "schema id" for the field where we are now.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,612 @@
...
+              }
+
+              if (tree_stack.size() <= 0) {
+                done = true;
+              }
+          }
+        }
+      }
+      if (tree_stack.size() != 0 || num_children_stack.size() != 0) {
+        throw std::invalid_argument("DIDN'T CONSUME EVERYTHING...");
```
abellina, line 344: should this message be more specific for parquet?

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,612 @@
...
+  if (start_index < min_start_index) {
+    // a bad offset detected, try first column's offset
+    // can not use minStartIndex in case of padding
+    invalid = true;
+  }
+
+  return invalid;
+}
+
+static int64_t get_offset(parquet::format::ColumnChunk & column_chunk) {
```
abellina, line 376:
```suggestion
static int64_t get_offset(parquet::format::ColumnChunk const& column_chunk) {
```

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,612 @@
...
+  auto md = column_chunk.meta_data;
+  int64_t offset = md.data_page_offset;
+  if (md.__isset.dictionary_page_offset && offset > md.dictionary_page_offset) {
+    offset = md.dictionary_page_offset;
+  }
+  return offset;
+}
+
+static std::vector<parquet::format::RowGroup> filter_groups(parquet::format::FileMetaData & meta,
```
abellina, line 385:
```suggestion
static std::vector<parquet::format::RowGroup> filter_groups(parquet::format::FileMetaData const& meta, 
```

abellina: though I am not 100% sure if that will compile with all the requirements here. It would just be nice.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,612 @@
...
+        final_num_children_map.push_back(it->second);
+      }
+
+      std::vector<int> final_chunk_map;
+      final_chunk_map.reserve(chunk_map.size());
+      for (auto it = chunk_map.begin(); it != chunk_map.end(); it++) {
+        final_chunk_map.push_back(it->second);
+      }
+
+      return column_pruning_maps{final_schema_map, final_num_children_map, final_chunk_map};
```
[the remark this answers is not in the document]
revans2: I thought that the C++ code was smart enough to do the move automatically. I could be wrong but on the CUDF code I keep tell people not to put the std::move on a return, but because we are putting it into a struct I am not 100% sure on it.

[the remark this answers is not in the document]
abellina: I wrote sample code with my own struct with operator= and the copy ctor, then I had a container struct like yours, but with std::vector<my_struct>. Without std::move there was an extra copy when constructing the return value. There wasn't an extra copy outbound of the function (so the return-value-optimization worked, e.g the return by value was optimized out).

src/main/cpp/src/NativeParquetJni.cpp
revans2: `try_emplace` does not replace the value if it is already there. Technically there is nothing stopping someone from having a malformed schema with multiple entries that have the same name. If that happens the duplicate names are combined together instead of last one wins and stops over the top of it. I could instead throw an exception if I see duplicates. Probably should add that in as some of the follow on work.

src/main/cpp/src/NativeParquetJni.cpp
revans2: It should never really happen unless there is a bug in the java code that sends things to this. Because we control both I am fine with a cryptic message.

revans2:
@abellina I think I have addressed most if not all of your review comments. Please take another look.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,625 @@
...
+        }
+        tree_stack.back()->children.try_emplace(name, local_s_id, tmp_c_id);
+        if (num_c > 0) {
+          tree_stack.push_back(&tree_stack.back()->children[name]);
+          num_children_stack.push_back(num_c);
+        } else {
+          // go back up the stack/tree removing children until we hit one with more children
+          bool done = false;
+          while (!done) {
+              int parent_children_left = num_children_stack.back() - 1;
```
abellina, line 341: nit 2 space indentation.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,625 @@
...
+  std::vector<int> chunk_map;
+};
+
+/**
+ * This class will handle processing column pruning for a schema. It is written as a class because
+ * of JNI we are sending the names of the columns as a depth first list, like parquet does internally.
+ */
+class column_pruner {
+public:
+    /**
```
abellina, line 102: nit 2 space indentation.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -0,0 +1,625 @@
...
+  int64_t offset = md.data_page_offset;
+  if (md.__isset.dictionary_page_offset && offset > md.dictionary_page_offset) {
+    offset = md.dictionary_page_offset;
+  }
+  return offset;
+}
+
+static std::vector<parquet::format::RowGroup> filter_groups(parquet::format::FileMetaData const& meta, 
+        int64_t part_offset, int64_t part_length) {
+    CUDF_FUNC_RANGE();
```
abellina, line 400: nit indentation is off in this function

abellina (commented):
All nits at this point that could also be handled later.

abellina (dismissed):

revans2:
@jlowe could you please take another look?

revans2:
build

abellina (approved):

PR #225, opened 2022-05-05
Use ccache in build-in-docker by default

### LINKED ISSUE #222: [FEA] Use ccache in build-in-docker
I wish RAPIDS Accelerator JNI for Apache Spark would be built with `ccache` support introduced in rapidsai/cudf#10790

Provide an option or environment variable to easily turn on/off ccache via `build-in-docker`  

A one-off personal script

**Additional context**
Sometimes one needs to remove the build directory to fix build issues. Always being able to clean build without taking a productivity hit prevents issues from occurring to begin with.

gerashegalov:
This PR enables `ccache` in build-in-docker script by default, closes #222. 

A repeated no-change build with the native build dir cleaned is ~90 seconds
A repeated no-change build without removing the native build dir is ~30 seconds

The PR proposes to change the default to not skipping the clean of libcudf build unless CCACHE_DISABLE=1 is set.

Manual override of `libcudf.clean.skip` is honored.
 
### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,8 +53,8 @@ Due to the lengthy build of libcudf, it is **not cleaned** during a normal Maven
 unless built using `build/build-in-docker`. `build/build-in-docker` uses `ccache` by default
 unless CCACHE_DISABLE=1 is set in the environment.
 
-Boolean property `-Dlibcudf.clean.skip=false` can also be set manually to clean libcudf
-build area in addition to the normal clean of `target/` directories.
+`-Dlibcudf.clean.skip=false` can also be specified on the Maven command-line to force
+libcudf to be cleaned during the Maven clean phase.
 
 Currently libcudf is only configured once and the build relies on cmake to re-configure as needed.
 This is because libcudf currently is rebuilding almost entirely when it is configured with the same
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,9 +49,12 @@ so it should be invoked as one would invoke Maven, e.g.: `build/build-in-docker 
 ### cudf Submodule and Build
 
 [RAPIDS cuDF](https://github.com/rapidsai/cudf) is being used as a submodule in this project.
-Due to the lengthy build of libcudf, it is **not cleaned** during a normal Maven clean phase.
-Use `-Dlibcudf.clean.skip=true` to clean the libcudf build area in addition to the normal clean
-of `target/` directories.
+Due to the lengthy build of libcudf, it is **not cleaned** during a normal Maven clean phase
+unless built using `build/build-in-docker`. `build/build-in-docker` uses `ccache` by default
+unless CCACHE_DISABLE=1 is set in the environment.
+
+Boolean property `-Dlibcudf.clean.skip=false` can also be set manually to clean libcudf
+build area in addition to the normal clean of `target/` directories.
 
 Currently libcudf is only configured once and the build relies on cmake to re-configure as needed.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -49,9 +49,12 @@ so it should be invoked as one would invoke Maven, e.g.: `build/build-in-docker
...
 [RAPIDS cuDF](https://github.com/rapidsai/cudf) is being used as a submodule in this project.
-Due to the lengthy build of libcudf, it is **not cleaned** during a normal Maven clean phase.
-Use `-Dlibcudf.clean.skip=true` to clean the libcudf build area in addition to the normal clean
-of `target/` directories.
+Due to the lengthy build of libcudf, it is **not cleaned** during a normal Maven clean phase
+unless built using `build/build-in-docker`. `build/build-in-docker` uses `ccache` by default
+unless CCACHE_DISABLE=1 is set in the environment.
+
+Boolean property `-Dlibcudf.clean.skip=false` can also be set manually to clean libcudf
+build area in addition to the normal clean of `target/` directories.
```
jlowe, line 57:
```suggestion
`-Dlibcudf.clean.skip=false` can also be specified on the Maven command-line to force
libcudf to be cleaned during the Maven clean phase.
```

build/build-in-docker
```diff
@@ -46,21 +46,44 @@ $DOCKER_CMD build -f $REPODIR/thirdparty/cudf/java/ci/Dockerfile.centos7 \
...
   -v "/etc/sudoers.d:/etc/sudoers.d:ro" \
   -v "$REPODIR:$WORKSPACE_REPODIR:rw" \
   -v "$LOCAL_MAVEN_REPO:$WORKSPACE_MAVEN_REPODIR:rw" \
+  -v "$HOME/.ccache:$HOME/.ccache:rw" \
   --workdir "$WORKSPACE_REPODIR" \
   -e CMAKE_GENERATOR="$CMAKE_GENERATOR" \
   -e CUDA_VISIBLE_DEVICES \
   -e PARALLEL_LEVEL \
   -e VERBOSE \
+  -e CCACHE_DISABLE \
```
jlowe, line 81: Nit: It'd be nice to keep the variable export list sorted lexicographically.

build/build-in-docker
```diff
@@ -46,21 +46,44 @@ $DOCKER_CMD build -f $REPODIR/thirdparty/cudf/java/ci/Dockerfile.centos7 \
...
   $IMAGE_NAME \
   scl enable devtoolset-9 "mvn \
     -Dmaven.repo.local=$WORKSPACE_MAVEN_REPODIR \
     -DPER_THREAD_DEFAULT_STREAM=$PER_THREAD_DEFAULT_STREAM \
     -DUSE_GDS=$USE_GDS \
+    -Dcmake.ccache.opts=\"${CCACHE_OPTS[*]}\" \
+    "$_CUDF_CLEAN_SKIP" \
```
jlowe, line 88: The quotes around this are not escaped, so they're causing the quoting of the scl enable command argument to be terminated early which is probably not what was intended.

Also normally by quoting this variable, it's forcing an argument to appear at this position which we don't want.  It _might_ work in practice since this is essentially triple-evaluated by the docker run cmd and scl enable devtoolset, but since we know there aren't any tricky spaces in the value, I suggest ensuring this doesn't end up as an invalid empty argument on the Maven cmdline.
```suggestion
    $_CUDF_CLEAN_SKIP \
```

jlowe (approved):

jlowe:
build

PR #243, opened 2022-05-13
Add lib to HINTS for libparquet

sperlingxx:
Add hints for 32bit system to find libparquet. 

I tried to build spark-rapids-jni on my local machine. I got the a CMake error: could not find `libparquet.a`. Then, I checked the build directionary `./target/libcudf-install` and found `libparquet.a` is in the folder `./target/libcudf-install/lib` instead of `./target/libcudf-install/lib64` along with other static libs.

Signed-off-by: sperlingxx <lovedreamf@gmail.com>


sperlingxx:
build

jlowe:
@sperlingxx please add a description to PRs.  For example, what bug is this fixing and what environments are affected?  Paste some sample error so others could find this fix if they encounter the same error.  Also, why would we want to search for 32-bit libs?  In what environment is that desired?

sperlingxx:
Hi @jlowe, I added my propose as the description, while I am not sure whether it is necessary to add as a general config.

jlowe (approved):

PR #253, opened 2022-05-17
Exclude CuFileTest unless user specifies -DUSE_GDS=ON

jlowe:
This updates the spark-rapids-jni pom to exclude CuFileTest unless the `USE_GDS` property is `ON`.  The cudf pom has the same logic, and this updates the spark-rapids-jni pom to match that behavior.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,6 +154,27 @@
 
   <profiles>
     <profile>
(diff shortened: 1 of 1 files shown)


jlowe:
build

gerashegalov (approved):

PR #261, opened 2022-05-20
moving rapids cmake include to pull from rapids version

hyperbolic2346:
I noticed today that the cmake file was pulling an out of date RAPIDS.cmake and upon some investigation I found that cudf had moved this functionality into a file in the root directory. Changing this to pull the cmake file from cudf instead of trying to roll the version as cudf changes.

I wanted to use `${CUDF_DIR}`, but I couldn't get it to work as I believe `${PROJECT_BINARY_DIR}` is not initialized this early.

### DIFF
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -14,11 +14,7 @@
 
 cmake_minimum_required(VERSION 3.20.1 FATAL_ERROR)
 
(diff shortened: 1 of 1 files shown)


jbrennan333 (approved):
+1 this looks good to me.  It is done the same way in the cudf jni CMakeLists.txt.

jlowe (approved):

jlowe:
@hyperbolic2346 looks like there needs to be a [signoff](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-22.06/CONTRIBUTING.md#sign-your-work).

jlowe:
build

PR #323, opened 2022-06-08
Merge remote-tracking branch-22.06 to main

NvTimLiu:
Merge remote-tracking branch-22.06 to main

Change version to 22.06.0 

Signed-off-by: Tim Liu <timl@nvidia.com>


NvTimLiu:
build

jlowe:
This is still based on a cudf commit that is 22.06.0-SNAPSHOT

jlowe (approved):
I verified that thirdparty/cudf is on v22.06.00, but unfortunately that version has cudf/java/pom.xml still at 22.06.0-SNAPSHOT.  It's relatively minor, as this version number doesn't make it into the cudf-java-version-info.properties file used by the plugin as a sanity check.  So we should be good for the 22.06.0 release.

It would be nice to have the proper cudf version number in the official cudf version tag, but that's an issue to be solved in the cudf repository.

pxLi:
> but unfortunately that version has cudf/java/pom.xml still at 22.06.0-SNAPSHOT

@jlowe hi, this is the same as we used to release cudfjni. As cudf tagging process, no one helped update jni pom version, we were not allowed to update tagging after they already updated cpp and python versions. So we used to provide the release version directly when we do the cudfjni release candidate build. 

Since this won't affect the pluginjni version check, and in cudfjni release it should have the correct release version in properties file, it might be fine to keep this?

pxLi (approved):
+1, for plugin jni build

GaryShen2008 (approved):

PR #327, opened 2022-06-16
Use same paths in Docker build container as used outside the container

jlowe:
@hyperbolic2346 noted that when building with `build/build-in-docker` the container uses directory paths not used when working outside of the container (i.e.: everything is mounted under `/rapids` in the container which typically does not exist outside of the container).  This can lead to issues when building a native binary in the container and then trying to run it outside of the container, as container-specific paths under `/rapids` could be encoded in shared libraries that will not be found outside of the container.

This updates the `build-in-docker` script to use the same paths in the container for any paths bind-mounted into the container.  If parent directories are missing for the destination mount path, Docker will create them as needed with root:root 755 permissions.  By using the same paths, it makes it easier for binaries built in the container to run outside of the container.

### DIFF
--- a/build/build-in-docker
+++ b/build/build-in-docker
@@ -33,10 +33,6 @@ CUDF_USE_PER_THREAD_DEFAULT_STREAM=${CUDF_USE_PER_THREAD_DEFAULT_STREAM:-ON}
 USE_GDS=${USE_GDS:-ON}
 
 SPARK_IMAGE_NAME="spark-rapids-jni-build:${CUDA_VERSION}-devel-centos7"
-WORKSPACE_DIR=/rapids
-WORKSPACE_REPODIR="$WORKSPACE_DIR/spark-rapids-jni"
-WORKSPACE_CCACHE_REPODIR="$WORKSPACE_DIR/.ccache"
-WORKSPACE_MAVEN_REPODIR="$WORKSPACE_DIR/.m2/repository"
 
 if (( $# == 0 )); then
   echo "Usage: $0 <Maven build arguments>"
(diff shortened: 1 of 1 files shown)


jlowe:
build

gerashegalov:
will this make ccache serve binary objects created under centos7 docker devtoolset to host builds and vice versa?

jlowe:
> will this make ccache serve binary objects created under centos7 docker devtoolset to host builds and vice versa?

I'm not a ccache expert, so I think that depends on whether ccache tracks which underlying compiler is being used as part of its decision to reuse an artifact rather than rebuild.  From [the docs](https://ccache.dev/manual/4.2.html#config_compiler_check) it appears it will detect this and handle it properly:
```
By default, ccache includes the modification time (“mtime”) and size of the compiler in the hash to ensure that results retrieved from the cache are accurate.
```

However I think the concern is orthogonal to this change.  Even before this change it was bind-mounting the ccache directory into the container, just under a different path.  End result is ccache in the container shares a cache that exists outside of the container, and that part has not changed with this patch.  All this patch does in that respect is change the path to the cache, not which cache is used.

hyperbolic2346:
The motivation here was that I was unable to run tests generated by the container because the executables were unable to find the libraries needed to run. This is still the case for things like base cuda libraries, but if the host machine matches the major version of cuda inside the container this should work. As an additional piece of work, I'm going to add scripts to run inside the docker container as well in another PR.

hyperbolic2346:
If it wasn't clear, I have tested this and it worked well for me.

hyperbolic2346 (approved):

build/build-in-docker
```diff
@@ -72,12 +68,12 @@ $DOCKER_CMD run $DOCKER_GPU_OPTS -it -u $(id -u):$(id -g) --rm \
...
   -v "/etc/shadow:/etc/shadow:ro" \
   -v "/etc/sudoers.d:/etc/sudoers.d:ro" \
-  -v "$REPODIR:$WORKSPACE_REPODIR:rw" \
-  -v "$LOCAL_CCACHE_DIR:$WORKSPACE_CCACHE_REPODIR:rw" \
-  -v "$LOCAL_MAVEN_REPO:$WORKSPACE_MAVEN_REPODIR:rw" \
-  --workdir "$WORKSPACE_REPODIR" \
+  -v "$REPODIR:$REPODIR:rw" \
+  -v "$LOCAL_CCACHE_DIR:$LOCAL_CCACHE_DIR:rw" \
+  -v "$LOCAL_MAVEN_REPO:$LOCAL_MAVEN_REPO:rw" \
+  --workdir "$REPODIR" \
```
gerashegalov, line 74: @jlowe The reason I asked about ccache is that now we have the same input paths between host and docker container

gerashegalov:
> The motivation here was that I was unable to run tests generated by the container because the executables were unable to find the libraries needed to run. This is still the case for things like base cuda libraries, but if the host machine matches the major version of cuda inside the container this should work. As an additional piece of work, I'm going to add scripts to run inside the docker container as well in another PR.

it would be great to see a precise description as an issue. I assume you are not talking about cudart that we link statically with CUDA_STATIC_RUNTIME=ON

Besides what I am trying to figure out . Is it always a good idea to have identical paths inside and outside the container or should it be optional

hyperbolic2346:
Opened https://github.com/NVIDIA/spark-rapids-jni/issues/329 to document the issue.

hyperbolic2346:
> I assume you are not talking about cudart that we link statically with CUDA_STATIC_RUNTIME=ON

The host library I was having trouble with later was `libcudart.so.11.0` as I had `libcudart.so.10.0` on my local host. I'm not sure if this answers your question, but this will not 100% solve any issues with libraries, but it is a gives the host a chance.

> Besides what I am trying to figure out . Is it always a good idea to have identical paths inside and outside the container or should it be optional

I think matching the host directory all the time is a good solution. I don't know if ccache will have trouble with this if the cache directory is re-used on the host, but I do know this will solve the issue of library resolution on the host.

gerashegalov (approved):
LGTM

PR #362, opened 2022-07-08
Add new native parquet footer API and deprecate the old one

revans2:
This adds a new API for parsing and filtering a parquet footer using native code. This will allow us to fix the issues associated with legacy parquet files. It does not fix any of those issues yet. This just adds in the API so we can start to use it without breaking anyone.

PR #365, opened 2022-07-11
Add in better native parquet footer implementation and remove the old one

revans2:
This uses the new API to provide a good implementation that fixes the issues older parquet files (legacy ones).

This depends on https://github.com/NVIDIA/spark-rapids-jni/pull/362

I think this will fix https://github.com/NVIDIA/spark-rapids/issues/5493 because it does not reorder the columns like the old implementation did.


revans2:
build

revans2:
build

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -93,6 +93,25 @@ struct column_pruning_maps {
...
+    tag_name = "value";
+  } else if (tag == STRUCT_TAG) {
+    tag_name = "struct";
+  } else if (tag == LIST_TAG) {
+    tag_name = "list";
+  } else if (tag == MAP_TAG) {
+    tag_name = "map";
+  }
+  return tag_name;
+}
```
ttnghia, line 113: Let return `std::string` instead:
```suggestion
std::string get_tag_name(const int tag) {
 switch(tag) {
  case VALUE_TAG: return "value";
  case STRUCT_TAG: return "struct";
  case LIST_TAG: return "list";
  case MAP_TAG: return "map";
  default: return "unknown";
 }
```

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -106,215 +125,272 @@ class column_pruner {
      */
     column_pruner(const std::vector<std::string> & names, 
             const std::vector<int> & num_children, 
-            int parent_num_children): children(), s_id(0), c_id(-1) {
-      add_depth_first(names, num_children, parent_num_children);
+            const std::vector<int> & tags,
```
ttnghia, line 128: I'm not sure about the code format in this repo. If we try to follow rapids, we should use clangformat config from cudf. If we have our own standard then we should also define our own clangformat config.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -106,215 +125,272 @@ class column_pruner {
...
+        return elem.num_children;
+      }
+      return 0;
+    }
+
+    parquet::format::SchemaElement & get_schema_item(std::vector<parquet::format::SchemaElement> & schema, const std::size_t index) {
+      if (index >= schema.size()) {
+        throw std::runtime_error("walked off the end of the schema some how...");
+      }
+      return schema[index];
```
ttnghia, line 187: The entire of this function is not necessary. If you want to access with bound check then just use `schema.get(index)` at the caller. If you want to have some alias function like this then just:
```suggestion
    parquet::format::SchemaElement & get_schema_item(std::vector<parquet::format::SchemaElement> & schema, const std::size_t index) {
      return schema.get(index);
```

hyperbolic2346 (changes requested):
A few comments. I'll look at this more later, but wanted to get these down.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -93,6 +93,25 @@ struct column_pruning_maps {
...
+  if (tag == VALUE_TAG) {
+    tag_name = "value";
+  } else if (tag == STRUCT_TAG) {
+    tag_name = "struct";
+  } else if (tag == LIST_TAG) {
+    tag_name = "list";
+  } else if (tag == MAP_TAG) {
+    tag_name = "map";
+  }
+  return tag_name;
```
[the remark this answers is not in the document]
ttnghia: Humnn, maybe? So `default: throw exception(...);`?

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -106,215 +125,272 @@ class column_pruner {
...
-      std::vector<int> num_children_stack;
-      std::vector<column_pruner*> tree_stack;
-      tree_stack.push_back(this);
-      if (schema.size() == 0) {
-        throw std::invalid_argument("a root schema element must exist");
+    void filter_schema(std::vector<parquet::format::SchemaElement> & schema, const bool ignore_case,
+            std::size_t & current_input_schema_index, std::size_t & next_input_chunk_index,
+            std::vector<int> & chunk_map, std::vector<int> & schema_map, std::vector<int> & schema_num_children) {
+      if (current_input_schema_index >= schema.size()) {
+        throw std::runtime_error("walked off the end of the schema some how...");
```
ttnghia, line 201: I'm not sure if this bound check is still relevant, as it will be checked again later when using that index to access.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -106,215 +125,272 @@ class column_pruner {
...
-
-      // num_children_stack and tree_stack hold the current state as we walk though schema
-      std::vector<int> num_children_stack;
-      std::vector<column_pruner*> tree_stack;
-      tree_stack.push_back(this);
-      if (schema.size() == 0) {
-        throw std::invalid_argument("a root schema element must exist");
+    void filter_schema(std::vector<parquet::format::SchemaElement> & schema, const bool ignore_case,
+            std::size_t & current_input_schema_index, std::size_t & next_input_chunk_index,
+            std::vector<int> & chunk_map, std::vector<int> & schema_map, std::vector<int> & schema_num_children) {
```
ttnghia, line 199: A "good" practice is to split such giant function into multiple smaller functions (like `filter_schema_value`, `filter_schema_struct`, etc.). Not necessarily required here but it's recommended to do so.

ttnghia:
In addition to my left comments, I have no idea if we can have unit tests in this repo?

hyperbolic2346 (commented):
c++ code could in general get a smattering of const. I can add some comments in places if desired. All told, this seems reasonable. I agree with Nghia's comments as well.

As for testing, we have tests for cpp and java. I would assume this would be tested via the java side only and those are setup to run in CI I believe.

src/main/cpp/src/NativeParquetJni.cpp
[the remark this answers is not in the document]
mythrocks: Get 3 C++ programmers in a room, you get 4 opinions. :p
```c++
std::string get_tag_name(const int tag)
{
  static auto const names = std::vector{ "value", "struct", "list", "map" };
  return ( i>=VALUE_TAG && i<=MAP_TAG ) ? names[tag] : "unknown";
}
```
I'm kidding, of course.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -106,215 +125,272 @@ class column_pruner {
...
+      }
     }
 
-    column_pruner(): children(), s_id(0), c_id(-1) {
+    column_pruner * find_child(std::string name) {
+      auto found_it = children.find(name);
+      if (found_it != children.end()) {
+        return &(found_it->second);
+      }
+      return nullptr;
```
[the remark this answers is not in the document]
revans2: Yes, but I am not sure how to do it with the iterators in a clean way. Suggestions are welcome.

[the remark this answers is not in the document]
ttnghia: The solution here would be returning a reference instead of a pointer. But by doing so, this function must throw a not found exception if the element doesn't exist. As such, you can do something like:
```
const column_pruner& find_child(std::string name) {
      auto found_it = children.find(name);
      if (found_it != children.end()) {
        throw exception(...);
      }
      return found_it->second;
}

if(exist_child(name)) {
  auto const& child = find_child(name);
  // process child
}

```

revans2:
The problem with tests is that the only way to test it is with parquet files. We could do some round trip tests and see that it works properly, but I found it was much simpler to test it in the plugin itself. I know there are some big downsides to this and if we really want to be alerted earlier if there is a regression I can add tests here.

src/main/cpp/src/NativeParquetJni.cpp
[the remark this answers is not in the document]
hyperbolic2346: my first thought was to simply return the iterator, but then you have the issue of the caller having to know the .end() of children, which is an implementation detail. I'm leaning towards an optional string return value right now.

hyperbolic2346:
> I know there are some big downsides to this and if we really want to be alerted earlier if there is a regression I can add tests here.

The critical thing for me is that it is tested regularly and hopefully in an automated way. If we don't know about a breakage until the spark level, I'm not that concerned as long as testing is done.

src/main/cpp/src/NativeParquetJni.cpp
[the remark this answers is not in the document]
revans2: std::optional does not work for references. Apparently it is a union and you cannot store a reference inside a union. At least that is the error message I got when I tried to use it.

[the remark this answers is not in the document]
ttnghia: Yes, in order to use a reference in `<optional>` you need to wrap it in `reference_wrapper`. Here is an example: https://godbolt.org/z/o3MhMfczY

revans2:
build

revans2:
@hyperbolic2346 and @ttnghia could you take another look

ttnghia (dismissed):

revans2:
build

revans2:
@hyperbolic2346 can you please take a look at the latest changes?

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -104,217 +116,285 @@ class column_pruner {
...
+        case Tag::MAP:
+          filter_schema_map(schema, ignore_case, current_input_schema_index, next_input_chunk_index, chunk_map, schema_map, schema_num_children);
+          break;
+        default:
+          throw std::runtime_error(std::string("INTERNAL ERROR UNEXPECTED TAG FOUND ") + std::to_string(static_cast<int>(tag)));
+      }
+    }
 
+    /**
+     * Do a depth first traversal to build up this.
```
hyperbolic2346, line 386:
```suggestion
     * Do a depth first traversal to build up schema elements.
```
Is this building schema elements? Can we replace "this" with something more descriptive?

src/main/java/com/nvidia/spark/rapids/jni/ParquetFooter.java
```diff
@@ -202,55 +202,27 @@ public static ParquetFooter readAndFilter(HostMemoryBuffer buffer,
     int parentNumChildren = schema.children.length;
     ArrayList<String> names = new ArrayList<>();
     ArrayList<Integer> numChildren = new ArrayList<>();
-    depthFirstNames(schema, ignoreCase, names, numChildren);
+    ArrayList<Integer> tags = new ArrayList<>();
```
hyperbolic2346, line 205: This smells an awful lot like a struct. I assume we have a benefit to keeping this in discrete arrays, but wanted to make sure.

revans2: Java does not have structs. They have objects. The only way to access members of these objects in C/CPP involves a lot of JNI code. Doing it efficiently is even more complicated where you have to get the method/field ID of what you want to access, and even then it is slow compared to doing it this way.  Oddly this is faster and more efficient then trying to reach back into the JVM to do the processing.

revans2:
@hyperbolic2346 thanks for the new review I think I have addressed all of your comments. Please take another look

revans2:
build

hyperbolic2346 (approved):

PR #381, opened 2022-07-15
Document local cross-repo testing for spark-rapids, spark-rapids-jni, and cudf [skip ci]

gerashegalov:
This PR documents how to test changes in cudf, spark-rapids-jni, and spark-rapids together locally end-to-end.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,9 +83,9 @@ to control aspects of the build:
 
 ### Local testing of cross-repo contributions cudf, spark-rapids-jni, and spark-rapids
 
-When we work on a feauture or a bug fix across repositories, it is beneficial to be able to
+When we work on a feature or a bug fix across repositories, it is beneficial to be able to
 run manual and integration tests end to end on the full stack from Apache Spark
-with spark-rapids Plugin upfront before merging the PRs. So we are dealing with a subset of the following:
+with spark-rapids plugin upfront before merging the PRs. So we are dealing with a subset of the following:
 
 Local PR branches for
 - ~/repos/rapidsai/cuDF, branch pr1
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,10 +114,12 @@ $ git submodule update --init --recursive --remote
 ```
 
 spark-rapids will consume spark-rapids-jni pr2 changes from the local Maven cache after
-we run `mvn install` via `build/build-in-docker` in ~/repos/NVIDIA/spark-rapids-jni skipping the submodule check.
+we run `mvn install` via `build/build-in-docker` in ~/repos/NVIDIA/spark-rapids-jni.
+Make sure to stage thirdparty/cudf with `git add` to satifsfy build's submodule check.
 
 ```bash
-$ ./build/build-in-docker install -Dsubmodule.check.skip=true ...
+$ git add thirdparty/cudf
+$ ./build/build-in-docker install ...
 ```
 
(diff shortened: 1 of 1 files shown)


hyperbolic2346:
I like the documentation update. I'm not sure the best language is to phrase it with we though like "When we build ...". I'm not overly critical of this though and willing to live with it if it is acceptable for others. Thank you for taking the time to do this. I believe this will help new users acclimate to the process.

CONTRIBUTING.md
```diff
@@ -80,6 +80,61 @@ to control aspects of the build:
...
+
+spark-rapids will consume spark-rapids-jni pr2 changes from the local Maven cache after
+we run `mvn install` via `build/build-in-docker` in ~/repos/NVIDIA/spark-rapids-jni skipping the submodule check.
+
+```bash
+$ ./build/build-in-docker install -Dsubmodule.check.skip=true ...
+```
+
+Now cd to ~/repos/NVIDIA/spark-rapids and build per
+[spark-rapids instructions](https://github.com/NVIDIA/spark-rapids/blob/branch-22.08/CONTRIBUTING.md#building-from-source).
```
jlowe, line 124: Would be good to point out here that one should look closely at the portion of the build where the spark-rapids-jni is downloaded from Maven Central and make sure it is _not_ downloaded, otherwise the build is not running with your changes.  Instructions for verifying the build contains the changes would also be helpful here (i.e.: extracting the properties files and verifying the cudf git hash matches expectations).

gerashegalov:
build

CONTRIBUTING.md
```diff
@@ -129,9 +150,43 @@ Now cd to ~/repos/NVIDIA/spark-rapids and build per
 $ ./build/buildall
 ```
 
-Prior to switching to another spark-rapids-jni or submiting a PR to NVIDIA/spark-rapids-jni,
-we should undo the cudf submodule modification.
+Since we rely on local Maven cache we need to pay extra attention to make sure that
+the final rapids-4-spark artifact included locally built dependencies as opposed to
+CI-built snapshot dependencies from the remote Maven repo. This may happen if Maven
+is invoked with `--offline` or `--no-snapshot-updates` option due to IDE-Maven
+interactions in the background. To confirm that the artifact is correct we can either enable INFO logging in Spark or directly inspect the resulting jar for build info:
```
jlowe: Nit: This line isn't wrapped which sticks out (literally) when viewing it outside of the github render (e.g.: from a terminal):
```suggestion
interactions in the background. To confirm that the artifact is correct we can either
enable INFO logging in Spark or directly inspect the resulting jar for build info:
```

CONTRIBUTING.md
```diff
@@ -129,9 +150,43 @@ Now cd to ~/repos/NVIDIA/spark-rapids and build per
 $ ./build/buildall
 ```
 
-Prior to switching to another spark-rapids-jni or submiting a PR to NVIDIA/spark-rapids-jni,
-we should undo the cudf submodule modification.
+Since we rely on local Maven cache we need to pay extra attention to make sure that
+the final rapids-4-spark artifact included locally built dependencies as opposed to
+CI-built snapshot dependencies from the remote Maven repo. This may happen if Maven
+is invoked with `--offline` or `--no-snapshot-updates` option due to IDE-Maven
```
jlowe, line 156: The wording of this implies it is only a possibility if specifying `--offline` or `--no-snapshot-updates`.  Maybe "This may happen even if"?

jlowe (approved):

Issue #386, opened 2022-07-18
[BUG] ant exec errors logged as info in the Maven log, not included in the error 


gerashegalov:
**Describe the bug**
Standard implementation in antrun plugin does not propagate error output to Maven. Therefore error output is logged as info.

The actual ERROR log from Maven only includes the return code 1 without explanation.

**Steps/Code to reproduce bug**

Run build when cudf submodule is stale

```
[INFO] ------------------------------------------------------------------------
[INFO] Building RAPIDS Accelerator JNI for Apache Spark 22.08.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-antrun-plugin:3.0.0:run (submodule check) @ spark-rapids-jni ---
[INFO] Executing tasks
[INFO]      [exec] ERROR: submodules out of date: +dba4eea4a5db9e1b3ceb8ceb8f2762cf86b91170 thirdparty/cudf (v0.12.0-16695-gdba4eea4a5). To fix: git submodule update --init --recursive
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 0.741s
[INFO] Finished at: Mon Jul 18 19:06:16 UTC 2022
[INFO] Final Memory: 15M/477M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.0.0:run (submodule check) on project spark-rapids-jni: An Ant BuildException has occured: exec returned: 1
[ERROR] around Ant part ...<exec failonerror="true" dir="/home/gshegalov/gits/NVIDIA/spark-rapids-jni" executable="/home/gshegalov/gits/NVIDIA/spark-rapids-jni/build/submodule-check"></exec>... @ 4:161 in /home/gshegalov/gits/NVIDIA/spark-rapids-jni/target/antrun/build-main.xml
[ERROR] -> [Help 1]
```
**Expected behavior**

ERROR output should include diagnostics beyond the result code.

This can be achieved by turning off failonerror, capturing result code and error in properties of exec and  having an additional check
```xml
        <fail message="$Exit code: {exitCode}, Error message: ${errorMsg}">
            <condition>
                <not>
                    <equals arg1="${exitCode}" arg2="0"/>
                </not>
            </condition>
        </fail>
```
**Environment details (please complete the following information)**
 - local build

[...]

PR #389, opened 2022-07-19
Enable cpp tests

SrikarVanavasam:
This PR enables cpp tests to run as part of the CI. 

Signed-off-by: Srikar Vanavasam <srikar.vanavasam@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -221,6 +221,45 @@
         </plugins>
       </build>
     </profile>
+      <profile>
+      <id>test-cpp</id>
+      <activation>
+        <property>
+          <name>BUILD_TESTS</name>
+          <value>ON</value>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-antrun-plugin</artifactId>
+            <version>3.0.0</version>
+            <executions>
+              <execution>
+                <id>test-sparkrapidsjni</id>
+                <phase>test</phase>
+                <configuration>
+                  <target>
+                    <exec dir="${project.basedir}" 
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -221,6 +221,45 @@
...
+          <name>BUILD_TESTS</name>
+          <value>ON</value>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-antrun-plugin</artifactId>
+            <version>3.0.0</version>
```
jlowe, line 237: Nit: it would be nice to have a property like `antrun.version` or use dependencyMgmt to help keep the two antrun-plugin setups in sync, in case we need to update the plugin version in the future.

jlowe (commented):
The nightly CI scripts should also be updated to run the tests so they are run regularly rather than only when PRs are built.

jlowe:
build

pom.xml
```diff
@@ -221,14 +222,53 @@
...
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-antrun-plugin</artifactId>
+            <version>${antrun.version}</version>
+            <executions>
+              <execution>
+                <id>test-sparkrapidsjni</id>
+                <phase>test</phase>
+                <configuration>
+                  <target>
+                    <exec dir="${project.basedir}"
```
gerashegalov, line 245: It's optional , just wanted to point out that there is an exec-related issue #386 . If you  happen to create a reusable ant script wrapping exec, that we can call with `<ant >` task , we can reuse it to fix it consistently everywhere in this pom.

gerashegalov (approved):
LGTM

PR #399, opened 2022-07-22
Initial commit for CUDA fault injection tool

gerashegalov:
This PR is the initial version of CUDA fault injection tool to explore and test for correctness of CUDA error handling in fault-tolerant CUDA applications.

The tool is designed with automated testing and interactive testing use cases in mind. The tool is a dynamically linked library `libcufaultinj.so` that is loaded by the CUDA process via CUDA Driver API `cuInit` if it's provided via the `CUDA_INJECTION64_PATH` environment variable.

As an example it can be used to test RAPIDS Accelerator for Apache Spark. 

### Local Mode
```bash
CUDA_INJECTION64_PATH=$PWD/target/cmake-build/faultinj/libcufaultinj.so \ 
FAULT_INJECTOR_CONFIG_PATH=src/test/cpp/faultinj/test_faultinj.json \
$SPARK_HOME/bin/pyspark \ 
  --jars $SPARK_RAPIDS_REPO/dist/target/rapids-4-spark_2.12-22.08.0-SNAPSHOT-cuda11.jar \ 
  --conf spark.plugins=com.nvidia.spark.SQLPlugin
```
### Distributed Mode
```bash
$SPARK_HOME/bin/spark-shell \
  --jars $SPARK_RAPIDS_REPO/dist/target/rapids-4-spark_2.12-22.08.0-SNAPSHOT-cuda11.jar \
  --conf spark.plugins=com.nvidia.spark.SQLPlugin \
  --files ./target/cmake-build/faultinj/libcufaultinj.so,./src/test/cpp/faultinj/test_faultinj.json \
  --conf spark.executorEnv.CUDA_INJECTION64_PATH=./libcufaultinj.so \
  --conf spark.executorEnv.FAULT_INJECTOR_CONFIG_PATH=test_faultinj.json \
  --conf spark.rapids.memory.gpu.minAllocFraction=0 \
  --conf spark.rapids.memory.gpu.allocFraction=0.2 \
  --master spark://hostname:7077 
```
When we configure the executor environment spark.executorEnv.CUDA_INJECTION64_PATH we have to use a path separator in the value ./libcufaultinj.so with the leading dot to make sure that dlopen loads the library file submitted. Otherwise it will assume a locally installed library accessible to the dynamic linker via LD_LIBRARY_PATH and similar mechanisms. See [dlopen man page](https://man7.org/linux/man-pages/man3/dlopen.3.html)

### Fault injection configuration 

Fault injection configuration is provided via the `FAULT_INJECTOR_CONFIG_PATH` environment variable. It's a set of rules to apply fault injection when CUDA Drvier or Runtime is matched by function name or callback id with a given probability.

There are currently three types of fault injection:
- launch a kernel with the PTX `trap` instruction
- launch a kernel with a device assert
- replace the return code for the CUDA Runtime call

Example config:
```json
{
    "logLevel": 1,
    "dynamic": true,
    "cudaRuntimeFaults": {
        "cudaLaunchKernel_ptsz": {
            "percent": 0,
            "injectionType": 0,
            "injectionType_comment": "PTX trap = 0, C assert = 1",
            "interceptionCount": 1
        }
    },
    "cudaDriverFaults": {
        "cuMemFreeAsync_ptsz": {
            "percent": 0,
            "injectionType": 2,
            "injectionType_comment": "substitute return code",
            "substituteReturnCode": 999,
            "interceptionCount": 1
        }
    }
}
```


src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,469 @@
...
+
+static void
+registerAtExitHandler(void) {
+    // Register atExitHandler
+    atexit(&atExitHandler);
+}
+
+
+static void
+atExitHandler(void) {
```
jlowe, line 119: Curious if this is being pedantic, or is there something that is being cleaned up here that normal OS process teardown will not automatically handle?

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,469 @@
...
+    //
+    if (cbInfo->callbackSite != CUPTI_API_EXIT) {
+        return;
+    }
+
+    // Check last error
+    CUPTI_CALL(cuptiGetLastError());
+    boost::optional<const boost::property_tree::ptree&> matchedFaultConfig = boost::none;
+
+    // TODO make a function, switch to read lock after debugging
```
jlowe, line 213: Is this still relevant?

jlowe (commented):
I didn't get into the deep details in the code.  It would be good to have someone with more C++/CUDA experience take a look as well.

I'd like to see the nice writeup in the PR description made into a markdown doc and checked in with this code.  That way there's a user guide sitting next to the code rather than hoping someone does the software archeology to trace the code back to this PR in hopes there's a description of what it does.

Are there plans for packaging this?  It currently builds alongside spark-rapids-jni but doesn't really need it. It's not a tool that requires a Spark/Java application, so wondering if this is the best long-term place for this code.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,453 @@
...
+do {                                                                           \
+    CUptiResult _status = call;                                                \
+    if (_status != CUPTI_SUCCESS) {                                            \
+        const char *errstr;                                                    \
+        cuptiGetResultString(_status, &errstr);                                \
+        spdlog::error("function {} failed with error {}", #call, errstr);      \
+    }                                                                          \
+} while (0)
+
+#define PTHREAD_CALL(call)                                                         \
```
mythrocks, line 52: Not a deal breaker, but I'm curious why we went with a macro, when we had lambdas at our disposal (`CMAKE_CXX_STANDARD 17`).
I'm guessing it's because the call site is simplified if we use a macro?

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,453 @@
...
+
+// Function Declarations
+static void atExitHandler(void);
+
+static void CUPTIAPI faultInjectionCallbackHandler(void *userdata, CUpti_CallbackDomain domain, CUpti_CallbackId cbid, void *cbInfo);
+
+extern int STDCALL InitializeInjection(void);
+
+
+#if defined(__cplusplus)
```
mythrocks, line 89: This is a CUDA file, no? And we have `CMAKE_CXX_STANDARD 17`? Perhaps we could dispense with the `__cplusplus` checks?

mythrocks (changes requested):

src/main/cpp/faultinj/faultinj.cu
mythrocks: Actually, I should read this more closely. I'm not sure if the usage model is:
- tag a separately running Spark RAPIDS process to inject a fault, or,
- inject fault from within the same Spark process.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,453 @@
...
+ */
+int STDCALL
+InitializeInjection(void) {
+    spdlog::info("InitializeInjection");
+    if (globalControl.initialized) {
+        return 1;
+    }
+    // Init globalControl
+    globalControlInit();
+    globalControl.initialized = 1;
```
mythrocks, line 326: Nitpick:

I wonder if we might streamline the initialization/management of `globalControl` state from one place. It looks like `globalControl.initialized` is first set to `0`, then changed to `1` *after* (and *outside*) `globalControlInit()`. 

Would there be value in using a `static` constructor to initialize this singleton? It might obviate the need for `initialized`:

```c++
injGlobalControl& STDCALL
￼globalControl(void) {
￼    spdlog::info("InitializeInjection");
     
     static injGlobalControl instance;
     // Guaranteed to be initialized, thread-safely.

     // Move PTRHEAD_CALL(), `cuptiInitialize()`, etc. into body of constructor.

     return &instance;
}
```

`instance` would be initialized with thread-safety, on first call.

gerashegalov: I think there are many opportunities to make the initial code look more C++ than C. Is it OK if we handle as a separate epic with various suggestions you are making

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,453 @@
...
+        spdlog::error("specify convig via environment {}", FAULT_INJECTOR_CONFIG_PATH);
+        return;
+    }
+    std::ifstream jsonStream(configFilePath);
+    if (!jsonStream.good()){
+        spdlog::error("check file exists {}", configFilePath);
+        return;
+    }
+
+    PTHREAD_CALL(pthread_rwlock_wrlock(&globalControl.configLock));
```
mythrocks, line 407: Thank you for this bit: I had to go look up how we'd do this in C++17: [`std::shared_mutex`](https://en.cppreference.com/w/cpp/thread/shared_mutex), coupled with [`std::unique_lock`](https://en.cppreference.com/w/cpp/thread/unique_lock) for RAII.

Looks like this is doable with the standard library. We can address this later (should we choose to address it at all).

src/main/cpp/faultinj/faultinj.cu
mythrocks: Yes, sir. Of course. None of what I'm pointing out warrants blocking this.

src/main/cpp/faultinj/faultinj.cu
mythrocks: More than anything else, this is a reflection of my prejudice against macros. Your code looks correct, at least to my eye.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,453 @@
...
+    if (inotifyFd < 0) {
+        spdlog::error("inotify_init() failed");
+        return NULL;
+    }
+    spdlog::debug("config watcher thread: inotify_add_watch {}", configFilePath);
+    const int watchFd = inotify_add_watch(inotifyFd, configFilePath, IN_MODIFY);
+    if (watchFd < 0) {
+        spdlog::error("config watcher thread: inotify_add_watch {} failed", configFilePath);
+        return NULL;
+    }
```
mythrocks, line 469: TIL about `inotify`, I'm ashamed to admit.

src/main/cpp/faultinj/faultinj.cu
mythrocks: > dynamically linked to the process under test

Ah, then @jlowe might have a point about CUPTI cleanup likely happening on process exit.
That said, my vote would be to keep the `atExitHandler()` anyway.

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,98 @@
...
+- error handling logic does not cause deadlocks and other sort of unresponsiveness.
+
+This tool allows creating test scenarios and triggering CUDA errors at will
+with some degree of configurability to determine whether an individual CUDA process
+or the higher-level framework such as Apache Spark remain usable when it should
+or gracefully exits with an actionable error message when the errors are irrecoverable.
+
+## Deployment
+
+The tool is designed with automated testing and interactive testing use cases in mind. The tool is a dynamically linked library `libcufaultinj.so` that is loaded by the CUDA process via CUDA Driver API `cuInit` if it's provided via the `CUDA_INJECTION64_PATH` environment variable.
```
jlowe, line 20: This should be wrapped to a reasonable length like the other lines for viewing in a terminal

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+As an example it can be used to test RAPIDS Accelerator for Apache Spark. Consult documnetation to find
```
jlowe, line 22:
```suggestion
As an example it can be used to test RAPIDS Accelerator for Apache Spark. Consult documentation to find
```

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,98 @@
+# Fault Injection Tool for CUDA Runtime and CUDA Driver API
+
+The goal of this tool is to increase testability of the failure handling logic
+in CUDA applications.
+
+It's especially important when CUDA is embedded in a higher-level fault-tolerant framework
+to ensure that CUDA failures are handled correctly in that
+- fatal errors leaving a GPU in unusbale state are detected and such GPUs are prevented from executing retries for
```
mythrocks, line 8:
```suggestion
- fatal errors leaving a GPU in unusable state are detected and such GPUs are prevented from executing retries for
```

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,98 @@
...
+a failed computation
+- non-fatal errors are retried without losing valuable compute resources
+- error handling logic does not cause deadlocks and other sort of unresponsiveness.
+
+This tool allows creating test scenarios and triggering CUDA errors at will
+with some degree of configurability to determine whether an individual CUDA process
+or the higher-level framework such as Apache Spark remain usable when it should
```
mythrocks, line 15:
```suggestion
or the higher-level framework such as Apache Spark remains usable when it should,
```

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,98 @@
...
+  --jars $SPARK_RAPIDS_REPO/dist/target/rapids-4-spark_2.12-22.08.0-SNAPSHOT-cuda11.jar \
+  --conf spark.plugins=com.nvidia.spark.SQLPlugin \
+  --files ./target/cmake-build/faultinj/libcufaultinj.so,./src/test/cpp/faultinj/test_faultinj.json \
+  --conf spark.executorEnv.CUDA_INJECTION64_PATH=./libcufaultinj.so \
+  --conf spark.executorEnv.FAULT_INJECTOR_CONFIG_PATH=test_faultinj.json \
+  --conf spark.rapids.memory.gpu.minAllocFraction=0 \
+  --conf spark.rapids.memory.gpu.allocFraction=0.2 \
+  --master spark://hostname:7077
+```
+When we configure the executor environment spark.executorEnv.CUDA_INJECTION64_PATH
```
mythrocks, line 47: Should the code bits be quoted?
```suggestion
When we configure the executor environment `spark.executorEnv.CUDA_INJECTION64_PATH`
```

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,98 @@
+# Fault Injection Tool for CUDA Runtime and CUDA Driver API
+
+The goal of this tool is to increase testability of the failure handling logic
```
mythrocks, line 3: Brilliant work here. Thank you for documenting this.

mythrocks (dismissed):
I'm largely :+1: on this. There are some minor fixes pending in documentation.

There are parts of this we should probably revisit in the future:
- RAII for (RW)locks, closing files, etc.
- Maybe manage singleton state differently, for `globalConfig`.
- Maybe switch out of using `pthreads` directly, in favour of C++ standard threads.

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,98 @@
...
+in CUDA applications.
+
+It's especially important when CUDA is embedded in a higher-level fault-tolerant framework
```
mythrocks, line 6: Super pedantic nitpick: I would avoid this sort of contraction in documentation, if possible:
```suggestion
It is especially important when CUDA is embedded in a higher-level fault-tolerant framework
```

src/main/cpp/faultinj/README.md
gerashegalov: thank you!

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,506 @@
...
+        PTHREAD_CALL(pthread_rwlock_wrlock(&globalControl.configLock));
+        const int interceptionCount = (*matchedFaultConfig).get<int>("interceptionCount");
+        (*matchedFaultConfig).put("interceptionCount", interceptionCount - 1);
+        PTHREAD_CALL(pthread_rwlock_unlock(&globalControl.configLock));
+    }
+
+    switch (injectionType)
+    {
+    case FI_TRAP:
+        faultInjectorKernelTrap<<<1,1>>>();
```
mythrocks, line 323: Apologies for not catching this sooner: I think we might need a `cudaStreamSynchronize(0)` or something here. It's a tiny kernel.

mythrocks (changes requested):
I think we need a stream sync right after the kernel launches, given how short the kernel are. Otherwise, we run the risk of exiting before the kernel has launched.

src/main/cpp/CMakeLists.txt
```diff
@@ -253,3 +253,5 @@ if(SPARK_RAPIDS_JNI_BUILD_BENCHMARKS)
   rapids_cpm_nvbench()
   add_subdirectory(benchmarks)
 endif()
+
+add_subdirectory(faultinj)
```
ttnghia, line 257: Maybe everything called `fault_injection` is more expressive and better than `faultinj`.

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+The goal of this tool is to increase testability of the failure handling logic
+in CUDA applications.
+
+It is especially important when CUDA is embedded in a higher-level fault-tolerant
+framework to ensure that CUDA failures are handled correctly in that
+- fatal errors leaving a GPU in unusable state are detected and such GPUs are
+prevented from executing retries for a failed computation
+- non-fatal errors are retried without losing valuable compute resources
+- error handling logic does not cause deadlocks and other sort of unresponsiveness.
```
ttnghia, line 11:
```suggestion
framework to ensure that CUDA failures are handled correctly in that:
 * Fatal errors leaving a GPU in unusable state are detected and such GPUs are
prevented from executing retries for a failed computation
 * Non-fatal errors are retried without losing valuable compute resources
 * Error handling logic does not cause deadlocks and other sort of unresponsiveness.
```

src/main/cpp/faultinj/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Fault injection configuration
+
+Fault injection configuration is provided via the `FAULT_INJECTOR_CONFIG_PATH`
+environment variable.
+It is a set of rules to apply fault injection when CUDA Drvier or Runtime is matched by either
+- function name such as [`cudaLaunchKernel_ptsz`](https://docs.nvidia.com/nsight-systems/UserGuide/index.html#cuda-default-cli)
+- or callback id such as [`214`](https://gitlab.com/nvidia/headers/cuda-individual/cupti/-/blob/main/cupti_runtime_cbid.h#L224)
+- or wildcard `*` to match all API function names
+with a given probability.
```
ttnghia, line 69: Similar to above.

src/main/cpp/faultinj/faultinj.cu
ttnghia: If you want to print the crash line using `__LINE__` native macro then a new macro is necessary. Otherwise either macro or function will work. But a function is preferred.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,506 @@
...
+static void atExitHandler(void);
+
+static void CUPTIAPI faultInjectionCallbackHandler(void *userdata, CUpti_CallbackDomain domain, CUpti_CallbackId cbid, void *cbInfo);
+
+extern int STDCALL InitializeInjection(void);
+
+}
+
+const std::string configFilePathEnv = "FAULT_INJECTOR_CONFIG_PATH";
+const char *configFilePath = std::getenv(configFilePathEnv.c_str());
```
ttnghia, line 88: Global variables are discouraged as they may have some potential side effects with their initialization order.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,506 @@
...
+
+#define PTHREAD_CALL(call)                                                         \
+do {                                                                               \
+    int _status = call;                                                            \
+    if (_status != 0) {                                                            \
+        spdlog::error("function {} failed with error code {}", #call, _status);    \
+    }                                                                              \
+} while (0)
+
+typedef enum {
```
ttnghia, line 56: To prevent name clash, all the symbols, except the one that will be called outside of this file, should be wrapped in anonymous namespace:
```
namespace {
...
}
```

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,506 @@
...
+    pthread_rwlock_t configLock = PTHREAD_RWLOCK_INITIALIZER;
+
+    boost::property_tree::ptree configRoot;
+    boost::optional<boost::property_tree::ptree&> driverFaultConfigs = boost::none;
+    boost::optional<boost::property_tree::ptree&> runtimeFaultConfigs = boost::none;
+} injGlobalControl;
+injGlobalControl globalControl;
+
+// Function Declarations
+static void atExitHandler(void);
```
ttnghia, line 79: Functions without attaching to any class can be just functions, don't need to have `static` keyword as they are already static.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,506 @@
...
+    if (watchFd < 0) {
+        spdlog::error("config watcher thread: inotify_add_watch {} failed", configFilePath);
+        return NULL;
+    }
+
+    constexpr auto MAX_EVENTS = 1024;
+    const auto configFilePathStr = std::string(configFilePath);
+    constexpr auto EVENT_SIZE = sizeof(struct inotify_event);
+    const auto BUF_LEN = MAX_EVENTS * (EVENT_SIZE + configFilePathStr.length());
+    char eventBuffer[BUF_LEN];
```
ttnghia, line 475: Please group `constexpr` and other variables in different groups, don't mix them.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,506 @@
...
+    if (watchFd >= 0) {
+        spdlog::debug("config watcher thread: inotify_rm_watch {} {}", inotifyFd, watchFd);
+        inotify_rm_watch(inotifyFd, watchFd);
+    }
+    if (inotifyFd >= 0) {
+        spdlog::debug("config watcher thread: close {}", inotifyFd);
+        close(inotifyFd);
+    }
+    spdlog::info("exiting dynamic reconfig thread: terminateThread={}", globalControl.terminateThread);
+    return NULL;
```
ttnghia, line 505: In C++, `NULL` is discouraged. Use `nullptr` for pointer instead.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,506 @@
...
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include <spdlog/spdlog.h>
+#include <sys/inotify.h>
+#include <sys/time.h>
+
+#define STDCALL
+
+extern "C" {
```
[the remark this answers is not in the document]
gerashegalov: at the very least we need to export InitializeInjection for cuInit. There may be currently also unnecessary symbols in this scope

src/main/cpp/faultinj/faultinj.cu
gerashegalov: I had a bunch of iterations here https://github.com/NVIDIA/spark-rapids-jni/pull/399/commits/36d2678c590d001f614a42d251b26ce159a7dfcf . Let us do cudaStreamSynchronize then :)

src/main/cpp/CMakeLists.txt
gerashegalov: We can reconsider, it feels too long for a directory name at the moment. It contains a README.md with enough background

src/main/cpp/faultinj/faultinj.cu
gerashegalov: after we start using anonymous namespace which I'll try

gerashegalov:
build

mythrocks (dismissed):
I'm :+1:. Thank you for accommodating the changes requested.

There might be future iterations where we might make C++-related changes. E.g. RIAA for mutex locking, `std::mutex` and threads instead of `pthreads` primitives. But we needn't bother with those now.

mythrocks:
> [mythrocks](https://github.com/mythrocks) approved these changes [9 minutes ago](https://github.com/NVIDIA/spark-rapids-jni/pull/399#pullrequestreview-1055874081)
...
> Review required
At least 1 approving review is required by reviewers with write access.

Notice that my :+1: is ignored here. :]

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,515 @@
...
+void CUPTIAPI faultInjectionCallbackHandler(void *userdata, CUpti_CallbackDomain domain, CUpti_CallbackId cbid, void *cbInfo);
+
+const std::string configFilePathEnv = "FAULT_INJECTOR_CONFIG_PATH";
+CUptiResult cuptiInitialize(void);
+void readFaultInjectorConfig(void);
+void traceConfig(boost::property_tree::ptree const& pTree);
+void *dynamicReconfig(void *args);
+
+void
+globalControlInit(void) {
```
ttnghia, line 90: Style needs to be consistent. As a more widely used style: please do not break line between return type and function name.

ttnghia: I wonder if you use any formatter? We should use clang-format.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,515 @@
...
+                spdlog::debug("rejecting fake launch functionName={} symbol={}",
+                    cbInfo->functionName, cbInfo->symbolName);
+                break;
+            }
+            // intentional fallthrough
+        default:
+            matchedFaultConfig = lookupConfig(globalControl.driverFaultConfigs, cbInfo->functionName, cbid);
+        }
+    }
+    if (domain == CUPTI_CB_DOMAIN_RUNTIME_API && globalControl.runtimeFaultConfigs) {
```
ttnghia, line 226: Should be `else if`.

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,515 @@
...
+    case FI_RETURN_VALUE:
+        if (domain == CUPTI_CB_DOMAIN_DRIVER_API) {
+            CUresult *cuResPtr = static_cast<CUresult*>(cbInfo->functionReturnValue);
+            *cuResPtr = static_cast<CUresult>(substituteReturnCode);
+        } else if (domain == CUPTI_CB_DOMAIN_RUNTIME_API) {
+            cudaError_t *cudaErrPtr = static_cast<cudaError_t*>(cbInfo->functionReturnValue);
+            spdlog::error("updating runtime return value DOES NOT WORK, use trap or assert");
+            *cudaErrPtr = static_cast<cudaError_t>(substituteReturnCode);
+            break;
+        }
```
ttnghia, line 341:
```suggestion
    case FI_RETURN_VALUE:
        if (domain == CUPTI_CB_DOMAIN_DRIVER_API) {
            CUresult *cuResPtr = static_cast<CUresult*>(cbInfo->functionReturnValue);
            *cuResPtr = static_cast<CUresult>(substituteReturnCode);
        } else if (domain == CUPTI_CB_DOMAIN_RUNTIME_API) {
            cudaError_t *cudaErrPtr = static_cast<cudaError_t*>(cbInfo->functionReturnValue);
            spdlog::error("updating runtime return value DOES NOT WORK, use trap or assert");
            *cudaErrPtr = static_cast<cudaError_t>(substituteReturnCode);
        }
        break;
```

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,515 @@
...
+
+    default:
+        break;
```
ttnghia, line 344:
```suggestion
    default: ;
```

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,515 @@
...
+    spdlog::debug("readFaultInjectorConfig from {} DONE", globalControl.configFilePath);
+}
+
+void
+traceConfig(boost::property_tree::ptree const& pTree) {
+    boost::property_tree::ptree::const_iterator end = pTree.end();
+    for (boost::property_tree::ptree::const_iterator it = pTree.begin(); it != end; ++it) {
+        spdlog::trace("congig key={} value={}",  it->first, it->second.get_value<std::string>());
+        traceConfig(it->second);
+    }
```
ttnghia, line 416: Don't have to cache `end`.

```
for(auto it = pTree.begin(); it != pTreen.end(); ++it)
```

src/main/cpp/faultinj/faultinj.cu
gerashegalov: We need pre-commit hool like cudf to do that like in cudf.

ttnghia: I would recommend to use the clang-format config from cudf. Our IDE can just load the config and do its job.

gerashegalov: any style works for me, especially if it is just done automatically. Tieing it to the editor is ok but not ideal.

src/main/cpp/faultinj/faultinj.cu
gerashegalov: I find it more readable the way it's symmetrically written, but either way is fine with me in this case.

src/main/cpp/faultinj/faultinj.cu
gerashegalov: any style works for me, especially if it is just done automatically. Tieing it to the editor is ok but not ideal.

gerashegalov:
> There might be future iterations where we might make C++-related changes. E.g. RIAA for mutex locking, `std::mutex` and threads instead of `pthreads` primitives. But we needn't bother with those now.

Thanks for the review @mythrocks 

absolutely, once the PR is merged I'll file an epic for these improvements

gerashegalov:
build

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -0,0 +1,498 @@
...
+    case FI_RETURN_VALUE:
+      if (domain == CUPTI_CB_DOMAIN_DRIVER_API) {
+        CUresult* cuResPtr = static_cast<CUresult*>(cbInfo->functionReturnValue);
+        *cuResPtr          = static_cast<CUresult>(substituteReturnCode);
+      } else if (domain == CUPTI_CB_DOMAIN_RUNTIME_API) {
+        cudaError_t* cudaErrPtr = static_cast<cudaError_t*>(cbInfo->functionReturnValue);
+        spdlog::error("updating runtime return value DOES NOT WORK, use trap or assert");
+        *cudaErrPtr = static_cast<cudaError_t>(substituteReturnCode);
+        break;
+      }
```
ttnghia, line 337:
```suggestion
      }
```

ttnghia (approved):

jlowe (approved):

PR #460, opened 2022-08-05
Allow providing a seed value for std::rand in faultinj config

### LINKED ISSUE #452: [FEA] Allow seed for the random number generator in config
I wish the Fault Injector would allow specifying the seed for the random number generator.
And it should log the default time(0) value. If incorrect failure handling scenario is discovered during an automated run, the developer will be able to reproduce the sequence of interceptions if the CUDA app is otherwise deterministic.
 
top level config "seed"

Hope that the failure is not rare enough as to depend on the exact sequence of faults

**Additional context**
N/A

gerashegalov:
Add ability to provide a seed value in config. Closes #452  
    
### DIFF
--- a/src/main/cpp/faultinj/faultinj.cu
+++ b/src/main/cpp/faultinj/faultinj.cu
@@ -281,7 +281,7 @@ void CUPTIAPI faultInjectionCallbackHandler(void*,
 
   if (injectionProbability < 100) {
     if (injectionProbability <= 0) { return; }
-    const int rand10000     = rand() % 10000;
+    const int rand10000     = std::rand() % 10000;
     const int skipThreshold = injectionProbability * 10000 / 100;
     spdlog::trace("rand1000={} skipThreshold={}", rand10000, skipThreshold);
     if (rand10000 >= skipThreshold) { return; }
@@ -336,7 +336,7 @@ void CUPTIAPI faultInjectionCallbackHandler(void*,
         break;
       }
 
-    default: ;
+    default:;
   }
 }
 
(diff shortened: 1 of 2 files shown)


gerashegalov:
build

gerashegalov:
build

jlowe (approved):

src/main/cpp/faultinj/faultinj.cu
```diff
@@ -366,6 +366,11 @@ void readFaultInjectorConfig(void)
   //
   const std::string dynamicConfigKey = "dynamic";
 
+  // An unsigned int to seed the random number generator to deterministically
+  // recreate a fault sequence
+  //
+  const std::string seedKey = "seed";
```
hyperbolic2346, line 372: seems odd that this isn't an unsigned int.

hyperbolic2346 (commented):
Are we seeding with an unsigned int or a string?

src/main/cpp/faultinj/faultinj.cu
gerashegalov: this is just the name of the key in the config JSON. The value is an unsigned int https://github.com/NVIDIA/spark-rapids-jni/pull/460/files#diff-8718439339d6c151ff2ae2c04e7642b1c81e6a7c3c02397d34bfb525d9cf4323R389-R390

hyperbolic2346: I get it now. I was missing the key part wasn't saying read this string as an unsigned int, but instead read the json and find a key named this string and treat the value as an unsigned int. I see. I thought it was a default value that wouldn't properly evaluate into an unsigned int to force the `value_or` or path.

Thanks for the clarification.

hyperbolic2346 (approved):

src/main/cpp/faultinj/faultinj.cu
gerashegalov: oh I see this can be confusing

PR #463, opened 2022-08-05
Pulling row_conversion.cu over from cudf

hyperbolic2346:
This code should live here in this repository rather than cudf. Since there is a calm in the churn of that file, it seems like a good time to bring it over here. My plan:

1. update row_conversion.cu in jni this with this PR
2. update the plugin to use jni version of row_conversion.cu
3. remove row_conversion.cu in cudf.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/RowConversion.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/RowConversion.java
@@ -25,6 +25,23 @@ public class RowConversion {
   }
 
   /**
+   * For details about how this method functions refer to
+   * {@link #convertToRowsFixedWidthOptimized()}.
+   *
+   * The only thing different between this method and {@link #convertToRowsFixedWidthOptimized()}
+   * is that this can handle rougly 250M columns while {@link #convertToRowsFixedWidthOptimized()}
+   * can only handle columns less than 100
+   */
+  public static ColumnVector[] convertToRows(Table table) {
+    long[] ptrs = convertToRows(table.getNativeView());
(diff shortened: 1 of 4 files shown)


jlowe (commented):
Code that is here looks OK, but cudf's TableTest.fixedWidthRowsRoundTripWide should be ported over to the RowConversionTest in spark-rapids-jni, so when we remove the cudf versions (and the corresponding cudf tests) we don't lose test coverage.  I think there's only one other cudf row conversion test that is already covered in spark-rapids-jni's RowConversionTest.

src/main/java/com/nvidia/spark/rapids/jni/RowConversion.java
```diff
@@ -121,6 +167,8 @@ public static Table convertFromRows(ColumnView vec, DType ... schema) {
   }
 
   private static native long[] convertToRows(long nativeHandle);
+  private static native long[] convertToRowsFixedWidthOptimized(long nativeHandle);
 
   private static native long[] convertFromRows(long nativeColumnView, int[] types, int[] scale);
+  private static native long[] convertFromRowsFixedWidthOptimized(long nativeColumnView, int[] types, int[] scale);
```
jlowe, line 173: I see two new JNI methods being declared here but no corresponding changes to RowConversionJni.cpp.

hyperbolic2346: Good catch. I believe I have this sorted, but I'm still having trouble testing myself.

hyperbolic2346:
build

jlowe:
build

jlowe (approved):

PR #481, opened 2022-08-16
add DOCKER_BUILD_EXTRA_ARGS and DOCKER_RUN_EXTRA_ARGS variable [skip ci]

wbo4958:
`./build/build-in-docker` is not working if vpn is enabled. And it's inconvenient to change script directly, So the PR add a DOCKER_NETWORKING to pass the network parameter, for example

```
DOCKER_NETWORKING="--network host" ./build/build-in-docker ...
```

### DIFF
--- a/build/run-in-docker
+++ b/build/run-in-docker
@@ -26,7 +26,8 @@ REPODIR=$SCRIPTDIR/..
 
 CUDA_VERSION=${CUDA_VERSION:-11.5.0}
 DOCKER_CMD=${DOCKER_CMD:-docker}
-DOCKER_NETWORKING=${DOCKER_NETWORKING:-""}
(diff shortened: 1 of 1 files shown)


jlowe (commented):
Rather than add a very specific variable for a very specific docker argument (and thus likely we would end up adding another variable for another docker option to control later), why not have something more general like `DOCKER_EXTRA_ARGS`?

wbo4958:
> Rather than add a very specific variable for a very specific docker argument (and thus likely we would end up adding another variable for another docker option to control later), why not have something more general like `DOCKER_EXTRA_ARGS`?

Thx @jlowe, done.

jlowe (approved):

jlowe:
build

jlowe:
build

gerashegalov:
should host network be used by default for robustness?

PR #496, opened 2022-08-19
A Spark compatible implementation of decimal128 multiply.

revans2:
This is the native portion of doing an Apache Spark compatible full decimal multiply that requires up to a 256-bit intermediate value.

The code here is highly unoptimized. The divide is horribly slow. There are very basic optimizations we could do to almost any part of the code.  But even without all of those optimizations it is still a lot faster than the CPU, and it even appears to be faster than the existing multiply implementation. I have not done any profiling of the benchmarks to know why, but I suspect that it has something to do with having all of the overflow checks and scaling operations being done in a single kernel, instead of in multiple different kernels.

For the benchmark I ran the following query on a 12 thread/6-core CPU with an a6000 GPU in it. The batch size was set to 1 GiB because the batches get to be kind of big after casing a long to a 128-bit decimal value. I set the GPU parallelism to 4.

```
spark.time(spark.range(10000000000L).selectExpr("CAST(id as DECIMAL($PRECISION,0))").selectExpr("id * id as p").filter("p > 0").count())
```

precision | GPU | CPU | speedup
-- | -- | -- | --
9 | 6,413 | 63,248 | 9.86
10 | 8,062 | 151,103 | 18.74
18 | 8,027 | 175,387 | 21.85
19 | 4,844 | 174,887 | 36.10
32 | 4,853 | 172,636 | 35.57
38 | 4,859 | 173,061 | 35.62

The drop from about 8 seconds to 4.8 seconds at precision 19 appears to be when this new kernel takes over.

I will file follow on issues to actually optimize the code, and some others to try to migrate all decimal multiply operations to this, or a variant that is more specific to the given types involved.

Some of the core code is based off of an initial patch from @jlowe, but it has been highly modified.

### DIFF
--- a/src/main/cpp/src/decimal_utils.cu
+++ b/src/main/cpp/src/decimal_utils.cu
@@ -102,10 +102,6 @@ struct chunked256 {
     }
   }
 
-  inline __device__ bool is_zero() {
-    return chunks[0] == 0 && chunks[1] == 0 && chunks[2] == 0 && chunks[3] == 0;
-  }
-
 private:
   uint64_t chunks[4];
 };
@@ -482,7 +478,6 @@ struct dec128_multiplier : public thrust::unary_function<cudf::size_type, __int1
       overflows[i] = true;
       return;
     }
-    float test_scale_div = std::pow(10, exponent);
     __int128_t scale_divisor = std::pow(10, exponent);
 
     // scale and round to target scale
--- a/src/main/cpp/src/decimal_utils.cu
+++ b/src/main/cpp/src/decimal_utils.cu
@@ -1,4 +1,3 @@
-
 /*
  * Copyright (c) 2022, NVIDIA CORPORATION.
  *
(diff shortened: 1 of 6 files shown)


revans2:
build

revans2:
build

src/main/cpp/src/decimal_utils.cu
```diff
@@ -0,0 +1,542 @@
...
+    int mult_scale = a_scale + b_scale;
+    if (first_div_precision > 0) {
+      __int128_t first_div_scale_divisor = std::pow(10, first_div_precision);
+      divmod256 div_result = divide(product, first_div_scale_divisor);
+      product = div_result.quotient;
+
+      bool const needIncrement = (div_result.remainder < 0 ? -div_result.remainder : div_result.remainder) 
+          >= (first_div_scale_divisor >> 1);
+      int64_t const sign = product.sign();
+      int const roundIncrement = (needIncrement ? sign < 0 ? -1 : 1 : 0);
```
jlowe, line 470: Nit: IMHO some parenthesis around the inner ternary would improve readability

src/main/cpp/src/decimal_utils.cu
```diff
@@ -0,0 +1,542 @@
...
+      // a_scale and b_scale are negative. first_div_precision is not
+      mult_scale = a_scale + b_scale + first_div_precision;
+    }
+
+    int exponent = prod_scale - mult_scale;
+    if (exponent < 0) {
+      overflows[i] = true;
+      return;
+    }
+    __int128_t scale_divisor = std::pow(10, exponent);
```
jlowe, line 481: Seems like we could refactor to leverage some common code with `pow_ten` here, maybe a TODO?

src/main/cpp/src/decimal_utils.cu
```diff
@@ -0,0 +1,542 @@
...
+    // scale and round to target scale
+    if (scale_divisor != 1) {
+      divmod256 div_result = divide(product, scale_divisor);
+      product = div_result.quotient;
+
+      bool const needIncrement = (div_result.remainder < 0 ? -div_result.remainder : div_result.remainder) 
+          >= (scale_divisor >> 1);
+      int64_t const sign = product.sign();
+      int const roundIncrement = (needIncrement ? sign < 0 ? -1 : 1 : 0);
+      product.add(roundIncrement);
```
jlowe, line 492: Seems like we could factor this code out that also appears above

revans2:
Switching to draft because Spark supports negative scalar values (positive ones here), and this conflicts with the source and target scales checks where the source must be larger than the target. Will see if there are things I can do to make this work. Otherwise I might need to make some adjustments. Partly because this goes counter to the games we play with changing a 100 into a 1e2 to make the number of items we can work with in a multiply larger.

revans2:
build

jlowe (approved):

PR #503, opened 2022-08-25
Adding test framework for casting and integer casting kernel and tests

### LINKED ISSUE #444: [FEA] Add a kernel for integer casting
The main issue is https://github.com/NVIDIA/spark-rapids/issues/5639 and it lays out kernels we would like to create to improve performance of casting in spark-rapids. This issue is for the integer kernel.

A kernel should be created to convert strings to integers for reading CSV and JSON.

hyperbolic2346:
This change adds the cpp test framework for string casting kernels and adds the integer casting kernel along with JNI changes to make this work.

This kernel is significantly faster than the CPU, but I believe there is still performance on the table.

```
scala> val df = spark.range(100000000).selectExpr("CAST(id as STRING)")
df: org.apache.spark.sql.DataFrame = [id: string]

scala> spark.time(df.selectExpr("CAST(id as INT)").selectExpr("id as p").filter("p!=0").count())
Time taken: 1681 ms                                                             
res0: Long = 99999999

scala> spark.conf.set("spark.rapids.sql.enabled", "false")

scala> spark.time(df.selectExpr("CAST(id as INT)").selectExpr("id as p").filter("p!=0").count())
Time taken: 9495 ms                                                             
res2: Long = 99999999
```

closes #444


hyperbolic2346:
build

src/main/cpp/src/cast_string.cu
```diff
@@ -0,0 +1,188 @@
+#include "cast_string.hpp"
```
jlowe, line 1: License and copyrights missing

jlowe (changes requested):

jlowe:
If this is not going to include the full plumbing to invoke this kernel from Java then it should not be marked as closing #444.   Personally I'd rather not see a PR put in a kernel that is uncallable from Java.  The JNI code is minimal and should be part of this PR (along with at least a smoke test in Java to verify the JNI plumbing works at least at some level).

hyperbolic2346:
I have no problem with that. I didn't want to step on Dave's toes since he is doing the float kernel and he has jni code for these kernels, but I didn't want to sit on this either. I can add that here.

hyperbolic2346:
Ok, I've updated the initial description with the performance numbers comparing CPU and GPU casts. I have finished the JNI plumbing and added tests for this and verified the tests are passing. I have also added the `spark-rapids` changes I did to test this to this [PR](https://github.com/NVIDIA/spark-rapids/pull/6436). I'm unsure if this is sufficient for the `spark-rapids` side or if more tests/changes are needed.

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -0,0 +1,32 @@
...
+
+import ai.rapids.cudf.*;
+
+/** Utility class for casting between string columns and native type columns */
+public class CastStrings {
+  static {
+    NativeDepsLoader.loadNativeDeps();
+  }
+
+  public static ColumnVector stringToInteger(ColumnView cv, boolean ansiMode, DType type) {
```
jlowe, line 27: Nit: `strings` seems a bit redundant here since it's a static method of the `CastStrings` class, so it's already known  we're casting from a string.
```suggestion
  public static ColumnVector toInteger(ColumnView cv, boolean ansiMode, DType type) {
```

jlowe, line 27: This is a public API and needs Javadoc

src/main/cpp/src/cast_string.cu
```diff
@@ -0,0 +1,204 @@
...
+      if (chars[i + row_start] == '+' || chars[i + row_start] == '-') {
+        if (chars[i + row_start] == '-') { sign = -1; }
+        i++;
+      }
+    }
+
+    bool truncating = false;
+    for (int c = i; c < len; ++c) {
+      auto const chr = chars[c + row_start];
+      if (chr == '.') {
```
jlowe, line 90: It doesn't look like this handles multiple `.` characters appearing in the string properly.  Seems like it needs to be:
```suggestion
      if (!truncating && chr == '.') {
```

src/main/cpp/src/cast_string.cu
```diff
@@ -0,0 +1,204 @@
...
+        // Values are truncated after a decimal point. However, invalid characters AFTER this
+        // decimal point will still invalidate this entry.
+        truncating = true;
+      } else {
+        if (chr > '9' || chr < '0') {
+          // invalid character in string!
+          valid = false;
```
jlowe, line 97: What about trailing whitespace?  That should be ignored.

src/main/cpp/tests/cast_string.cpp
```diff
@@ -0,0 +1,109 @@
...
+TYPED_TEST(StringToIntegerTests, Ansi)
+{
+  auto const strings =
+    test::strings_column_wrapper({"+1",
+                                  "-0",
+                                  "4.2",
+                                  "asdf",
+                                  "98fe",
+                                  "  00012",
+                                  ".--e-37602.n",
```
jlowe, line 56: There should be a test for a case where the only thing wrong is an extra decimal (e.g.: `3..5` or `3.54.0`.  There should also be tests for values with only trailing whitespace and both leading and trailing whitespace.

src/main/cpp/src/cast_string.cu
hyperbolic2346: Thanks, nice catch!

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -0,0 +1,40 @@
...
+import ai.rapids.cudf.*;
+
+/** Utility class for casting between string columns and native type columns */
+public class CastStrings {
+  static {
+    NativeDepsLoader.loadNativeDeps();
+  }
+
+  /**
+   * Convert a string column to an integer column of type type
```
jlowe, line 28: Nit: "type type" looks odd and could be confusing.
```suggestion
   * Convert a string column to an integer column of a specified type.
```

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class CastStringsTest {
+  @Test
+  void castToIntegerTest() {
+    Table.TestBuilder tb = new Table.TestBuilder();
+    tb.column(3l, 9l, 4l, 2l, 20l, null);
+    tb.column(5, 1, 0, 2, 7, null);
+    tb.column(new Byte[]{2, 3, 4, 5, 9, null});
```
jlowe, line 36: Nit: It'd be nice to add an input that expects the cast to generate a null that wasn't in the input to make sure validity of the result is plumbed through properly.

hyperbolic2346: an invalid string, you mean?

jlowe: yes, an invalid string to cast to an integer, should generate null.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -0,0 +1,57 @@
...
+    tb2.column("3", "9", "4", "2", "20", null);
+    tb2.column("5", "1", "0", "2", "7", null);
+    tb2.column("2", "3", "4", "5", "9", null);
+
+    List<ColumnVector> result = new ArrayList<>();
+
+    try (Table origTable = tb2.build()) {
+      for (int i = 0; i < origTable.getNumberOfColumns(); i++) {
+        ColumnVector string_col = origTable.getColumn(i);
+        result.add(CastStrings.toInteger(string_col, false,
```
jlowe, line 49: Should also have a test for a `true` parameter to make sure that's plumbed through to the native code properly.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -0,0 +1,84 @@
...
+    Table.TestBuilder tb = new Table.TestBuilder();
+    tb.column(3l, 9l, 4l, 2l, 20l);
+    tb.column(5, 1, 0, 2, 7);
+    tb.column(new Byte[]{2, 3, 4, 5, 9});
+    Table expected = tb.build();
+
+    Table.TestBuilder tb2 = new Table.TestBuilder();
+    tb2.column("3", "9", "4", "2", "20");
+    tb2.column("5", "1", "0", "2", "7");
+    tb2.column("2", "3", "4", "5", "9");
```
jlowe, line 69: Do these values tell us anything about whether the ANSI flag is being passed properly?  Seems like these values would pass whether or not we properly pass `true` to the native code.

hyperbolic2346: Yeah, I was thinking I should test a failure, but I am not entirely sure how to catch the exception and validate it is the correct exception. I want to add that though, so I will play around with it later tonight or tomorrow.

jlowe: Couple of strategies:

If you just want to check that a specific exception type is thrown, `assertThrows` is appropriate, see an example at https://github.com/rapidsai/cudf/blob/branch-22.10/java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java#L847.

If you want to examine the exception (e.g.: check what's in the message or other extended checking) then this works:
```
try {
   do_the_operation();
   Assertions.fail("Should have thrown!");
} catch (ExpectedException e) {
   Assertions.assertTrue(e.getMessage().contains("some expected message string")
}
```

hyperbolic2346:
@jlowe I feel like there is a better way to do the overflow detection here, but I'm not sure right now. I need to whiteboard it a bit to ensure this is going to work all the time.

hyperbolic2346:
Just realized I need to account for overflow when changing the sign.

hyperbolic2346:
After thinking about this some more, this overflow method will not work at all really. I will need to better this with a different algorithm.

hyperbolic2346:
build

jlowe:
CI failed because cmake is 3.22.3 instead of 3.23.3 [as stated in the premerge Dockerfile](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-22.10/ci/Dockerfile#L38):
```
[2022-09-06T17:59:17.073Z] [INFO] --- maven-antrun-plugin:3.0.0:run (build-libcudf) @ spark-rapids-jni ---
[2022-09-06T17:59:17.073Z] [INFO] Executing tasks
[2022-09-06T17:59:17.073Z] [INFO]     [mkdir] Created dir: /home/jenkins/agent/workspace/jenkins-rapids-jni_premerge-github-98/target/libcudf-install
[2022-09-06T17:59:17.073Z] [INFO]     [mkdir] Created dir: /home/jenkins/agent/workspace/jenkins-rapids-jni_premerge-github-98/thirdparty/cudf/cpp/build
[2022-09-06T17:59:23.638Z] [INFO]      [exec] -- Configuring incomplete, errors occurred!
[2022-09-06T17:59:23.638Z] [INFO]      [exec] CMake Error at build/_deps/rapids-cmake-src/CMakeLists.txt:28 (cmake_minimum_required):
[2022-09-06T17:59:23.638Z] [INFO]      [exec]   CMake 3.23.1 or higher is required.  You are running version 3.22.3
[2022-09-06T17:59:23.638Z] [INFO]      [exec] 
[2022-09-06T17:59:23.638Z] [INFO]      [exec] 
[2022-09-06T17:59:23.638Z] [INFO] ------------------------------------------------------------------------
[2022-09-06T17:59:23.638Z] [INFO] BUILD FAILURE
[2022-09-06T17:59:23.638Z] [INFO] ------------------------------------------------------------------------
```

@pxLi is the premerge Dockerfile somehow out of date?

src/main/cpp/src/cast_string.cu
```diff
@@ -140,58 +133,24 @@ __global__ void string_to_integer_kernel(T* out,
...
-          if ((T)thread_val > (T)original_val) {
-            if (debug)
-              printf("%d %d - overflow because %d is > %d\n",
-                     threadIdx.x,
-                     blockIdx.x,
-                     thread_val,
-                     original_val);
+          auto const c       = chr - '0';
+          auto const min_val = std::numeric_limits<T>::min() + c;
+          if (min_val > thread_val) {
```
jlowe, line 139: Nit: The similar code above uses `thread_val < miin_val` and t would be nice to keep it consistent

src/main/cpp/src/cast_string.cu
```diff
@@ -140,58 +133,24 @@ __global__ void string_to_integer_kernel(T* out,
...
-          if (thread_val < original_val) {
-            if (debug)
-              printf("%d %d - overflow because %d is < %d\n",
-                     threadIdx.x,
-                     blockIdx.x,
-                     thread_val,
-                     original_val);
+          auto const c       = chr - '0';
+          auto const max_val = std::numeric_limits<T>::max() - c;
+          if (max_val < thread_val) {
```
jlowe, line 148: Nit: Same consistency issue with similar overflow check above.

hyperbolic2346:
build

jlowe (approved):

jlowe:
build

pxLi:
build

hyperbolic2346:
build

jlowe (approved):

PR #509, opened 2022-08-26
Fix handling of older encoding of Parquet arrays

jlowe:
Relates to NVIDIA/spark-rapids#6410.  An older encoding of array data in Parquet lacked any surrounding grouping type, but the native Parquet parser was throwing an error in that case.  This updates the native Parquet footer parsing code to allow list data without a surrounding group type as long as it's a repeating type.


jlowe:
build

revans2:
build

revans2 (approved):

PR #512, opened 2022-08-29
changing cmake to 3.23.3

hyperbolic2346:
Ran into a problem with cmake versions when building:
```
[INFO]      [exec] CMake Error at /home/knobby/code/spark-rapids-jni/target/libcudfjni/_deps/rapids-cmake-src/CMakeLists.txt:28 (cmake_minimum_required):
[INFO]      [exec]   CMake 3.23.1 or higher is required.  You are running version 3.22.3
```

Bumped to 3.24.2, which is latest and ran into a regression finding `cupti_static`. This was fixed in [this PR,](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7608) but it isn't released yet, so the suggestion was to use 3.23.3.

Signed-off-by: Mike Wilson <knobby@burntsheep.com>

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -35,7 +35,7 @@ RUN scl enable rh-python38 "pip install requests"
 RUN mkdir /usr/local/rapids && mkdir /rapids && chmod 777 /usr/local/rapids && chmod 777 /rapids
 
 # 3.22.3: CUDA architecture 'native' support + flexible CMAKE_<LANG>_*_LAUNCHER for ccache
-ARG CMAKE_VERSION=3.22.3
+ARG CMAKE_VERSION=3.23.3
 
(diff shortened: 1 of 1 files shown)


jlowe (approved):
Note that separately we're going to need the same change at https://github.com/rapidsai/cudf/blob/branch-22.10/java/ci/Dockerfile.centos7#L37

jlowe:
build

hyperbolic2346:
> Note that separately we're going to need the same change at https://github.com/rapidsai/cudf/blob/branch-22.10/java/ci/Dockerfile.centos7#L37

Added a PR for that as well [here](https://github.com/rapidsai/cudf/pull/11619).

jbrennan333 (approved):

PR #533, opened 2022-09-08
Fixing ansi casting to integer with decimal string values

hyperbolic2346:
During testing with the plugin, I found that I was incorrect with handling of decimal strings like "1.2" in ansi mode. These values are converted to 1 in non-ansi mode, but are invalid in ansi. This change fixes that.

Signed-off-by: Mike Wilson <knobby@burntsheep.com>


hyperbolic2346:
build

PR #562, opened 2022-09-26
Adding single kernel for string to decimal casting

hyperbolic2346:
This adds a kernel to support casting a string to a decimal.

This is currently still WIP as there are some known edge cases from the plugin tests. I wanted to get this up to get input and decide about 22.10 vs 22.12 release.

### DIFF
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -357,9 +357,7 @@ __device__ thrust::optional<thrust::tuple<bool, int, int>> validate_and_exponent
     auto const last_state = state;
     state                 = validate_char(state, chr, char_num);
 
-    if (state == ST_INVALID) {
-      return thrust::nullopt;
-    }
+    if (state == ST_INVALID) { return thrust::nullopt; }
 
     if (last_state == ST_DIGITS && state != ST_DIGITS && state != ST_DECIMAL_POINT) {
       // past digits, save location
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -389,8 +389,7 @@ __global__ void string_to_decimal_kernel(T* out,
                                          bitmask_type const* incoming_null_mask,
                                          size_type num_rows,
                                          int32_t scale,
-                                         int32_t precision,
-                                         bool ansi_mode)
+                                         int32_t precision)
 {
   auto const group = cooperative_groups::this_thread_block();
   auto const warp  = cooperative_groups::tiled_partition<cudf::detail::warp_size>(group);
@@ -718,8 +717,7 @@ struct string_to_decimal_impl {
       string_col.null_mask(),
       string_col.size(),
       dtype.scale(),
-      precision,
-      ansi_mode);
+      precision);
 
     auto col =
       std::make_unique<column>(dtype, string_col.size(), data.release(), null_mask.release());
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -102,22 +102,13 @@ bool __device__ will_overflow(T const lhs, T const rhs, bool adding)
 {
   if constexpr (std::is_signed_v<T>) {
     if (!adding) {
-      auto const min_val = cuda::std::numeric_limits<T>::min() + rhs;
-      if (lhs < min_val) {
-        // overflow
-        return true;
-      }
-      return false;
+      auto const minval = cuda::std::numeric_limits<T>::min() + rhs;
+      return lhs < minval;
     }
   }
 
-  auto const max_val = cuda::std::numeric_limits<T>::max() - rhs;
-  if (lhs > max_val) {
-    // overflow
-    return true;
-  }
-
-  return false;
+  auto const maxval = cuda::std::numeric_limits<T>::max() - rhs;
+  return lhs > maxval;
 }
 
 /**
@@ -131,15 +122,16 @@ bool __device__ will_overflow(T const lhs, T const rhs, bool adding)
  * @return true if success, false if overflow
  */
 template <typename T>
-bool __device__ process_value(bool first_value, T& current_val, T const new_digit, bool adding)
+thrust::pair<bool, T> __device__
+process_value(bool first_value, T current_val, T const new_digit, bool adding)
 {
   if (!first_value) {
-    if (will_overflow(current_val, adding)) { return false; }
+    if (will_overflow(current_val, adding)) { return {false, current_val}; }
 
     current_val *= 10;
   }
 
-  if (will_overflow(current_val, new_digit, adding)) { return false; }
+  if (will_overflow(current_val, new_digit, adding)) { return {false, current_val}; }
 
   if (adding) {
     current_val += new_digit;
@@ -147,7 +139,7 @@ bool __device__ process_value(bool first_value, T& current_val, T const new_digi
     current_val -= new_digit;
   }
 
-  return true;
+  return {true, current_val};
 }
 
 /**
@@ -229,11 +221,13 @@ void __global__ string_to_integer_kernel(T* out,
       }
 
       if (!truncating && !trailing_whitespace) {
-        T const new_digit = chr - '0';
-        if (!process_value(c == i, thread_val, new_digit, sign > 0)) {
+        T const new_digit             = chr - '0';
+        auto const [success, new_val] = process_value(c == i, thread_val, new_digit, sign > 0);
+        if (!success) {
           valid = false;
           break;
         }
+        thread_val = new_val;
       }
     }
 
@@ -356,9 +350,10 @@ __device__ thrust::optional<thrust::tuple<bool, int, int>> validate_and_exponent
 
     if (state == ST_EXPONENT) {
       T const new_digit = chr - '0';
-      if (!process_value(exponent_val == 0, exponent_val, new_digit, exponent_positive)) {
-        return thrust::nullopt;
-      }
+      auto const [success, new_val] =
+        process_value(exponent_val == 0, exponent_val, new_digit, exponent_positive);
+      if (!success) { return thrust::nullopt; }
+      exponent_val = new_val;
     }
   }
 
@@ -516,10 +511,13 @@ __global__ void string_to_decimal_kernel(T* out,
           num_precise_digits++;
         }
 
-        if (!process_value(i == first_digit, thread_val, new_digit, positive)) {
+        auto const [success, new_val] =
+          process_value(i == first_digit, thread_val, new_digit, positive);
+        if (!success) {
           valid = false;
           break;
         }
+        thread_val = new_val;
       }
     }
 
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -334,18 +334,16 @@ __device__ thrust::optional<thrust::tuple<bool, int, int>> validate_and_exponent
   processing_state state = ST_DIGITS;
 
   // skip leading whitespace
-  while (i < len && is_whitespace(chars[i + row_start])) {
+  while (i < len && is_whitespace(chars[i])) {
     i++;
   }
 
-  if (i < len) {
-    // check for leading +-
-    if (chars[i] == '-') {
-      positive = false;
-      i++;
-    } else if (chars[i] == '+') {
-      i++;
-    }
+  // check for leading +-
+  if (chars[i] == '-') {
+    positive = false;
+    i++;
+  } else if (chars[i] == '+') {
+    i++;
   }
 
   // if there is no data left, this is invalid
@@ -462,11 +460,10 @@ __global__ void string_to_decimal_kernel(T* out,
     // number of precise digits we have encountered
     int num_precise_digits = 0;
     // number of digits we have encountered, even leading 0's
-    int total_digits = 0;
-
-    T thread_val = 0;
-
+    int total_digits     = 0;
+    T thread_val         = 0;
     bool found_sig_digit = false;
+    int rounding_digits  = 0;
 
     // march string starting at first_digit and build value
     for (int i = first_digit; i < len && valid; ++i) {
@@ -482,6 +479,7 @@ __global__ void string_to_decimal_kernel(T* out,
       if (num_precise_digits + 1 > precision || total_digits + 1 > last_digit) {
         // more digits than required, but we need to round
         if (new_digit >= 5) {
+          auto const orig_val = thread_val;
           if (will_overflow(thread_val, static_cast<T>(1), positive)) {
             valid = false;
             break;
@@ -489,6 +487,29 @@ __global__ void string_to_decimal_kernel(T* out,
             thread_val++;
           } else {
             thread_val--;
+          }
+          // we need to know if the first digit overflowed and added a new digit
+          // this can only happen if the first digit is lower now than before
+          // rounding added a digit. There may be a faster route, but it has to work
+          // with __int128_t as well.
+          auto count_digits = [](T val) {
+            int count = 0;
+            while (val != 0) {
+              count++;
+              val /= 10;
+            }
+            return count;
+          };
+
+          auto before_digits = count_digits(orig_val);
+          auto after_digits  = count_digits(thread_val);
+
+          if (count_digits(thread_val) > count_digits(orig_val)) {
+            // more digits now than before rounding
+            total_digits++;
+            num_precise_digits++;
+            decimal_location++;
+            rounding_digits++;
           }
         }
         break;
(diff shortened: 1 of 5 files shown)


revans2:
@hyperbolic2346 you need to sign off on your changes in one of the commits.

src/main/cpp/tests/cast_string.cpp
```diff
@@ -229,4 +231,212 @@ TYPED_TEST(StringToIntegerTests, Overflow)
...
+TEST_F(StringToDecimalTests, OverPrecise)
+{
+  auto const strings = test::strings_column_wrapper({"123456", "999999", "-123456", "-999999"});
+  strings_column_view scv{strings};
+
+  auto const result =
+    spark_rapids_jni::string_to_decimal(5, 0, scv, false, rmm::cuda_stream_default);
+
+  test::fixed_point_column_wrapper<int32_t> expected(
+    {12346, 100000, -12346, -100000}, {0, 0, 0, 0}, numeric::scale_type{0});
```
revans2, line 260: nit: This test is a little confusing. With a precision of 5 and a scale of 0 all of the numbers overflow and you get null back.  All of that is great. The confusing part is that you don't have an XXX like in other code that indicates it is a NULL placeholder so I don't spend a few mins trying to understand why these values look like they are rounded almost correctly.

src/main/cpp/tests/cast_string.cpp
```diff
@@ -229,4 +231,212 @@ TYPED_TEST(StringToIntegerTests, Overflow)
...
+
+  test::print(result->view());
```
revans2, line 262: nit: do we still need/want this print?

src/main/cpp/src/cast_string.cu
```diff
@@ -271,4 +756,37 @@ std::unique_ptr<column> string_to_integer(data_type dtype,
...
+std::unique_ptr<column> string_to_decimal(int32_t precision,
+                                          int32_t scale,
+                                          strings_column_view const& string_col,
+                                          bool ansi_mode,
+                                          rmm::cuda_stream_view stream,
+                                          rmm::mr::device_memory_resource* mr)
+{
+  data_type dtype = [precision, scale]() {
+    if (precision <= cuda::std::numeric_limits<int32_t>::digits10)
+      return data_type(type_id::DECIMAL32, -scale);
```
revans2, line 780: nit: can we have the scale passed in be the CUDF scale and not the java/scala/spark scale?  This is for consistency. All of the other CUDF java APIs take the CUDF scale.

src/main/cpp/tests/cast_string.cpp
```diff
@@ -229,4 +231,212 @@ TYPED_TEST(StringToIntegerTests, Overflow)
...
+TEST_F(StringToDecimalTests, Rounding)
+{
+  auto const strings = test::strings_column_wrapper({"1.23456", "9.99999", "-1.23456", "-9.99999"});
+  strings_column_view scv{strings};
+
+  auto const result =
+    spark_rapids_jni::string_to_decimal(5, 4, scv, false, rmm::cuda_stream_default);
+
+  test::fixed_point_column_wrapper<int32_t> expected(
+    {12346, 100000, -12346, -100000}, {1, 1, 1, 1}, numeric::scale_type{-4});
```
revans2, line 276: So it is expected that the java code still needs to do an overflow check on the output of this? The second and fourth numbers technically do not fit in a precision of 5. The precision is 6 after rounding. If this is expected that is fine, we just need to make sure that we have documented it. Just because it is not what I would have expected.

src/main/cpp/tests/cast_string.cpp
```diff
@@ -229,4 +231,212 @@ TYPED_TEST(StringToIntegerTests, Overflow)
...
+    strings_column_view scv{str};
+
+    auto const result =
+      spark_rapids_jni::string_to_decimal(6, 6, scv, false, rmm::cuda_stream_default);
+    test::fixed_point_column_wrapper<int32_t> expected({0}, {0}, numeric::scale_type{-6});
+
+    test::print(result->view());
+
+    CUDF_TEST_EXPECT_COLUMNS_EQUAL(result->view(), expected);
+  }
```
revans2, line 441: Does this support values with a negative java/scala/spark scale?  If so we are missing tests for this. If not then we should document it.  Support for it is turned off by default in Spark, so I don't think it is a big deal for us to not support it right now.

src/main/cpp/tests/cast_string.cpp
hyperbolic2346: I agree and this is a result of copy/paste. I have changed this to 0's, but would you prefer a null variable name instead?

src/main/cpp/src/cast_string.cu
hyperbolic2346: This is actually helpful for understanding the code as well. I'll make this change.

src/main/cpp/tests/cast_string.cpp
revans2: That is good enough for me.

src/main/cpp/tests/cast_string.cpp
hyperbolic2346: Not at all. My goal is to match `spark-shell` in the handling. I thought I was doing that, but I see that spark will return null for these tests. I will update the code to ensure it matches the output of spark.

src/main/cpp/tests/cast_string.cpp
hyperbolic2346: I believe it will, but I haven't tested it. I need to do that and then decide about fixing or adding comments about it. If this doesn't support negative scale, I would need to adjust the plugin to still be able to call the original code in those cases, right?

revans2: Yes

src/main/cpp/tests/cast_string.cpp
hyperbolic2346: Updated these to match spark-shell and updated the code to produce valid results.

src/main/cpp/tests/cast_string.cpp
hyperbolic2346: I added a test for negative scales, but I'm not 100% these match spark. I tried to enable negative scales, but it still complained. Need to figure this out so I can verify negative scales are working as intended.
```
scala> spark.conf.set("spark.sql.legacy.allowNegativeScaleOfDecimal", true)

scala> spark.sql("""select cast("1.234e-1" as decimal(6,-2))""").show
org.apache.spark.sql.catalyst.parser.ParseException:
extraneous input '-' expecting INTEGER_VALUE(line 1, pos 36)

== SQL ==
select cast("1.234e-1" as decimal(6,-2))
------------------------------------^^^

  at org.apache.spark.sql.catalyst.parser.ParseException.withCommand(ParseDriver.scala:266)
  at org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parse(ParseDriver.scala:127)
  at org.apache.spark.sql.execution.SparkSqlParser.parse(SparkSqlParser.scala:51)
  at org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parsePlan(ParseDriver.scala:77)
  at org.apache.spark.sql.SparkSession.$anonfun$sql$2(SparkSession.scala:616)
  at org.apache.spark.sql.catalyst.QueryPlanningTracker.measurePhase(QueryPlanningTracker.scala:111)
  at org.apache.spark.sql.SparkSession.$anonfun$sql$1(SparkSession.scala:616)
  at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:775)
  at org.apache.spark.sql.SparkSession.sql(SparkSession.scala:613)
  ... 47 elided
```

hyperbolic2346: Finished testing positive scale values thanks to Bobby's help:
```
spark.conf.set("spark.sql.legacy.allowNegativeScaleOfDecimal", "true")
Seq("-7.691546687437984E16").toDF().withColumn("dec", col("value").cast(DecimalType(10, -1))).show()
```

hyperbolic2346:
build

hyperbolic2346:
build

hyperbolic2346:
build

src/main/cpp/src/cast_string.cu
```diff
@@ -37,6 +37,12 @@ namespace detail {
 
 constexpr auto NUM_THREADS{256};
 
+/**
+ * @brief identifies if a character is whitespace
```
ttnghia, line 41:
```suggestion
 * @brief Identify if a character is whitespace.
```

src/main/cpp/src/cast_string.cu
```diff
@@ -48,8 +54,126 @@ constexpr bool is_whitespace(char const chr)
...
+
+template <typename T, std::enable_if_t<not cuda::std::is_signed_v<T>>* = nullptr>
+constexpr T __device__ generic_abs(T value)
+{
+  return value;
+}
+
+/**
+ * @brief determines if overflow will occur when multiplying
+ * a value by 10
```
ttnghia, line 71:
```suggestion
 * @brief Determine if overflow will occur when multiplying a value by 10.
```

src/main/cpp/src/cast_string.cu
```diff
@@ -48,8 +54,126 @@ constexpr bool is_whitespace(char const chr)
...
+bool __device__ will_overflow(T const val, bool adding)
+{
+  if constexpr (std::is_signed_v<T>) {
+    if (!adding) {
+      auto constexpr minval = cuda::std::numeric_limits<T>::min() / 10;
+      if (val < minval) {
+        // overflow
+        return true;
+      }
+      return false;
```
ttnghia, line 88:
```suggestion
      return val < minval;
```

src/main/cpp/src/cast_string.cu
ttnghia: This also needs to clarify why do we need the `adding` parameter.

src/main/cpp/src/cast_string.cu
```diff
@@ -48,8 +54,126 @@ constexpr bool is_whitespace(char const chr)
...
+    }
+  }
+
+  auto constexpr maxval = cuda::std::numeric_limits<T>::max() / 10;
+  if (val > maxval) {
+    // overflow
+    return true;
+  }
+
+  return false;
```
ttnghia, line 98:
```suggestion
  return val > maxval;
```

src/main/cpp/src/cast_string.cu
```diff
@@ -48,8 +54,126 @@ constexpr bool is_whitespace(char const chr)
...
+}
+
+/**
+ * @brief determines if overflow will occur when adding or subtracting values
```
ttnghia, line 102:
```suggestion
 * @brief Determine if overflow will occur when adding or subtracting values.
```

src/main/cpp/src/cast_string.cu
```diff
@@ -48,8 +54,126 @@ constexpr bool is_whitespace(char const chr)
...
+bool __device__ will_overflow(T const lhs, T const rhs, bool adding)
+{
+  if constexpr (std::is_signed_v<T>) {
+    if (!adding) {
+      auto const min_val = cuda::std::numeric_limits<T>::min() + rhs;
+      if (lhs < min_val) {
+        // overflow
+        return true;
+      }
+      return false;
```
ttnghia, line 120: Simplify code similar to above.

src/main/cpp/src/cast_string.cu
```diff
@@ -48,8 +54,126 @@ constexpr bool is_whitespace(char const chr)
...
+ *
+ * @tparam T type of the input data
+ * @param first_value true if this is the first value added
+ * @param current_val current computed value
+ * @param new_digit digit to append to the computed value
+ * @param adding true if adding, false if subtracting
+ * @return true if success, false if overflow
+ */
+template <typename T>
+bool __device__ process_value(bool first_value, T& current_val, T const new_digit, bool adding)
```
ttnghia, line 144: We should not use ref in input parameter. Instead, return a pair of `<ret, bool>`.

hyperbolic2346: Updated to fix this, but I went with `<bool, ret>` for no particular reason.

hyperbolic2346:
build

src/main/cpp/src/cast_string.cu
```diff
@@ -115,48 +221,350 @@ __global__ void string_to_integer_kernel(T* out,
...
+template <typename T>
+__global__ void string_to_decimal_kernel(T* out,
+                                         bitmask_type* validity,
+                                         const char* const chars,
+                                         offset_type const* offsets,
+                                         bitmask_type const* incoming_null_mask,
+                                         size_type num_rows,
+                                         int32_t scale,
+                                         int32_t precision,
+                                         bool ansi_mode)
```
ttnghia, line 393: I don't see `ansi_mode` is used anywhere in this function?

src/main/cpp/src/cast_string.cu
```diff
@@ -176,7 +584,59 @@ struct row_valid_fn {
...
+                                             thrust::make_counting_iterator(col.size()),
+                                             row_valid_fn{col.null_mask(), source_col.null_mask()});
+
+    offset_type string_bounds[2];
+    cudaMemcpyAsync(&string_bounds,
+                    &source_col.offsets().data<offset_type>()[*first_error],
+                    sizeof(offset_type) * 2,
+                    cudaMemcpyDeviceToHost,
+                    stream.value());
+    stream.synchronize();
```
ttnghia, line 613: If we need copy then sync stream then why not just use `cudaMemcpy` (sync copy)?

src/main/cpp/src/cast_string.cu
```diff
@@ -176,7 +584,59 @@ struct row_valid_fn {
...
+ */
+void validate_ansi_column(column_view const& col,
+                          strings_column_view const& source_col,
+                          rmm::cuda_stream_view stream)
+{
+  auto const num_nulls      = col.null_count();
+  auto const incoming_nulls = source_col.null_count();
+  auto const num_errors     = num_nulls - incoming_nulls;
+  if (num_errors > 0) {
+    auto const first_error = thrust::find_if(rmm::exec_policy(stream),
```
ttnghia, line 602: `thrust::find_if` is known to have performance issue. Use `count_if(...) > 0`.

ttnghia, line 602: What if `first_error` iter is out of bound?

src/main/cpp/src/cast_string.cu
```diff
@@ -202,42 +662,14 @@ struct string_to_integer_impl {
...
-                        &string_col.chars().data<char const>()[string_bounds[0]],
-                        string_bounds[1] - string_bounds[0],
-                        cudaMemcpyDeviceToHost,
-                        stream.value());
-        stream.synchronize();
-
-        throw cast_error(*first_error, dest);
-      }
-    }
+    if (ansi_mode) { validate_ansi_column(col->view(), string_col, stream); }
```
ttnghia, line 665: Can we validate while casting, instead of calling to a separate kernel? We can make `ansi_mode` a template parameter to reduce performance impact.

src/main/cpp/src/cast_string.cu
```diff
@@ -248,6 +680,70 @@ struct string_to_integer_impl {
...
+    using Type = device_storage_type_t<T>;
+
+    rmm::device_uvector<Type> data(string_col.size(), stream, mr);
+    auto const num_words = bitmask_allocation_size_bytes(string_col.size()) / sizeof(bitmask_type);
+    rmm::device_uvector<bitmask_type> null_mask(num_words, stream, mr);
+
+    dim3 const blocks(util::div_rounding_up_unsafe(string_col.size(), detail::NUM_THREADS));
+    dim3 const threads{detail::NUM_THREADS};
+
+    detail::string_to_decimal_kernel<<<blocks, threads, 0, stream.value()>>>(
```
ttnghia, line 713: Can we maximize code share between `string_to_integer_kernel` and `string_to_decimal_kernel`? Like extracting some helper functions that can be called from both places.

src/main/cpp/src/cast_string.cu
hyperbolic2346: if `num_errors` > 0 then `first_error` will be valid. I can check this anyway, but it would be a sanity check.

src/main/cpp/src/cast_string.cu
hyperbolic2346: I was torn on this myself, but I assume the majority of cases will be non-ansi and as a result, I didn't want to impact performance of the common path. If we have an ansi mode request and there are errors, I'm not concerned about the kernel launch for it. It does introduce possibly 2 kernel launches to determine if we have errors though when we count the nulls, but I chalked that up as the cost of admission.

If we pushed this content into the kernel itself, we would lose some of the commonality for it. Each thread would need to either atomic set a bit somewhere or atomic increment or we each row would have a mask similar to the null mask. The problem there is the extra memory required over the common case. This was the least-intrusive way to address this I could find.

src/main/cpp/src/cast_string.cu
hyperbolic2346: That is 100% the goal here and I did pull out some common functions like `will_overflow`, `is_whitespace`, and `process_value`. I believe there is more sharing going on here and we can pull more out. My plan was to continue to factor out common code as more casts are added. There is a list we are going down for these.

ttnghia:
LGTM now (may have more nitpick later) but I would like to stress test this with more integration tests (i.e., various random input) in https://github.com/NVIDIA/spark-rapids/pull/6616. If we can run a lot of tests before merging this then we will not have to go back and fix bugs showing up later.

hyperbolic2346:
This has been tested in `spark-rapids` and passes all tests now. The tests in this project are issues that were found in testing and known edges. The real extensive testing is done in the plugin. Are you asking for more tests to be added there?

revans2:
build

PR #607, opened 2022-10-04
Adding workflow automation for spark-rapids-jni [skip ci]

mattahrens:
Signed-off-by: Matt Ahrens <matthewahrens@gmail.com>

Automatically adds new issues into global spark-rapids Github Project

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -1,0 +1,31 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/add-to-project.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  add-to-project:
+    if: github.repository == 'NVIDIA/spark-rapids-jni'
+    name: Add new issues to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v0.3.0
+        with:
+          project-url: https://github.com/orgs/NVIDIA/projects/4
```
pxLi, line 30: Q:  so we could add issues from different repos to the same github project, right?

pxLi (approved):
LGTM, just one question

pxLi:
build

.github/workflows/add-to-project.yml
mattahrens: Yes.  That's the benefit of the new GitHub Projects capability, in that it allows us to bring together issues across repos for a single view.

PR #631, opened 2022-10-12
Adding cast string to float kernel

### LINKED ISSUE #632: [FEA] Support cast string to float
Is your feature request related to a problem? Please describe.
The main issue is https://github.com/NVIDIA/spark-rapids/issues/5639 and it lays out kernels we would like to create to improve performance of casting in spark-rapids. This issue is for the float/double kernel.

Describe the solution you'd like
A kernel should be created to convert strings to floats

hyperbolic2346:
This adds a string to float kernel to improve performance of this conversion.

This is one of the casting kernels for https://github.com/NVIDIA/spark-rapids/issues/5639.

Closes #632

### DIFF
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -144,6 +144,7 @@ add_library(
   src/CastStringJni.cpp
   src/NativeParquetJni.cpp
   src/cast_string.cu
+  src/cast_string_to_float.cu
   src/row_conversion.cu
   src/DecimalUtilsJni.cpp
   src/decimal_utils.cu
--- a/src/main/cpp/src/cast_string.hpp
+++ b/src/main/cpp/src/cast_string.hpp
@@ -92,4 +92,23 @@ std::unique_ptr<cudf::column> string_to_decimal(
   bool ansi_mode,
   rmm::cuda_stream_view stream,
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Convert a string column into an float column.
+ *
+ * @param dtype Type of column to return.
+ * @param string_col Incoming string column to convert to integers.
+ * @param ansi_mode If true, strict conversion and throws on erorr.
+ *                  If false, null invalid entries.
+ * @param stream Stream on which to operate.
+ * @param mr Memory resource for returned column
+ * @return std::unique_ptr<column> Integer column that was created from string_col.
+ */
+std::unique_ptr<cudf::column> string_to_float(
+  cudf::data_type dtype,
+  cudf::strings_column_view const& string_col,
+  bool ansi_mode,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
 }  // namespace spark_rapids_jni
--- a/src/main/cpp/src/cast_string_to_float.cu
+++ b/src/main/cpp/src/cast_string_to_float.cu
@@ -618,9 +618,8 @@ __global__ void string_to_float_kernel(T* out,
   __syncthreads();
 
   // convert
-  string_to_float<T, block_size> convert(
-    out, validity, ansi_except, valid_count, chars, offsets, ipow, incoming_null_mask, num_rows);
-  convert();
+  string_to_float<T, block_size>{
+    out, validity, ansi_except, valid_count, chars, offsets, ipow, incoming_null_mask, num_rows}();
 }
 
 }  // namespace detail
@@ -650,8 +649,8 @@ std::unique_ptr<column> string_to_float(data_type dtype,
   auto out = cudf::make_numeric_column(dtype, string_col.size(), mask_state::ALL_NULL, stream, mr);
 
   using ScalarType = cudf::scalar_type_t<size_type>;
-  auto valid_count = cudf::make_numeric_scalar(cudf::data_type(cudf::type_id::INT32));
-  auto ansi_count  = cudf::make_numeric_scalar(cudf::data_type(cudf::type_id::INT32));
+  auto valid_count = cudf::make_numeric_scalar(cudf::data_type(cudf::type_to_id<size_type>()));
+  auto ansi_count  = cudf::make_numeric_scalar(cudf::data_type(cudf::type_to_id<size_type>()));
   static_cast<ScalarType*>(valid_count.get())->set_value(0, stream);
   if (ansi_mode) { static_cast<ScalarType*>(ansi_count.get())->set_value(-1, stream); }
 
--- a/src/main/cpp/src/cast_string_to_float.cu
+++ b/src/main/cpp/src/cast_string_to_float.cu
@@ -1,0 +1,655 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#include <cub/warp/warp_reduce.cuh>
+#include <cudf/column/column.hpp>
+#include <cudf/column/column_factories.hpp>
+#include <cudf/detail/utilities/cuda.cuh>
+#include <cudf/detail/utilities/integer_utils.hpp>
+#include <cudf/utilities/bit.hpp>
+#include <cudf/scalar/scalar_factories.hpp>
+#include <cudf/strings/convert/convert_floats.hpp>
+
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/column_utilities.hpp>
+#include <cudf_test/column_wrapper.hpp>
+
+using namespace cudf;
+
+__device__ __inline__ bool is_digit(char c)
+{
+  return c >= '0' && c <= '9';
+}
+
+template<typename T, size_type block_size>
+class string_to_float {
+public:
+  __device__ 
+  string_to_float(T* _out, 
+                  bitmask_type* _validity, 
+                  int32_t *_ansi_except, 
+                  size_type* _valid_count, 
+                  const char* const _chars, 
+                  offset_type const* _offsets, 
+                  int _warp_id,
+                  uint64_t const*const _ipow) 
+    : out(_out),
+      validity(_validity),
+      ansi_except(_ansi_except),
+      valid_count(_valid_count),
+      chars(_chars),
+      offsets(_offsets),
+      tid(threadIdx.x + (blockDim.x * blockIdx.x)),
+      warp_id(tid / 32),
+      row(warp_id),
+      warp_lane(tid % 32),
+      row_start(offsets[row]),
+      len(offsets[row+1] - row_start),
+      ipow(_ipow)
+  {    
+  }
+
+  __device__ void operator()()
+  {
+    bstart = 0;                             // start position of the current batch
+    blen = min(32, len);                    // length of the batch
+    bpos = 0;                               // current position within the current batch of chars for the warp  
+    c = warp_lane < blen ? chars[row_start + warp_lane] : 0;
+
+    // printf("(%d): bstart(%d), blen(%d), bpos(%d), tpos(%d), c(%c)\n", tid, bstart, blen, bpos, tpos, c);
+
+    // check for leading nan
+    if(check_for_nan()){
+      compute_validity(valid, except);
+      return;
+    }
+
+    // check for + or -
+    int sign = check_for_sign();
+
+    // check for inf / infinity
+    if(check_for_inf()){
+      if(warp_lane == 0){
+        out[row] = sign > 0? std::numeric_limits<T>::infinity() : -std::numeric_limits<T>::infinity();
+      }
+      compute_validity(valid, except);
+      return;
+    }
+
+    // parse the remainder as floating point. 
+    auto const [digits, exp_base] = parse_digits();
+    if(!valid){
+      compute_validity(valid, except);
+      return;
+    }
+
+    // 0 / -0.
+    if(digits == 0){      
+      if(warp_lane == 0){
+        out[row] = sign * static_cast<double>(0);
+      }      
+      compute_validity(valid, except);
+      return;
+    }
+
+    // parse any manual exponent
+    auto const manual_exp = parse_manual_exp();
+    if(!valid){
+      compute_validity(valid, except);
+      return;
+    }
+        
+    // construct the final float value
+    if(warp_lane == 0){
+      // base value
+      double digitsf = sign * static_cast<double>(digits);
+
+      // exponent
+      int exp_ten = exp_base + manual_exp;
+
+      // final value
+      if (exp_ten > std::numeric_limits<double>::max_exponent10){
+        out[row] = sign > 0 ? std::numeric_limits<double>::infinity()
+                            : -std::numeric_limits<double>::infinity();      
+      } else {
+        // make sure we don't produce a subnormal number. 
+        // - a normal number is one where the leading digit of the floating point rep is not zero. 
+        //      eg:   0.0123  represented as  1.23e-2
+        //
+        // - a denormalized number is one where the leading digit of the floating point rep is zero.
+        //      eg:   0.0123 represented as   0.123e-1
+        //
+        // - a subnormal number is a denormalized number where if you tried to normalize it, the exponent
+        //   required would be smaller then the smallest representable exponent. 
+        // 
+        // https://en.wikipedia.org/wiki/Denormal_number
+        //
+
+        double const exponent = exp10(static_cast<double>(std::abs(exp_ten)));
+        double const result = exp_ten < 0 ? digitsf / exponent : digitsf * exponent;    
+
+        /*
+        if(warp_lane == 0){
+          printf("row(%d), %lf, %d, %lf\n", row, digitsf, exp_ten, exponent);
+        }
+        */
+        
+        out[row] = result;
+      }
+    } 
+    compute_validity(valid, except);
+  }
+
+private:
+  // returns true if we encountered 'nan'
+  // potentially changes:  valid/except
+  __device__ bool check_for_nan()
+  {
+    auto const nan_mask = __ballot_sync(0xffffffff, (warp_lane == 0 && (c == 'N' || c == 'n')) ||
+                                                    (warp_lane == 1 && (c == 'A' || c == 'a')) ||
+                                                    (warp_lane == 2 && (c == 'N' || c == 'n')));
+    if(nan_mask == 0x7){    
+      // if we start with 'nan', then even if we have other garbage character, this is a null row.
+      //
+      // if we're in ansi mode and this is not -precisely- nan, report that so that we can throw
+      // an exception later.
+      valid = false;
+      except = len != 3;
+      return true;
+    }
+    return false;
+  }
+
+  // returns 1 or -1 to indicate sign
+  __device__ int check_for_sign()
+  {
+    auto const sign_mask = __ballot_sync(0xffffffff, warp_lane == 0 && (c == '+' || c == '-'));
+    int sign = 1;
+    if(sign_mask){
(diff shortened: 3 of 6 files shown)


hyperbolic2346:
build

src/main/cpp/src/cast_string.hpp
```diff
@@ -92,4 +92,23 @@ std::unique_ptr<cudf::column> string_to_decimal(
   bool ansi_mode,
   rmm::cuda_stream_view stream,
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Convert a string column into an float column.
+ *
+ * @param dtype Type of column to return.
+ * @param string_col Incoming string column to convert to integers.
```
jlowe, line 100:
```suggestion
 * @param string_col Incoming string column to convert to floating point.
```

src/main/cpp/src/cast_string.hpp
```diff
@@ -92,4 +92,23 @@ std::unique_ptr<cudf::column> string_to_decimal(
...
+ * @param ansi_mode If true, strict conversion and throws on erorr.
+ *                  If false, null invalid entries.
+ * @param stream Stream on which to operate.
+ * @param mr Memory resource for returned column
+ * @return std::unique_ptr<column> Integer column that was created from string_col.
```
jlowe, line 105:
```suggestion
 * @return std::unique_ptr<column> Floating point column that was created from string_col.
```

src/main/cpp/src/cast_string.hpp
```diff
@@ -92,4 +92,23 @@ std::unique_ptr<cudf::column> string_to_decimal(
   bool ansi_mode,
   rmm::cuda_stream_view stream,
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Convert a string column into an float column.
+ *
+ * @param dtype Type of column to return.
+ * @param string_col Incoming string column to convert to integers.
+ * @param ansi_mode If true, strict conversion and throws on erorr.
```
jlowe, line 101:
```suggestion
 * @param ansi_mode If true, strict conversion and throws on error.
```

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -48,6 +48,19 @@ public static ColumnVector toDecimal(ColumnView cv, boolean ansiMode, int precis
     return new ColumnVector(toDecimal(cv.getNativeView(), ansiMode, precision, scale));
   }
 
+    /**
+   * Convert a string column to an float column of a specified type.
```
jlowe, line 52:
```suggestion
  /**
   * Convert a string column to a float column of a specified type.
```

jlowe (commented):
Would be good to have someone with more CUDA experience review cast_string_to_float.cu.

src/main/cpp/CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ add_library(
   src/CastStringJni.cpp
   src/NativeParquetJni.cpp
   src/cast_string.cu
+  src/cast_string_to_float.cu
```
ttnghia, line 147: What is `cast_string.cu`? This new file name is more meaningful, thus we may want to rename the old file to make it more meaningful too.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -0,0 +1,730 @@
...
+#include <cudf/column/column_factories.hpp>
+#include <cudf/detail/utilities/cuda.cuh>
+#include <cudf/detail/utilities/integer_utils.hpp>
+#include <cudf/scalar/scalar_factories.hpp>
+#include <cudf/strings/convert/convert_floats.hpp>
+#include <cudf/utilities/bit.hpp>
+
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/column_utilities.hpp>
+#include <cudf_test/column_wrapper.hpp>
```
ttnghia, line 30: Remove them?

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -0,0 +1,730 @@
...
+  __device__ string_to_float(T* _out,
+                             bitmask_type* _validity,
+                             int32_t* _ansi_except,
+                             size_type* _valid_count,
+                             const char* const _chars,
+                             offset_type const* _offsets,
+                             int _warp_id,
+                             uint64_t const* const _ipow,
+                             bitmask_type const* _incoming_null_mask,
+                             size_type const _num_rows)
```
ttnghia, line 69: I feel something's wrong: This functor takes too many parameters. It would be great if we can decouple them by using multiple functors instead.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -0,0 +1,730 @@
...
+
+    // 0th thread in each warp updates the validity
+    size_type const row_id = warp_id;
+    if (threadIdx.x % 32 == 0 && valid) {
+      // uses atomics
+      cudf::set_bit(validity, row_id);
+    }
+  }
+
+  T* out;
```
ttnghia, line 562: Since these members are not public, please add prefix (`_`) so they can be recognized in the code so we can distinguish between member vars and local vars.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -0,0 +1,730 @@
...
+                                       int32_t* ansi_except,
+                                       size_type* valid_count,
+                                       const char* const chars,
+                                       offset_type const* offsets,
+                                       bitmask_type const* incoming_null_mask,
+                                       size_type const num_rows)
+{
+  size_type const tid     = threadIdx.x + (blockDim.x * blockIdx.x);
+  size_type const warp_id = tid / 32;
+  size_type const row     = warp_id;
```
ttnghia, line 599: They seem redundant: The `string_to_float` functor later will compute them.

src/main/cpp/src/cast_string_to_float.cu
hyperbolic2346: The design choice was to use a functor to prevent passing so many variables around to each function, but that leaves us passing it all in at the beginning to feed into that object. I don't know how much we could free up as this is a single kernel invocation for speed reasons.

hyperbolic2346: Thinking about this more, we can pull the base functor into this function and then call the private functions with parameters, but it feels like too much for this function. If we push that out to a function we're back in the same boat as now as we have to pass all that into that next function.

src/main/cpp/src/cast_string_to_float.cu
hyperbolic2346: Found a few of these and removed them. Thanks!

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -0,0 +1,716 @@
...
+        // exponent
+        //   required would be smaller then the smallest representable exponent.
+        //
+        // https://en.wikipedia.org/wiki/Denormal_number
+        //
+
+        double const exponent = exp10(static_cast<double>(std::abs(exp_ten)));
+        double const result   = exp_ten < 0 ? digitsf / exponent : digitsf * exponent;
+
+        _out[_row] = result;
```
ttnghia, line 179: This may do a narrow cast and issue a compiler warning. So we need explicit `static_cast<T>` here.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -0,0 +1,716 @@
...
+    ipow[16] = 10000000000000000;
+    ipow[17] = 100000000000000000;
+    ipow[18] = 1000000000000000000;
+  }
+  __syncthreads();
+
+  // convert
+  string_to_float<T, block_size> convert(
+    out, validity, ansi_except, valid_count, chars, offsets, ipow, incoming_null_mask, num_rows);
+  convert();
```
ttnghia, line 623:
```suggestion
  string_to_float<T, block_size>{
    out, validity, ansi_except, valid_count, chars, offsets, ipow, incoming_null_mask, num_rows}();
```

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -0,0 +1,716 @@
...
+std::unique_ptr<column> string_to_float(data_type dtype,
+                                        strings_column_view const& string_col,
+                                        bool ansi_mode,
+                                        rmm::cuda_stream_view stream,
+                                        rmm::mr::device_memory_resource* mr)
+{
+  if ((string_col.size() == 0) ||
+      (dtype != data_type{type_id::FLOAT32} && dtype != data_type{type_id::FLOAT64})) {
+    return std::make_unique<column>();
+  }
```
ttnghia, line 648: This should throw an exception instead, because non-float target types are invalid input.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -0,0 +1,716 @@
...
+
+  auto out = cudf::make_numeric_column(dtype, string_col.size(), mask_state::ALL_NULL, stream, mr);
+
+  using ScalarType = cudf::scalar_type_t<size_type>;
+  auto valid_count = cudf::make_numeric_scalar(cudf::data_type(cudf::type_id::INT32));
+  auto ansi_count  = cudf::make_numeric_scalar(cudf::data_type(cudf::type_id::INT32));
```
ttnghia, line 654: Nit: Should use this style:
```suggestion
  auto valid_count = cudf::make_numeric_scalar(cudf::data_type(cudf::type_to_id<size_type>()));
  auto ansi_count  = cudf::make_numeric_scalar(cudf::data_type(cudf::type_to_id<size_type>()));
```

src/main/cpp/tests/cast_string.cpp
```diff
@@ -473,3 +481,138 @@ TEST_F(StringToDecimalTests, Edges)
...
+                                        "18446744073709551619999999999999",
+                                        "-18446744073709551609",
+                                        "-18446744073709551610",
+                                        "-184467440737095516199999999999997"};
+
+  std::vector<bool> valids{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
+
+  auto expected =
+    cudf::strings::to_floats(strings_column_view(in), cudf::data_type{type_to_id<TypeParam>()});
+  expected->set_null_mask(cudf::test::detail::make_null_mask(valids.begin(), valids.end()));
```
ttnghia, line 510: This is less error-prone:
```suggestion
  auto valid_iter = cudf::test::iterators::no_nulls();
  auto expected =
    cudf::strings::to_floats(strings_column_view(in), cudf::data_type{type_to_id<TypeParam>()});
  expected->set_null_mask(cudf::test::detail::make_null_mask(valid_iter, valid_iter + expected->size()));
```

src/main/cpp/tests/cast_string.cpp
```diff
@@ -473,3 +481,138 @@ TEST_F(StringToDecimalTests, Edges)
...
+
+TYPED_TEST(StringToFloatTests, InfNaN)
+{
+  cudf::test::strings_column_wrapper in{"NaN", "-Infinity", "inf", "Infinity", "-inf", "-nan"};
+
+  std::vector<bool> valids{1, 1, 1, 1, 1, 0};
+
+  auto expected =
+    cudf::strings::to_floats(strings_column_view(in), cudf::data_type{type_to_id<TypeParam>()});
+  expected->set_null_mask(cudf::test::detail::make_null_mask(valids.begin(), valids.end()));
```
ttnghia, line 526: Similar to above, but using `cudf::test::iterators::null_at(5)`.

src/main/cpp/tests/cast_string.cpp
```diff
@@ -473,3 +481,138 @@ TEST_F(StringToDecimalTests, Edges)
...
+
+TYPED_TEST(StringToFloatTests, InvalidValues)
+{
+  cudf::test::strings_column_wrapper in{"A", "null", "na7.62", "e", ".", "", "f"};
+
+  std::vector<bool> valids{0, 0, 0, 0, 0, 0, 0, 0};
+
+  auto expected =
+    cudf::strings::to_floats(strings_column_view(in), cudf::data_type{type_to_id<TypeParam>()});
+  expected->set_null_mask(cudf::test::detail::make_null_mask(valids.begin(), valids.end()));
```
ttnghia, line 542: Similar to above, using `cudf::test::iterators::all_nulls()`.

src/main/cpp/tests/cast_string.cpp
```diff
@@ -473,3 +481,138 @@ TEST_F(StringToDecimalTests, Edges)
...
+  cudf::test::strings_column_wrapper in[] = {{"A"}, {"."}, {"e"}};
+
+  std::vector<bool> valids{0};
+
+  for (auto& col : in) {
+    try {
+      auto const result = spark_rapids_jni::string_to_float(data_type{type_to_id<TypeParam>()},
+                                                            strings_column_view{col},
+                                                            true,
+                                                            rmm::cuda_stream_default);
```
ttnghia, line 561: All stream values should be `cudf::default_stream_value`.

src/main/cpp/tests/cast_string.cpp
```diff
@@ -473,3 +481,138 @@ TEST_F(StringToDecimalTests, Edges)
...
+    data_type{type_to_id<TypeParam>()}, strings_column_view{in}, false, rmm::cuda_stream_default);
+
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(result->view(), expected->view());
+}
+
+TYPED_TEST(StringToFloatTests, ANSIInvalids)
+{
+  cudf::test::strings_column_wrapper in[] = {{"A"}, {"."}, {"e"}};
+
+  std::vector<bool> valids{0};
```
ttnghia, line 554: Unused var!

src/main/cpp/tests/cast_string.cpp
```diff
@@ -473,3 +481,138 @@ TEST_F(StringToDecimalTests, Edges)
...
+
+  auto const result = spark_rapids_jni::string_to_float(
+    data_type{type_to_id<TypeParam>()}, strings_column_view{in}, false, rmm::cuda_stream_default);
+
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(result->view(), expected->view());
+}
+
+TYPED_TEST(StringToFloatTests, ANSIInvalids)
+{
+  cudf::test::strings_column_wrapper in[] = {{"A"}, {"."}, {"e"}};
```
ttnghia, line 552: I think that we should add more cases for the ansi test. Here are only very simple cases.

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -48,6 +48,19 @@ public static ColumnVector toDecimal(ColumnView cv, boolean ansiMode, int precis
     return new ColumnVector(toDecimal(cv.getNativeView(), ansiMode, precision, scale));
   }
 
+  /**
+   * Convert a string column to a float column of a specified type.
```
ttnghia, line 52:
```suggestion
   * Convert a string column to a given floating-point type column.
```

ttnghia (changes requested):
Finally I have read through the code. My head was about to explode without calling to any `xxx::explode`.

src/main/cpp/tests/cast_string.cpp
revans2, line 552: When are we going to support hex as input? Not exactly the same here, but it is something that Spark supports and we cannot/do not support.

ie `0x1p0 => 2.0`

ttnghia: We can always construct a string from hex, but with some simple hacks (the working code may not be exactly like this):
```
char data[][] { { 0x0a, 0x0b, 0x01, 0x04 }, { 0x0a, 0x0b, 0x01, 0x04 }, .....};
cudf::test::strings_column_wrapper in[] = { {std::string(data[0], 4)}, {std::string(data[1], 4)}, .....};
```

ttnghia: If you want to use hex frequently, you can write a utility function for simplifying the work in test too.

revans2: The reason I ask is that I thought one of the goals of this was to fix the issues with casting a string to floating point.  It is off by default because of these issues.  Many of them are corner cases that CUDF does not support, but we need to in order to have this operation on by default.

Hex strings is one of the reasons it is off by default. I didn't see any tests for hex strings so that is why I asked. I probably put this comment in the wrong place with the ASCII test. It is not really related.

```
        willNotWorkOnGpu("Currently hex values aren't supported on the GPU. Also note " +
            "that casting from string to float types on the GPU returns incorrect results when " +
            "the string represents any number \"1.7976931348623158E308\" <= x < " +
            "\"1.7976931348623159E308\" and \"-1.7976931348623159E308\" < x <= " +
            "\"-1.7976931348623158E308\" in both these cases the GPU returns Double.MaxValue " +
            "while CPU returns \"+Infinity\" and \"-Infinity\" respectively. To enable this " +
            s"operation on the GPU, set ${RapidsConf.ENABLE_CAST_STRING_TO_FLOAT} to true.")
```

If we are not fixing these issues can we have a follow on issue filed to fix them, now that we have a custom kernel? Also what are the performance differences vs the existing kernel. The bug just listed this as a performance fix, so if that is the case then have we measured the performance improvement vs the existing code? I assume it is a lot better because it is a single kernel instead of multiple etc..., but it would be nice to know.

If we did fix the issues can we have some tests to verify that we did?

ttnghia: Sorry I misunderstood what you asked. I meant we can **test** string constructed from hex **value** in the code (like the snippet above: `string` from `char` given as hex values). For converting hex string like `0x41973333` then we didn't support yet and definitely need a custom kernel.

ttnghia: @revans2 I just realized that cudf already has `hex_to_integers` API that can convert strings into integer numbers. I believe that we can just do a `bit_cast` from the output integers into the target doubles, which is very straightforward. If there is anything missing here then we can easily add that.

revans2: Not really. The hex format is not just parse the hex to get the bit representation.  The values fall under a specific format that appears to be the same for both C, C++, and Java

https://www.exploringbinary.com/hexadecimal-floating-point-constants/  

From what I can tell the form is `0x[HEX VALUE]p[EXP]` where the result is the parsed `HEX VALUE * 2 ^ EXP`. and oddly the `EXP` is represented in decimal not hex.

Also we have the issue that we need to be able to support these intermixed in the same column.

`["1.0", "0x1p0"]` so we would need to move that kind of logic into our custom kernel.

hyperbolic2346: Added an issue for this.

hyperbolic2346:
Performance information added to [spark-rapids PR](https://github.com/NVIDIA/spark-rapids/pull/6761) since the original code path was entirely inside that code base.

src/main/cpp/tests/cast_string.cpp
hyperbolic2346: I added tests from the plugin that failed to ensure they didn't regress. The plugin tests are VERY extensive.

hyperbolic2346:
build

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -641,10 +641,9 @@ std::unique_ptr<column> string_to_float(data_type dtype,
                                         rmm::cuda_stream_view stream,
                                         rmm::mr::device_memory_resource* mr)
 {
-  if ((string_col.size() == 0) ||
-      (dtype != data_type{type_id::FLOAT32} && dtype != data_type{type_id::FLOAT64})) {
-    return std::make_unique<column>();
-  }
+  CUDF_EXPECTS(dtype == data_type{type_id::FLOAT32} || dtype == data_type{type_id::FLOAT64},
+               "invalid float data type");
+  if (string_col.size() == 0) { return std::make_unique<column>(); }
```
ttnghia, line 646: Probably empty column of floating point type?

ttnghia:
Please be aware of https://github.com/rapidsai/cudf/pull/11875. You may need to change the code in this PR (and throughout this repo, in another separate PR).

ttnghia (approved):
I changed my mind. Please finalize this and merge it ASAP then make a separate PR to adopt that breaking change.

src/main/cpp/src/cast_string_to_float.cu
hyperbolic2346: Good point, I should. Thanks.

hyperbolic2346:
build

ttnghia (approved):

PR #633, opened 2022-10-12
Fixing empty columns when casting to integer or decimal crashing

hyperbolic2346:
As outlined in https://github.com/NVIDIA/spark-rapids/issues/6748 passing a completely empty string column into the casting kernel crashes. This fixes that since an empty column is valid.

closes https://github.com/NVIDIA/spark-rapids/issues/6748

### DIFF
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -643,6 +643,10 @@ struct string_to_integer_impl {
                                      rmm::cuda_stream_view stream,
                                      rmm::mr::device_memory_resource* mr)
   {
+    if (string_col.size() == 0) {
+      return std::make_unique<column>(data_type{type_to_id<T>()}, 0, rmm::device_buffer{});
+    }
+
     rmm::device_uvector<T> data(string_col.size(), stream, mr);
     auto const num_words = bitmask_allocation_size_bytes(string_col.size()) / sizeof(bitmask_type);
     rmm::device_uvector<bitmask_type> null_mask(num_words, stream, mr);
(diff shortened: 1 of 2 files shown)


src/main/cpp/src/cast_string.cu
```diff
@@ -762,6 +762,8 @@ std::unique_ptr<column> string_to_integer(data_type dtype,
                                           rmm::cuda_stream_view stream,
                                           rmm::mr::device_memory_resource* mr)
 {
+  if (string_col.size() == 0) { return std::make_unique<column>(); }
```
jlowe, line 765: Doesn't this need to setup the data type of the resulting column?

src/main/cpp/tests/cast_string.cpp
```diff
@@ -515,3 +527,13 @@ TEST_F(StringToDecimalTests, Edges)
...
 }
+
+TEST_F(StringToDecimalTests, Empty)
+{
+  auto empty = std::make_unique<column>(data_type{type_id::STRING}, 0, rmm::device_buffer{});
+
+  auto const result = spark_rapids_jni::string_to_decimal(
+    8, 2, strings_column_view{empty->view()}, false, rmm::cuda_stream_default);
+
+  EXPECT_EQ(result->size(), 0);
```
jlowe, line 538: Need to also check resulting column type

src/main/cpp/src/cast_string.cu
```diff
@@ -785,6 +787,7 @@ std::unique_ptr<column> string_to_decimal(int32_t precision,
...
```
jlowe, line 790: Need to setup column data type

hyperbolic2346:
build

revans2 (approved):

jlowe (approved):

PR #643, opened 2022-10-14
Verify automerge fix of Token permission

pxLi:
Signed-off-by: Peixin Li <pxli@nyu.edu>

as title


pxLi:
not sure if this could fix the issue. Going to force merge and verify the automerge status

PR #650, opened 2022-10-14
Add in support for hilbert index as a part of zorder work

revans2:
This is related to the interleave bits work done for zorder before. This is work to support the databricks internal implementation for zorder.

### DIFF
--- a/src/main/cpp/src/ZOrderJni.cpp
+++ b/src/main/cpp/src/ZOrderJni.cpp
@@ -35,4 +35,20 @@ Java_com_nvidia_spark_rapids_jni_ZOrder_interleaveBits(JNIEnv *env, jclass, jlon
   CATCH_STD(env, 0);
 }
 
+JNIEXPORT jlong JNICALL
+Java_com_nvidia_spark_rapids_jni_ZOrder_hilbertIndex(JNIEnv *env, jclass, jint num_bits, jlongArray input_columns) {
+  JNI_NULL_CHECK(env, input_columns, "input is null", 0);
+
+  try {
+    cudf::jni::auto_set_device(env);
+    cudf::jni::native_jpointerArray<cudf::column_view> n_input_columns(env, input_columns);
+    cudf::table_view tbl(n_input_columns.get_dereferenced());
+
+    return cudf::jni::ptr_as_jlong(spark_rapids_jni::hilbert_index(num_bits, tbl).release());
+  }
+  CATCH_STD(env, 0);
+}
+
+
+
 }
--- a/src/main/cpp/src/zorder.cu
+++ b/src/main/cpp/src/zorder.cu
@@ -255,12 +255,12 @@ std::unique_ptr<cudf::column> hilbert_index(
 
   auto const output_dv_ptr = cudf::mutable_column_device_view::create(*output_data_col, stream);
 
-  thrust::for_each_n(
+  thrust::transform(
     rmm::exec_policy(stream),
     thrust::make_counting_iterator<cudf::size_type>(0),
-    num_rows,
-    [output_col = *output_dv_ptr,
-     num_bits_per_entry,
+    thrust::make_counting_iterator<cudf::size_type>(0) + num_rows,
+    output_dv_ptr->begin<int64_t>(),
+    [num_bits_per_entry,
      num_columns,
      input = *input_dv] __device__ (cudf::size_type row_index) {
        long_backed_array row(num_bits_per_entry);
@@ -271,8 +271,8 @@ std::unique_ptr<cudf::column> hilbert_index(
        }
 
        auto const transposed_index = hilbert_transposed_index(row, num_bits_per_entry, num_columns);
-       output_col.data<uint64_t>()[row_index] =
-         to_hilbert_index(transposed_index, num_bits_per_entry, num_columns);
+       return static_cast<int64_t>(
+         to_hilbert_index(transposed_index, num_bits_per_entry, num_columns));
      });
 
   return output_data_col;
--- a/src/main/cpp/src/zorder.cu
+++ b/src/main/cpp/src/zorder.cu
@@ -30,13 +30,13 @@
 namespace {
 
 // pretends to be an array of uint32_t, but really only stores
-// the data in a long with a set number of bits allocated for
+// the data in an int type with a set number of bits allocated for
 // each item (num_bits_per_entry)
-struct long_backed_array {
-  using data_type = uint64_t;
-  long_backed_array() = delete;
-  __device__ explicit long_backed_array(int32_t num_bits_per_entry): data(0),
-    num_bits_per_entry(num_bits_per_entry),  mask(static_cast<data_type>((1L << num_bits_per_entry) - 1)) {}
+template <typename data_type>
+struct uint_backed_array {
+  uint_backed_array() = delete;
+  __device__ explicit uint_backed_array(int32_t num_bits_per_entry): data(0),
+    num_bits_per_entry(num_bits_per_entry),  mask(static_cast<uint32_t>((1L << num_bits_per_entry) - 1)) {}
 
   __device__ uint32_t operator[](int32_t i) const {
     int32_t offset = num_bits_per_entry * i;
@@ -50,7 +50,7 @@ struct long_backed_array {
   }
 
 private:
-  uint64_t data;
+  data_type data;
   int32_t const num_bits_per_entry;
   uint32_t const mask;
 };
@@ -64,7 +64,7 @@ private:
 // With thanks also to Paul Chernoch who published a C# algorithm for Skilling's
 // work on StackOverflow and
 // <a href="https://github.com/paulchernoch/HilbertTransformation">GitHub</a>.
-__device__ uint64_t to_hilbert_index(long_backed_array const & transposed_index,
+__device__ uint64_t to_hilbert_index(uint_backed_array<uint64_t> const & transposed_index,
         int32_t const num_bits_per_entry, int32_t const num_dimensions) {
   uint64_t b = 0;
   int32_t const length = num_bits_per_entry * num_dimensions;
@@ -83,7 +83,7 @@ __device__ uint64_t to_hilbert_index(long_backed_array const & transposed_index,
   return b;
 }
 
-__device__ long_backed_array hilbert_transposed_index(long_backed_array const & point,
+__device__ uint_backed_array<uint64_t> hilbert_transposed_index(uint_backed_array<uint64_t> const & point,
         int32_t const num_bits_per_entry, int32_t const num_dimensions) {
   uint32_t const M = 1L << (num_bits_per_entry - 1);
   int32_t const n = num_dimensions;
@@ -247,7 +247,7 @@ std::unique_ptr<cudf::column> hilbert_index(
     [num_bits_per_entry,
      num_columns,
      input = *input_dv] __device__ (cudf::size_type row_index) {
-       long_backed_array row(num_bits_per_entry);
+       uint_backed_array<uint64_t> row(num_bits_per_entry);
        for (cudf::size_type column_index = 0; column_index < num_columns; column_index++) {
          auto const column = input.column(column_index);
          uint32_t const data = column.is_valid(row_index) ? column.data<uint32_t>()[row_index] : 0;
--- a/src/main/cpp/src/zorder.cu
+++ b/src/main/cpp/src/zorder.cu
@@ -33,12 +33,13 @@ namespace {
 // the data in a long with a set number of bits allocated for
 // each item (num_bits_per_entry)
 struct long_backed_array {
+  using data_type = uint64_t;
   long_backed_array() = delete;
   ~long_backed_array() = default;
   long_backed_array(long_backed_array const&) = default;  ///< Copy constructor
   long_backed_array(long_backed_array&&) = default;  ///< Move constructor
-  inline __device__ explicit long_backed_array(int32_t num_bits_per_entry): data(0),
-    num_bits_per_entry(num_bits_per_entry),  mask(static_cast<uint64_t>((1L << num_bits_per_entry) - 1)) {}
+  __device__ explicit long_backed_array(int32_t num_bits_per_entry): data(0),
+    num_bits_per_entry(num_bits_per_entry),  mask(static_cast<data_type>((1L << num_bits_per_entry) - 1)) {}
 
   /**
    * @brief Copy assignment operator
@@ -53,21 +54,21 @@ struct long_backed_array {
    */
   long_backed_array& operator=(long_backed_array&&) = default;
 
-  inline __device__ uint32_t operator[](int i) const {
+  __device__ uint32_t operator[](int32_t i) const {
     int32_t offset = num_bits_per_entry * i;
     return (data >> offset) & mask;
   }
 
-  inline __device__ void set(int i, uint32_t value) {
+  __device__ void set(int32_t i, uint32_t value) {
     int32_t offset = i * num_bits_per_entry;
-    uint64_t masked_data = data & ~(static_cast<uint64_t>(mask) << offset);
-    data = masked_data | (static_cast<uint64_t>(value & mask) << offset);
+    data_type masked_data = data & ~(static_cast<data_type>(mask) << offset);
+    data = masked_data | (static_cast<data_type>(value & mask) << offset);
   }
 
 private:
   uint64_t data;
-  int32_t num_bits_per_entry;
-  uint32_t mask;
+  int32_t const num_bits_per_entry;
+  uint32_t const mask;
 };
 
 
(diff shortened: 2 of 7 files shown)


revans2:
build

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
...
+// pretends to be an array of uint32_t, but really only stores
+// the data in a long with a set number of bits allocated for
+// each item
+struct long_backed_array {
+  long_backed_array() = delete;
+  ~long_backed_array() = default;
+  long_backed_array(long_backed_array const&) = default;  ///< Copy constructor
+  long_backed_array(long_backed_array&&) = default;  ///< Move constructor
+  inline __device__ explicit long_backed_array(int32_t num_bits): data(0), 
+    num_bits(num_bits),  mask(static_cast<uint64_t>((1L << num_bits) - 1)) {}
```
hyperbolic2346, line 41: Should `num_bits` be named something more like `num_bits_per_entry` or some such? It took me a while to make the connection that it wasn't the total number of bits in the array. I also don't see any verification that the number of bits desired per element * number of elements is < sizeof(long).

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
...
+
+// Most of the hilbert index code is based off of the work done by David Moten at
+// https://github.com/davidmoten/hilbert-curve, which has the following Note in
+// the code too
+// This algorithm is derived from work done by John Skilling and published
+// in "Programming the Hilbert curve". (c) 2004 American Institute of Physics.
+// With thanks also to Paul Chernoch who published a C# algorithm for Skilling's
+// work on StackOverflow and
+// <a href="https://github.com/paulchernoch/HilbertTransformation">GitHub</a>.
+__device__ uint64_t to_hilbert_index(const long_backed_array & transposed_index, const int num_bits, const int num_dimensions) {
```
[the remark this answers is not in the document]
revans2: It is the same `num_bits`. I called it this to match closely with the original code.

https://github.com/davidmoten/hilbert-curve/blob/758058c0b86ff01edffb46317f04a619d861ba14/src/main/java/org/davidmoten/hilbert/HilbertCurve.java#L26

but I added `num_` to it because I thought `bits` was not descriptive enough. Happy to make it clearer.

revans2:
build

src/main/cpp/src/ZOrderJni.cpp
```diff
@@ -35,4 +35,20 @@ Java_com_nvidia_spark_rapids_jni_ZOrder_interleaveBits(JNIEnv *env, jclass, jlon
   CATCH_STD(env, 0);
 }
 
+JNIEXPORT jlong JNICALL
+Java_com_nvidia_spark_rapids_jni_ZOrder_hilbertIndex(JNIEnv *env, jclass, jint num_bits, jlongArray input_columns) {
```
ttnghia, line 39: Why not pass in a table from Java instead?

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
 #include <thrust/iterator/constant_iterator.h>
 #include <thrust/iterator/counting_iterator.h>
 
+namespace {
+
+// pretends to be an array of uint32_t, but really only stores
+// the data in a long with a set number of bits allocated for
+// each item (num_bits_per_entry)
+struct long_backed_array {
+  long_backed_array() = delete;
```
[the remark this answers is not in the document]
ttnghia: If we declare `data_type` inside this struct, we should call `long_backed_array::data_type` all the time. Otherwise, we should put it somewhere else (probably above this struct).

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
...
+  ~long_backed_array() = default;
+  long_backed_array(long_backed_array const&) = default;  ///< Copy constructor
+  long_backed_array(long_backed_array&&) = default;  ///< Move constructor
```
ttnghia, line 39: I believe that you don't have to declare these default methods. Only declare what you need.

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
...
+   *
+   * @return Reference to this object
+   */
+  long_backed_array& operator=(long_backed_array const&) = default;
+  /**
+   * @brief Move assignment operator
+   *
+   * @return Reference to this object (after transferring ownership)
+   */
+  long_backed_array& operator=(long_backed_array&&) = default;
```
ttnghia, line 54: Similar to above.

src/main/cpp/src/zorder.cu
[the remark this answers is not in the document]
hyperbolic2346: That is the gist of the change. I was so tempted to create ```
template <typename T>
struct backed_array {
}```

I worry I'm too far in the weeds though and missing the point.

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
...
+
+// pretends to be an array of uint32_t, but really only stores
+// the data in a long with a set number of bits allocated for
+// each item (num_bits_per_entry)
+struct long_backed_array {
+  long_backed_array() = delete;
+  ~long_backed_array() = default;
+  long_backed_array(long_backed_array const&) = default;  ///< Copy constructor
+  long_backed_array(long_backed_array&&) = default;  ///< Move constructor
+  inline __device__ explicit long_backed_array(int32_t num_bits_per_entry): data(0),
```
ttnghia, line 40: Also we can avoid redundant `inline` keyword. When a function only has definition (implemented without a separate declaration) then it is implicitly inline. So all the `inline` keyword in this file can be removed.

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
...
+  long_backed_array& operator=(long_backed_array const&) = default;
+  /**
+   * @brief Move assignment operator
+   *
+   * @return Reference to this object (after transferring ownership)
+   */
+  long_backed_array& operator=(long_backed_array&&) = default;
+
+  inline __device__ uint32_t operator[](int i) const {
+    int32_t offset = num_bits_per_entry * i;
```
ttnghia, line 57: Please try to enforce consistency: use `int32_t` all the time and avoid `int`.

src/main/cpp/src/zorder.cu
```diff
@@ -114,4 +226,53 @@ std::unique_ptr<cudf::column> interleave_bits(
...
 }
 
+std::unique_ptr<cudf::column> hilbert_index(
+  int32_t const num_bits_per_entry,
+  cudf::table_view const& tbl,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr) {
+ 
+  auto num_rows = tbl.num_rows();
+  auto num_columns = tbl.num_columns();
```
ttnghia, line 236: Use `const` as much as possible.

src/main/cpp/src/zorder.cu
```diff
@@ -114,4 +226,53 @@ std::unique_ptr<cudf::column> interleave_bits(
...
+  CUDF_EXPECTS(num_bits_per_entry * num_columns <= 64, "we only support up to 64 bits of output right now.");
+  CUDF_EXPECTS(num_columns > 0, "at least one column is required.");
+
+  CUDF_EXPECTS(
+    std::all_of(tbl.begin(),
+                tbl.end(),
+                [](cudf::column_view const& col) { return col.type().id() == cudf::type_id::INT32; }),
+    "All columns of the input table must be INT32.");
+
+  auto input_dv = cudf::table_device_view::create(tbl, stream);
```
ttnghia, line 248: Also `auto const`.

src/main/cpp/src/zorder.cu
```diff
@@ -114,4 +226,53 @@ std::unique_ptr<cudf::column> interleave_bits(
...
+     num_columns,
+     input = *input_dv] __device__ (cudf::size_type row_index) {
+       long_backed_array row(num_bits_per_entry);
+       for (cudf::size_type column_index = 0; column_index < num_columns; column_index++) {
+         auto const column = input.column(column_index);
+         uint32_t const data = column.is_valid(row_index) ? column.data<uint32_t>()[row_index] : 0;
+         row.set(column_index, data);
+       }
+
+       auto transposed_index = hilbert_transposed_index(row, num_bits_per_entry, num_columns);
```
ttnghia, line 270: `auto const`.

src/main/cpp/src/zorder.cu
```diff
@@ -114,4 +226,53 @@ std::unique_ptr<cudf::column> interleave_bits(
...
+
+  thrust::for_each_n(
+    rmm::exec_policy(stream),
+    thrust::make_counting_iterator<cudf::size_type>(0),
+    num_rows,
+    [output_col = *output_dv_ptr,
+     num_bits_per_entry,
+     num_columns,
+     input = *input_dv] __device__ (cudf::size_type row_index) {
+       long_backed_array row(num_bits_per_entry);
```
ttnghia, line 263: `const`.

src/main/cpp/src/zorder.cu
```diff
@@ -114,4 +226,53 @@ std::unique_ptr<cudf::column> interleave_bits(
...
+
+  auto output_data_col = cudf::make_numeric_column(
+      cudf::data_type{cudf::type_id::INT64}, num_rows, cudf::mask_state::UNALLOCATED, stream, mr);
+
+  auto output_dv_ptr = cudf::mutable_column_device_view::create(*output_data_col, stream);
+
+  thrust::for_each_n(
+    rmm::exec_policy(stream),
+    thrust::make_counting_iterator<cudf::size_type>(0),
+    num_rows,
```
ttnghia, line 258: Nit: This can be rewritten using `thrust::transform`.

src/main/cpp/src/zorder.hpp
```diff
@@ -29,4 +29,10 @@ std::unique_ptr<cudf::column> interleave_bits(
   rmm::cuda_stream_view stream        = rmm::cuda_stream_default,
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
+std::unique_ptr<cudf::column> hilbert_index(
+  int32_t const num_bits,
+  cudf::table_view const& tbl,
+  rmm::cuda_stream_view stream        = rmm::cuda_stream_default,
```
ttnghia, line 35: I think all the default stream should be accessed through `cudf::default_stream_value`.

src/main/java/com/nvidia/spark/rapids/jni/ZOrder.java
```diff
@@ -54,5 +54,34 @@ public static ColumnVector interleaveBits(int numRows, ColumnVector ... inputCol
     return new ColumnVector(interleaveBits(addrs));
   }
 
+  /**
+   * Yes a hilbertIndex is not technically used in zorder, but it is an alternative way to
+   * cluster the data that databricks uses, and that is why we have it here. Please note that
+   * this currently only supports indexes where numBits * inputColumns.length <= 64.
```
ttnghia, line 60:
```suggestion
   * this currently only supports indexes where numBits * inputColumns.length <= 64.
   
```

src/main/java/com/nvidia/spark/rapids/jni/ZOrder.java
```diff
@@ -54,5 +54,34 @@ public static ColumnVector interleaveBits(int numRows, ColumnVector ... inputCol
...
+   * @param numRows the number of rows. Used if inputColumns is empty. I think this is also a corner
+   *                case that can never happen in practice, but I am just covering my bases here.
+   *                a column of 0 is returned in this case.
+   * @param inputColumns The columns to intermix.
+   * @return the corresponding indexes stored as long values.
+   */
+  public static ColumnVector hilbertIndex(int numBits, int numRows, ColumnVector ... inputColumns) {
+    if (inputColumns.length == 0) {
+      try (Scalar zero = Scalar.fromLong(0)) {
+        return ColumnVector.fromScalar(zero, numRows);
```
ttnghia, line 72: Since the column returned from cuda code is `UINT64` type, this should also be `fromUnsignedLong`.

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
...
+  long_backed_array& operator=(long_backed_array&&) = default;
+
+  inline __device__ uint32_t operator[](int i) const {
+    int32_t offset = num_bits_per_entry * i;
+    return (data >> offset) & mask;
+  }
+
+  inline __device__ void set(int i, uint32_t value) {
+    int32_t offset = i * num_bits_per_entry;
+    uint64_t masked_data = data & ~(static_cast<uint64_t>(mask) << offset);
```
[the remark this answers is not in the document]
revans2: I don't think we can drop the static_cast. I need to be able to shift the mask above what a 32-bit value can hold.

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,118 @@
...
+
+// Most of the hilbert index code is based off of the work done by David Moten at
+// https://github.com/davidmoten/hilbert-curve, which has the following Note in
+// the code too
+// This algorithm is derived from work done by John Skilling and published
+// in "Programming the Hilbert curve". (c) 2004 American Institute of Physics.
+// With thanks also to Paul Chernoch who published a C# algorithm for Skilling's
+// work on StackOverflow and
+// <a href="https://github.com/paulchernoch/HilbertTransformation">GitHub</a>.
+__device__ uint64_t to_hilbert_index(const long_backed_array & transposed_index, const int num_bits_per_entry, const int num_dimensions) {
```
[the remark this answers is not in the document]
revans2: I would prefer to keep the uint64_t here separate from the data_type in the long_backed_array. Conceptually they are different types that are not related to each other.

src/main/cpp/src/zorder.cu
```diff
@@ -114,4 +226,53 @@ std::unique_ptr<cudf::column> interleave_bits(
...
+  CUDF_EXPECTS(
+    std::all_of(tbl.begin(),
+                tbl.end(),
+                [](cudf::column_view const& col) { return col.type().id() == cudf::type_id::INT32; }),
+    "All columns of the input table must be INT32.");
+
+  auto input_dv = cudf::table_device_view::create(tbl, stream);
+
+  auto output_data_col = cudf::make_numeric_column(
+      cudf::data_type{cudf::type_id::INT64}, num_rows, cudf::mask_state::UNALLOCATED, stream, mr);
```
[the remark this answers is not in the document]
revans2: I want an `INT64` so it matches what Spark expects.  Java does not have uint types so I want the output to actually be an int.

src/main/cpp/src/ZOrderJni.cpp
revans2: It is minor but this reduces the number of JNI calls and in my opinion fits better with how Spark uses it. If we pass a table down from java then I have to create a table in java, which will require a JNI call.  I use the table internally to verify that the length of the columns are all the same, and so I can reuse the existing device_table_view.  Why does java need to know that is how I am doing it.

src/main/java/com/nvidia/spark/rapids/jni/ZOrder.java
revans2: Like I said. Spark/Java do not use/support `uint64`. The value returned was computed using a uint64_t to avoid shift issues, but in order to match exactly what the java code does will be returned as an `int64`/`long`.

src/main/cpp/src/zorder.cu
revans2: We modify the row. Why would I mark it as `const` to say that we cannot modify it?

revans2:
@hyperbolic2346 and @ttnghia I think I have addressed or replied to all of your comments. There were a few `const` that caused compile errors, but I think I fixed most of it. Please take another look.

revans2:
build

src/main/cpp/src/zorder.cu
ttnghia: They are just to dilute your code. Except for the overloaded constructor (you provide one then the compiler will not provide the default constructor), other methods are implicitly generated the same as you write `default`. So explicitly writing them `... = default` is redundant.

revans2:
build

ttnghia (dismissed):

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,105 @@
...
+  }
+
+  __device__ void set(int32_t i, uint32_t value) {
+    int32_t offset = i * num_bits_per_entry;
+    data_type masked_data = data & ~(static_cast<data_type>(mask) << offset);
+    data = masked_data | (static_cast<data_type>(value & mask) << offset);
+  }
+
+private:
+  uint64_t data;
```
hyperbolic2346, line 53:
```suggestion
  data_type data;
```

src/main/cpp/src/zorder.cu
```diff
@@ -27,6 +27,105 @@
...
+// https://github.com/davidmoten/hilbert-curve, which has the following Note in
+// the code too
+// This algorithm is derived from work done by John Skilling and published
+// in "Programming the Hilbert curve". (c) 2004 American Institute of Physics.
+// With thanks also to Paul Chernoch who published a C# algorithm for Skilling's
+// work on StackOverflow and
+// <a href="https://github.com/paulchernoch/HilbertTransformation">GitHub</a>.
+__device__ uint64_t to_hilbert_index(long_backed_array const & transposed_index,
+        int32_t const num_bits_per_entry, int32_t const num_dimensions) {
+  uint64_t b = 0;
```
hyperbolic2346, line 69: these are sized to match the `long_backed_array`, correct? I would expect these to be
`long_backed_array::data_type` instead of `uint64_t`. It is a bit pedantic, but if the backing store changes, it would be easy to miss changing these.

hyperbolic2346 (changes requested):
I'm ok with this change, but have a couple of questions.

src/main/cpp/src/zorder.cu
revans2: They are but also they are not. I picked the size of long_backed_array because of the size of the output needed in this case. So for me you are pushing for the dependency to go in the wrong direction. The size of long_backed_array is set by the desired output size, not the other way around.  Perhaps I should just change it to be templated. But I don't really want to because what we are trying to replace does not have more options than a long.

revans2:
build

hyperbolic2346 (approved):

PR #676, opened 2022-10-25
Adding float casting benchmark

hyperbolic2346:
Only supports float for now. Double should be added but this is what I have right now.


hyperbolic2346:
build

jlowe (approved):

PR #678, opened 2022-10-26
Dynamic link libcupti for aarch64

wjxiz1992:
To close: #680
This PR adds in a conditional link method for libcupti. 
Due to the lack of static library of libcupti in arm64 cuda Docker image, we must link cupti dynamically.
This PR is verified on an AWS ARM instance inside RockyLinux8 docker container.
Please note, dynamic link will require users to install cuda toolkit in their environment when they need to use faultinj module.

### DIFF
--- a/src/main/cpp/faultinj/CMakeLists.txt
+++ b/src/main/cpp/faultinj/CMakeLists.txt
@@ -36,6 +36,8 @@ include_directories(
   "${SPDLOG_INCLUDE}"
 )
 
-target_link_libraries(
-  cufaultinj CUDA::cupti_static
-)
+if (CMAKE_HOST_SYSTEM_PROCESSOR MATCHES "aarch64")
+    target_link_libraries(cufaultinj CUDA::cupti)
+else ()
+    target_link_libraries(cufaultinj CUDA::cupti_static)
+endif()


jlowe (approved):

wjxiz1992:
build

src/main/cpp/faultinj/CMakeLists.txt
```diff
@@ -36,6 +36,8 @@ include_directories(
   "${SPDLOG_INCLUDE}"
 )
 
-target_link_libraries(
-  cufaultinj CUDA::cupti_static
-)
+if (CMAKE_HOST_SYSTEM_PROCESSOR MATCHES "aarch64")
```
gerashegalov, line 39: Whereas the first use case for dynamic linking may be a particular architecture aarch64 but I would just make it a boolean option that can be passed on the CLI to the tune of -DCUPTI_STATIC=ON

wjxiz1992:
Some update on this PR:
On Amazon Linux image(image number: `ami-0fadfbbdddef5a4fd`) where there's no official package managed installer provided, I have to use `runfile` to install Nvidia Driver and CUDA Toolkit. In this case the CMake tool is unable to find `cupti.h` and its shared library.

I applied the following change to make it pass compile:
```
diff --git a/src/main/cpp/faultinj/CMakeLists.txt b/src/main/cpp/faultinj/CMakeLists.txt
index a40a76e..3fc681b 100644
--- a/src/main/cpp/faultinj/CMakeLists.txt
+++ b/src/main/cpp/faultinj/CMakeLists.txt
@@ -32,10 +32,15 @@ add_library(
 find_path(SPDLOG_INCLUDE "spdlog"
     HINTS "$ENV{RMM_ROOT}/_deps/spdlog-src/include")
 
+find_library(CUPTI_LIB "cupti" REQUIRED NO_DEFAULT_PATH
+    HINTS "/usr/local/cuda/extras/CUPTI/lib64"
+)
+
 include_directories(
   "${SPDLOG_INCLUDE}"
+  "/usr/local/cuda/extras/CUPTI/include"
 )
 
 target_link_libraries(
-  cufaultinj CUDA::cupti_static
+       cufaultinj ${CUPTI_LIB}
 )
```

I haven't got time figuring out a proper way to handle all cases, so share this info for some eyes.
cc: @GaryShen2008

wjxiz1992:
According to our decision to build JNI library on Amazon Linux2 image directly (Where cupti library was not installed in `$CUDA_HOME/lib64`, but another `$CUDA_HOME/extras/CUPTI/lib64`) this PR doesn't help. We are now building the JNI library on ARM64 by modifying the CMakeList to hard-code the searching path for cupti. I'll mark this PR as draft to hold on the change until CUDA team gives more clear clarifications for their CUDA11.5 support OS Docker images.

PR #764, opened 2022-11-22
Decimal 128 add and subtract

razajafri:
This PR handles the Decimal 128 addition for Spark 3.4. 

In previous versions of Spark, we were casting the data before the add operation in all cases. In Spark 3.4, this has changed, in cases where there is no overflow, we still follow the old way, but in cases where there is an overflow that can be avoided, Spark 3.4 does the add and then does the cast.


revans2:
Please sign off on your PR if you want it to go in.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +520,111 @@ inline __device__ int precision10(chunked256 value) {
...
+
+  __device__ __int128_t operator()(cudf::size_type const i) const {
+    chunked256 const a(a_data[i]);
+    chunked256 const b(b_data[i]);
+
+    chunked256 working_a(a_data[i]);
+    chunked256 working_b(b_data[i]);
+
+    /*
+    * The way Spark 3.4 does add is first it rescales the original numbers
```
revans2, line 553: I am confused by this. The Spark/Java code does not appear to do what you are claiming, and your own code is making assumptions that are not true.

https://github.com/apache/spark/blob/25133684ef9991d0b0414d97ac64c70e878b4d4d/sql/catalyst/src/main/scala/org/apache/spark/sql/types/Decimal.scala#L461-L472

Spark has a special case if the value is stored as a long and the scales are the same, but in all other cases it becomes a Java BigDecimal add, which according to the docs will output a scale that is the max of LHS.scale() and RHS.scale().

After that the result is rounded to the final scale of the answer.

The current code assumes that LHS.scale() is the same as RHS.scale(), which is not guaranteed to be true, and will only adjust the scales if the LHS.scale() != the output scale.

We should follow what java does.

```
int intermediate_scale = max(a_scale, b_scale);
if (a_scale != intermediate_scale) {
  a.set_scale_and_round(intermediate_scale); // you can use the code on lines 560 to 570 for this
}

if (b_scale != intermediate_scale) {
  b.set_scale_and_round(intermediate_scale);
}

// Or just use a directly and return it.
chunked256 sum = a;
sum.add(b);
if (sum_scale != intermediate_scale) {
  sum.set_scale_and_round(sum_scale);
}

overflows[i] = !sum.fits_in_128_bits();
add_data[i] = sum.as_128_bits();
```

We just need to be sure that we test this all out with negative scale decimal values.

revans2 (changes requested):

src/main/cpp/src/decimal_utils.cu
razajafri: So, here is what I see from REPL 

```
scala> val df = Seq(("-9191008513307131620269245301.1615457290", "-9447850332473678680446404122.5624623187")).toDF("a", "b")
df: org.apache.spark.sql.DataFrame = [a: string, b: string]

scala> df.selectExpr("cast(a as decimal(38,10)) + cast(b as decimal(38,10))").show(truncate=false) 
+-------------------------------------------------------+
|(CAST(a AS DECIMAL(38,10)) + CAST(b AS DECIMAL(38,10)))|
+-------------------------------------------------------+
|-18638858845780810300715649423.724008048               |
+-------------------------------------------------------+


scala> val left = BigDecimal("-9191008513307131620269245301.1615457290") 
left: scala.math.BigDecimal = -9191008513307131620269245301.1615457290

scala> val right = BigDecimal ("-9447850332473678680446404122.5624623187")
right: scala.math.BigDecimal = -9447850332473678680446404122.5624623187

scala> left + right 
res6: scala.math.BigDecimal = -18638858845780810300715649423.7240080477
```

Why is there a discrepancy in the sum when doing it in a dataframe vs BigDecimal?

revans2: The difference is that Spark will round it to the desired scale when it is done. That is the 

```
if (sum_scale != intermediate_scale) {
  sum.set_scale_and_round(sum_scale);
}
```
in the pseudo code I added.

In Spark it does it by calling `setScale`, with half up rounding.

```
scala>  val left = BigDecimal("-9191008513307131620269245301.1615457290") 
left: scala.math.BigDecimal = -9191008513307131620269245301.1615457290

scala>  val right = BigDecimal ("-9447850332473678680446404122.5624623187")
right: scala.math.BigDecimal = -9447850332473678680446404122.5624623187

scala> val tmp = left + right
tmp: scala.math.BigDecimal = -18638858845780810300715649423.7240080477

scala> tmp.setScale(9, BigDecimal.RoundingMode.HALF_UP)
res3: scala.math.BigDecimal = -18638858845780810300715649423.724008048
```

razajafri:
build

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,72 @@ inline __device__ int precision10(chunked256 value) {
...
 
+__device__ chunked256 set_scale_and_round(chunked256 data, int old_scale, int new_scale) {
+    if (old_scale != new_scale) {
+        if (new_scale < old_scale) {
+            int raise = old_scale - new_scale;
+            int multiplier = pow_ten(raise).as_128_bits();
+            data = multiply(data, chunked256(multiplier));
+        } else {
+            int drop = new_scale - old_scale;
+            auto const diviser = pow_ten(drop).as_128_bits();
```
revans2, line 530: nit: I think this should be divisor, not diviser. (I cannot spell but Firefox is highlighting it saying something is wrong).

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,72 @@ inline __device__ int precision10(chunked256 value) {
...
+            data = divide_and_round(data, chunked256(diviser).as_128_bits());
```
revans2, line 531: Why are we converting diviser from __int128_t to chunked256 and back to __int128_t?

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,72 @@ inline __device__ int precision10(chunked256 value) {
...
+struct dec128_add: public thrust::unary_function<cudf::size_type, __int128_t> {
+  dec128_add(bool *overflows, cudf::mutable_column_view const &sum_view,
+                    cudf::column_view const &a_col, cudf::column_view const &b_col)
+      : overflows(overflows), a_data(a_col.data<__int128_t>()), b_data(b_col.data<__int128_t>()),
+        add_data(sum_view.data<__int128_t>()),
+        a_scale(a_col.type().scale()), b_scale(b_col.type().scale()),
+        sum_scale(sum_view.type().scale()) {}
+
+  __device__ __int128_t operator()(cudf::size_type const i) const {
+    chunked256 const a(a_data[i]);
```
revans2, line 547: nit: why do we need `a` and `working_a`? or `b` and `working_b`? `a` and `b` are never used, except to set `working_a` and `working_b`.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,72 @@ inline __device__ int precision10(chunked256 value) {
...
+    chunked256 const b(b_data[i]);
+
+    chunked256 working_a = a;
+    chunked256 working_b = b;
+
+    int intermediate_scale = min(a_scale, b_scale);
```
revans2, line 553: I think this needs to be `max` not `min`.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,72 @@ inline __device__ int precision10(chunked256 value) {
...
+    if (a_scale != intermediate_scale) {
+    printf("converting a_scale \n");
```
revans2, line 555: Remove the debug prints.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,72 @@ inline __device__ int precision10(chunked256 value) {
...
+        working_a = set_scale_and_round(working_a, a_scale, intermediate_scale);
+    }
+    if (b_scale != intermediate_scale) {
+    printf("converting b_scale \n");
```
revans2, line 559: Remove the debug prints

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,72 @@ inline __device__ int precision10(chunked256 value) {
...
+        working_b = set_scale_and_round(working_b, b_scale, intermediate_scale);
+    }
+
+    chunked256 sum = working_a;
+    sum.add(working_b);
+
+    if (sum_scale != intermediate_scale) {
+        sum = set_scale_and_round(sum, intermediate_scale, sum_scale);
```
revans2, line 567: nit: the indentation appears to be off in a few places. We have been using 2 spaces for indentation.

revans2 (commented):
Looks a lot better.

src/main/cpp/src/decimal_utils.cu
razajafri: scales are negative in cudf which is why I have chosen min. 

I could just reverse the `if` condition in `set_scale_and_round` if you really want

revans2: right. Good catch.

revans2 (approved):
This looks good now.  Just need to make sure that we have tests somewhere for mixed scale addition.

razajafri:
> This looks good now. Just need to make sure that we have tests somewhere for mixed scale addition.

Thanks for the review. I added one test in this PR. Will add more thorough tests on the plugin side

razajafri:
build

razajafri:
build

razajafri:
merge

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -56,7 +56,28 @@ public static Table divide128(ColumnView a, ColumnView b, int quotientScale) {
...
 
+  /**
+   * Add two DECIMAL128 columns and produce a DECIMAL128 result rounded to the specified
+   * scale with overflow detection.
+   *
+   * NOTE: This is very specific to Spark 3.4. This method won't work with previous versions
+   * of Spark. We don't need this for versions prior to Spark 3.4
+   *
+   * @param a            input, must match row count of the other input
+   * @param b            input, must match row count of the other input
```
ttnghia, line 67:
```suggestion
   * @param a input, must match row count of the other input
   * @param b input, must match row count of the other input
```

src/main/cpp/src/DecimalUtilsJni.cpp
```diff
@@ -53,4 +53,21 @@ JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_DecimalUtils_divid
...
+                                                                                     jlong j_view_a,
+                                                                                     jlong j_view_b,
+                                                                                     jint j_target_scale) {
+  JNI_NULL_CHECK(env, j_view_a, "column is null", 0);
+  JNI_NULL_CHECK(env, j_view_b, "column is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
+    auto view_a = reinterpret_cast<cudf::column_view const *>(j_view_a);
+    auto view_b = reinterpret_cast<cudf::column_view const *>(j_view_b);
+    auto scale = static_cast<int>(j_target_scale);
```
ttnghia, line 66:
```suggestion
    auto const view_a = reinterpret_cast<cudf::column_view const *>(j_view_a);
    auto const view_b = reinterpret_cast<cudf::column_view const *>(j_view_b);
    auto const scale = static_cast<int>(j_target_scale);
```

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,67 @@ inline __device__ int precision10(chunked256 value) {
     return -1;
 }
 
+__device__ chunked256 set_scale_and_round(chunked256 data, int old_scale, int new_scale) {
+  if (old_scale != new_scale) {
+    if (new_scale < old_scale) {
+      int raise = old_scale - new_scale;
```
ttnghia, line 525: Use `const` for immutable variables as much as possible. Not just here but everywhere else if possible.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -732,4 +792,25 @@ divide_decimal128(cudf::column_view const &a, cudf::column_view const &b, int32_
...
+  CUDF_EXPECTS(num_rows == b.size(), "inputs have mismatched row counts");
+  auto [result_null_mask, result_null_count] = cudf::detail::bitmask_and(cudf::table_view{{a, b}}, stream);
+  std::vector<std::unique_ptr<cudf::column>> columns;
+  // copy the null mask here, as it will be used again later
+  columns.push_back(cudf::make_fixed_width_column(cudf::data_type{cudf::type_id::BOOL8}, num_rows,
+                                                  rmm::device_buffer(result_null_mask, stream), result_null_count, stream));
+  columns.push_back(cudf::make_fixed_width_column(cudf::data_type{cudf::type_id::DECIMAL128, target_scale}, num_rows, std::move(result_null_mask), result_null_count, stream));
+  auto overflows_view = columns[0]->mutable_view();
+  auto sum_view = columns[1]->mutable_view();
+  thrust::transform(rmm::exec_policy(stream), thrust::make_counting_iterator<cudf::size_type>(0),
```
ttnghia, line 810: If using `thrust::make_` iterator factory then don't need to specify the template type. We only need the template type if using `thrust::counting_iterator` directly.

src/main/cpp/src/decimal_utils.cu
mythrocks: My rule of thumb is to define as `auto const`, and then remove the const when the need to modify presents itself, not unlike default `val` in Scala.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -520,6 +519,109 @@ inline __device__ int precision10(chunked256 value) {
     return -1;
 }
 
+__device__ chunked256 set_scale_and_round(chunked256 data, int old_scale, int new_scale) {
+  if (old_scale != new_scale) {
+    if (new_scale < old_scale) {
+      int const raise = old_scale - new_scale;
+      int const multiplier = pow_ten(raise).as_128_bits();
```
revans2, line 526: I just realized in the general case that raise and/or drop could be much larger than `pow_ten` supports. This should be a very extreme corner case that would require negative scale (positive scale in CUDF) decimal values which spark has off by default. In those cases I think it would happen if pone side is so much smaller than the other side that adding or subtracting them would be a NOOP, so we might want to special case it inside of the DecimalAdd or DecimalSubtract operator.  We probably want to check for this case before ever calling these kernels and have it either do the right thing, or throw an error on the CPU before we ever reach the GPU.

revans2 (changes requested):
looks good, just the one issue with the odd corner cases. Also this needs to move to 23.02

src/main/cpp/src/decimal_utils.cu
razajafri: Sounds to me, that this special case will only affect add/subtract. 
Will division/multiplication be also affected? 

Regardless of above, can/should we try to fix this here or create a follow-on?

razajafri: This should be handled on the plugin side?

razajafri:
@revans2 I found a case where we weren't reporting overflow for mul/div/add/subtract. PTAL

src/main/cpp/src/decimal_utils.cu
revans2: That is fine, but I would feel better if we documented it in the javadocs for the methods and ideally had some checks on the CPU before calling this, so we don't hit the assert.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -766,7 +774,7 @@ struct dec128_divider : public thrust::unary_function<cudf::size_type, __int128_
 
       auto const result = divide_and_round(n, d);
 
-      overflows[i] = !result.fits_in_128_bits();
+      overflows[i] = is_greater_than_decimal_38(result);
```
revans2, line 777: We explicitly had multiply and divide check that it fits inside of a decimal128, not that it fits in a precision of 38, because of some odd corner cases that are possible with spark versions prior to 3.4. I don't remember off of the top of my head exactly what it was, or if it was just because the CheckOverflow would run afterwards and fix it there. I am going to have to think about this more to be sure that it is not going to mess something up for the older versions of Spark.

revans2: Talked to @jlowe and I think this is fine. I was remembering something related to the SUM aggregation.

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -56,7 +56,49 @@ public static Table divide128(ColumnView a, ColumnView b, int quotientScale) {
...
+   *         DECIMAL128 result value. A null input row will result in a corresponding null output
+   *         row.
+   */
+
+  public static Table subtract128(ColumnView a, ColumnView b, int targetScale) {
+    return new Table(subtract128(a.getNativeView(), b.getNativeView(), targetScale));
+  }
+  /**
+   * Add two DECIMAL128 columns and produce a DECIMAL128 result rounded to the specified
+   * scale with overflow detection.
```
revans2, line 80: Can we say what kind of overflow checks it is doing explicitly. Here and in all of the methods that were modified?

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -56,7 +56,49 @@ public static Table divide128(ColumnView a, ColumnView b, int quotientScale) {
...
+   *
+   * NOTE: This is very specific to Spark 3.4. This method won't work with previous versions
```
revans2, line 82: nit: can we reword this to say that this add is not compatible with versions of spark prior to 3.4 instead of it "won't work" with them?

src/main/cpp/src/decimal_utils.cu
razajafri: I don't think I understand your original comment. The max scale supported by Spark is 38, and the min is -38. Considering that the `min_scale = -38` and the raise or drop will be 76. Which case will make the raise or drop be more than 76 for `pow_ten` to hit the assert?

razajafri: I stand corrected. Thanks @revans2 
`Seq("1E100", "5E100").toDF.withColumn("dec", col("value").cast(DecimalType(38, -100))).show()`

razajafri:
@revans2 does this work?

revans2 (dismissed):

revans2:
build

razajafri:
> This looks good to me. I would feel a little better about checking it in if you ran through the arithmetic integration tests using this patch for spark 3.3, as it does make some very small subtile changes to how we handle overflow detection, but I think it is fine.

I can do this as soon as I can build this locally. It's stopped for me ever since I switched to branch-23.02

revans2:
okay I built it just fine. Not sure what you problem is then. I do see warnings about 

```
decimal_utils.cu(107): warning #177-D: function "<unnamed>::chunked256::fits_in_128_bits" was declared but never referenced
```

Could you please remove the `fits_in_128_bits` method from `chunkced256`?

All of the other tests passed for me.

razajafri:
build

revans2 (approved):

PR #796, opened 2022-12-05
Fix run-in-docker for interactive default shell

jlowe:
Fixes run-in-docker to pass the flags for interactive TTY when running in the no-args default mode of an interactive shell.  Without these flags, the shell immediately terminates since there is no stdin.


jlowe:
build

hyperbolic2346 (approved):

revans2 (approved):

PR #797, opened 2022-12-05
Make strip optional for castToInteger and castToDecimal

revans2:
Hive parsing does not strip out leading or trailing white space (for integer and decimal types, float it does because java does it for them).

This makes that stripping optional so we can reuse the existing kernels with hive parsing.


jlowe (approved):

revans2:
CI failed because of an spdlog issue that I think is unrelated. Looking into it...

revans2:
Looks like the spdlog issue is related to a change to RMM. https://github.com/rapidsai/rmm/pull/1173#issuecomment-1337440524 and because CUDF pulls in the latest RMM and we have not fixed https://github.com/NVIDIA/spark-rapids/issues/6445 yet so we let CUDF pull in the latest RMM, which makes our builds not reproducible, we are failing yet again.

hyperbolic2346 (approved):

mythrocks (approved):

revans2:
build

revans2:
@jlowe @hyperbolic2346 @mythrocks could you all take another look. I forgot to update the C++ tests, which should be fixed now.

jlowe (approved):

PR #803, opened 2022-12-07
Add an integer divide kernel for adding 128-bit numbers

razajafri:
This PR repurposes an existing kernel to return a Long output after dividing two decimal columns

Signed-off-by: Raza Jafri <rjafri@nvidia.com>

### DIFF
--- a/src/main/cpp/src/decimal_utils.cu
+++ b/src/main/cpp/src/decimal_utils.cu
@@ -104,13 +104,6 @@ struct chunked256 {
     }
   }
 
-  inline __device__ bool fits_in_64_bits() const {
-    // check for overflow by ensuring no significant bits will be lost when truncating to 64-bits
-    int64_t sign = static_cast<int64_t>(chunks[0]) >> 63;
-    return sign == static_cast<int64_t>(chunks[1]) && sign == static_cast<int64_t>(chunks[2])
-       && sign == static_cast<int64_t>(chunks[3]);
-  }
-
   inline __device__ __int128_t as_128_bits() const {
     return (static_cast<__int128_t>(chunks[1]) << 64) | chunks[0];
   }
@@ -118,6 +111,7 @@ struct chunked256 {
   inline __device__ uint64_t as_64_bits() const {
     return chunks[0];
   }
+
 private:
   uint64_t chunks[4];
 };
@@ -755,7 +749,7 @@ struct dec128_divider : public thrust::unary_function<cudf::size_type, __int128_
       auto const result = is_int_div ? integer_divide(first_div_result.quotient, scale_divisor)
                             : divide_and_round(first_div_result.quotient, scale_divisor);
 
-      overflows[i] = is_int_div ? !result.fits_in_64_bits() : is_greater_than_decimal_38(result);
+      overflows[i] = is_greater_than_decimal_38(result);
       quotient_data[i] = is_int_div ? result.as_64_bits() : result.as_128_bits();
     } else if (n_shift_exp < -38) {
       // We need to do a multiply before we can divide, but the multiply might
@@ -784,7 +778,7 @@ struct dec128_divider : public thrust::unary_function<cudf::size_type, __int128_
         result = round_from_remainder(result, second_div_result.remainder, scaled_div_r, d);
       }
 
-      overflows[i] = is_int_div ? !result.fits_in_64_bits() : is_greater_than_decimal_38(result);
+      overflows[i] = is_greater_than_decimal_38(result);
       quotient_data[i] = is_int_div ? result.as_64_bits() : result.as_128_bits();
     } else {
       // Regular multiply followed by a divide
@@ -794,7 +788,7 @@ struct dec128_divider : public thrust::unary_function<cudf::size_type, __int128_
 
       auto const result = is_int_div ? integer_divide(n, d) : divide_and_round(n, d);
 
-      overflows[i] = is_int_div ? !result.fits_in_64_bits() : is_greater_than_decimal_38(result);
+      overflows[i] = is_greater_than_decimal_38(result);
       quotient_data[i] = is_int_div ? result.as_64_bits() : result.as_128_bits();
     }
   }
--- a/src/main/cpp/src/decimal_utils.cu
+++ b/src/main/cpp/src/decimal_utils.cu
@@ -204,7 +204,7 @@ __device__ chunked256 round_from_remainder(chunked256 const &q, __int128_t const
       (d < 0 ? -d : d); // abs divisor
 
   // To know which way to round, more specifically when the quotient is 0
-  // we keed to know what the sign of the quotient would have been. In this
+  // we need to know what the sign of the quotient would have been. In this
   // case that happens if only one of the inputs was negative (xor)
   bool const is_n_neg = n.sign() < 0;
   bool const is_d_neg = d < 0;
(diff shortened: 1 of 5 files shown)


razajafri:
I still need to fix the overflow detection as it's still detecting for 128-bit where in fact it should flag numbers more than 64-bit as overflows

razajafri:
I still need to fix the overflow detection as it's still detecting for 128-bit where in fact it should flag numbers more than 64-bit as overflows

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -55,9 +55,23 @@ public static Table multiply128(ColumnView a, ColumnView b, int productScale) {
    *         row.
    */
   public static Table divide128(ColumnView a, ColumnView b, int quotientScale) {
-    return new Table(divide128(a.getNativeView(), b.getNativeView(), quotientScale));
+    return new Table(divide128(a.getNativeView(), b.getNativeView(), quotientScale, false));
   }
 
+  /**
+   * Divide two DECIMAL128 columns and produce a INT64 quotient with overflow detection.
+   * This method considers a precision greater than 19 as overflow even if the number still fits in
```
revans2, line 63: There are a few things wrong with this.  A Decimal(19,0) can hold all INT64 values plus some.  When I first suggested that we should do a regular divide and set the result to Decimal(19,0) and then cast it to a long I thought that the divide used in integral divide would do rounding. But it does not. We want to be sure that the output fits in 64 bits, not Decimal(19,0).

razajafri: I will cast the output to INT64 as per docs. It was late last night when I pushed, I wanted people to see the approach before I did a cleanup. Thanks for the feedback

razajafri:
build

razajafri:
build

razajafri:
build

src/main/cpp/src/decimal_utils.cu
```diff
@@ -696,13 +711,15 @@ private:
 };
 
 // Functor to divide two DECIMAL128 columns with rounding and overflow detection.
+template <typename T>
 struct dec128_divider : public thrust::unary_function<cudf::size_type, __int128_t> {
```
revans2, line 715: I just realized that I made a big mistake doing the processing this way and I need to fix this. We should not be doing a thrust::transform.  I am going to put up a PR and see if I can fix this.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -729,10 +746,11 @@ struct dec128_divider : public thrust::unary_function<cudf::size_type, __int128_
       auto const scale_divisor = pow_ten(n_shift_exp).as_128_bits();
 
       // The second divide gets the result into the scale that we care about and does the rounding.
-      auto const result = divide_and_round(first_div_result.quotient, scale_divisor);
+      auto const result = is_int_div ? integer_divide(first_div_result.quotient, scale_divisor)
+                            : divide_and_round(first_div_result.quotient, scale_divisor);
 
       overflows[i] = is_greater_than_decimal_38(result);
-      quotient_data[i] = result.as_128_bits();
+      quotient_data[i] = is_int_div ? result.as_64_bits() : result.as_128_bits();
```
revans2, line 753: I have some problems with the code now that you are using templates and is_int_div.  We are mixing concerns, but we didn't fully define them.  Let me fix the problem with thrust::transform, and then we can talk about what the right way to fix this would be.

src/main/cpp/src/decimal_utils.cu
revans2: Okay I put up a PR for this fix

https://github.com/NVIDIA/spark-rapids-jni/pull/820

please take a look and upmerge as needed.

src/main/cpp/src/decimal_utils.cu
ttnghia: Yeah using template classes is best for performance, as you can know all input types at compile time. Implementation would not be very flexible but you will get everything very clearly working.

The issue I see from `quotient_data[i] = is_int_div ? result.as_64_bits() : result.as_128_bits();` is that, `quotient_data[i]` is going to be assigned values of different types depending on `is_int_div`. This is prohibited in C++. If you really want to do that, `is_int_div` must be a template argument (i.e., compiled time constant).

src/main/cpp/src/decimal_utils.cu
```diff
@@ -778,10 +798,11 @@ private:
   // output column for overflow detected
   bool * const overflows;
 
+  bool const is_int_div;
```
hyperbolic2346, line 801: I don't understand `is_int_div` and why that changes the result from 64 bits to 128 bits. Should this be tied to `T`?

revans2: So there are two types of divide that Spark supports `/` and `DIV`.  The output of `\` is a decimal value where the scale and precision is determined by Spark. The output of `DIV` is a long. `is_int_div` is an attempt to implement `DIV`.  The problem is that to make `DIV` work properly multiple different things need to be set. `T` must be an `int64_t`, scale must be `0`, and `is_int_div` must be `true`. This is all hidden from the end user, so it is mostly okay to me, but if you have suggestions to clean it up that would be great.

The biggest problem that I have is that the kernels have dead paths. The kernel for `DIV` still has `is_int_div` in it which means the paths for both true and false are still compiled and runnable.

razajafri: I have added `static_assert` which should alleviate the problem of instantiating the class with "illegal" types that we don't support.

razajafri:
build

src/main/cpp/src/decimal_utils.cu
```diff
@@ -746,11 +753,15 @@ struct dec128_divider {
...
-                            : divide_and_round(first_div_result.quotient, scale_divisor);
-
+      chunked256 result;
+      if constexpr (is_int_div) {
+        result = integer_divide(first_div_result.quotient, scale_divisor);
+        quotient_data[i] = result.as_64_bits();
+      } else {
+        result = divide_and_round(first_div_result.quotient, scale_divisor);
+        quotient_data[i] = result.as_128_bits();
+      }
```
ttnghia, line 763:
```suggestion
      if constexpr (is_int_div) {
        auto const result = integer_divide(first_div_result.quotient, scale_divisor);
        quotient_data[i] = result.as_64_bits();
      } else {
        auto const result = divide_and_round(first_div_result.quotient, scale_divisor);
        quotient_data[i] = result.as_128_bits();
      }
```

src/main/cpp/src/decimal_utils.cu
```diff
@@ -773,32 +784,36 @@ struct dec128_divider {
...
-      auto const result = is_int_div ? integer_divide(n, d) : divide_and_round(n, d);
-
+      chunked256 result;
+      if constexpr (is_int_div) {
+        result = integer_divide(n, d);
+        quotient_data[i] = result.as_64_bits();
+      } else {
+        result = divide_and_round(n, d);
+        quotient_data[i] = result.as_128_bits();
+      }
```
ttnghia, line 808:
```suggestion
      if constexpr (is_int_div) {
        auto const result = integer_divide(n, d);
        quotient_data[i] = result.as_64_bits();
      } else {
        auto const result = divide_and_round(n, d);
        quotient_data[i] = result.as_128_bits();
      }
```

src/main/cpp/src/decimal_utils.cu
ttnghia: Yeah `static_assert` is good.

src/main/cpp/src/decimal_utils.cu
razajafri: result is being used outside of the if block. This suggestion will cause a compile error

src/main/cpp/src/decimal_utils.cu
razajafri: Same comment as above

PR #812, opened 2022-12-08
Fixing subnormal values for string to float casting kernel

hyperbolic2346:
Found that values such as `-1.6721969836937668E-304` would fail to produce a number due to the mechanism for generating the doubles making it subnormal. In this case, the exponent was -320 and the code was throwing it out, but could have scaled the value down to produce the correct double.

Added code to handle it and a test to verify.


jlowe:
Commit needs a signoff

jlowe (approved):

jlowe:
Looks good other than the signoff check.

hyperbolic2346:
build

PR #820, opened 2022-12-09
Fix decimal utils transform vs for_each

revans2:
This fixes an issue with decimal utils where we did a thrust:transform, but the function passed in didn't return anything, but instead wrote the data directly to where it needed to be. Because C does not define what to do in this case I think we were just really lucky that we didn't get data corruption.  This switches things over to for_each and avoids the return problem all together.


revans2:
build

jlowe (approved):

PR #828, opened 2022-12-13
Fixing issues with floats that had exactly 19 characters before the exponent

### LINKED ISSUE #827: [BUG] string to float cast kernel incorrectly casts "-2.21363921575273728E17"
**Describe the bug**
CPU code is able to convert the string "-2.21363921575273728E17" without issue, but the kernel produces a null value.

hyperbolic2346:
Problems showed up with the value -2.21363921575273728E17 and it was found that it was exactly 19 characters before the exponent. This was causing the loop to start the second time on the E17 part of the string and incorrectly flagging this as an invalid value.

While working through this I thought about some other edges as well including:

- enough leading zeros to push to another loop iteration on a value 
- 19 leading zeros with e0
- more than 19 digits in the float.

These are all fixed now and tests added to verify.

closes #827


hyperbolic2346:
build

jlowe (approved):

PR #835, opened 2022-12-15
Fixing issue with strings that only have exponents

### LINKED ISSUE #834: [BUG] string to float kernel fails parsing string with only exponent
**Describe the bug**
The string to float code fails when parsing the string "E15" resulting in a value of 0 instead of a null value.

hyperbolic2346:
Strings like "E15" were incorrectly being considered valid. This PR fixes that and adds it as a test.

closes #834


hyperbolic2346:
build

ttnghia (approved):

PR #842, opened 2022-12-20
Use thread safe API for lower case native parquet footer

### LINKED ISSUE #832: [BUG] use `r` form of multibyte conversion
**Describe the bug**
In the native parquet footer we try to convert UTF-8 characters to lower case.  The only way to do this I found was to convert the data to wide characters, do the lower case conversion, and convert them back to multi-byte (UTF-8).

Recently when looking at the code and docs again I found the following...

https://en.cppreference.com/w/cpp/string/multibyte/mbstowcs

> In most implementations, this function updates a global static object of type [std::mbstate_t](https://en.cppreference.com/w/cpp/string/multibyte/mbstate_t) as it processes through the string, and cannot be called simultaneously by two threads, [std::mbsrtowcs](https://en.cppreference.com/w/cpp/string/multibyte/mbsrtowcs) should be used in such cases.

We are using the non `r` versions to do the conversions. This is only for a small amount of metadta, but we should switch to the r versions of these functions to be safe.

revans2:
This switches a few APIs over to thread safe versions of them. We have seen no issues in practice right now, but better to be safe about this.

This fixes #832

### DIFF
--- a/src/main/cpp/src/NativeParquetJni.cpp
+++ b/src/main/cpp/src/NativeParquetJni.cpp
@@ -43,8 +43,10 @@ namespace jni {
  * for now.
  */
 std::string unicode_to_lower(std::string const& input) {
+  std::mbstate_t to_wc_state = std::mbstate_t();
+  const char * mbstr = input.data();
   // get the size of the wide character result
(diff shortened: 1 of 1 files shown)


revans2:
build

jlowe (approved):

PR #844, opened 2022-12-21
Implement a simplified version of `from_json`

ttnghia:
This PR tries to implement a simple but efficient version of `from_json` SQL function. in particular, it simply extracts the key-value pairs from the input json strings. These pairs are extracted by copying the substrings of the input without any data conversion or validation.

Demo:
```
Input:
[
"{\"Zipcode\" : 704 , \"ZipCodeType\" : \"STANDARD\" , \"City\" : \"PARC PARQUE\" , \"State\" : \"PR\"}",
"{\"Integer\":12345,\"String\":\"ABCXYZ\",\"Double\":1.1245}",
"{\"category\": \"reference\", \"index\": [4,{},null,{\"a\":[{ }, {}] } ], \"author\": \"Nigel Rees\", \"title\": \"{}[], <=semantic-symbols-string\", \"price\": 8.95}"
]

Output:
List 0: [0, 4]
	"Zipcode" : "704"
	"ZipCodeType" : "STANDARD"
	"City" : "PARC PARQUE"
	"State" : "PR"
List 1: [4, 7]
	"Integer" : "12345"
	"String" : "ABCXYZ"
	"Double" : "1.1245"
List 2: [7, 12]
	"category" : "reference"
	"index" : "[4,{},null,{"a":[{ }, {}] } ]"
	"author" : "Nigel Rees"
	"title" : "{}[], <=semantic-symbols-string"
	"price" : "8.95"
```

Here is a simple performance comparison with fully parsing json from file (which has output is just a table, not yet the desired map type for `from_json`). The comparison shows around 4X-5X difference in performance:
```
## Fully parse json:

| string_size  | Samples |  CPU Time  | Noise |  GPU Time  | Noise | Elem/s | bytes_per_second | peak_memory_usage |
|--------------|---------|------------|-------|------------|-------|--------|------------------|-------------------|
| 2^15 = 32768 |     35x | 438.905 ms | 2.30% | 438.821 ms | 2.30% | 1.090G |       1090446756 |         3.695 GiB |

## Simplified `from_json`:

| string_size  | Samples | CPU Time  | Noise | GPU Time  | Noise | Elem/s | bytes_per_second | peak_memory_usage |
|--------------|---------|-----------|-------|-----------|-------|--------|------------------|-------------------|
| 2^15 = 32768 |    171x | 87.881 ms | 5.11% | 87.859 ms | 5.10% | 5.446G |       5445602432 |         4.082 GiB |
```

Depends on:
 * https://github.com/rapidsai/cudf/pull/12432
 * https://github.com/rapidsai/cudf/pull/12455

Closes:
 * https://github.com/rapidsai/cudf/issues/12051


ttnghia:
Note that `from_json` doesn't check for key duplicate:
```
val jsonString="""{"Double": 2.23, "Integer":12345, "String": "ABCXYZ", "Double": a}"""
val data = Seq((1, jsonString))
import spark.implicits._
val df=data.toDF("id","value")
df.show(false)

import org.apache.spark.sql.functions.{from_json,col}
import org.apache.spark.sql.types.{MapType, StringType}
val df2=df.withColumn("map",from_json(col("value"),MapType(StringType,StringType)))
df2.printSchema()
df2.show(false)
```
Output:
```
df2: org.apache.spark.sql.DataFrame = [id: int, value: string ... 1 more field]
root
 |-- id: integer (nullable = false)
 |-- value: string (nullable = true)
 |-- map: map (nullable = true)
 |    |-- key: string
 |    |-- value: string (valueContainsNull = true)

+---+------------------------------------------------------------------+----+
|id |value                                                             |map |
+---+------------------------------------------------------------------+----+
|1  |{"Double": 2.23, "Integer":12345, "String": "ABCXYZ", "Double": a}|null|
+---+------------------------------------------------------------------+----+
```

ttnghia:
Just hit a performance issue with substring operation. I've filed https://github.com/rapidsai/cudf/issues/12445 for it.

ttnghia:
Assuming that the input doesn't have UTF8 character, the implementation can be simplified and will have a very good performance. With input json strings having around 250 fields:

 * Parsing json into cudf column (similar to the old way from https://github.com/NVIDIA/spark-rapids/pull/6211):
```
Run:  [1/7] nested_json_gpu_parser [Device=0 string_size=2^0]
Fail: Unexpected error: std::bad_alloc: out_of_memory: RMM failure at:/home/nghiat/Devel/cudf/1/cpp/build/_deps/rmm-src/include/rmm/mr/device/pool_memory_resource.hpp:193: Maximum pool size exceeded
```
It fails immediately without any success!


With this new implementation:
```

| string_size  | Samples |  CPU Time  | Noise  |  GPU Time  | Noise  |  Elem/s  | bytes_per_second | peak_memory_usage |
|--------------|---------|------------|--------|------------|--------|----------|------------------|-------------------|
|      2^0 = 1 |   1600x |   2.836 ms | 26.84% |   2.831 ms | 26.76% |   5.157M |          5157188 |       140.688 KiB |
|     2^4 = 16 |   1792x |   3.933 ms |  2.47% |   3.929 ms |  2.46% |  59.456M |         59456135 |         2.197 MiB |
|    2^8 = 256 |   1632x |   8.150 ms |  8.16% |   8.146 ms |  8.15% | 458.799M |        458798726 |        35.142 MiB |
|  2^12 = 4096 |    727x |  20.637 ms |  5.05% |  20.631 ms |  5.05% |   2.898G |       2898447128 |       562.277 MiB |
| 2^16 = 65536 |     30x | 513.180 ms |  0.85% | 513.170 ms |  0.85% |   1.864G |       1864412713 |         8.786 GiB |
```

revans2 (commented):
I started to look at this, but I am a little confused about the algorithm being used. We only care about a very specific level to extract. The top level of each element in the list passed in.  So we would have a `ListBegin` as the very first token and a `ListEnd` as the very last token. That means each row of input data would also be wrapped in a `StructBegin` and a `StructEnd`. Not totally sure about the `StructMember{Begin,End}`, so the actual depth may be off by a little bit. But it sounds like for the keys that we would want to look for the indexes of `FieldNameBegin` and `FieldNameEnd` at a depth of either 2 or 3 depending on how the `StructMember` tokenization happens. For the values we would look for the token after `FieldNameEnd` for the start of the value, and the end of the value would be `StructMemberEnd` at the level above it? or perhaps the next end value after the `FieldNameEnd` at the same level.

it just feels to me like we are doing a lot more work to practically parse the entire JSON when all we need to do is to find the top level items. I could be wrong though, I have not dug deeply into the code, and if it works I am happy to keep it.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,754 @@
...
+    ss << static_cast<U>(h_input[i]);
+    if (separator.size() > 0 && i + 1 < h_input.size()) {
+      ss << separator;
+    }
+  }
+  std::cerr << ss.str() << std::endl;
+}
+
+// Convert the token value into string name, for debugging purpose.
+std::string token_to_string(PdaTokenT const token_type) {
```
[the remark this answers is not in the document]
ttnghia: I think we should keep all the debug code. They were very helpful for me when working on this. In case there are new bugs uncovered in the future, we can use such debug code to track them down.

Such code will be disabled by default and will never be compiled thus there is nothing to worry about.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,754 @@
...
+  }
+}
+
+#endif
+
+// Unify the input json strings by:
+// 1. Append one comma character (',') to the end of each input string, except the last one.
+// 2. Concatenate all input strings into one string.
+// 3. Add a pair of bracket characters ('[' and ']') to the beginning and the end of the output.
+rmm::device_uvector<char> unify_json_strings(cudf::column_view const &input,
```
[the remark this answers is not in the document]
ttnghia: I think so. We have a previous issue asking to parse json directly from a String column, this new issue can also be the prerequisite for it.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,754 @@
...
+    std::cerr << ss.str() << std::endl;
+  }
+#endif
+  return node_token_ids;
+}
+
+// This is copied from cudf's `json_tree.cu`.
+template <typename KeyType, typename IndexType = cudf::size_type>
+std::pair<rmm::device_uvector<KeyType>, rmm::device_uvector<IndexType>>
+stable_sorted_key_order(rmm::device_uvector<KeyType> const &keys, rmm::cuda_stream_view stream) {
```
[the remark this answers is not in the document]
ttnghia: Yeah it seems that, so probably we don't need stable sort here. This code was copied from the json parser code. Let me ask for it to confirm.

ttnghia:
> I started to look at this, but I am a little confused about the algorithm being used. We only care about a very specific level to extract. The top level of each element in the list passed in. So we would have a `ListBegin` as the very first token and a `ListEnd` as the very last token. That means each row of input data would also be wrapped in a `StructBegin` and a `StructEnd`. 
>  .....

Your logic is right. But we are process characters (i.e., tokens) in parallel. If we want to find a match token `_End` to each `Begin_` token directly from the list a tokens, we have to do linear search. My (complex) implementation involves some sort of computation for node levels, node parents etc... so we can maximize parallelism. Even that, I still have to do linear search for nested values to find the value boundary: https://github.com/NVIDIA/spark-rapids-jni/pull/844/files#diff-60a759906a1fdb92b9b286d781d66a2f3a25e7f7232b6ad27f75309cdcc78e2aR470.

src/main/cpp/src/map_utils.cu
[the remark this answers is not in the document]
ttnghia: I'll file an issue for that.

revans2:
I get that you are trying to process the tokens in parallel. That is why I wanted to just look for the level. It looks like we could do a few very simple kernels with almost no walking, just a few set offsets from known locations. But I would have to do a lot more digging into how the tokenization code works to know exactly what the levels need to be and exactly what the tokens are that we would need to look for. I think all we want is something like the following psudo code.

```
key_start = copy offsets[index] if level[index] == 3 and token[index] == FieldNameBegin;
key_end = copy offsets[index] if level[index] == 3 and token[index] == FieldNameEnd;
value_start = copy_offsets[index] if level[index] == 3 and token[index - 1] == FieldNameEnd;
value_end = copy_offsets[index] if level[index] == 2 and token[index] == StructMemberEnd; 
```

But like I said before, if this works I am good with how it is right now. This could be an optimization that we look into as a follow on issue.

src/main/cpp/src/map_utils.cu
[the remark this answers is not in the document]
ttnghia: I got confirmation. This indeed needs stable sort to keep the order of children having the same parent for further processing.

revans2 (approved):

revans2:
Even though I approved this I am not enough of a CUDA expert to really find anything itnerresting.  @hyperbolic2346 and @nvdbaranec could one or both of you also take a look?

ttnghia:
FYI: I'm implementing the plugin work to hook up with this so I can run a benchmark to tell you how much performance gain we can get with this.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,633 @@
...
+// 3. Add a pair of bracket characters ('[' and ']') to the beginning and the end of the output.
+rmm::device_uvector<char> unify_json_strings(cudf::column_view const &input,
+                                             rmm::cuda_stream_view stream) {
+  if (input.is_empty()) {
+    return cudf::detail::make_device_uvector_async<char>(std::vector<char>{'[', ']'}, stream);
+  }
+
+  auto const d_strings = cudf::column_device_view::create(input, stream);
+  auto const transform_it = thrust::make_counting_iterator(0);
+  auto const output_size =
```
hyperbolic2346, line 75: Why not look at the last offset for the total characters in the column? Is it due to nulls needing 2 characters? Is this faster than `null_count() * 2`?

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,633 @@
...
+  print_debug<char, char>(output, "Processed json string", "", stream);
+#endif
+  return output;
+}
+
+// Check and throw exception if there is any parsing error.
+void throw_if_error(rmm::device_uvector<PdaTokenT> const &tokens,
+                    rmm::device_uvector<SymbolOffsetT> const &token_indices,
+                    rmm::cuda_stream_view stream) {
+  auto const error_count =
```
hyperbolic2346, line 124: We don't use error count outside of knowing if there was an error, why not use the find below instead to reduce the number of operations? I know this is only in an error state and may not be worth it, so feel free to push back on this.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,633 @@
...
+      case token_t::ErrorBegin: return true;
+      default: return false;
+    };
+  }
+};
+
+// Compute the level of each token node.
+// The top json node (top json object level) has level 0.
+// Each row in the input column should have levels starting from 1.
+// This is copied from cudf's `json_tree.cu`.
```
hyperbolic2346, line 158: Did we need to copy this? Has it been changed?

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,633 @@
...
+  print_map_debug(node_token_ids, "Node-to-token-index map", stream);
+#endif
+  return node_token_ids;
+}
+
+// This is copied from cudf's `json_tree.cu`.
+template <typename KeyType, typename IndexType = cudf::size_type>
+std::pair<rmm::device_uvector<KeyType>, rmm::device_uvector<IndexType>>
+stable_sorted_key_order(rmm::device_uvector<KeyType> const &keys, rmm::cuda_stream_view stream) {
+  // Buffers used for storing intermediate resuls during sorting.
```
hyperbolic2346, line 226:
```suggestion
  // Buffers used for storing intermediate results during sorting.
```

hyperbolic2346 (changes requested):
Still processing this. Have some questions. The largest is why are we pulling code from cudf when we can directly call it.

src/main/cpp/src/map_utils.cu
ttnghia: Hah, good point. I'll replace this with a trivial computation.

src/main/cpp/src/map_utils.cu
ttnghia: The `thrust::find(+_if)`  APIs are known to have performance regression. We talked about that many times in cudf PRs. As such, it is called only when we will throw for sure.

src/main/cpp/src/map_utils.cu
ttnghia: If I mark "This is copied" then there is no changes. These are helper functions that we need from cudf json code, but we can use them from cudf because it makes no sense to expose such minor helper functions into a public header.

ttnghia:
Some performance tests with a dataset of 670k rows, each row has >100 fields:

## CPU only:
```
Run time: 0:00:16.046214
```

## Old GPU implementation (https://github.com/NVIDIA/spark-rapids/pull/6211):
```
Run time:  <to be updated>
```


## This GPU implementation:
```
Run time:  <to be updated>
```

src/main/cpp/src/map_utils.cu
hyperbolic2346: ok, so these are locked behind an anonymous or detail namespace, hence the copy. Ok.

hyperbolic2346 (approved):
This looks like a good change to me. I worry about the overhead of copying code out of cudf, but I don't see a way around it.

ttnghia:
build

ttnghia:
build

ttnghia:
build

Issue #857, opened 2023-01-05
[FEA] Sign spark-rapids-jni jars with 3S signature service


NvTimLiu:
**Is your feature request related to a problem? Please describe.**
Change to sign spark-rapids-jni jars with 3S signature service, as 3S signing is more security than current gpg-sign method. 

**Describe the solution you'd like**
Change to sign spark-rapids-jni jars with 3S signature service, as 3S signing is more security than current gpg-sign method. 

**Additional context**
[3S: Securely sign Spark Rapids packages using 3S](https://jirasw.nvidia.com/browse/SGN-118)

NvTimLiu:
PR merged, close

PR #858, opened 2023-01-05
Sign jars with nvsec 3S services [skip ci]

NvTimLiu:
For feature: https://github.com/NVIDIA/spark-rapids-jni/issues/857

Sign artifacts and push them to Sonatype staging one by one

Signed-off-by: Tim Liu <timl@nvidia.com>

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -23,22 +23,26 @@
 #   SIGN_FILE: true/false, whether to sign the jar/pom file to de deployed
 #
 # Used environment(s):
+#   OUT:            The path where jar files are, relative to $WORKSPACE
 #   CLASSIFIERS:    The classifier list of the jars to be deployed
 #   SERVER_ID:      The repository id for this deployment.
 #   SERVER_URL:     The url where to deploy artifacts.
-#   GPG_PASSPHRASE: The passphrase used to sign files, only required when <SIGN_FILE> is true.
+#   NVSEC_CFG_FILE: The nvsec credentials used to sign via 3S service, only required when <SIGN_FILE> is true.
 ###
 
 set -ex
 
 SIGN_FILE=$1
+WORKSPACE=${WORKSPACE:-`pwd`}
+OUT=${OUT:-'out'}
+#Set the absolute path for 'out'
 OUT_PATH=$WORKSPACE/$OUT
 
 cd $WORKSPACE/
 REL_VERSION=$(mvn exec:exec -q --non-recursive -Dexec.executable=echo -Dexec.args='${project.version}')
 
 echo "REL_VERSION: $REL_VERSION, OUT_PATH: $OUT_PATH \
-SERVER_URL: $SERVER_URL, SERVER_ID: $SERVER_ID"
+        SERVER_URL: $SERVER_URL, SERVER_ID: $SERVER_ID"
 
 ###### Build types/files from classifiers ######
 FPATH="$OUT_PATH/spark-rapids-jni-$REL_VERSION"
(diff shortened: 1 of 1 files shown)


ci/deploy.sh
```diff
@@ -62,18 +66,37 @@ cp -f "$FIRST_FILE" "$FPATH.jar"
 
 ###### Build the deploy command ######
 MVN="mvn -Dmaven.wagon.http.retryHandler.count=3 -DretryFailedDeploymentCount=3 -B"
+DEPLOY_CMD="$MVN -B deploy:deploy-file -Durl=$SERVER_URL -DrepositoryId=$SERVER_ID \
+    -DgroupId=com.nvidia -DartifactId=spark-rapids-jni -Dversion=$REL_VERSION -s ci/settings.xml"
+SIGN_CMD='eval nvsec sign --job-name "Spark Jar Signing" --description "Sign artifact with 3s"'
```
jlowe, line 71: The `eval` here is odd, and if truly necessary a comment would help clarify why this is needed.

NvTimLiu: `eval`  treat the Env SIGN_CMD string containing quotations interpreted as one cmd, otherwise ￥SIGN_CMD is interpret as `nvsec sign --job-name Spark Jar Signing --description Sign artifact with 3s`

As `eval` is not recommended, I will find other ways to fix the problem

jlowe: IMHO the most straightforward way to solve this is to create a function that does the signing.  For example:
```bash
function sign_jar() {
  nvsec sign --job-name "Spark Jar Signing" --description "Sign artifact with 3s" "$@"
}

[...]

sign_jar --file $FPATH.jar --out-dir $OUT_PATH
```

NvTimLiu: Good suggestion, let me update it.

ci/deploy.sh
```diff
@@ -23,22 +23,26 @@
 #   SIGN_FILE: true/false, whether to sign the jar/pom file to de deployed
```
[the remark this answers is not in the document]
NvTimLiu: Good catch, updated

jlowe (approved):
lgtm, although I'm not familiar with the requirements that drove this PR and whether this change meets them.

revans2:
I am a little concerned about how this implemented. This is using a proprietary signing API and is dropping support for an open alternative. If we are going to use the proprietary version, that is fine, but I don't want us to lock others from being able to sign jars that they build themselves.  Could we try to redesign this so the user passes in a script or command that is used to do the signing instead?

NvTimLiu:
> I am a little concerned about how this implemented. This is using a proprietary signing API and is dropping support for an open alternative. If we are going to use the proprietary version, that is fine, but I don't want us to lock others from being able to sign jars that they build themselves. Could we try to redesign this so the user passes in a script or command that is used to do the signing instead?

Good suggestion.

Updated, support both open `gpg` sign and `nvsec` proprietary sign.

Keep `gpg` sign as default in the script, we will overwrite to use `nvsec` proprietary sign in our release CI

ci/deploy.sh
```diff
@@ -61,19 +70,33 @@ FIRST_FILE=${CLASS_FILES%%,*}
 cp -f "$FIRST_FILE" "$FPATH.jar"
 
 ###### Build the deploy command ######
-MVN="mvn -Dmaven.wagon.http.retryHandler.count=3 -DretryFailedDeploymentCount=3 -B"
 if [ "$SIGN_FILE" == true ]; then
-    DEPLOY_CMD="$MVN -B gpg:sign-and-deploy-file -s ci/settings.xml \
-      -Dsources=$FPATH-sources.jar -Djavadoc=$FPATH-javadoc.jar -Dgpg.passphrase=$GPG_PASSPHRASE"
+    case $SIGN_TOOL in
```
pxLi, line 74: nit: can we also put a default `gpg` for SIGN_TOOL?

NvTimLiu: Good catch! updated

revans2 (approved):

NvTimLiu:
build

NvTimLiu:
build

PR #859, opened 2023-01-05
Work around missing symbols when linking tests

revans2:
Test could not link because `libparquet.a` had references to some symbols that `libarrow.a` didn't provide. I am not sure why. It looks like we do not need those symbols, but it is a little scary so we should look more deeply into what is happening here.

I plan on running some more tests to be sure that this is still working well.


revans2:
build

jlowe (approved):
The unresolved Arrow symbols from Arrow-provided libs is very odd.  However I don't see a reason why we want to whole-link Parquet and Thrift libs, so I'm definitely +1 for this change from that perspective.

PR #867, opened 2023-01-09
[submodule-sync] bot-submodule-sync-branch-23.02 to branch-23.02 [skip ci] [bot]

nvauto:
submodule-sync to create a PR keeping thirdparty/cudf up-to-date.
HEAD commit SHA: e38c66f0550882f3fc065e78be1efcc7f342db9e, cudf commit SHA: https://github.com/rapidsai/cudf/commit/d24bf11838863739da014d828156e8bf638430ff

This PR will be auto-merged if test passed. If failed, it will remain open until test pass or manually fix.

PR #868, opened 2023-01-10
Update to latest cudf changes on branch-23.02

jlowe:
This fixes the submodule sync update failure reported in #867.  rapidsai/cudf#12382 changed `make_offsets_child_column` to relocate which namespace it was in and now returns the size along with the offsets column.  The use in zorder.cu was updated to use the new namespace location and extract the column from the new result type.


jlowe:
build

revans2 (approved):

PR #930, opened 2023-02-07
Provide a basic API for spark specific RMM handling with some error injection for testing

revans2:
This really is just an initial patch to let others start making changes to the spark rapids plugin in relation to reliability.

It adds in some new OOM errors that can be caught for specific situations and also provides a way to inject them for unit tests.

Actually doing something beyond injecting the errors will come later on.  Please note that the API might change slightly as time goes on as well.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -1,0 +1,250 @@
+/*
+ * Copyright (c) 2023, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <exception>
+#include <map>
+#include <set>
+#include <pthread.h>
+
+#include <rmm/mr/device/device_memory_resource.hpp>
+
+#include <cudf_jni_apis.hpp>
+
+namespace {
+
+constexpr char const *RMM_EXCEPTION_CLASS = "ai/rapids/cudf/RmmException";
+
+class thread_state {
+public:
+    bool retry_oom_injected = false;
+    bool split_and_retry_oom_injected = false;
+    long task_id = -1;
+};
+
+template<typename T>
+class rollback {
+public:
+    T on_error;
+
+    rollback(T & on_error): on_error(on_error) {}
+
+    ~rollback() {
+      if (std::uncaught_exceptions() > 0) {
+        on_error();
+      }
+    }
+};
+
+class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
+public:
+  spark_resource_adaptor(JNIEnv *env, rmm::mr::device_memory_resource *mr)
+      : resource{mr} {
+    if (env->GetJavaVM(&jvm) < 0) {
+      throw std::runtime_error("GetJavaVM failed");
+    }
+  }
+
+  rmm::mr::device_memory_resource *get_wrapped_resource() { return resource; }
+
+  bool supports_get_mem_info() const noexcept override { return resource->supports_get_mem_info(); }
+
+  bool supports_streams() const noexcept override { return resource->supports_streams(); }
+
+  void associate_thread_with_task(long thread_id, long task_id) {
+    std::scoped_lock lock(state_mutex);
+    if (threads.find(thread_id) != threads.end()) {
+      throw std::invalid_argument("a thread can only be added if it is in the unknown state");
+    }
+    threads.insert({thread_id, {false, false, task_id}});
+    auto recover = [this, thread_id]() {
+      threads.erase(thread_id);
+    };
+    {
+      rollback rb(recover);
+
+      auto tasks_it = task_to_threads.find(task_id);
+      if (tasks_it == task_to_threads.end()) {
+        task_to_threads.insert({task_id, {thread_id}});
+      } else {
+        tasks_it->second.insert(thread_id);
+      }
+    }
+  }
+
+  void remove_thread_association(long thread_id) {
+    std::scoped_lock lock(state_mutex);
+    auto threads_at = threads.find(thread_id);
+    if (threads_at != threads.end()) {
+      auto task_id = threads_at->second.task_id;
+      if (task_id > 0) {
+        auto task_at = task_to_threads.find(task_id);
+        if (task_at != task_to_threads.end()) {
+          task_at->second.erase(thread_id);
+        }
+      }
+      threads.erase(threads_at);
+    }
+  }
+
+  void force_retry_oom(long thread_id) {
+    std::scoped_lock lock(state_mutex);
+    auto threads_at = threads.find(thread_id);
+    if (threads_at != threads.end()) {
+      threads_at->second.retry_oom_injected = true;
+    } else {
+      throw std::invalid_argument("the thread is not associated with any task.");
+    }
+  }
+
+  void force_split_and_retry_oom(long thread_id) {
+    std::scoped_lock lock(state_mutex);
+    auto threads_at = threads.find(thread_id);
+    if (threads_at != threads.end()) {
+      threads_at->second.split_and_retry_oom_injected = true;
+    } else {
+      throw std::invalid_argument("the thread is not associated with any task.");
+    }
+  }
+
+
+private:
+  rmm::mr::device_memory_resource *const resource;
+  // The state mutex must be held when modifying the state of threads or tasks
+  // it must never be held when calling into the child resource or after returning
+  // from an operation.
+  std::mutex state_mutex;
+  std::map<long, thread_state> threads;
+  std::map<long, std::set<long>> task_to_threads;
+  JavaVM *jvm;
+
+  void *do_allocate(std::size_t num_bytes, rmm::cuda_stream_view stream) override {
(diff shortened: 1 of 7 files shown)


revans2:
build

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+class thread_state {
+public:
+    bool retry_oom_injected = false;
+    bool split_and_retry_oom_injected = false;
+    long task_id = -1;
+};
+
+template<typename T>
+class rollback {
```
abellina, line 37: I would prefer `std::function<void()>` as the type instead of the template (but I do not know if other C++ experts agree). It seems we always want this to be a void function that takes no arguments.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -0,0 +1,250 @@
...
+public:
+    T on_error;
```
abellina, line 39: `on_error` member could be in `private`

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -0,0 +1,250 @@
...
+        task_to_threads.insert({task_id, {thread_id}});
+      } else {
+        tasks_it->second.insert(thread_id);
+      }
+    }
+  }
+
+  void remove_thread_association(long thread_id) {
+    std::scoped_lock lock(state_mutex);
+    auto threads_at = threads.find(thread_id);
```
abellina, line 88: if we are removing a `thread_id` from `thread`, and the thread wasn't there before, is that bad?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -0,0 +1,250 @@
...
+  std::mutex state_mutex;
+  std::map<long, thread_state> threads;
+  std::map<long, std::set<long>> task_to_threads;
+  JavaVM *jvm;
+
+  void *do_allocate(std::size_t num_bytes, rmm::cuda_stream_view stream) override {
+    auto tid = static_cast<long>(pthread_self());
+    {
+      // pre allocate checks
+      auto thread = threads.find(tid);
```
abellina, line 136: this is calling into `threads` without the `state_mutex` held, I believe we should lock `state_mutex` here.

abellina (commented):
First pass

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2: It is not what we expect, but is it bad enough to throw an exception over?  Probably a warning, but logging from c++ is hard, so I just ignore it.

abellina: Probably not worth that. I am guessing the calling code could call removeThreadAssociation many times and that's considered safe/ok.

revans2:
build

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -33,18 +33,18 @@ class thread_state {
     long task_id = -1;
 };
 
-template<typename T>
 class rollback {
+    using T = const std::function<void()>;
```
abellina, line 37: sorry could this be not T also? e.g. `callback_type` or just use `std::function` directly (since T is normally reserved for a template argument, and this isn't a template argument now.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -0,0 +1,250 @@
...
+  bool supports_streams() const noexcept override { return resource->supports_streams(); }
+
+  void associate_thread_with_task(long thread_id, long task_id) {
+    std::scoped_lock lock(state_mutex);
+    if (threads.find(thread_id) != threads.end()) {
+      throw std::invalid_argument("a thread can only be added if it is in the unknown state");
+    }
+    threads.insert({thread_id, {false, false, task_id}});
+    {
+      rollback rb([this, thread_id]() {
```
abellina, line 72: so when do we expect the rollback to trigger? e.g. why do we need to have this RAII object?

Perhaps a class header on `rollback` may help me.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -0,0 +1,250 @@
...
+  try {
+    cudf::jni::auto_set_device(env);
+    auto mr = reinterpret_cast<spark_resource_adaptor *>(ptr);
+    mr->associate_thread_with_task(thread_id, task_id);
+  }
+  CATCH_STD(env, )
+}
+
+JNIEXPORT void JNICALL Java_com_nvidia_spark_rapids_jni_SparkResourceAdaptor_removeThreadAssociation(
+        JNIEnv *env, jclass, jlong ptr, jlong thread_id) {
```
abellina, line 219: For `ptr` cast here and in the next two functions, should we use `JNI_NULL_CHECK`?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2: It is not about when do I expect this to happen. It is about the fact that an exception can happen. Any of the `insert` calls that happen after this could throw an exception. If that happens `threads` would point to something that is not inside of `task_to_threads`. The spec for `erase` says 

> Unless the container's [comparison object](https://cplusplus.com/map::key_comp) throws, this function never throws exceptions (no-throw guarantee).


So this code should be able to handle anything that is thrown at it. Do I think that this is going to happen in practice? No insert should only throw if memory allocation failed, which on modern Linux running a 64-bit process should really never happen thanks to virtual memory.

revans2:
build

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  bool supports_get_mem_info() const noexcept override { return resource->supports_get_mem_info(); }
+
+  bool supports_streams() const noexcept override { return resource->supports_streams(); }
+
+  void associate_thread_with_task(long thread_id, long task_id) {
+    std::scoped_lock lock(state_mutex);
+    if (threads.find(thread_id) != threads.end()) {
+      throw std::invalid_argument("a thread can only be added if it is in the unknown state");
+    }
```
jlowe, line 69: Instead of separately doing a lookup and then trying to insert, we can just use the result of `threads.insert` to tell whether or not the entry already existed and fail if it did.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -0,0 +1,252 @@
...
+      rollback rb([this, thread_id]() {
+        threads.erase(thread_id);
+      });
+
+      auto tasks_it = task_to_threads.find(task_id);
+      if (tasks_it == task_to_threads.end()) {
+        task_to_threads.insert({task_id, {thread_id}});
+      } else {
+        tasks_it->second.insert(thread_id);
+      }
```
jlowe, line 81: Same thing here.  We can blindly attempt to do the insert as if it were new, then use the result of `task_to_threads.insert` to tell if the insert occurred, and if not, update the already existing entry if the element already existed with the iterator pointing to the element to update.

jlowe (commented):
Partially through the code, here's what I've got so far.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
abellina: This is for my own edification, I am just curious if the below would be equivalent. If so that's fine, I am just trying to understand if we prefer RAII for this, instead of try-catch-rethrow.

```
threads.insert({thread_id, {false, false, task_id}});
try {
  auto tasks_it = task_to_threads.find(task_id);
  if (tasks_it == task_to_threads.end()) {
    task_to_threads.insert({task_id, {thread_id}});
  } else {
    tasks_it->second.insert(thread_id);
  }
} catch (const std::exception& x) {
  threads.erase(thread_id);
  throw;
}
```

src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
```diff
@@ -0,0 +1,215 @@
...
+      }
+    }
+  }
+
+  /**
+   * Clears the active RMM event handler and anything else that is needed for spark to behave
+   * properly.
+   */
+  public static void clearEventHandler() throws RmmException {
+    synchronized (Rmm.class) {
```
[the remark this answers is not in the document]
revans2: This is synchronized to Rmm not to RmmSpark so that it is consistent with what the RMM code is doing.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2: I did it. I hope it is faster because the code is not more readable, at least in my opinion.

revans2:
build

src/main/cpp/src/SparkResourceAdaptorJni.cpp
jbrennan333: +1 on this.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
jbrennan333: I kind of prefer the try/catch block here.  The rollback is emulating finally, but it is is explicitly checking for uncaught exceptions, so isn't it equivalent to a catch?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
jbrennan333: +1 again.

jlowe (approved):

jbrennan333 (commented):
This looks good to me.  My comments were just me-too's on some of the other review comments.

abellina (approved):

revans2:
I'll merge this now and revisit the rollback in my next PR.

PR #936, opened 2023-02-09
Add new API to RmmSpark and do some cleanup from last review

revans2:
This adds a new API to RmmSpark so that we can recover from a retry attempt without acquiring any spillable memory handles.

It also addresses the concerns about `rollback` by getting rid of it.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
@@ -93,6 +93,9 @@ public class RmmSpark {
     }
   }
 
+  /**
+   * Get the id of the current thread as used by RmmSpark.
+   */
   public static long getCurrentThreadId() {
     return SparkResourceAdaptor.getCurrentThreadId();
   }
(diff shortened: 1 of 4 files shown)


revans2:
build

src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
```diff
@@ -186,6 +211,29 @@ public static void threadDoneWithShuffle() {
...
   }
 
+  /**
+   * This should be called as a part of handling any RetryOOM or SplitAndRetryOOM exception.
+   * The order should be something like.
+   * <ol>
+   *   <li>Catch Exception</li>
+   *   <li>Mark any GPU input as spillable, (should have already had contig split called on it)</li>
+   *   <li>call blockUntilReady</li>
+   *   <li>split the input data if SplitAndRetryOOM</li>
```
abellina, line 221: It is in the header above, this line should include `RetryOOM`

abellina (commented):
minor comment

jlowe (approved):

src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
revans2: This is an ordered list. We only want to split the data if we are asked to.

abellina (approved):

PR #950, opened 2023-02-15
Added way to throw multiple RetryOOMs and an injection point for CudfException

abellina:
This adds a `numOOMs` argument to `forceRetryOOM` and `forceSplitAndRetryOOM`. It also adds a `forceCudfException` to throw something other than the OOM exceptions, such that we can test that we can handle other types of exceptions gracefully.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -28,8 +28,9 @@ constexpr char const *RMM_EXCEPTION_CLASS = "ai/rapids/cudf/RmmException";
 
 class thread_state {
 public:
-    bool retry_oom_injected = false;
-    bool split_and_retry_oom_injected = false;
+    int retry_oom_injected = 0;
+    int split_and_retry_oom_injected = 0;
+    int cudf_exception_injected = 0;
     long task_id = -1;
 };
 
@@ -50,7 +51,7 @@ public:
 
   void associate_thread_with_task(long thread_id, long task_id) {
     std::scoped_lock lock(state_mutex);
-    auto was_threads_inserted = threads.insert({thread_id, {false, false, task_id}});
+    auto was_threads_inserted = threads.insert({thread_id, {0, 0, 0, task_id}});
     if (was_threads_inserted.second == false) {
       throw std::invalid_argument("a thread can only be added if it is in the unknown state");
     }
@@ -84,21 +85,31 @@ public:
     }
   }
 
-  void force_retry_oom(long thread_id) {
+  void force_retry_oom(long thread_id, int num_ooms) {
     std::scoped_lock lock(state_mutex);
     auto threads_at = threads.find(thread_id);
     if (threads_at != threads.end()) {
-      threads_at->second.retry_oom_injected = true;
+      threads_at->second.retry_oom_injected = num_ooms;
     } else {
       throw std::invalid_argument("the thread is not associated with any task.");
     }
   }
 
-  void force_split_and_retry_oom(long thread_id) {
+  void force_split_and_retry_oom(long thread_id, int num_ooms) {
     std::scoped_lock lock(state_mutex);
     auto threads_at = threads.find(thread_id);
     if (threads_at != threads.end()) {
-      threads_at->second.split_and_retry_oom_injected = true;
+      threads_at->second.split_and_retry_oom_injected = num_ooms;
+    } else {
+      throw std::invalid_argument("the thread is not associated with any task.");
+    }
+  }
+
+  void force_cudf_exception(long thread_id, int num_times) {
+    std::scoped_lock lock(state_mutex);
+    auto threads_at = threads.find(thread_id);
+    if (threads_at != threads.end()) {
+      threads_at->second.cudf_exception_injected = num_times;
     } else {
       throw std::invalid_argument("the thread is not associated with any task.");
     }
@@ -127,8 +138,8 @@ private:
       // pre allocate checks
       auto thread = threads.find(tid);
       if (thread != threads.end()) {
-        if (thread->second.retry_oom_injected) {
-          thread->second.retry_oom_injected = false;
+        if (thread->second.retry_oom_injected > 0) {
+          thread->second.retry_oom_injected--;
           JNIEnv *env = cudf::jni::get_jni_env(jvm);
           // TODO cache what is needed for this...
           jclass ex_class = env->FindClass("com/nvidia/spark/rapids/jni/RetryOOM");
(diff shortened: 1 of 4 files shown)


src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -138,8 +149,8 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
           throw cudf::jni::jni_exception("injected RetryOOM");
         }
 
-        if (thread->second.split_and_retry_oom_injected) {
-          thread->second.split_and_retry_oom_injected = false;
+        if (thread->second.split_and_retry_oom_injected > 0) {
+          thread->second.split_and_retry_oom_injected--;
           JNIEnv *env = cudf::jni::get_jni_env(jvm);
           // TODO cache what is needed for this...
```
abellina, line 155: Let me push a commit to handle this TODO

abellina: https://github.com/NVIDIA/spark-rapids-jni/pull/950/commits/9b4b1219bd017f549ea97a2abe940e320881cd7c

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -40,6 +41,9 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     if (env->GetJavaVM(&jvm) < 0) {
       throw std::runtime_error("GetJavaVM failed");
     }
+    retry_oom_class = env->FindClass("com/nvidia/spark/rapids/jni/RetryOOM");
+    split_and_retry_oom_class = env->FindClass("com/nvidia/spark/rapids/jni/SplitAndRetryOOM");
+    cudf_exception_class = env->FindClass("ai/rapids/cudf/CudfException");
```
jlowe, line 46: This risks the JVM garbage collecting these classes unless we create an explicit reference (either directly or indirectly).

abellina:
I got a segfault after trying to handle the class caching here https://github.com/NVIDIA/spark-rapids-jni/pull/950/commits/9b4b1219bd017f549ea97a2abe940e320881cd7c in my spark-rapids tests, not the JNI ones oddly enough. Adding some debug to figure out what happened, but yeah don't merge this.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -118,35 +132,40 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
   std::map<long, std::set<long>> task_to_threads;
+  jclass retry_oom_class;
+  jclass split_and_retry_oom_class;
+  jclass cudf_exception_class;
   JavaVM *jvm;
 
+  void throw_java_exception(jclass ex_class, const char* msg) {
+    JNIEnv *env = cudf::jni::get_jni_env(jvm);
+    if (ex_class != nullptr) {
+      env->ThrowNew(ex_class, msg);
```
jlowe, line 143: Was the intent of this method to return back to the caller if there was an exception class?  The way it's used seems to imply it's not expecting this method to return but always throw a C++-level exception.

abellina: No that's not the intent. We want this to throw to java, we also want the C++ code to see an exception, I just got confused with `ThrowNew` thinking it did both. Also realizing that's not what we do in RmmJni.cpp for example, comparing that with what I did here to see where I went wrong.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
abellina: Yeah I think this is related to what I am seeing.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
abellina: ok should befixed

src/main/cpp/src/SparkResourceAdaptorJni.cpp
abellina: also should be fixed

abellina:
build

abellina:
build

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -120,33 +134,36 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
   JavaVM *jvm;
 
+  void throw_java_exception(const char * ex_class_name, const char* msg) {
+    JNIEnv *env = cudf::jni::get_jni_env(jvm);
+    //TODO: we may want to cache these in the future
+    jclass ex_class = env->FindClass(ex_class_name);
+    if (ex_class != nullptr) {
+      env->ThrowNew(ex_class, msg);
+    }
+    throw cudf::jni::jni_exception(msg);
```
jlowe, line 144: Nit: Now that this isn't trying to cache classes anymore, we can reuse the method from cudf::jni.
```suggestion
    cudf::jni::throw_java_exception(env, ex_class_name, msg);
```

jlowe (approved):
Minor nit, lgtm.

revans2 (dismissed):
I agree with Jason's comment about reusing some existing code.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
abellina: Done https://github.com/NVIDIA/spark-rapids-jni/pull/950/commits/ecc302daf5d2b4c3d33fe1277ce1c139e8d80a51

abellina:
build

jlowe (dismissed):

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkTest.java
```diff
@@ -79,6 +80,71 @@ public void testInsertOOMs() {
...
+      }
+
+      // Allocate something small and verify that it works...
+      Rmm.alloc(100).close();
+    } finally {
+      RmmSpark.removeThreadAssociation(threadId);
+    }
+  }
+
+  public void testCudfException() {
```
revans2, line 122: Sorry I just noticed. IS this test intended to be run as a test? Shouldn't it be tagged with `@Test`???

abellina: yikes, indeed it was not running. It is now though. Nice catch

abellina:
build

abellina:
@jlowe @revans2 this should be ready for another look.

revans2 (approved):

jlowe (approved):

PR #959, opened 2023-02-17
Add in implementation for OOM retry

revans2:
This is an initial implementation of the OOM retry framework. There is still a need for a lot more testing, but this is good enough to get started.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -13,10 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <chrono>
 #include <exception>
 #include <map>
 #include <set>
+#include <sstream>
 #include <pthread.h>
+
+#include <spdlog/common.h>
+#include <spdlog/sinks/ostream_sink.h>
+#include <spdlog/sinks/null_sink.h>
+#include <spdlog/spdlog.h>
 
 #include <rmm/mr/device/device_memory_resource.hpp>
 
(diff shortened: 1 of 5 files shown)


revans2:
build

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -13,37 +13,210 @@
...
+  long get_thread_id() const {
+    return thread_id;
+  }
+
+  long get_task_id() const {
+    return task_id;
+  }
+
+  bool operator<(const thread_priority &other) const {
+    long task_priority = std::numeric_limits<long>::max() - task_id + 1;
```
jbrennan333, line 136: I think you need parens here:  max() - (task_id + 1)
Same in following lines.

abellina: nit, may be nice to put this formula in a function so that we call that to calculate the priority.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -52,121 +225,832 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+      if (thread != threads.end()) {
+        switch (thread->second.state) {
+        case TASK_BLOCKED:
+        // fall through
+        case TASK_BUFN:
+        // fall through
+        case SHUFFLE_BLOCKED:
+          log_status("WAITING", thread_id, thread->second.task_id, thread->second.state);
+          task_wake_condition.wait(lock);
+          task_has_woken_condition.notify_all();
```
jlowe, line 598: This doesn't seem to account for spurious wakeups.  Seems we need to check the state before notifying others that this thread is waking up.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -52,121 +225,832 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+          break;
+        case TASK_BUFN_THROW:
+          transition(thread->second, thread_state::TASK_BUFN_WAIT);
+          throw_retry_oom("rollback and retry operation", thread->second, lock);
+          break;
+        case TASK_BUFN_WAIT:
+          transition(thread->second, thread_state::TASK_BUFN);
+          log_status("WAITING", thread_id, thread->second.task_id, thread->second.state);
+          task_wake_condition.wait(lock);
+          task_has_woken_condition.notify_all();
```
jlowe, line 612: Same spurious wakeup comment here.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -52,121 +225,832 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+   * and if there are no blocked threads, then we wake up all BUFN threads.
+   * Hopefully the frees have already woken up all the blocked threads anyways.
+   */
+  void wake_up_threads_after_task_finishes(std::unique_lock<std::mutex> & lock) {
+    bool are_any_tasks_just_blocked = false;
+    for (auto thread = threads.begin(); thread != threads.end(); thread++) {
+      switch(thread->second.state) {
+        case TASK_BLOCKED:
+          transition(thread->second, thread_state::TASK_RUNNING);
+          task_wake_condition.notify_all();
```
jlowe, line 649: It's a bit odd that we're notifying multiple times on the same condition variable, once per task in a blocked state.  Given the way the condition variable is being used and that I think we're always doing a state transition on the task before waking it up, seems like we could avoid a lot of unnecessary wakeups by having a condition variable per task rather than a global one.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -52,121 +225,832 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+   * Check to see if any threads need to move to BUFN. This should be
+   * called when a task or shuffle thread becomes blocked so that we can 
+   * check to see if one of them needs to becom BUFN or do a split and rollback.
+   */
+  void check_and_update_for_bufn(std::unique_lock<std::mutex> & lock) {
+    // We want to know if all active tasks have at least one thread that
+    // is effectivley blocked or not.  We could change the definitions here,
+    // but for now this sounds like a good starting point.
+    std::set<long> tasks_with_threads;
+    std::set<long> tasks_with_threads_effectively_blocked;
```
abellina, line 851: should this just be a counter? Judging by how it's used for the deadlock busting logic.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -52,121 +225,832 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+
+    bool need_to_break_deadlock = tasks_with_threads.size() == tasks_with_threads_effectively_blocked.size();
+    if (need_to_break_deadlock) {
+      // Find the task thread with the lowest priority that is not already BUFN
+      thread_priority to_bufn(-1, -1);
+      bool is_to_bufn_set = false;
+      for (auto thread = threads.begin(); thread != threads.end(); thread++) {
+        switch (thread->second.state) {
+          case TASK_BLOCKED:
+            {
```
abellina, line 902: style nit, shouldn't really need this top level curly bracket.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -52,121 +225,832 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+        thread->second.cudf_exception_injected--;
+        throw_java_exception(CUDF_EXCEPTION_CLASS, "injected CudfException");
+      }
+
+      if (thread->second.split_and_retry_oom_injected > 0) {
+        thread->second.split_and_retry_oom_injected--;
+        throw_java_exception(SPLIT_AND_RETRY_OOM_CLASS, "injected SplitAndRetryOOM");
+      }
+
+      block_thread_until_ready(lock);
```
abellina, line 744: nit, we could pass `thread_id` here... save us a lookup and check against `threads.end`  `block_thread_until_ready`

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2: This is tasks, not threads. So if we have multiple threads on a single task a counter will not work.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2: I need the top level curly to prevent `thread_priority current` from escaping that scope and causing issues in other places. I get compile errors without it.

abellina: Looks like switch/case doesn't affect the scope, so that makes sense!

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -52,121 +225,832 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
     }
-    return resource->allocate(num_bytes, stream);
+    // we should never reach this point, but just in case
+    throw std::bad_alloc();
   }
 
   void do_deallocate(void *p, std::size_t size, rmm::cuda_stream_view stream) override {
     resource->deallocate(p, size, stream);
+    // deallocate success
+    if (size > 0) {
```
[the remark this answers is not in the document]
revans2: I am going to decline this right now. do_deallocate is not big enough for me to fell like I need to break it up.

revans2:
I think I have addressed all of the comments. Please take another look.  Sorry. To fix the formatting question I ran the auto-format tool from CUDF and it kind of changed a lot.

jlowe (approved):

abellina (dismissed):

revans2:
build

revans2:
@abellina @jlowe just pushed a fix for two possible live lock situations. One that is for a OOM when running from a non task thread and another for a CUDA fatal error that might cause all allocations to fail which would then try to retry the alloc and always loop. Please take another look

abellina (approved):

jlowe (approved):

pxLi:
build

PR #977, opened 2023-02-27
Add in fuzz testing/monte carlo simulation for OOM Retry

revans2:
This is far from perfect. There is no documentation for the code yet. There are no nightly tests to verify that we didn't break anything.  It just is a program that you can run. I wrote it so that we can do different experiments without having to get Spark all the way updated to deal with retry. Also it should be a lot faster than trying to run thousands of spark jobs.

I am happy to add more docs/etc to this. But I thought I would put it up here anyways.


jlowe (approved):
Seems like we should have a followup issue to run the fuzz tests nightly/weekly for regression testing

revans2:
build

PR #982, opened 2023-02-28
Fix an issue where RmmSpark could deadlock if UCX shuffle is used

revans2:
This was found with the fuzz testing in https://github.com/NVIDIA/spark-rapids-jni/pull/977

essentially it shows up if the shuffle thread is already blocked and then the task thread says that it might block on shuffle, and that task thread was the last task thread before block until further notice would be needed to break a deadlock.  This adds the block until further notice checks after a task says it may block on shuffle.


revans2:
build

jlowe (approved):

abellina (approved):

PR #995, opened 2023-03-08
Add in a GpuOOM exception so running out of GPU memory is not a fatal to Spark

revans2:
In Spark if an OutOfMemory error is thrown, then the entire executor is torn down. This is not technically required when it is GPU memory that we ran out of. The upside is that if the task is setup correctly and memory can be recovered from it, other tasks in the same process will not be killed, and shuffle data will not be lost.  But if memory ends up being leaked by a task, then it is more likely to cause other tasks and even possibly later jobs to fail.


revans2:
build

abellina (approved):

PR #1016, opened 2023-03-22
Add -DBUILD_TESTS=ON to submodule sync [skip ci]

### LINKED ISSUE #1010: [BUG] submodule sync job can succeed when nightly build will fail
Currently the submodule sync job is regularly passing while the nightly build is regularly failing. This should not be the case, as one of the main points of the submodule sync job is to not move to a new cudf version if that will break the build.  That means somehow the submodule sync job is building in a different way than the nightly build, and that should be fixed.

pxLi:
fix #1010 

Lets add `-DBUILD_TESTS=ON` to submodule sync first.
we can revisit maven args of pre-merge, nightly, submodule-sync,
maybe try add a common args for all of them later


pxLi:
build

jlowe (approved):

PR #1034, opened 2023-03-29
Log injected exceptions to spdlog

abellina:
Adds a log message when throwing an injected exception. Note it doesn't log for all exceptions, as the idea is to detect in the tests whether we were able to fake an OOM happening. That said, if folks see value in logging this for all exceptions, we can, and I can unify the code a bit.

I have not run this at all, but I am running a build in the background.


abellina:
The output looks like this:
```
13:20:30.273583,DEALLOC,140286787307264,140286787307264,21,TASK_RUNNING,,
13:20:30.304895,INJECTED_RETRY_OOM,140286787307264,140286787307264,21,TASK_RUNNING,,
13:20:30.306643,TRANSITION,140286787307264,140286787307264,21,TASK_RUNNING,TASK_ALLOC,
```

abellina:
build

revans2 (approved):

PR #1064, opened 2023-04-11
Update GPU_ARCHS default to RAPIDS

### LINKED ISSUE #1062: Need to use GPU_ARCHS=RAPIDS rather than ALL
https://github.com/rapidsai/rapids-cmake/pull/397 changed the build system so that requesting "ALL" will break the build.  It needs to be "RAPIDS" instead.  See https://github.com/rapidsai/cudf/pull/13117 for the corresponding change in cudf.

jlowe:
Fixes #1062.  Updates the pom and contributors guide to use RAPIDS instead of ALL for the default GPU architectures to build.


jlowe:
build

pxLi (approved):

PR #1074, opened 2023-04-13
Change exception text ensuring we mention GPU OOM

abellina:
This changes the exception text for RetryOOM, SplitAndRetryOOM and Java's OutOfMemory exception to make sure we say that this is a GPU out of memory. They also remove the task language in Retry and SplitAndRetry because the plugin may not catch the exception at all, so a user may be the end consumer.


revans2 (approved):

revans2:
build

jlowe (approved):

sameerz (approved):
Good for 23.04, as there is no functionality change.

PR #1080, opened 2023-04-15
[submodule-sync] bot-submodule-sync-branch-23.06 to branch-23.06 [skip ci] [bot]

nvauto:
submodule-sync to create a PR keeping thirdparty/cudf up-to-date.
HEAD commit SHA: b55db80ae707065e2ca0fb464619e87718f61705, cudf commit SHA: https://github.com/rapidsai/cudf/commit/daf3ac0cfd538f6e5d63e3064c71b96076b1e530

This PR will be auto-merged if test passed. If failed, it will remain open until test pass or manually fix.

PR #1081, opened 2023-04-17
Update to latest cudf 23.06 and handle new result from make_null_mask

jlowe:
Resolves the submodule sync failure reported in #1080.  make_null_mask was recently updated to return a mask, null_count pair instead of just the mask, and this method was being used in a C++ test.


jlowe:
build

PR #1087, opened 2023-04-19
Disable non-empty nulls assertions 

razajafri:
This PR disables the assertions added in https://github.com/rapidsai/cudf/pull/13071 when performing unit tests

contributes to https://github.com/NVIDIA/spark-rapids/issues/6680

### DIFF
--- a/ci/nightly-build.sh
+++ b/ci/nightly-build.sh
@@ -30,5 +30,5 @@ ${MVN} clean package ${MVN_MIRROR}  \
   -Psource-javadoc \
   -DCPP_PARALLEL_LEVEL=${PARALLEL_LEVEL} \
   -Dlibcudf.build.configure=true \
-  -DUSE_GDS=${USE_GDS} -Dtest=*,!CuFileTest,!CudaFatalTest \
+  -DUSE_GDS=${USE_GDS} -Dtest=*,!CuFileTest,!CudaFatalTest,!ColumnViewNonEmptyNullsTest \
   -DBUILD_TESTS=ON -Dcuda.version=$CUDA_VER
--- a/pom.xml
+++ b/pom.xml
@@ -546,7 +546,7 @@
             </goals>
             <configuration>
               <argLine>-da:ai.rapids.cudf.AssertEmptyNulls</argLine>
-              <test>**/ColumnViewNonEmptyNullsTest.java</test>
+              <test>ColumnViewNonEmptyNullsTest</test>
             </configuration>
           </execution>
           <execution>
--- a/pom.xml
+++ b/pom.xml
@@ -535,7 +535,18 @@
             <configuration>
               <excludes>
                 <exclude>**/CudaFatalTest.java</exclude>
+                <exclude>**/ColumnViewNonEmptyNullsTest.java</exclude>
               </excludes>
+            </configuration>
+          </execution>
+          <execution>
+            <id>non-empty-null-test</id>
+            <goals>
+              <goal>test</goal>
+            </goals>
+            <configuration>
+              <argLine>-da:ai.rapids.cudf.AssertEmptyNulls</argLine>
+              <test>**/ColumnViewNonEmptyNullsTest.java</test>
             </configuration>
           </execution>
           <execution>
--- a/ci/premerge-build.sh
+++ b/ci/premerge-build.sh
@@ -26,5 +26,5 @@ PARALLEL_LEVEL=${PARALLEL_LEVEL:-4}
 ${MVN} verify ${MVN_MIRROR} \
   -DCPP_PARALLEL_LEVEL=${PARALLEL_LEVEL} \
   -Dlibcudf.build.configure=true \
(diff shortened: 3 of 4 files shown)


jlowe:
Are we turning _off_ assertions in unit tests?  Why would we do that?

razajafri:
> Are we turning _off_ assertions in unit tests? Why would we do that?

Yes, but only assert in the class `ai.rapids.cudf.AssertEmptyNulls`. The reason for that is that we have 4 unit tests that explicitly introduce non-empty nulls which will fail this assertion. Either we disable asserts in that class for all unit-tests or we skip those 4 unit-tests that will fail unless we run the tests with `-da:ai.rapids.cudf.AssertEmptyNulls`.

jlowe:
> Either we disable asserts in that class for all unit-tests or we skip those 4 unit-tests that will fail unless we run the tests with -da:ai.rapids.cudf.AssertEmptyNulls

Turning these assertions off for the unit tests means they never run in practice.  Java assertions are not on by default, they are explicitly enabled during our unit tests runs.  Turning them off here means these assertions in essence never will fire, which makes one wonder why we're adding them.

thirdparty/cudf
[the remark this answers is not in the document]
jlowe: Moving the PR to draft, as we don't want to pull in cudf changes from an unmerged pull request.

pom.xml
```diff
@@ -535,9 +535,20 @@
...
             </configuration>
           </execution>
+          <execution>
+            <id>non-empty-null-test</id>
+            <goals>
+              <goal>test</goal>
+            </goals>
+            <configuration>
+              <argLine>-da:ai.rapids.cudf.AssertEmptyNulls</argLine>
+              <test>**/ColumnViewNonEmptyNullsTest.java</test>
```
jlowe, line 549: I think build.sh also needs to be updated to handle this test that shouldn't run with other tests.  That script is explicitly pulling out CuFileTest and CudaFatalTest, probably needs to pull out this one as well.

revans2 (dismissed):

sameerz:
build

jlowe (approved):

razajafri:
build

pom.xml
```diff
@@ -535,9 +535,20 @@
...
+          <execution>
+            <id>non-empty-null-test</id>
+            <goals>
+              <goal>test</goal>
+            </goals>
+            <configuration>
+              <argLine>-da:ai.rapids.cudf.AssertEmptyNulls</argLine>
+              <test>ColumnViewNonEmptyNullsTest</test>
+            </configuration>
+          </execution>
```
gerashegalov, line 551: This new execution is added however all tests are excluded:
```
3:56:31  [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.0:test (non-empty-null-test) on project spark-rapids-jni: No tests were executed!  (Set -DfailIfNoTests=false to ignore this error.) -> [Help 1]
```

razajafri: How come? The documentation says it will add each test https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#test

gerashegalov: CLI properties override pom

razajafri: hmm... but that would not even run CudaFatalTest but we see in previous successful premerge that it does run the [CudaFatalTest](https://blossom.nvidia.com/sw-gpu-spark-jenkins/job/rapids-jni_premerge-github/243/console)

ci/nightly-build.sh
```diff
@@ -30,5 +30,5 @@ ${MVN} clean package ${MVN_MIRROR}  \
   -Psource-javadoc \
   -DCPP_PARALLEL_LEVEL=${PARALLEL_LEVEL} \
   -Dlibcudf.build.configure=true \
-  -DUSE_GDS=${USE_GDS} -Dtest=*,!CuFileTest,!CudaFatalTest \
+  -DUSE_GDS=${USE_GDS} -Dtest=*,!CuFileTest,!CudaFatalTest,!ColumnViewNonEmptyNullsTest \
```
gerashegalov, line 33: Try to undo this and check if it gets rid of the exclude override for ColumnViewNonEmptyNullsTest 

```suggestion
  -DUSE_GDS=${USE_GDS} -Dtest=*,!CuFileTest,!CudaFatalTest \
```

and check if you can manually run with the flags before your PR

razajafri: OK, I will try that next.

pom.xml
razajafri: I think it cannot find the Test because it's pointing to an older cudf which doesn't have that Test. I wonder how we expect this to work if we aren't supposed to commit the submodule change

razajafri:
build

pom.xml
razajafri: OK. So I got confused when @jlowe said we aren't supposed to push changes to the submodule. 
Thanks @pxLi for clarifying

pxLi (approved):

PR #1122, opened 2023-05-08
Detect recursion in pre_allocate and skip state transitions

abellina:
Due to some of the recent null_count changes in cuDF I am seeing a case I didn't cover well while implementing https://github.com/NVIDIA/spark-rapids/issues/7672.

In summary, I've come to the realization that it's really brittle to assume that cuDF operations on columns won't require allocations from the default per-device memory resource. Because the work in https://github.com/NVIDIA/spark-rapids/issues/7672 is looking to spill (handling an allocation failure), it means we could allocate while we are trying to spill. Note this didn't happen before because prior to this we would just cudaMemcpy the contiguous buffer, which we no longer have.

This change is to fix two exceptions that the state machine can raise if we are setting up a spill while we are already handling an allocation failure:

```
ai.rapids.cudf.CudfException: thread 139871119120128 in unexpected state pre alloc TASK_ALLOC_FREE
```
and

```
ai.rapids.cudf.CudfException: thread 140005702866688 in unexpected state pre alloc TASK_ALLOC
```

`TASK_ALLOC` makes perfect sense to me. We were allocating and failed due to OOM, so we called into the spill code. Then we ended up back in the same place, so we are still in `TASK_ALLOC`. This should be a pass-through and we should not transition states given this recursive call.

`TASK_ALLOC_FREE`, if I understand it correctly, can happen due to a race condition. All threads that were in `TASK_ALLOC` will go to `TASK_ALLOC_FREE` once a free succeeds in any thread, so we should treat this as a special case here, because we wouldn't have transitioned to `TASK_ALLOC_FREE` unless we were also in this recursive scenario.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -837,12 +837,29 @@ private:
    * Called prior to processing an alloc attempt. This will throw any injected exception and
    * wait until the thread is ready to actually do/retry the allocation. That blocking API may
    * throw other exceptions if rolling back or splitting the input is considered needed.
+   * 
+   * @return true if the call finds our thread in an ALLOC state, meaning that we recursively
+   *         entered the state machine. The only known case is GPU memory required for setup in
+   *         cuDF for a spill operation.
    */
-  void pre_alloc(long thread_id) {
+  bool pre_alloc(long thread_id) {
     std::unique_lock<std::mutex> lock(state_mutex);
 
     auto thread = threads.find(thread_id);
     if (thread != threads.end()) {
+      switch(thread->second.state) {
+        // If the thread is in one of the ALLOC or ALLOC_FREE states, we have detected a loop
+        // likely due to spill setup required in cuDF. We will treat this allocation differently
+        // and skip transitions.
+        case TASK_ALLOC:
+        case SHUFFLE_ALLOC:
+        case TASK_ALLOC_FREE:
+        case SHUFFLE_ALLOC_FREE:
+          return true;
+
+        default: break;
+      }
+
       if (thread->second.retry_oom_injected > 0) {
         thread->second.retry_oom_injected--;
         thread->second.num_times_retry_throw++;
@@ -870,8 +887,9 @@ private:
         case SHUFFLE_RUNNING:
           transition(thread->second, thread_state::SHUFFLE_ALLOC);
           break;
-          // TODO I don't think there are other states that we need to handle, but
-          // this needs more testing.
+
+        // TODO I don't think there are other states that we need to handle, but
+        // this needs more testing.
         default: {
           std::stringstream ss;
           ss << "thread " << thread_id << " in unexpected state pre alloc "
@@ -881,6 +899,7 @@ private:
         }
       }
     }
+    return false;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


abellina:
@revans2 let me know if the above logic makes sense. I'll add tests in a subsequent commit, but wanted some feedback.

abellina:
Also, the plugin needs to change as well to detect loops here. If we are spilling while spilling, we are going to throw an unrecoverable OOM

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1087,14 +1109,21 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
     auto thread = threads.find(thread_id);
     // only retry if this was due to an out of memory exception.
     bool ret = true;
-    if (thread != threads.end()) {
+    if (!likely_spill && thread != threads.end()) {
       switch (thread->second.state) {
-        case TASK_ALLOC_FREE: transition(thread->second, thread_state::TASK_RUNNING); break;
+        case TASK_ALLOC_FREE: 
+          if (is_oom) {
+            transition(thread->second, thread_state::TASK_BLOCKED);
```
revans2, line 1121: Why are we blocking on an OOM if a free happened while in the middle of the allocation? That is the entire point of the TASK_ALLOC_FREE state.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1163,10 +1192,15 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
 
       std::unique_lock<std::mutex> lock(state_mutex);
       for (auto thread = threads.begin(); thread != threads.end(); thread++) {
-        switch (thread->second.state) {
-          case TASK_ALLOC: transition(thread->second, thread_state::TASK_ALLOC_FREE); break;
-          case SHUFFLE_ALLOC: transition(thread->second, thread_state::SHUFFLE_ALLOC_FREE); break;
-          default: break;
+        // only update state for _other_ threads
```
revans2, line 1195: Can you add more of an explanation why this is only for other threads.

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkTest.java
```diff
@@ -774,6 +780,46 @@ public void retryWatchdog() {
     System.err.println("Took " + (endTime - startTime) + "ns to retry 500 times...");
   }
 
+  @Test
+  public void testAllocationDuringSpill() {
```
revans2, line 784: Can we add some more explanation as to why this is possible or we want to be able to deal with it.

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkTest.java
```diff
@@ -799,4 +845,53 @@ public boolean onAllocFailure(long sizeRequested, int retryCount) {
...
+    public AllocatingRmmEventHandler(long allocSize) {
+      this.allocSize = allocSize;
+    }
+
+    @Override
+    public boolean onAllocFailure(long sizeRequested, int retryCount) {
+      // Catch all JVM exceptions in this handlers to prevent loops:
+      // - If we throw a Java exception here, it will get caught in the RmmJni on_alloc_fail
+      //   method. This causes a native std::runtime_error to be thrown.
+      // - The std::runtime_error is then caught in spark_resource_adaptor::do_allocate and
```
revans2, line 870: Isn't the issue that we retry on any exception that is not an OOM exception? I am not sure why we would want to retry for other exception types. That feels wrong to me.  And I think would be a lot better than catching a Throwable and turning it into a boolean that loses all of the information about it.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
abellina: oh I am sorry, I think this was something I should have removed before pushing, since we have the `likely_spill` boolean that helps us skip this whole block in this recursive scenario.

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkTest.java
abellina: I can remove special handling of `std::exception` here: https://github.com/NVIDIA/spark-rapids-jni/pull/1122/files#diff-936903e8604f063e1364ecc9ad83dcd4d5552d6c26d93073b469f99ca562d7a3R1171 or perhaps leave the handling but always throw (instead of conditionally). Does that make sense? Or should it be removed entirely.

revans2 (approved):

PR #1128, opened 2023-05-10
Moves the lock around do_deallocate to protect threads.find

abellina:
This fixes a place where `do_deallocate` was using the STL container `threads` outside of a lock that is used in every other function I can find (I audited the rest of the usage).

Thanks to @revans2 and @jlowe they helped find this code because it looks to be pointed at by the hs_err_pid files we are getting in the crash reports (like this https://github.com/NVIDIA/spark-rapids/issues/8235).

We'd like to push this and run the CI pipelines to see if we have been able to fix.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -1174,6 +1174,8 @@ private:
     resource->deallocate(p, size, stream);
     // deallocate success
     if (size > 0) {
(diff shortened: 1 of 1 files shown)


abellina:
sorry, I forced push a signed commit.

jbrennan333 (approved):
Looks good to me.

abellina:
build

revans2 (approved):

jlowe (approved):

PR #1146, opened 2023-05-17
Avoid a race in retry where we might do a split when we don't need to

### LINKED ISSUE #1066: [BUG] Race before throwing SplitAndRetryOOM
**Describe the bug**
When we go to detect if we need to throw a SplitAndRetryOOM we are checking to see if any tasks could be blocked. This includes tasks that might still have memory allocated, but have not had a chance yet to rollback and clean up. This can cause us to throw this exception a lot more frequently than is strictly needed.  Also because the split case is not handled very well it is likely that this is causing a large number of task failures.

We should look into having the thread block if there are tasks that are in a state where they still might release more resources.  We should then wake up and throw this exception if a timeout happens or after all of the dependent tasks are in a state where we know that they have released all the resources that they are going to.

revans2:
This fixes #1066 but depends on https://github.com/NVIDIA/spark-rapids/pull/8306 to be merged first so it does not break retry in the plugin when a split is needed.

The old code had a race where if all of the threads were put into BUFN, but didn't finish releasing memory, then one of the threads would switch over and throw a `SplitAndRetryOOM` exception. Ideally what we want is to have all of the threads finish rolling back and then we retry the highest priority thread and only then do we throw a `SplitAndRetryOOM`. 

I didn't do that exactly here. Instead what I did is to only throw a `SplitAndRetryOOM` if all of the tasks are `TASK_BUFN` or are blocked on shuffle. This has the effect of delaying the `SplitAndRetryOOM` until all tasks are done rolling back. This shifts the API that would throw this exception to `blockThreadUntilReady` after a `RetryOOM` is already thrown.

Then to let the highest priority BUFN thread wake up early I have it restart iff all of the threads are blocked and memory is freed by some thread that is not this thread. In the best case all of the memory is freed before the BUFN thread needs it. But in the worst case it would retry again as it waits for more memory to be freed.

I ran some tests to see if this actually fixed the issue and using the Monte Carlo simulator I think I was able to reproduce the problem.  Without this patch when I run the following I will see a number of splits, typically between 1 and 5.  But if I run with the patch the number of splits is 0. Also in many cases I saw the performance end up being slightly better (because it was able to avoid the splits that can be costly).

```
java -cp ./target/spark-rapids-jni-23.06.0-SNAPSHOT-tests.jar:./target/spark-rapids-jni-23.06.0-SNAPSHOT-cuda11.jar:./target/dependency/\* com.nvidia.spark.rapids.jni.RmmSparkMonteCarlo --seed=11511967777172538 --taskMaxMiB=950
```

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -713,14 +713,18 @@ private:
             // and the other threads didn't get unblocked by this, so we need to
             // check again to see if this was fixed or not.
             check_and_update_for_bufn(lock);
-            log_status("WAITING", thread_id, thread->second.task_id, thread->second.state);
-            thread->second.before_block();
-            do {
-              thread->second.wake_condition->wait(lock);
-              thread = threads.find(thread_id);
-            } while (thread != threads.end() && is_blocked(thread->second.state));
-            thread->second.after_block();
-            task_has_woken_condition.notify_all();
+            // If that caused us to transition to a new state, then we need to adjust to it
+            // appropriately...
+            if (is_blocked(thread->second.state)) {
+              log_status("WAITING", thread_id, thread->second.task_id, thread->second.state);
+              thread->second.before_block();
+              do {
+                thread->second.wake_condition->wait(lock);
+                thread = threads.find(thread_id);
+              } while (thread != threads.end() && is_blocked(thread->second.state));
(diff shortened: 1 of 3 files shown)


revans2:
build

revans2:
Just merged in the patch that this depends on so now this is ready for review.

jlowe (approved):

abellina (approved):

PR #1148, opened 2023-05-17
Ensure columns have valid null counts.

mythrocks:
This is in prep for https://github.com/rapidsai/cudf/issues/11968 and https://github.com/rapidsai/cudf/pull/13372.

`libcudf` will soon require that all CUDF columns are created with a known null-count. `UNKNOWN_NULL_COUNT` will no longer be supported, or even available as a code constant.

This change replicates part of https://github.com/rapidsai/cudf/pull/13355, as it applies to `row_conversion.cu`. The (single) reference to the unknown-null-count is replaced with a pre-calculated value.


mythrocks:
Build

revans2 (approved):

mythrocks:
Thank you for reviewing, @revans2. I have merged this change.

PR #1155, opened 2023-05-19
Followup for null count fixup in row_conversion.cu.

mythrocks:
This is a followup to #1148.

`row_conversion.cu` was modified in https://github.com/rapidsai/cudf/pull/13372 to explicitly calculate null-counts for output columns.

This commit replicates the changes in rapidsai/cudf/pull/13372, and adds explicit null-count calculation for the string offsets column.

### DIFF
--- a/src/main/cpp/src/row_conversion.cu
+++ b/src/main/cpp/src/row_conversion.cu
@@ -2332,6 +2332,9 @@ std::unique_ptr<table> convert_from_rows_fixed_width_optimized(
         num_rows, num_columns, size_per_row, dev_column_start.data(), dev_column_size.data(),
         dev_output_data.data(), dev_output_nm.data(), child.data<int8_t>());
 
+    // Set null counts, because output_columns are modified via mutable-view,
+    // in the kernel above.
+    // TODO(future): Consider setting null count in the kernel itself.
     for (auto &col : output_columns) {
       col->set_null_count(cudf::null_count(col->view().null_mask(), 0, col->size()));
     }
--- a/src/main/cpp/src/row_conversion.cu
+++ b/src/main/cpp/src/row_conversion.cu
@@ -2276,9 +2276,6 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
     }
   }
 
-  for (auto &col : output_columns) {
-    col->set_null_count(cudf::null_count(col->view().null_mask(), 0, col->size()));
-  }
   return std::make_unique<table>(std::move(output_columns));
 }
 
--- a/src/main/cpp/src/row_conversion.cu
+++ b/src/main/cpp/src/row_conversion.cu
@@ -2259,8 +2259,14 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
     for (int i = 0; i < static_cast<int>(schema.size()); ++i) {
       if (schema[i].id() == type_id::STRING) {
         // stuff real string column
-        auto const null_count = string_row_offset_columns[string_idx]->null_count();
         auto string_data = string_row_offset_columns[string_idx].release()->release();
+        auto const null_count = [&] {
+          auto const &null_mask = *string_data.null_mask;
+          return null_mask.data() ?
+                     cudf::null_count(static_cast<bitmask_type const *>(null_mask.data()), 0,
+                                      num_rows) :
+                     0;
+        }();
         output_columns[i] =
             make_strings_column(num_rows, std::move(string_col_offsets[string_idx]),
                                 std::move(string_data_cols[string_idx]),
@@ -2270,6 +2276,9 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
     }
   }
 
+  for (auto &col : output_columns) {
+    col->set_null_count(cudf::null_count(col->view().null_mask(), 0, col->size()));
+  }
   return std::make_unique<table>(std::move(output_columns));
 }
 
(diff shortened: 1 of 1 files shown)


mythrocks:
Build

src/main/cpp/src/row_conversion.cu
```diff
@@ -2259,8 +2259,14 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
...
         // stuff real string column
-        auto const null_count = string_row_offset_columns[string_idx]->null_count();
         auto string_data = string_row_offset_columns[string_idx].release()->release();
+        auto const null_count = [&] {
+          auto const &null_mask = *string_data.null_mask;
+          return null_mask.data() ?
+                     cudf::null_count(static_cast<bitmask_type const *>(null_mask.data()), 0,
+                                      num_rows) :
+                     0;
+        }();
```
ttnghia, line 2269: Similar to my previous comment in previous PR, put this above `string_data`:
```suggestion
auto const null_count = string_row_offset_columns[string_idx]->null_count();
```

src/main/cpp/src/row_conversion.cu
```diff
@@ -2270,6 +2276,9 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
     }
   }
 
+  for (auto &col : output_columns) {
+    col->set_null_count(cudf::null_count(col->view().null_mask(), 0, col->size()));
+  }
```
ttnghia, line 2281: This is not needed, because null mask/count need to be set during column construction in line 2272 above.

src/main/cpp/src/row_conversion.cu
```diff
@@ -2325,6 +2334,9 @@ std::unique_ptr<table> convert_from_rows_fixed_width_optimized(
         num_rows, num_columns, size_per_row, dev_column_start.data(), dev_column_size.data(),
         dev_output_data.data(), dev_output_nm.data(), child.data<int8_t>());
 
+    for (auto &col : output_columns) {
+      col->set_null_count(cudf::null_count(col->view().null_mask(), 0, col->size()));
+    }
```
ttnghia, line 2339: This is similar.

I'm also not sure if we can compute null mask and null count in the kernel above  this.

src/main/cpp/src/row_conversion.cu
mythrocks: That is the problem that I'm trying to rectify. You'll notice that your suggestion aligns with how the code is currently. That's what we went with in #1148.

The problem is that calling `string_row_offset_columns[string_idx]->null_count()` will not automatically calculate the null-count if none was provided, after [#13372](https://github.com/rapidsai/cudf/pull/13372/files#diff-0d9ff1eeb036035ec9552be6a9f182c150233a8d68ae714de574bb8c309b1172L130) is merged.

Perhaps I have missed something?

src/main/cpp/src/row_conversion.cu
mythrocks: Based on prior discussion with @hyperbolic2346, it might  well be possible to track the null-mask and null-count in the kernel. I think it would be better not to change that kernel till after 23.06. (Unless @hyperbolic2346 would like to have a go at it? :])

src/main/cpp/src/row_conversion.cu
mythrocks: Yes, sorry. That's not required. I'll remove that.

src/main/cpp/src/row_conversion.cu
ttnghia: Oh so you meant that the column `string_row_offset_columns[string_idx]` was actually constructed without the `null_count` number provided? Then that makes sense. So please add a comment to clarify that.

mythrocks: Added docs for this.

src/main/cpp/src/row_conversion.cu
mythrocks: Removed redundant calc.

src/main/cpp/src/row_conversion.cu
mythrocks: I've documented why this is being set externally, and added a to-do to move this to the kernel if possible.

mythrocks:
Build

src/main/cpp/src/row_conversion.cu
```diff
@@ -2259,8 +2259,15 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
     for (int i = 0; i < static_cast<int>(schema.size()); ++i) {
       if (schema[i].id() == type_id::STRING) {
         // stuff real string column
-        auto const null_count = string_row_offset_columns[string_idx]->null_count();
         auto string_data = string_row_offset_columns[string_idx].release()->release();
+        auto const null_count = [&] {
+          // Null-count not set previously. Calculate, on the fly.
+          auto const &null_mask = *string_data.null_mask;
+          return null_mask.data() ?
+                     cudf::null_count(static_cast<bitmask_type const *>(null_mask.data()), 0,
```
ttnghia, line 2267:
```suggestion
                     cudf::null_count(reinterpret_cast<bitmask_type const *>(null_mask.data()), 0,
```

mythrocks: `null_mask.data()` is a `void*`. There is a standards-defined conversion from `void*` to `T*`. Is there a reason here to use `reinterpret_cast` when `static_cast` suffices?

mythrocks: Oh, interesting: 
1. Per [some readings](http://jdebp.info/FGA/reinterpret-cast-and-pointer-to-void.html) of the Standard, using `reinterpret_cast` here would constitute bad code. TIL. I don't have a copy of the standards to confirm.
2. Stroustrup seems to have [changed](https://groups.google.com/g/comp.lang.c++.moderated/c/rcgB5ZRooIE/m/rvv5UQhyDCUJ) from using `reinterpret_cast` to `static_cast`, for `void* -> T*`, in the "Special Edition" of TC++PL.

Looks like `static_cast` is the right call here.

mythrocks:
Build

mythrocks:
Thanks for reviewing, @hyperbolic2346. I'm ready to merge this, if @ttnghia agrees on my assessment on the remaining issue.

src/main/cpp/src/row_conversion.cu
mythrocks: Keeping this as `static_cast`. I'm open to discussing this if required.

PR #1168, opened 2023-05-24
Adding non-ansi decimal to string conversion

hyperbolic2346:
This change adds a custom kernel for decimal to string conversion. This is identical to the cudf kernel with the exception of a zero with more than 5 decimal places. Spark < 3.4 creates a string in scientific notation for this case as described in https://github.com/NVIDIA/spark-rapids/issues/7870
 
This is the kernel side of the work and doesn't close that issue. There will be a PR on the plugin side to change the default behavior based on spark version and configuration.

### DIFF
--- a/src/main/cpp/src/cast_decimal_to_string.cu
+++ b/src/main/cpp/src/cast_decimal_to_string.cu
@@ -1,0 +1,203 @@
+/*
+ * Copyright (c) 2023, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cast_string.hpp"
+
+#include <cudf/column/column_device_view.cuh>
+#include <cudf/column/column_factories.hpp>
+#include <cudf/detail/iterator.cuh>
+#include <cudf/detail/null_mask.hpp>
+#include <cudf/detail/nvtx/ranges.hpp>
+#include <cudf/strings/detail/convert/int_to_string.cuh>
+#include <cudf/strings/detail/converters.hpp>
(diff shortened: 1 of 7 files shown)


src/main/cpp/src/cast_decimal_to_string.cu
```diff
@@ -0,0 +1,203 @@
...
+   *
+   * Output format is [-]integer.fraction
+   */
+  __device__ int32_t compute_output_size(DecimalType value)
+  {
+    auto const scale = d_decimals.type().scale();
+
+    if (scale >= 0) return strings::detail::count_digits(value) + scale;
+
+    auto const abs_value = numeric::detail::abs(value);
```
revans2, line 70: Same comment here about overflow.

revans2 (approved):
Looks good to me. Would like to see some comment about abs being used here and why it is okay, or fix the abs issue so it is not going to be a problem.

revans2:
Oh you also need to sign off on the PR.

src/main/cpp/src/cast_decimal_to_string.cu
hyperbolic2346: So both of these overflows are in the kernel running now from cudf. The cudf approach doesn't verify data and lets the user do undefined things if they want and things just don't work as expected. I don't know if that is the approach we want to take or not. I'm fine either way.

revans2: The overflow I am worried about is the min value for `DecimalType` lets take `int32_t` as an example because it is the smallest one and a scale of -1, to expose a problem. In this case `abs(-2147483648)` is `-2147483648`

This is going to mess up both the size logic and the string itself.

When calculating the size we get:

`exp_ten(1)` will be 10

Then to compute `fraction` we are doing a `count_digits` on `abs_value % exp_ten` `-2147483648 % 10` is `-8`

`count_digits` of -8 is 2, because it includes a space for the "-".

num_zeros is now 0 because `max(0, 1 - 2)` is 0

Now we calculate the output size. `value < 0` + count digits(-214748364) + 1 + 0 + 2

1 + 10 + 1 + 0 + 2 => 14


When outputting the result we get

"--214748364.-8" That is a total of 14 characters just like the count wanted, so no data overflows here, but I don't think that is the correct answer.

revans2: To be clear I don't think this will show up in practice because the smallest decimal32 value we should ever see is `-999999999`

revans2 (dismissed):
The only nit I might have is pointing to the docs that this code is based off of.

revans2:
build

revans2 (approved):

hyperbolic2346:
build

PR #1172, opened 2023-05-25
Fixing float to string not setting null count

hyperbolic2346:
This is a fix from the fallout of the cudf change to not allow unknown validity. We computed, incorrectly, the null count in this kernel, but then didn't use that value at all. I've updated it to properly compute the validity and actually store it in the column. This isn't the fastest approach, but it works. We can investigate performance later if it becomes an issue.

### DIFF
--- a/src/main/cpp/src/cast_string_to_float.cu
+++ b/src/main/cpp/src/cast_string_to_float.cu
@@ -555,22 +555,14 @@ class string_to_float {
   // outgoing ansi_except field
   __device__ void compute_validity(bool const valid, bool const except = false)
   {
-    // compute null count for the block. each warp processes one string, so lane 0
-    // from each warp contributes 1 bit of validity
-    size_type const block_valid_count =
-      cudf::detail::single_lane_block_sum_reduce<block_size, 0>(valid ? 1 : 0);
-    // 0th thread in each block updates the validity count and (optionally) the ansi_except flag
-    if (threadIdx.x == 0) {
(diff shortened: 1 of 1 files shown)


src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -555,22 +555,14 @@ class string_to_float {
...
-      if (_ansi_except && except) { atomicMax(_ansi_except, _num_rows - _row); }
-    }
+    if (threadIdx.x == 0 && _ansi_except && except) { atomicMax(_ansi_except, _num_rows - _row); }
 
     // 0th thread in each warp updates the validity
     size_type const row_id = _warp_id;
     if (threadIdx.x % 32 == 0 && valid) {
       // uses atomics
       cudf::set_bit(_validity, row_id);
+      atomicAdd(_valid_count, 1);
```
mythrocks, line 565: Ah, ok. This makes sense, phrased this way.

mythrocks (approved):
LGTM! The sync-test now runs cleanly.

pxLi:
@hyperbolic2346 can you sign the commit? thanks~

pxLi:
build

pxLi:
build

nvdbaranec:
An alternative solution to consider:  don't have `compute_validity` compute null count.  Just have it update the validity vector.  Instead, have the operator() itself return 1 or 0 for null/non null and then do one big `single_lane_block_sum_reduce()` and the end of the kernel and store the null count once, without the atomic.

PR #1175, opened 2023-05-26
Implement Remainder for Decimal128 that handles when operands overflow

NVnavkumar:
Part of a fix for https://github.com/NVIDIA/spark-rapids/issues/8330.

This implements a native GPU implementation of the Remainder computation for 2 DECIMAL128 columns. This patterns the same algorithm that the JDK uses when computing the remainder for 2 BigDecimal values (which is the algorithm that Apache Spark relies on). This is used when trying the compute the remainder for 2 columns that would otherwise need to be upconverted to a higher precision than is supported by the Decimal128 format (higher than 38). 

The algorithm is quite simple for `a % b` (where a and b are 2 DECIMAL128 values/columns): 
1) Compute the integer division of the 2 columns: `r_i = a // b`.
2) Multiply that result by `b`: `a_int = r_i * b`.
3) The remainder is the result of subtracting that value from `a`: `r = a - a_int`.

This PR does this in a single custom kernel for optimization, and returns a result in the requested scale.

A corresponding PR in https://github.com/NVIDIA/spark-rapids/ will be created to implement the Spark plugin support for this.

### DIFF
--- a/src/main/cpp/src/DecimalUtilsJni.cpp
+++ b/src/main/cpp/src/DecimalUtilsJni.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/cpp/src/DecimalUtilsJni.cpp
+++ b/src/main/cpp/src/DecimalUtilsJni.cpp
@@ -58,6 +58,22 @@ JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_DecimalUtils_divid
   CATCH_STD(env, 0);
 }
 
+JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_DecimalUtils_remainder128(JNIEnv *env, jclass,
+                                                                                     jlong j_view_a,
+                                                                                     jlong j_view_b,
+                                                                                     jint j_remainder_scale) {
+  JNI_NULL_CHECK(env, j_view_a, "column is null", 0);
+  JNI_NULL_CHECK(env, j_view_b, "column is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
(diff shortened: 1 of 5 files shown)


NVnavkumar:
build

revans2 (approved):
Looks good. Because it is so inefficient, it would be nice to have a follow on issue to see if we can speed this up. Or at least do a benchmark and see if it is good/bad.

src/main/cpp/src/DecimalUtilsJni.cpp
```diff
@@ -58,6 +58,22 @@ JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_DecimalUtils_divid
   CATCH_STD(env, 0);
 }
 
+JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_DecimalUtils_remainder128(JNIEnv *env, jclass,
```
abellina, line 61: file needs copyright updated.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -25,6 +25,9 @@
 #include <cmath>
 #include <cstddef>
 
+#define D19_ UINT64_C(10000000000000000000)
```
abellina, line 28: I don't see this used in this PR.

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -79,6 +79,26 @@ public static Table divide128(ColumnView a, ColumnView b, int quotientScale) {
...
+  /**
+   * Divide two DECIMAL128 columns and produce a DECIMAL128 remainder with overflow detection.
+   * Example:
+   * 451635271134476686911387864.48 % -961.110 = 775.233
+   * 
+   * Generally, this will never really overflow unless in the divide by zero case.
+   * But it will detect an overflow in any case.
+   *
+   * @param a factor input, must match row count of the other factor input
+   * @param b factor input, must match row count of the other factor input
```
abellina, line 92: missing `remainderScale` in the docs.

abellina (approved):

NVnavkumar:
build

revans2 (approved):

PR #1217, opened 2023-06-14
Adds in a metric for computation time lost due to retry.

revans2:
This adds in some new APIs that let us collect a metric for how much time has been spent in a retry block before the retry made us roll it back and start over. I will put up a corresponding change in the plugin to use these new APIs and collect the metric.  This is really helpful in seeing how much time we likely lost and need to redo when a retry happens.  It helped to point out that in one case I have been testing that the retry was causing a lot of lost computation and that we really would have benefitted from https://github.com/NVIDIA/spark-rapids/issues/7868 because we were getting the retry after doing a lot of computation in a very large project statement.

I am hoping that the changes are simple enough I don't need to add comments, but if you cannot follow them I am obviously wrong and am happy to add them. I did the work quickly as a part of debugging an issue so feel free to rip this apart.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -186,6 +186,11 @@ public:
   int num_times_retry_throw = 0;
   int num_times_split_retry_throw = 0;
   long time_blocked_nanos = 0;
+  long time_lost_nanos = 0;
+  long time_retry_running_nanos = 0;
+  std::chrono::time_point<std::chrono::steady_clock> retry_start_or_block_end;
+  bool is_in_retry = false;
+
 
   std::chrono::time_point<std::chrono::steady_clock> block_start;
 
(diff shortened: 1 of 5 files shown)


revans2:
build

revans2:
https://github.com/NVIDIA/spark-rapids/pull/8578 is the corresponding PR in the plugin

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -290,6 +333,37 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+
+  void end_retry_block(long thread_id) {
+    std::unique_lock<std::mutex> lock(state_mutex);
+    auto thread = threads.find(thread_id);
+    if (thread != threads.end()) {
+      thread->second.reset_retry_state(false);
+    }
+  }
+
+  long get_n_reset_lost_time(long task_id) {
```
jlowe, line 352: Nit: inconsistent use of `_n_` vs `_and_` in function names.  Would be nice to pick one and consistently use it.

jlowe (dismissed):

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -186,6 +186,11 @@ class full_thread_state {
   int num_times_retry_throw = 0;
   int num_times_split_retry_throw = 0;
   long time_blocked_nanos = 0;
+  long time_lost_nanos = 0;
+  long time_retry_running_nanos = 0;
+  std::chrono::time_point<std::chrono::steady_clock> retry_start_or_block_end;
```
abellina, line 191: could there be a comment for these times on what they mean? I think it would improve on figuring out how the code should be used.

revans2:
@jlowe @abellina please take another look

jlowe (approved):

abellina (approved):

PR #1269, opened 2023-07-14
Add bloom filter support.

nvdbaranec:
Adds support for Spark-style bloom filters via the `BloomFilter` class.  The gpu implementation is in `spark-rapids-jni` itself and not `cudf`.

Added benchmark for `bloom_filter_put`. On an A5000, we're getting 140 GB/s write-throughput for bloom filter sizes of 512k, 1MB, 2MB, 4MB and 8MB.  12.5 milliseconds for 150 million rows. So it's not lightning fast, but it's serviceable.

Also fixed several assorted benchmark build errors.  The cudf push for always providing null counts and specifying stream/mr broke a few of them.


src/main/cpp/src/BloomFilterJni.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "bloom_filter.hpp"
+#include "utilities.hpp"
+
+extern "C" {
+
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_BloomFilter_put(JNIEnv* env,
+                                                                         jclass,
+                                                                         jlong bloom_filter,
+                                                                         jlong bloom_filter_bytes,
+                                                                         long bloom_filter_bits,
```
jlowe, line 30: Curious why we're passing both the bit count and the byte count?  Seems like we could compute the byte count from the bit count trivially.

nvdbaranec: Right. My thinking was to distinguish between the bit count as being an explicit property of the bloom filter itself and the byte count as just being a side-effect of how the underlying buffer ends up being represented.  Also, since we're passing the buffer pointer itself, it felt slightly more correct to include the byte count as being explicitly related.   

All that said, I'm not married to that at all.

pxLi:
build

src/main/cpp/src/BloomFilterJni.cpp
revans2: nit: technically bloom_filter_bits should be a `jlong` too.

src/main/cpp/src/BloomFilterJni.cpp
```diff
@@ -0,0 +1,95 @@
...
+                                                                           jint num_hashes)
+{
+  try {
+    cudf::jni::auto_set_device(env);
+
+    cudf::column_view input_column{*reinterpret_cast<cudf::column_view const*>(cv)};
+    return cudf::jni::release_as_jlong(spark_rapids_jni::bloom_filter_probe(
+      input_column,
+      {reinterpret_cast<cudf::bitmask_type const*>(bloom_filter),
+       static_cast<std::size_t>(bloom_filter_bytes / 4)},
```
revans2, line 63: nit: the 4 magic number. Should that be `sizeof(cudf::butemask_type)`?  Also would it be good to have a common API to do the conversion because it happens everywhere.

src/main/java/com/nvidia/spark/rapids/jni/BloomFilter.java
```diff
@@ -0,0 +1,164 @@
...
+    for (int i = 0; i < bloomFilters.length; i++) {
+      buffers[i] = bloomFilters[i].getBuffer().getAddress();
+    }
+
+    int numHashes = bloomFilters[0].numHashes;
+    long bloomFilterBits = bloomFilters[0].bloomFilterBits;
+    long bloomFilterBytes = bloomFilterByteSize(bloomFilterBits);
+
+    long[] merged = merge(buffers, bloomFilterBytes);
+    return new BloomFilter(numHashes, bloomFilterBits, DeviceMemoryBuffer.fromRmm(merged[0], bloomFilterBytes, merged[1]));
```
revans2, line 147: nit: Technically if we made a mistake and allocated the merged bit/bytes wrong, the resulting bloom filter would be leaked because an exception would be thrown inside the constructor and nothing would be closed/cleaned up.

This is very minor.

revans2 (approved):
Just a few nits.

src/main/cpp/src/BloomFilterJni.cpp
```diff
@@ -0,0 +1,95 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cudf_jni_apis.hpp"
+#include "dtype_utils.hpp"
+#include "jni_utils.hpp"
+
+#include "bloom_filter.hpp"
+#include "utilities.hpp"
```
ttnghia, line 22: Should be the other way around:
```suggestion
#include "bloom_filter.hpp"
#include "utilities.hpp"

#include "cudf_jni_apis.hpp"
#include "dtype_utils.hpp"
#include "jni_utils.hpp"
```

src/main/cpp/src/BloomFilterJni.cpp
```diff
@@ -0,0 +1,95 @@
...
+                                                                         jlong bloom_filter,
+                                                                         jlong bloom_filter_bytes,
+                                                                         long bloom_filter_bits,
+                                                                         jlong cv,
+                                                                         jint num_hashes)
+{
+  try {
+    cudf::jni::auto_set_device(env);
+
+    cudf::column_view input_column{*reinterpret_cast<cudf::column_view const*>(cv)};
```
ttnghia, line 37: Oh no this is a copy construction.
```suggestion
    cudf::column_view input_column = reinterpret_cast<cudf::column_view const*>(cv);
    // use `*input_column`
```

src/main/cpp/src/BloomFilterJni.cpp
```diff
@@ -0,0 +1,95 @@
...
+                                                                           jlong cv,
+                                                                           jlong bloom_filter,
+                                                                           jlong bloom_filter_bytes,
+                                                                           long bloom_filter_bits,
+                                                                           jint num_hashes)
+{
+  try {
+    cudf::jni::auto_set_device(env);
+
+    cudf::column_view input_column{*reinterpret_cast<cudf::column_view const*>(cv)};
```
ttnghia, line 59: Similar to above.

src/main/cpp/src/bloom_filter.cu
```diff
@@ -0,0 +1,148 @@
...
+namespace spark_rapids_jni {
+
+namespace {
+
+using bloom_hash_type = spark_rapids_jni::murmur_hash_value_type;
+
+__device__ inline std::pair<cudf::size_type, cudf::bitmask_type> gpu_get_hash_mask(
+  bloom_hash_type h, cudf::size_type bloom_filter_bits)
+{
+  // https://github.com/apache/spark/blob/7bfbeb62cb1dc58d81243d22888faa688bad8064/common/sketch/src/main/java/org/apache/spark/util/sketch/BloomFilterImpl.java#L94
```
ttnghia, line 37: Love this. It is encouraging to adopt the same practice: including ref. code whenever possible.

src/main/cpp/src/bloom_filter.cu
```diff
@@ -0,0 +1,148 @@
...
+  return {word_index, (1 << bit_index)};
+}
+
+__global__ void gpu_bloom_filter_put(cudf::bitmask_type* const bloom_filter,
+                                     cudf::size_type bloom_filter_bits,
+                                     cudf::device_span<int64_t const> input,
+                                     cudf::size_type num_hashes)
+{
+  int const tid = threadIdx.x + blockIdx.x * blockDim.x;
+  if (tid >= input.size()) { return; }
```
ttnghia, line 50: Can `tid` be overflow when `input.size() == INT_MAX`?

src/main/cpp/src/bloom_filter.cu
```diff
@@ -0,0 +1,148 @@
...
+
+  // https://github.com/apache/spark/blob/7bfbeb62cb1dc58d81243d22888faa688bad8064/common/sketch/src/main/java/org/apache/spark/util/sketch/BloomFilterImpl.java#L87
+  bloom_hash_type const h1 = MurmurHash3_32<int64_t>(0)(input[tid]);
+  bloom_hash_type const h2 = MurmurHash3_32<int64_t>(h1)(input[tid]);
+
+  // set a bit in the bloom filter for each hashed value
+  for (auto idx = 1; idx <= num_hashes; idx++) {
+    bloom_hash_type combined_hash = h1 + (idx * h2);
```
ttnghia, line 58: Cab `idx * h2` overflow?

src/main/cpp/src/bloom_filter.cu
```diff
@@ -0,0 +1,148 @@
...
+    atomicOr(bloom_filter + word_index, mask);
+  }
+}
+
+struct bloom_probe_functor {
+  cudf::bitmask_type const* const bloom_filter;
+  cudf::size_type const bloom_filter_bits;
+  cudf::size_type const num_hashes;
+
+  __device__ bool operator()(int64_t input)
```
ttnghia, line 70:
```suggestion
  __device__ bool operator()(int64_t input) const
```

src/main/cpp/src/bloom_filter.cu
```diff
@@ -0,0 +1,148 @@
...
+    // https://github.com/apache/spark/blob/7bfbeb62cb1dc58d81243d22888faa688bad8064/common/sketch/src/main/java/org/apache/spark/util/sketch/BloomFilterImpl.java#L110
+    // this code could be combined with the very similar code in gpu_bloom_filter_put. i've
+    // left it this way since the expectation is that we will early out fairly often, whereas
+    // in the build case we never early out so doing the additional if() return check is pointless.
+    bloom_hash_type const h1 = MurmurHash3_32<int64_t>(0)(input);
+    bloom_hash_type const h2 = MurmurHash3_32<int64_t>(h1)(input);
+
+    // set a bit in the bloom filter for each hashed value
+    for (auto idx = 1; idx <= num_hashes; idx++) {
+      bloom_hash_type combined_hash = h1 + (idx * h2);
```
ttnghia, line 81: Similar to above: overflow concern.

src/main/cpp/src/bloom_filter.cu
```diff
@@ -0,0 +1,148 @@
...
+
+}  // anonymous namespace
+
+std::unique_ptr<rmm::device_buffer> bloom_filter_create(int64_t bloom_filter_bits,
+                                                        rmm::cuda_stream_view stream,
+                                                        rmm::mr::device_memory_resource* mr)
+{
+  std::unique_ptr<rmm::device_buffer> out = std::make_unique<rmm::device_buffer>(
+    cudf::num_bitmask_words(bloom_filter_bits) * sizeof(cudf::bitmask_type), stream, mr);
+  cudaMemsetAsync(out->data(), 0, out->size(), stream);
```
ttnghia, line 97: Please wrapt it in `CUDF_CUDA_TRY`.

src/main/cpp/src/bloom_filter.cu
```diff
@@ -0,0 +1,148 @@
...
+}
+
+void bloom_filter_put(cudf::device_span<cudf::bitmask_type> bloom_filter,
+                      int64_t bloom_filter_bits,
+                      cudf::column_view const& input,
+                      cudf::size_type num_hashes,
+                      rmm::cuda_stream_view stream)
+{
+  CUDF_EXPECTS(input.type() == cudf::data_type{cudf::type_id::INT64} && !input.nullable(),
+               "bloom filter input expects a non-nullable column of int64s");
```
ttnghia, line 108: Can we replace the interface with `device_uvector<int64_t>`?

src/main/cpp/src/utilities.cu
```diff
@@ -0,0 +1,73 @@
...
+    return std::make_unique<rmm::device_buffer>(rmm::device_buffer{0, stream, mr});
+  }
+
+  // move the pointers to the gpu
+  std::vector<cudf::bitmask_type const*> h_input(input.size());
+  std::transform(
+    input.begin(), input.end(), h_input.begin(), [](auto mask) { return mask.data(); });
+  rmm::device_uvector<cudf::bitmask_type const*> d_input(
+    h_input.size(), stream, rmm::mr::get_current_device_resource());
+  cudaMemcpyAsync(d_input.data(),
```
ttnghia, line 50: Wrap it in `CUDF_CUDA_TRY`.

src/main/cpp/src/utilities.cu
```diff
@@ -0,0 +1,73 @@
...
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/device_uvector.hpp>
+#include <rmm/device_vector.hpp>
+#include <rmm/exec_policy.hpp>
+#include <rmm/mr/device/device_memory_resource.hpp>
+
+namespace spark_rapids_jni {
+
+std::unique_ptr<rmm::device_buffer> bitmask_bitwise_or(
+  std::vector<cudf::device_span<cudf::bitmask_type const>> const& input,
```
ttnghia, line 30: We can be span of span of span of span of spannnnnnnnnnnnnnnnnnnnn.

src/main/cpp/src/utilities.cu
ttnghia: Actually we can use `make_device_uvector_async` to cover this instead of manual memcpy.

src/main/cpp/src/utilities.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <cudf/utilities/default_stream.hpp>
+#include <cudf/utilities/span.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/device_uvector.hpp>
+
+namespace spark_rapids_jni {
+
+/**
+ * @brief Bitwise-or an array of equally-sized bitmask buffers into a single output buffer
```
ttnghia, line 29: What is the difference of this vs `cudf::bitmask_or(table_view, ...)`?

src/main/java/com/nvidia/spark/rapids/jni/BloomFilter.java
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Probe the bloom filter with a column of longs. Returns a column of booleans. For 
+   * each row in the output; a value of true indicates that the corresponding input value
+   * -may- be in the set of values used to build the bloom filter; a value of false indicates
+   * that the corresponding input value is conclusively not in the set of values used to build
+   * the bloom filter.
+   * @param cv The column containing the values to check.
+   * @return A boolean column indicating the results of the probe.
+   */
+  public ColumnVector probe(ColumnView cv){
```
jlowe, line 106: It would be nice to have a static method form for this.  Currently the way this is working out in the RAPIDS Accelerator, it creates one of these objects just to call this method (in the filtering expression)  then throws the object away (because it handles making the buffer spillable outside of this object).

We may end up having something similar on the filter builder side as well, where we need to make the buffer spillable between put calls, which means this object only lives for just one method call.  At that point, we should just have this be one-shot static methods rather than a state-tracking class.  It looks like we want the functionality exposed as static methods.  We can keep the object around for now until the Spark-side gels a bit further.

src/main/java/com/nvidia/spark/rapids/jni/BloomFilter.java
```diff
@@ -0,0 +1,164 @@
...
+  }  
+
+  /**
+   * Construct a BloomFilter from a pre-existing buffer, using the specified number of
+   * hashes and bits.
+   * @param numHashes The number of hashes to use when adding values and probing.
+   * @param bloomFilterBits The size of the bloom filter in bits.
+   * @param buffer The pre-existing buffer of.
+   */
+  public BloomFilter(int numHashes, long bloomFilterBits, DeviceMemoryBuffer buffer){
```
jlowe, line 68: I think this should take a BaseDeviceMemoryBuffer instead.  That way it can take a zero-copy slice of another buffer rather than making the caller copy it first to get a DeviceMemoryBuffer.

src/main/cpp/src/BloomFilterJni.cpp
nvdbaranec: Copy constructor of a column_view isn't a big deal.  I can make it a `const&` though:

`cudf::column_view const& input_column = *reinterpret_cast<cudf::column_view const*>(cv);`

src/main/cpp/src/bloom_filter.cu
nvdbaranec: Yes, but this will never be passed a column bigger than a cudf column.  I'll change anyway though.  May as well futureproof.

src/main/cpp/src/bloom_filter.cu
nvdbaranec: In this case, the overflow is intentional.  It's just mixing bits up in an arbitrary way.  The link above shows the spark code doing it the same way.

src/main/cpp/src/bloom_filter.cu
nvdbaranec: Which interface? `gpu_bloom_filter_put`? The reason I don't like a span there is that it's passing useless information.  We don't need to know the size of the `bloom_filter` array itself.  We just care how many bits there are (which, importantly, can be different from length * 32)

src/main/cpp/src/utilities.cu
nvdbaranec: I'll make the change, but the counterpoint here is that `make_device_uvector_async` is a cudf detail function which is something we should generally avoid trying to call.

src/main/cpp/src/utilities.hpp
nvdbaranec: This works on raw device buffers, not tables/columns.

ttnghia: But are they doing the same thing? If so, can we create dummy columns with the given bitmasks then just call `cudf::bitmask_or`?

nvdbaranec: It's close, but the work to get everything wrapped up would end up being pretty much the same as just setting up the array of pointers and calling thrust like we're doing here. It would also involve more memory allocation (in terms of number of allocations).

nvdbaranec: Also I'd say that philosophically we're already down below the abstraction level of columns and tables. We've playing around with buffers.  It's somewhat odd to jump back up to the realm of things like columns and tables when the data we've got really isn't the same (meaning the columns we'd be creating wouldn't even have data in them.  We be just faking it so we can manipulate the bitmasks).

src/main/java/com/nvidia/spark/rapids/jni/BloomFilter.java
nvdbaranec: Done

src/main/java/com/nvidia/spark/rapids/jni/BloomFilter.java
```diff
@@ -0,0 +1,193 @@
...
+      if(numHashes <= 0){
+        throw new IllegalArgumentException("Bloom filters must have a positive hash count");
+      }
+      if(bloomFilterBits <= 0){
+        throw new IllegalArgumentException("Bloom filters must have a positive number of bits");
+      }
+      if(buffer.getLength() != bloomFilterByteSize(bloomFilterBits)){
+        throw new IllegalArgumentException("Invalid pre-existing buffer passed. Size mismatch");
+      }
+    } catch (Exception e) {
```
revans2, line 82: nit:
```
} catch (Throwable e) {
```

src/main/java/com/nvidia/spark/rapids/jni/BloomFilter.java
revans2: I think what you want to do is the following.

```
long merged = merge(buffers, bloomFilterBytes);
DeviceMemoryBuffer buffer = fromRMM(merged[0], bloomFilterBytes, merged[1]);
try {
  return new BloomFilter(numHashes, bloomFilterBits, buffer);
} catch (Throwable t) {
  buffer.close();
  buffer = null;
  throw t;
}
```

revans2 (approved):
After chatting with Dave offline I now think this is good to go.

ttnghia (dismissed):

src/main/cpp/src/bloom_filter.cu
```diff
@@ -0,0 +1,153 @@
...
+
+using bloom_hash_type = spark_rapids_jni::murmur_hash_value_type;
+
+__device__ inline std::pair<cudf::size_type, cudf::bitmask_type> gpu_get_hash_mask(
+  bloom_hash_type h, cudf::size_type bloom_filter_bits)
+{
+  // https://github.com/apache/spark/blob/7bfbeb62cb1dc58d81243d22888faa688bad8064/common/sketch/src/main/java/org/apache/spark/util/sketch/BloomFilterImpl.java#L94
+  auto const index      = (h < 0 ? ~h : h) % static_cast<bloom_hash_type>(bloom_filter_bits);
+  auto const word_index = cudf::word_index(index);
+  auto const bit_index  = cudf::intra_word_index(index);
```
jlowe, line 41: This almost works, but the data format we receive from Spark is actually big endian, because we receive the serialized form of the Bloom filter (see BloomFilterImpl.writeTo which uses DataOutputStream which in turn encodes big-endian).  We can operate on the big-endian longs directly like this:
```suggestion
  auto const word_index = cudf::word_index(index) ^ 0x1;
  auto const bit_index  = cudf::intra_word_index(index) ^ 0x18;
```

nvdbaranec: Done

jlowe (approved):

jlowe:
build

src/main/java/com/nvidia/spark/rapids/jni/BloomFilter.java
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Probe a bloom filter with a column of longs. Returns a column of booleans. For 
+   * each row in the output; a value of true indicates that the corresponding input value
+   * -may- be in the set of values used to build the bloom filter; a value of false indicates
+   * that the corresponding input value is conclusively not in the set of values used to build
+   * the bloom filter. It is expected that bloomFilterBits is a multiple of 64.
+   * @param cv The column containing the values to check.
+   * @return A boolean column indicating the results of the probe.
+   */
+  public static ColumnVector probe(int numHashes, int bloomFilterBits, BaseDeviceMemoryBuffer bloomFilter, ColumnView cv){
```
jlowe, line 152:
```suggestion
  public static ColumnVector probe(int numHashes, long bloomFilterBits, BaseDeviceMemoryBuffer bloomFilter, ColumnView cv){
```

jlowe (changes requested):

src/main/java/com/nvidia/spark/rapids/jni/BloomFilter.java
```diff
@@ -0,0 +1,205 @@
...
+   */
+  public BaseDeviceMemoryBuffer getBuffer(){
+    return bloomFilter;
+  }
+
+  /**
+   * Insert a column of longs into a bloom filter. It is expected that bloomFilterBits is a multiple of 64.
+   * @param cv The column containing the values to add.
+   */
+  public static void put(int numHashes, int bloomFilterBits, BaseDeviceMemoryBuffer bloomFilter, ColumnView cv){
```
jlowe, line 136:
```suggestion
  public static void put(int numHashes, long bloomFilterBits, BaseDeviceMemoryBuffer bloomFilter, ColumnView cv){
```

nvdbaranec:
Made obsolete by:  https://github.com/NVIDIA/spark-rapids-jni/pull/1303

PR #1303, opened 2023-07-28
Rework BloomFilter interface

nvdbaranec:
Adds support for Spark-style bloom filters via the BloomFilter class. The gpu implementation is in spark-rapids-jni itself and not cudf.

This version of the PR uses a different style of interface the encapsulates the entire Spark serialized blob of bloom filter data. It will probably render https://github.com/NVIDIA/spark-rapids-jni/pull/1269 obsolete.

Added benchmark for bloom_filter_put. On an A5000, we're getting 140 GB/s write-throughput for bloom filter sizes of 512k, 1MB, 2MB, 4MB and 8MB. 12.5 milliseconds for 150 million rows. So it's not lightning fast, but it's serviceable.

Also fixed several assorted benchmark build errors. The cudf push for always providing null counts and specifying stream/mr broke a few of them.


src/main/java/ai/rapids/cudf/CudfAccessor.java
```diff
@@ -0,0 +1,27 @@
...
+package ai.rapids.cudf;
+
+public class CudfAccessor {
+  public static long getScalarHandle(Scalar s) {
+    return s.getScalarHandle();
+  }
+
+  public static Scalar scalarFromHandle(DType type, long scalarHandle) {
+    return new Scalar(type, scalarHandle);
+  }
```
jlowe, line 26: Minimally we need to file a cudf issue to open up the Scalar API to allow doing this directly rather than via package hijacking.  Bonus points for linking the cudf issue here as TODO comments to use the cudf API when accessible.

jlowe (approved):
Tested with NVIDIA/spark-rapids#8775 as well as in-progress BloomFilterAggregate code.  Minor nit on better comments for CudfAccessor usage.

jlowe (approved):

jlowe:
build

PR #1314, opened 2023-08-03
Add JNI backend for Spark SQL function `conv` for (hexa)decimals

gerashegalov:
Contributes to NVIDIA/spark-rapids#8511

POC supporting form/to  radices 10 and 16 leveraging existing libcudf API 

### DIFF
--- a/src/main/cpp/src/CastStringJni.cpp
+++ b/src/main/cpp/src/CastStringJni.cpp
@@ -117,7 +117,7 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromDecimal
   CATCH_CAST_EXCEPTION(env, 0);
 }
 
-JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_toIntegerUsingBase(
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_toIntegersWithBase(
   JNIEnv* env, jclass, jlong input_column, jint base)
 {
   JNI_NULL_CHECK(env, input_column, "input column is null", 0);
@@ -148,7 +148,7 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_toIntegerUs
   CATCH_CAST_EXCEPTION(env, 0);
 }
 
-JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromIntegerUsingBase(
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromIntegersWithBase(
   JNIEnv* env, jclass, jlong input_column, jint base)
 {
   JNI_NULL_CHECK(env, input_column, "input column is null", 0);
@@ -166,7 +166,8 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromInteger
           return cudf::strings::integers_to_hex(input_view);
         }
         default: {
-          return std::unique_ptr<cudf::column>(nullptr);  // TODO all zeros
+          auto const error_msg = "Bases supported 10, 16; Actual: " + std::to_string(base);
+          throw spark_rapids_jni::cast_error(0, error_msg);
         }
       }
     }();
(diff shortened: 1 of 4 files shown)


src/main/cpp/src/CastStringJni.cpp
```diff
@@ -111,4 +112,46 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromDecimal
...
+    auto const uint64_cv = [&] {
+      switch (fromRadix) {
+        case 10: {
+          return spark_rapids_jni::string_to_integer(cudf::data_type(cudf::type_id::UINT64),
+                                                     cv,
+                                                     JNI_FALSE,
+                                                     JNI_TRUE,
+                                                     cudf::get_default_stream());
+        } break;
+        case 16: {
```
revans2, line 134: What happens if there is a different case? Does C++ throw an excetpion?

revans2 (commented):
Generally it looks fine. Just a little curious about an error case.

gerashegalov:
build

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -111,4 +118,72 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromDecimal
...
+    auto const input_view{*reinterpret_cast<column_view const*>(input_column)};
+    const strings_column_view input_strings{input_view};
+    auto integer_view = [&] {
+      switch (base) {
+        case 10: {
+          // TODO implement it in the kernel
+          auto const regex = strings::regex_program::create(R"(^\s*([0-9]+).*)");
+          auto const dec_str_table = strings::extract(input_strings, *regex);
+          const strings_column_view dec_str_view{dec_str_table->get_column(0)};
+          return strings::to_integers(dec_str_view, res_data_type);
```
revans2, line 140: It doesn't look like we have any overflow checking in here, but Spark does do overflow checking, and just returns a -1 in non-ansi mode. Is the something that will be handled by the plugin, or are we lucky and happen to get the right answer out of CUDF.

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -111,4 +118,72 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromDecimal
...
+    jni::auto_set_device(env);
+    auto const res_data_type = data_type(type_id::UINT64);
+
+    auto const input_view{*reinterpret_cast<column_view const*>(input_column)};
+    const strings_column_view input_strings{input_view};
+    auto integer_view = [&] {
+      switch (base) {
+        case 10: {
+          // TODO implement it in the kernel
+          auto const regex = strings::regex_program::create(R"(^\s*([0-9]+).*)");
```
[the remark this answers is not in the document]
gerashegalov: interestingly I started out with strip but dropped because of the hex_to_integers behavior but it might indeed work with the decimal str to_integers.

```Python
>>> import cudf
>>> gsr = cudf.Series('5g5')
>>> cudf._lib.string_casting.htoi(gsr._column)
<cudf.core.column.numerical.NumericalColumn object at 0x7f6dc54d3ec0>
[
  85
]
dtype: int64
>>> cudf._lib.string_casting.stoi(gsr._column)
<cudf.core.column.numerical.NumericalColumn object at 0x7f6dc54d3f40>
[
  5
]
dtype: int32
>>> 
```

src/main/cpp/src/CastStringJni.cpp
gerashegalov: At this point there is no overflow checking indeed. We currently support only the [happy path](https://github.com/NVIDIA/spark-rapids/pull/8925/files#diff-af93e5faff863709f2c9de84607d8bd79c6cc0e2454b6f982c693c10ad721402R784-R803) avoiding the overflow

revans2 (dismissed):
My only concern now is performance. I just want to be sure that the cost of the regular expressions/etc beats the CPU. But that can even wait for the plugin changes if needed, because that is where I actually care about the performance.

gerashegalov:
build

gerashegalov:
build

gerashegalov:
build

gerashegalov:
@revans2 PTAL, it should be ready since it passes https://github.com/NVIDIA/spark-rapids/pull/8925/files#diff-af93e5faff863709f2c9de84607d8bd79c6cc0e2454b6f982c693c10ad721402R784-R795

revans2 (approved):

Issue #1221, opened 2023-06-20
Run compute-sanitizer memcheck on tests


jlowe:
We should leverage compute-sanitizer in memcheck mode in our unit tests to help catch any kernels that may be doing something incorrectly.  Ideally the compute-sanitizer would run with no memory pooling, as memory allocation pooling can hide some invalid memory access patterns from the compute sanitizer.

firestarman:
Seems unit tests expecting errors/exceptions are not suitable to run with the sanitizer, e.g. `CudaFatalTest`. Because saniziter will report errors for this kind of tests but the errors are expected.

We are likely to split these tests away and enable sanitizer only for the tests not test exceptions.

@jlowe What's your idea on this?

firestarman:
@jlowe 
Currently we have two options for testing, with `CuFileTest` and without `CuFileTest` (a profile). If adding the sanitizer support by new maven profiles, we will be likely to have 4 options in total (two new options by maven profiles: sanitizer+CuFileTest, sanitizer+"no CuFileTest"), which seems a little complicated.

Shall we always enable the sanitizer for both of the current two cases, instead of using the maven profile for simplicity ?

jlowe:
I'm not sure it's really that complicated, especially if we work through how to properly inherit between the profiles such that one profile simply sets up the `jvm` tag in surefire to run the sanitizer and the other profile simply sets the tests to run.  Then you can independently enable surefire from which tests to run by simply specifying the corresponding profiles.

If we can't figure out how to merge the surefire settings between the profiles then I'm fine with hardcoding the sanitizer for now.

firestarman:
>especially if we work through how to properly inherit between the profiles such that one profile simply sets up the jvm tag in surefire to run the sanitizer and the other profile simply sets the tests to run. 

Good suggestion, I figoured out a way to do this. 
Could your help review https://github.com/NVIDIA/spark-rapids-jni/pull/1321 again ? 
And it now depends on a cudf PR https://github.com/rapidsai/cudf/pull/13872.

firestarman:
Please note it will take much more time to run the tests with the compute sanitizer. There are some numbers for some tests that the elapsed time became quite long.

|Test|No Sanitizer|With Sanitizer|
|--|--|--|
| Aggregation128UtilsTest| 2.54 s| 8.281 s |
| BinaryOpTest | 0.168 s  | 3.36 s |
| ColumnVectorTest  | 0.912 s | 21.192 s |
[...]

PR #1321, opened 2023-08-08
Enable memcheck for jni unit tests

firestarman:
closes https://github.com/NVIDIA/spark-rapids-jni/issues/1221

[compute-saniziter](https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html#using-compute-sanitizer) is a tool can detect some GPU memory relevant issues, this PR is to enable it in memory check mode for the JNI unit tests. 

Some explanation for the parameters of the Compute Sanitizer.
`--log-file` should be used to avoid a corrupting output issue from the surefire plugin.
`--error-exitcode` is used to fail the build process if any error is caught by the Compute Sanitizer. 
`--launch-timeout` is set to 10 minutes, and it should be enough since we monitor only the forked test processes.

Please note it will take much more time to run the tests with the compute sanitizer. There are some numbers for some tests that the elapsed time became quite long.

|Test|No Sanitizer|With Sanitizer|
|--|--|--|
| Aggregation128UtilsTest| 2.54 s| 8.281 s |
| BinaryOpTest | 0.168 s  | 3.36 s |
| ColumnVectorTest  | 0.912 s | 21.192 s |
| ReductionTest |0.104 s | 4.757 s|
|TableTest| 2.385 s  | 124.59 s|
| ColumnViewNonEmptyNullsTest | 2.615 s | 8.385 s |

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -227,13 +227,16 @@ arguments to get into an interactive shell inside the container.
 
 #### Testing with Compute Sanitizer
 [Compute Sanitizer](https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html) is a
-functional correctness checking suite included in the CUDA toolkit. And RAPIDS Accelerator JNI now
-supports leveraging Compute Sanitizer in memcheck mode in the unit tests to help catch any kernels
-that may be doing something incorrectly. To run the unit tests with Compute Sanitizer, append the
+functional correctness checking suite included in the CUDA toolkit. And the RAPIDS Accelerator JNI now
+supports leveraging the Compute Sanitizer in memcheck mode in the unit tests to help catch any kernels
+that may be doing something incorrectly. To run the unit tests with the Compute Sanitizer, append the
 `-DUSE_SANITIZER=ON` to the build command. e.g.
 ```
 >  ./build/build-in-docker clean package -DUSE_SANITIZER=ON
 ```
+
+The Compute Sanitizer will output its report into one or multiple log files named as
+`sanitizer_for_pid_<pid number>.log` under the current workspace root path.
 
 Please note not all the unit tests can run with Compute Sanitizer. For example, `RmmTest#testEventHandler`,
 a problematic test, intentionally tries an illegal allocation because of a too big size as part of the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -225,20 +225,20 @@ situation. A test can be run directly using this script or the script can be run
 arguments to get into an interactive shell inside the container.
 ```build/run-in-docker target/cmake-build/gtests/ROW_CONVERSION```
 
-#### Testing with Compute Sanitier
+#### Testing with Compute Sanitizer
 [Compute Sanitizer](https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html) is a
 functional correctness checking suite included in the CUDA toolkit. And RAPIDS Accelerator JNI now
 supports leveraging Compute Sanitizer in memcheck mode in the unit tests to help catch any kernels
-that may be doing something incorrectly. To run the unit tests with Compute Sanitier, append the
+that may be doing something incorrectly. To run the unit tests with Compute Sanitizer, append the
 `-DUSE_SANITIZER=ON` to the build command. e.g.
 ```
 >  ./build/build-in-docker clean package -DUSE_SANITIZER=ON
 ```
 
-Please note not all the unit tests can run with Compute Sanitier. For example, `RmmTest#testEventHandler`,
+Please note not all the unit tests can run with Compute Sanitizer. For example, `RmmTest#testEventHandler`,
 a problematic test, intentionally tries an illegal allocation because of a too big size as part of the
 test, but Compute Sanitier will still report the errors and fail the whole build process. And
-`UnsafeMemoryAccessorTest` is for host memory only, so no need to run with Compute Sanitier either.
+`UnsafeMemoryAccessorTest` is for host memory only, so there is no need to run it with Compute Sanitier either.
 
 If you think your tests are not suitable for Compute Sanitier, please add the JUnit5 tag (`@Tag("noSanitizer")`)
 to the tests or the test class.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -224,6 +224,36 @@ in errors finding libraries. The script `build/run-in-docker` was created to hel
 situation. A test can be run directly using this script or the script can be run without any
 arguments to get into an interactive shell inside the container.
 ```build/run-in-docker target/cmake-build/gtests/ROW_CONVERSION```
+
+#### Testing with Compute Sanitier
+[Compute Sanitizer](https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html) is a
+functional correctness checking suite included in the CUDA toolkit. And RAPIDS Accelerator JNI now
+supports leveraging Compute Sanitizer in memcheck mode in the unit tests to help catch any kernels
+that may be doing something incorrectly. To run the unit tests with Compute Sanitier, append the
+`-DUSE_SANITIZER=ON` to the build command. e.g.
+```
+>  ./build/build-in-docker clean package -DUSE_SANITIZER=ON
+```
+
+Please note not all the unit tests can run with Compute Sanitier. For example, `RmmTest#testEventHandler`,
+a problematic test, intentionally tries an illegal allocation because of a too big size as part of the
+test, but Compute Sanitier will still report the errors and fail the whole build process. And
+`UnsafeMemoryAccessorTest` is for host memory only, so no need to run with Compute Sanitier either.
+
+If you think your tests are not suitable for Compute Sanitier, please add the JUnit5 tag (`@Tag("noSanitizer")`)
+to the tests or the test class.
+```
+@Tag("noSanitizer")
+class ExceptionCaseTest { ... }
+
+# or for a single test
+class NormalCaseTest {
+
+  @Tag("noSanitizer")
+  public void testOneErrorCase(){ ... }
+}
+```
+
 ### Benchmarks
 Benchmarks exist for c++ benchmarks using NVBench and are in the `src/main/cpp/benchmarks` directory.
 To build these benchmarks requires the `-DBUILD_BENCHMARKS` build option. Once built, the benchmarks
--- a/build/sanitizer-java/bin/java
+++ b/build/sanitizer-java/bin/java
@@ -1,0 +1,5 @@
+compute-sanitizer --tool memcheck \
+    --launch-timeout 600 \
+    --error-exitcode -2 \
+    --log-file "./sanitizer_for_pid_%p.log" \
+    java "$@"
--- a/build/sanitizer-java/bin/java
+++ b/build/sanitizer-java/bin/java
@@ -1,4 +1,24 @@
-compute-sanitizer --tool memcheck \
+#!/bin/bash
+#
+# Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


ci/nightly-build.sh
```diff
@@ -26,7 +26,8 @@ MVN="mvn -Dmaven.wagon.http.retryHandler.count=3 -B"
 CUDA_VER=${CUDA_VER:-cuda`nvcc --version | sed -n 's/^.*release \([0-9]\+\)\..*$/\1/p'`}
 PARALLEL_LEVEL=${PARALLEL_LEVEL:-4}
 USE_GDS=${USE_GDS:-ON}
-${MVN} clean package ${MVN_MIRROR}  \
+SANITIZER_CMD="compute-sanitizer --tool memcheck --target-processes all $SANITIZER_OPTS"
```
pxLi, line 29: Q: 
So should all cuda runtimes have included the compute-sanitizer tooling? do we need to do some prerequisites to make sure the ENV has required pkgs for test run?

firestarman: I think yes.
Checked cuda 11.5 and 11.8, both have this tool.

pxLi:
some questions about the tooling, 
How should we expose the detected issues? Will the script error out if detect any issue like a memory leak? 
If yes, do we need to fix them all first before merging to make sure our regular CI won't be blocked instantly after enabling this tool? 
If not, how should we notify developers about the issues? 

thanks

firestarman:
>How should we expose the detected issues? Will the script error out if detect any issue like a memory leak?

You can specify where Compute Sanitizer will write all of its text output to by setting the `--log-file` option. By default, Compute Sanitizer will print all output to stdout. 
Per my local tests, This tool only records the errors into log, does not break the running process.

>If yes, do we need to fix them all first before merging to make sure our regular CI won't be blocked instantly after enabling this tool?
If not, how should we notify developers about the issues?

@GaryShen2008 may know what we should do.

ci/nightly-build.sh
jlowe, line 29: In the past we've needed to specify a hefty `--launch-timeout` to prevent compute-sanitizer from giving up on instrumenting a process that doesn't use the GPU shortly after startup.  That would certainly be the case here, as it could take close to an hour to complete building.  cc: @abellina who has more experience with this.

It would be nice to not have compute-sanitizer running for the main compiliation part of the build -- it's just a waste.  A rebuild without cleaning is pretty fast, and we could skip the Java tests on the first "clean package" build, but I don't know if there's a way to skip running the C++ tests on the first build and only on the subsequent test build.

abellina: +1 on not having this on during any part of compilation. It should only be used when running the tests. I also wouldn't run all the tests as it is likely to take a very long time.

Yes the `--launch-timeout` can be really high and it was needed to wait for that first cuda call, it is likely a requirement in every scenario (the default is 60 seconds for this case).

I do not know of a good way to have maven surefire replace JAVA_HOME or wrap its test process with this (there is also attach mode, but that seems complicated to script). I also think that we don't want to run this for all tests because it will be very slow, and it will alter timings that could hide other synchronization issues. 

I think we may need to curate the tests and run this only for certain things with a pytest mark. I would skip scala tests and do it all in the integration tests, also the integration tests should make this a lot easier to be opt in and configurable. Thoughts?

jlowe: > I think we may need to curate the tests and run this only for certain things with a pytest mark. I would skip scala tests and do it all in the integration tests, also the integration tests should make this a lot easier to be opt in and configurable. Thoughts?

This is spark-rapids-jni, there's no Python tests.  I think there's a small enough number of tests (and they're quick enough) to warrant running them on all the tests.  However it would be good to know how much slower this makes the test run.

firestarman:
Move to draft, the requirement is quite different than I thought.

firestarman:
> > > Per my local tests, This tool only records the errors into log, without breaking the running process.
> > 
> > 
> > Don't we want the premerge to fail if it detects an error? If it just records into a log on the side without failing the permerge or nightly build, nobody is going to think to check the log.
> 
> We could try `error-exitcode` set to 1 or something that isn't 0. It looks like if the application fails, compute-sanitizer echoes its error code. The error code in this parameter is for the case where the app succeeds (exit code 0), but compute-sanitizer detected an issue. Unfortunately, I do not see a "kill on first failure" option. I am a little afraid about that here since the CI tests run for a long time.

Thx and yes, this option works, setting it to -1 can let sanitizer fail if it detects any error even the application succeeds.

ci/nightly-build.sh
firestarman: Is it possible to use the [toolchain](https://maven.apache.org/surefire/maven-surefire-plugin/java9.html#head3) to replace the JAVA_HOME

firestarman:
Depends on https://github.com/rapidsai/cudf/pull/13872

firestarman:
Two follow-ups:
https://github.com/NVIDIA/spark-rapids-jni/issues/1336
https://github.com/NVIDIA/spark-rapids-jni/issues/1337

firestarman:
Plead do not triggre premerge until https://github.com/rapidsai/cudf/pull/13872 is merged.

CONTRIBUTING.md
```diff
@@ -224,6 +224,39 @@ in errors finding libraries. The script `build/run-in-docker` was created to hel
 situation. A test can be run directly using this script or the script can be run without any
 arguments to get into an interactive shell inside the container.
 ```build/run-in-docker target/cmake-build/gtests/ROW_CONVERSION```
+
+#### Testing with Compute Sanitizer
+[Compute Sanitizer](https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html) is a
+functional correctness checking suite included in the CUDA toolkit. And the RAPIDS Accelerator JNI now
```
jlowe, line 230:
```suggestion
functional correctness checking suite included in the CUDA toolkit. The RAPIDS Accelerator JNI
```

CONTRIBUTING.md
```diff
@@ -224,6 +224,39 @@ in errors finding libraries. The script `build/run-in-docker` was created to hel
...
+```
+>  ./build/build-in-docker clean package -DUSE_SANITIZER=ON
+```
+
+The Compute Sanitizer will output its report into one or multiple log files named as
+`sanitizer_for_pid_<pid number>.log` under the current workspace root path.
+
+Please note not all the unit tests can run with Compute Sanitizer. For example, `RmmTest#testEventHandler`,
+a problematic test, intentionally tries an illegal allocation because of a too big size as part of the
+test, but Compute Sanitier will still report the errors and fail the whole build process. And
```
jlowe, line 243:
```suggestion
test, but Compute Sanitizer will still report the errors and fail the whole build process.
```

CONTRIBUTING.md
```diff
@@ -224,6 +224,39 @@ in errors finding libraries. The script `build/run-in-docker` was created to hel
...
+`UnsafeMemoryAccessorTest` is for host memory only, so there is no need to run it with Compute Sanitier either.
```
jlowe, line 244:
```suggestion
`UnsafeMemoryAccessorTest` is for host memory only, so there is no need to run it with
Compute Sanitizer either.
```

CONTRIBUTING.md
```diff
@@ -224,6 +224,39 @@ in errors finding libraries. The script `build/run-in-docker` was created to hel
...
+
+If you think your tests are not suitable for Compute Sanitier, please add the JUnit5 tag (`@Tag("noSanitizer")`)
```
jlowe, line 246:
```suggestion
If you think your tests are not suitable for Compute Sanitizer, please add the JUnit5 tag (`@Tag("noSanitizer")`)
```

ci/premerge-build.sh
```diff
@@ -27,4 +27,4 @@ ${MVN} verify ${MVN_MIRROR} \
   -DCPP_PARALLEL_LEVEL=${PARALLEL_LEVEL} \
   -Dlibcudf.build.configure=true \
   -DUSE_GDS=ON -Dtest=*,!CuFileTest,!CudaFatalTest,!ColumnViewNonEmptyNullsTest \
-  -DBUILD_TESTS=ON
+  -DBUILD_TESTS=ON -DUSE_SANITIZER=ON
```
jlowe, line 30:
```suggestion
  -DBUILD_TESTS=ON \
  -DUSE_SANITIZER=ON
```

ci/premerge-build.sh
jlowe, line 1: 2023 copyrights

build/sanitizer-java/bin/java
jlowe, line 1: NVIDIA copyrights, ASLv2 license

jlowe, line 1: Curious about the '/bin/java' when we specify the full path to the jvm in the surefire-plugin.  Any reason this script can't simply be called `build/sanitizer-java`?  There should also be a comment towards the top of the file explaining why this file exists.

build/sanitizer-java/bin/java
```diff
@@ -0,0 +1,5 @@
+compute-sanitizer --tool memcheck \
```
jlowe, line 1:
```suggestion
#!/bin/sh
exec compute-sanitizer --tool memcheck \
```

pom.xml
```diff
@@ -199,14 +205,51 @@
...
+                </configuration>
+              </execution>
               <execution>
                 <id>non-empty-null-test</id>
                 <goals>
                   <goal>test</goal>
                 </goals>
                 <configuration>
-                  <argLine>-da:ai.rapids.cudf.AssertEmptyNulls</argLine>
-                  <test>ColumnViewNonEmptyNullsTest</test>
```
jlowe, line 209: Why are these settings removed -- are we still running this test anywhere?

pxLi:
@jlowe @abellina Liangcai helps record the test duration w/ sanitizer in [PR desc ](https://github.com/NVIDIA/spark-rapids-jni/pull/1321#issue-1840628246)
Currently, JNI does not have too many cases so the total time may not increase too much. But in the future after we add more cases, the significantly increased duration may affect regular CI and increase costs, do you think we are okay to add it to nightly only but not pre-merge CI?

build/sanitizer-java/bin/java
firestarman: It complains the following error if calling the script by `build/sanitizer-java`.
```
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.0:test (default-test) on project spark-rapids-jni: Given path does not end with java executor "/home/liangcail/work/projects/on_github/spark-rapids-jni/build/sanitizer-java". -> [Help 1]
```

pom.xml
firestarman: IIUC, this is a duplicate of the definition @ [line554](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-23.10/pom.xml#L554) .

build/sanitizer-java/bin/java
firestarman: Done

ci/premerge-build.sh
firestarman: done

firestarman:
build

jlowe:
@pxLi note that we should also enable the sanitizer builds on the cudf submodule sync job as well, since we don't want to pull in a cudf change that the sanitizer thinks is broken.

@firestarman this is running the sanitizer on the Java unit test, but I assume it is not running it on the spark-rapids-jni C++ unit tests, correct?  If so that can be a followup, since we should be running the sanitizer on those as well.

firestarman:
>@firestarman this is running the sanitizer on the Java unit test, but I assume it is not running it on the spark-rapids-jni C++ unit tests, correct? If so that can be a followup, since we should be running the sanitizer on those as well.

Yeah, this is only for the Java unit tests. BTW, Where can I find the the spark-rapids-jni C++ unit tests ? There is only a json file under https://github.com/NVIDIA/spark-rapids-jni/tree/branch-23.10/src/test/cpp/ .

pxLi (approved):
I will follow up on the update to submodule syncs-up CI

firestarman:
So I am going to merge this as-is and file follow up a PR.

firestarman:
>I will follow up on the update to submodule syncs-up CI

Here it is https://github.com/NVIDIA/spark-rapids-jni/pull/1347

jlowe:
> Where can I find the the spark-rapids-jni C++ unit tests ?

Under [src/main/cpp/tests](https://github.com/NVIDIA/spark-rapids-jni/tree/branch-23.10/src/main/cpp/tests).

firestarman:
Thx for filing https://github.com/NVIDIA/spark-rapids-jni/issues/1353

Issue #1368, opened 2023-08-24
[BUG] cufaultinj build failed in ARM64 environment


pxLi:
**Describe the bug**

```
[2023-08-24T02:00:18.087Z] [INFO]      [exec] -- Found the ArrowDataset static library: /home/jenkins/agent/workspace/peixinl-spark-rapids-jni/target/libcudf-install/lib64/libarrow_dataset.a

[2023-08-24T02:00:18.087Z] [INFO]      [exec] -- Configuring done (10.0s)

[2023-08-24T02:00:18.087Z] [INFO]      [exec] -- Generating done (0.0s)

[2023-08-24T02:00:18.087Z] [INFO]      [exec] CMake Error at faultinj/CMakeLists.txt:34 (target_link_libraries):

[2023-08-24T02:00:18.087Z] [INFO]      [exec]   Target "cufaultinj" links to:

[2023-08-24T02:00:18.087Z] [INFO]      [exec] 

[2023-08-24T02:00:18.087Z] [INFO]      [exec]     CUDA::cupti_static

[2023-08-24T02:00:18.087Z] [INFO]      [exec] 

[2023-08-24T02:00:18.087Z] [INFO]      [exec]   but the target was not found.  Possible reasons include:

[2023-08-24T02:00:18.087Z] [INFO]      [exec] 

[2023-08-24T02:00:18.087Z] [INFO]      [exec]     * There is a typo in the target name.

[2023-08-24T02:00:18.087Z] [INFO]      [exec]     * A find_package call is missing for an IMPORTED target.

[2023-08-24T02:00:18.087Z] [INFO]      [exec]     * An ALIAS target is missing.
```

**Steps/Code to reproduce bug**
build JNI in arm64 ENV

**Expected behavior**
Pass the faultinj build, or decide if we could skip faultinj build in arm64

PR #1385, opened 2023-08-30
Add arm64 build support

pxLi:
This PR is still in progress but want get more feedbacks in advance.

To re-enable build support for arm,
1. Build on internal ARM instance in blossom cluster
2. Provide artifacts w/ classifiers like cuda11-arm64, cuda12-arm64
3. Use dockerhub/cuda rockylinux image for arm and update corresponding tools to support the build and test (since there is no official arm support for centos7)
4. Disabled faultinj build https://github.com/NVIDIA/spark-rapids-jni/issues/1368
5. Would not run sanitizer check as it failed core dump in arm ENV (currently do not have time to check the details) https://github.com/NVIDIA/spark-rapids-jni/issues/1388

In this case, our plugin jar build would also need to be updated to support release w/ <cuda>-arm64 classifiers.
So the final releases would be like,
```
Default jar (maven required):
rapids-4-spark_2.12-<version>.jar (x86 cuda11 one)
X86 jars:
rapids-4-spark_2.12-<version>-cuda11.jar 
rapids-4-spark_2.12-<version>-cuda12.jar 
ARM64 jars:
rapids-4-spark_2.12-<version>-cuda11-arm64.jar 
rapids-4-spark_2.12-<version>-cuda12-arm64.jar 
```

### DIFF
--- a/ci/Dockerfile.multi
+++ b/ci/Dockerfile.multi
@@ -1,0 +1,76 @@
+#
+# Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+###
+# JNI CI image for multi-platform build
+#
+# Arguments: CUDA_VERSION=11.8.0
+#
+###
+ARG CUDA_VERSION=11.8.0
+ARG OS_RELEASE=8
+# multi-platform build with: docker buildx build --platform linux/arm64,linux/amd64 <ARGS> on either amd64 or arm64 host
+# check available offcial arm-based docker images at https://hub.docker.com/r/nvidia/cuda/tags (OS/ARCH)
+FROM --platform=$TARGETPLATFORM nvidia/cuda:$CUDA_VERSION-devel-rockylinux$OS_RELEASE
+ARG TOOLSET_VERSION=11
+### Install basic requirements
+RUN dnf install -y scl-utils
+RUN dnf install -y gcc-toolset-${TOOLSET_VERSION} python39
+RUN dnf --enablerepo=powertools install -y zlib-devel maven tar wget patch ninja-build
+# require git 2.18+ to keep consistent submodule operations
+RUN dnf install -y git
+## pre-create the CMAKE_INSTALL_PREFIX folder, set writable by any user for Jenkins
+RUN mkdir /usr/local/rapids && mkdir /rapids && chmod 777 /usr/local/rapids && chmod 777 /rapids
+
+# 3.22.3+: CUDA architecture 'native' support + flexible CMAKE_<LANG>_*_LAUNCHER for ccache
+ARG CMAKE_VERSION=3.26.4
+# default as arm64 release
+ARG CMAKE_ARCH=aarch64
+# aarch64 cmake for arm build
+RUN cd /usr/local && wget --quiet https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-linux-${CMAKE_ARCH}.tar.gz && \
+   tar zxf cmake-${CMAKE_VERSION}-linux-${CMAKE_ARCH}.tar.gz && \
+   rm cmake-${CMAKE_VERSION}-linux-${CMAKE_ARCH}.tar.gz
+ENV PATH /usr/local/cmake-${CMAKE_VERSION}-linux-${CMAKE_ARCH}/bin:$PATH
+
+# ccache for interactive builds
+ARG CCACHE_VERSION=4.6
+RUN cd /tmp && wget --quiet https://github.com/ccache/ccache/releases/download/v${CCACHE_VERSION}/ccache-${CCACHE_VERSION}.tar.gz && \
+   tar zxf ccache-${CCACHE_VERSION}.tar.gz && \
+   rm ccache-${CCACHE_VERSION}.tar.gz && \
+   cd ccache-${CCACHE_VERSION} && \
+   mkdir build && \
+   cd build && \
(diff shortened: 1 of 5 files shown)


pxLi:
@sameerz @GaryShen2008 @jlowe @revans2 @tgravescs @gerashegalov 
Please help review and let me know if we are good to continue the current idea, thanks!

Last time (a year ago) we also discussed having another option that would provide different <cpu_arch> based artifacts instead of classifiers like
```
rapids-4-spark_2.12-<version>.jar
or/and
rapids-4-spark_2.12_amd64-<version>.jar
and
rapids-4-spark_2.12_aarch64-<version>.jar
```
this may require more refactoring works to our existing project structure and CICD for both JNI and plugin repos.

pxLi:
build

ci/Dockerfile.multi
```diff
@@ -0,0 +1,76 @@
...
+# JNI CI image for multi-platform build
+#
+# Arguments: CUDA_VERSION=11.8.0
+#
+###
+ARG CUDA_VERSION=11.8.0
+ARG OS_RELEASE=8
+# multi-platform build with: docker buildx build --platform linux/arm64,linux/amd64 <ARGS> on either amd64 or arm64 host
+# check available offcial arm-based docker images at https://hub.docker.com/r/nvidia/cuda/tags (OS/ARCH)
+FROM --platform=$TARGETPLATFORM nvidia/cuda:$CUDA_VERSION-devel-rockylinux$OS_RELEASE
```
sameerz, line 27: I believe the GH systems will come with Ubuntu, can we build and test with Ubuntu 22.04?

pxLi: This would be more effort as we need to provide 2 toolsets to support developing on 2 kinds of OS distros (not only RHEL), or we are fine to just test the artifacts on ubuntu?

Both should be fine as we still have enough time to discuss and implement.
Currently I would like to know if the current classifier solution looks good, so I can continue the work, or I may require some help from developers before CICD updates, thanks ~

sameerz: Eventually we need to test on both Rocky9 and Ubuntu 22.04.  My request is to start with Ubuntu 22.04 if we can.

pxLi: > Eventually we need to test on both Rocky9 and Ubuntu 22.04. 

We did test on both Rocky9 and Ubuntu 22.04 for x86, you mean for arm we also need to cover all x86 scenarios?
sounds like we need fill much much more arm instances in our pool.

> My request is to start with Ubuntu 22.04 if we can.

Sure, I will prepare ubuntu ENV for arm build later

jlowe (commented):
Approach looks OK to me.  Ideally we would build an x86_64 for the non-ARM jar we have today, but I agree there's a lot of changes to CI and user education to accommodate that.  Since the "cuda11" classifier implies it runs on all supported processor architectures, I'm OK in the short term with that being equivalent to the cuda11-x86_64 classifier since the former is a superset of the latter, but eventually we should educate users here.

Maybe we should target deploying both the cuda11 and cuda11-x86_64 artifacts in the short term, which will be identical, and sstart to migrate existing users to the latter artifact.  Then we can, if desired, build an uber artifact in cuda11 that is all supported processors in one jar.

ci/Dockerfile.multi
jlowe: >> can we build and test with Ubuntu 22.04?

> Sure, I will prepare ubuntu ENV for arm build later

To ensure we're on the same page, the goal should not be to build on all supported OS's but instead build on _one_ OS and run that binary on all supported OS's.  That's why building on rocky makes sense to me, as that's closest to the build setup we had for x86_64 where we build once and run that everywhere.

ci/nightly-build.sh
```diff
@@ -26,10 +26,22 @@ MVN="mvn -Dmaven.wagon.http.retryHandler.count=3 -B"
 CUDA_VER=${CUDA_VER:-cuda`nvcc --version | sed -n 's/^.*release \([0-9]\+\)\..*$/\1/p'`}
 PARALLEL_LEVEL=${PARALLEL_LEVEL:-4}
 USE_GDS=${USE_GDS:-ON}
+USE_SANITIZER=${USE_SANITIZER:-ON}
+BUILD_FAULTINJ=${BUILD_FAULTINJ:-ON}
+ARM64=${ARM64:-false}
+
+profiles="source-javadoc"
+if [ "${ARM64}" == "true" ]; then
```
gerashegalov, line 34: I wonder if we should standardize on 'ON' 'OFF' for all Boolean switches.

ci/Dockerfile.multi
sameerz: Sorry, my misunderstanding.  Yes we can build on Rocky8 if that is the lowest common denominator, and then test on Ubuntu 22.04.

ci/nightly-build.sh
pxLi: Currently ON/OFF is for cmake config, 
the added `ARM64` ENV in CI script is for CI building only, it looks more clear (true/false) to me in the pipeline setup.
But if we decide to standardize all switches to follow cmake style, I am fine to update all.

pxLi:
> Approach looks OK to me. Ideally we would build an x86_64 for the non-ARM jar we have today, but I agree there's a lot of changes to CI and user education to accommodate that. Since the "cuda11" classifier implies it runs on all supported processor architectures, I'm OK in the short term with that being equivalent to the cuda11-x86_64 classifier since the former is a superset of the latter, but eventually we should educate users here.
> 
> Maybe we should target deploying both the cuda11 and cuda11-x86_64 artifacts in the short term, which will be identical, and sstart to migrate existing users to the latter artifact. Then we can, if desired, build an uber artifact in cuda11 that is all supported processors in one jar.

Thanks! For the short term I will just continue the work of classifiers, 
also filed https://github.com/NVIDIA/spark-rapids-jni/issues/1387 which may require help from developers

ci/Dockerfile.multi
pxLi: in default cuda11.8-ubuntu22 and cuda11.8-rocky8 arm64 images,

ubuntu22,
```
ldd --version
ldd (Ubuntu GLIBC 2.35-0ubuntu3.1) 2.35
```

rocky8,
```
ldd --version
ldd (GNU libc) 2.28
```

Lets keep rocky8 for building, and cover the tests in ubuntu22 ENV later

pxLi:
build

pxLi:
@jlowe can you help review current changes? 
Due to lack of arm resources, planning to merge and provide internal CI for developers fisrt, thanks

jlowe (approved):

PR #1445, opened 2023-09-26
Add 'dropLeadingZeros' parameter for integer to hexadecimal conversion

### LINKED ISSUE #1444: [FEA] Provide a way for converting integers col to hex string col without dropping the leading zeros.
Is your feature request related to a problem? Please describe.
We need a way to convert a string col to a hex string col without dropping the leading zeros. This is needed in cases when we just want to convert a binary col to a string col

Describe the solution you'd like
We already have a method in CastStrings.java that uses the `cudf:integers_to_hex` but drops the leading zeros. We just need to pass in a boolean to decide if we should drop the leading zeros or not

Describe alternatives you've considered
None

razajafri:
This PR adds a boolean parameter `dropLeadingZeros` to control whether we want to drop leading zeros while converting an integer column to a hex column. e.g 14 -> '0E' or 'E'

Testing: 
   Junit tests were added

Fixes #1444

### DIFF
--- a/src/main/cpp/src/CastStringJni.cpp
+++ b/src/main/cpp/src/CastStringJni.cpp
@@ -196,7 +196,7 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_toIntegersW
 }
 
 JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromIntegersWithBase(
-  JNIEnv* env, jclass, jlong input_column, jint base)
+  JNIEnv* env, jclass, jlong input_column, jint base, jboolean drop_leading_zeros)
 {
   JNI_NULL_CHECK(env, input_column, "input column is null", 0);
   using namespace cudf;
@@ -210,9 +210,13 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromInteger
         } break;
         case 16: {
           auto pre_res                = strings::integers_to_hex(input_view);
-          auto const regex            = strings::regex_program::create("^0?([0-9a-fA-F]+)$");
-          auto const wo_leading_zeros = strings::extract(strings_column_view(*pre_res), *regex);
-          return std::move(wo_leading_zeros->release()[0]);
+          if (drop_leading_zeros) {
+            auto const regex            = strings::regex_program::create("^0?([0-9a-fA-F]+)$");
+            auto const wo_leading_zeros = strings::extract(strings_column_view(*pre_res), *regex);
+            return std::move(wo_leading_zeros->release()[0]);
+          } else {
+            return pre_res;
+          }
         }
         default: {
           auto const error_msg = "Bases supported 10, 16; Actual: " + std::to_string(base);
--- a/src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
@@ -109,8 +109,9 @@ public class CastStrings {
       type.getTypeId().getNativeId()));
   }
 
-  public static ColumnVector fromIntegersWithBase(ColumnView cv, int base) {
-    return new ColumnVector(fromIntegersWithBase(cv.getNativeView(), base));
+  public static ColumnVector fromIntegersWithBase(ColumnView cv, int base,
+    boolean dropLeadingZeros) {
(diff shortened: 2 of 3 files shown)


razajafri:
build

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -193,18 +193,71 @@ void castToDecimalNoStripTest() {
     }
   }
 
-  private void convTestInternal(Table input, Table expected, int fromBase) {
+  private void baseConversionHelper(Table input, Table expected, int fromBase,
+    boolean dropLeadingZeros) {
```
mythrocks, line 197: Should we consider a more descriptive name for this helper?  This seems to be testing the round-trip from base-10 to both base-10 and base-16.

Perhaps `testRoundTripToBase10And16`?

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -193,18 +193,71 @@ void castToDecimalNoStripTest() {
...
+            "2621",
+            "50"
+        ).column(
+            "E",
+            "A3D",
+            "32"
+        ).build()
+    )
+    {
+      baseConversionHelper(input, expected, 10, true);
```
mythrocks, line 233: Should we consider also testing with a different value for `fromBase`? Say, `16`?

mythrocks:
LGTM: A couple of minor suggestions.

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -109,8 +109,9 @@ public static ColumnVector toIntegersWithBase(ColumnView cv, int base,
       type.getTypeId().getNativeId()));
   }
 
-  public static ColumnVector fromIntegersWithBase(ColumnView cv, int base) {
-    return new ColumnVector(fromIntegersWithBase(cv.getNativeView(), base));
+  public static ColumnVector fromIntegersWithBase(ColumnView cv, int base,
```
revans2, line 112: We need to maintain backwards compatibility. We need to keep a version that does not take the boolean parameter and defaults it to true. And while you are at it we should have some comments on this to explain what it is doing so the parameters are clear.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
razajafri: Testing that will not test the change and there is already an existing test for it [here](https://github.com/NVIDIA/spark-rapids-jni/blob/15b62685d1d965976644436507c54d4928652de6/src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java#L328)

Does that work?

razajafri:
@revans2 @mythrocks thanks for the review. PTAL I think I have addressed your concerns

revans2 (approved):

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
mythrocks: Ah, yes.  Sorry I missed that.

mythrocks (approved):

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -196,7 +196,7 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_toIntegersW
 }
 
 JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromIntegersWithBase(
-  JNIEnv* env, jclass, jlong input_column, jint base)
+  JNIEnv* env, jclass, jlong input_column, jint base, jboolean drop_leading_zeros)
```
gerashegalov, line 199: It looks odd that the new parameter applies only to base=16 but we pass it in the general API for all bases. I would make an extra method just for wiring a raw strings::integers_to_hex(input_view) to JNI

razajafri: That API should probably live in cudf repo if it's just a wrapper around `integers_to_hex`.

razajafri:
I am about to update this PR to remove the API change. The only other option is to just close it. I do think the documentation is good to have.

Thoughts?

razajafri:
I am closing this PR as it makes more sense to create this in cudf

PR #1475, opened 2023-10-04
Fix ParquetFooter parsing of legacy array-of-struct format

jlowe:
While testing with cudf and RAPIDS Accelerator fixes to allow proper parsing of the [parquet-testing repeated_no_annotation.parquet](https://github.com/apache/parquet-testing/blob/master/data/repeated_no_annotation.parquet) file, I discovered ParquetFooter had issues parsing the Parquet schema of this file.  The comments enumerating the various ways an array can appear state that a list could contain more than one child directly, but the code did not handle that case.

This PR updates the native Parquet parsing logic to handle the case where a list contains more than one child.  Verified this helps fix the RAPIDS Accelerator parsing of the problematic repeated_no_annotation.parquet file.


jlowe:
build

revans2 (approved):

PR #1477, opened 2023-10-05
handle rmm::out_of_memory instead of std::bad_alloc for retry

revans2:
This fixes https://github.com/NVIDIA/spark-rapids/issues/9364

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -1352,7 +1352,10 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
         void* ret = resource->allocate(num_bytes, stream);
         post_alloc_success(tid, likely_spill);
         return ret;
-      } catch (const std::bad_alloc& e) {
+      } catch (const rmm::out_of_memory& e) {
+        // rmm::out_of_memory is what is thrown when an allocation failed
+        // but there are other rmm::bad_alloc exceptions that could be
+        // thrown as well, which are handled by the std::exception case.
         if (!post_alloc_failed(tid, true, likely_spill)) { throw; }
       } catch (const std::exception& e) {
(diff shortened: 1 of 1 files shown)


revans2:
build

revans2:
The previous build timed out so going to retry it.

revans2:
build

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1352,15 +1352,18 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
         void* ret = resource->allocate(num_bytes, stream);
         post_alloc_success(tid, likely_spill);
         return ret;
-      } catch (const std::bad_alloc& e) {
+      } catch (const rmm::out_of_memory& e) {
+        // rmm::out_of_memory is what is thrown when an allocation failed
+        // but there are other rmm::bad_alloc exceptions that could be
+        // thrown as well, which are handled by the std::exception case.
```
abellina, line 1358: should the other exceptions handled in the std::exception catch also conditionaly throw if `post_alloc_failed` returns false?

revans2: The catch block for `std::exception` will always rethrow the original exception it caught.  Why would we want to hide the original exception if `post_alloc_failed` didn't return true? The boolean returned by `post_alloc_failed` indicates if we should retry the allocation or not. We don't want to retry an allocation unless the the first allocation failed for a reason we know we can potentially recover from by retrying.  That is the bug we are trying to fix with this patch. We saw an exception from RMM and misread it to be an allocation we should retry instead of a fatal error that we should not try to handle.  Also if it is not an OOM exception `post_alloc_failed` just moves the state of the thread back to RUNNING so it could recover if it needed to.

jlowe (approved):

abellina (approved):

PR #1502, opened 2023-10-13
Adding kernel to add parse_uri support for protocol extraction

### LINKED ISSUE #1501: Implement PROTOCOL url parsing as part of parse_url support
PROTOCOL parsing will consist of pulling out the protocol portion of the uri, such as `https` in `https://www.google.com/q/cats`. It will need to conform to the spark interpretation of what constitutes a hostname and not the RFC as there are discrepancies.

In some ways, the implementation appears naive, such as `www.google.com:8080` results in a `www.google.com` protocol. In other ways, the implementation is more complex such as allowing a different character set in the protocol than in the rest of the uri.

hyperbolic2346:
This adds support for parsing protocols from a uri on the GPU. This change models the kernel after `url_decode`, which dedicates a warp to a string. This is a better approach than a thread per byte due to removing all the `lower_bound` calls that hit global memory to figure out which row a byte belongs. This kernel copies the string from global memory into shared memory a block at a time for processing.

There is a potential optimization here by having the first kernel build a starting offset for each string. With that and the number of bytes the second kernel could turn into a series of `cudaMemcpyAsync` calls. This protocol kernel even starts at byte 0, so the starting offset is known. I don't know how this will fare as support for other extractions like HOST and PATH may add further complications. I am not sure if this will be a monolithic kernel or if it will be best to have a kernel per parse type yet.

This will evolve over time, but I don't know which direction yet so I have not done this optimization.

closes #1501

### DIFF
--- a/src/main/cpp/benchmarks/CMakeLists.txt
+++ b/src/main/cpp/benchmarks/CMakeLists.txt
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/main/cpp/benchmarks/CMakeLists.txt
+++ b/src/main/cpp/benchmarks/CMakeLists.txt
@@ -23,7 +23,7 @@ target_compile_options(
 )
 
 target_link_libraries(
-  spark_rapids_jni_datagen PUBLIC cudf::cudf 
+  spark_rapids_jni_datagen PUBLIC cudf::cudf cudf::cudftestutil
 )
 
 target_include_directories(
@@ -77,3 +77,6 @@ ConfigureBench(STRING_TO_FLOAT_BENCH
 
 ConfigureBench(BLOOM_FILTER_BENCH
   bloom_filter.cu)
+
+ConfigureBench(PARSE_URI_BENCH
+  parse_uri.cpp)
--- a/src/main/cpp/benchmarks/CMakeLists.txt
+++ b/src/main/cpp/benchmarks/CMakeLists.txt
@@ -23,7 +23,7 @@ target_compile_options(
 )
 
 target_link_libraries(
-  spark_rapids_jni_datagen PUBLIC cudf::cudf cudf::cudftestutil
+  spark_rapids_jni_datagen PUBLIC cudf::cudf
 )
 
 target_include_directories(
@@ -56,7 +56,7 @@ function(ConfigureBench CMAKE_BENCH_NAME)
                    CUDA_STANDARD_REQUIRED ON
         )
     target_link_libraries(${CMAKE_BENCH_NAME} nvbench::main spark_rapids_jni_datagen ${CUDF_BENCHMARK_COMMON}
-                                              cudf::cudf spark_rapids_jni Threads::Threads)
+                                              cudf::cudf spark_rapids_jni Threads::Threads cudf::cudftestutil)
     install(
         TARGETS ${CMAKE_BENCH_NAME}
         COMPONENT testing
--- a/src/main/cpp/src/parse_uri.cu
+++ b/src/main/cpp/src/parse_uri.cu
@@ -175,7 +175,7 @@ __global__ void parse_uri_protocol_char_counter(column_device_view const in_stri
       if (!found_token[local_warp_id]) {
         clear_bit(out_validity, row_idx);
         out_counts[row_idx] = 0;
-      } else {
+      } else if (valid) {
         out_counts[row_idx] = output_string_size;
       }
     }
--- a/src/main/cpp/src/parse_uri.cu
+++ b/src/main/cpp/src/parse_uri.cu
@@ -81,7 +81,8 @@ __global__ void parse_uri_protocol_char_counter(column_device_view const in_stri
   char* in_chars_shared = temporary_buffer[local_warp_id];
 
   // Loop through strings, and assign each string to a warp.
-  for (size_type row_idx = global_warp_id; row_idx < in_strings.size(); row_idx += nwarps) {
+  for (thread_index_type tidx = global_warp_id; tidx < in_strings.size(); tidx += nwarps) {
+    auto const row_idx = static_cast<size_type>(tidx);
     if (in_strings.is_null(row_idx)) {
       if (warp_lane == 0) out_counts[row_idx] = 0;
       continue;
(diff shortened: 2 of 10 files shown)


hyperbolic2346:
build

src/main/cpp/benchmarks/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ target_compile_options(
 )
 
 target_link_libraries(
-  spark_rapids_jni_datagen PUBLIC cudf::cudf 
+  spark_rapids_jni_datagen PUBLIC cudf::cudf cudf::cudftestutil
```
hyperbolic2346, line 26: This is added to support `cudf::test::get_default_stream`, which is needed for some test fixture things like string column wrappers

jlowe: Why not link this into the test/benchmark executable that needs it rather than into a library that doesn't?

src/main/cpp/src/parse_uri.cu
```diff
@@ -0,0 +1,366 @@
...
+
+  auto const global_thread_id = cudf::detail::grid_1d::global_thread_id();
+  auto const global_warp_id   = static_cast<size_type>(global_thread_id / cudf::detail::warp_size);
+  auto const local_warp_id    = static_cast<size_type>(threadIdx.x / cudf::detail::warp_size);
+  auto const warp_lane        = static_cast<size_type>(threadIdx.x % cudf::detail::warp_size);
+  auto const nwarps     = static_cast<size_type>(gridDim.x * blockDim.x / cudf::detail::warp_size);
+  char* in_chars_shared = temporary_buffer[local_warp_id];
+
+  // Loop through strings, and assign each string to a warp.
+  for (size_type row_idx = global_warp_id; row_idx < in_strings.size(); row_idx += nwarps) {
```
jlowe, line 84: Does this loop have the potential to overflow if in_strings.size is something close to maxint?

jlowe (commented):
Are the JNI bindings for this new kernel being postponed to a followup PR?

src/main/cpp/benchmarks/CMakeLists.txt
hyperbolic2346: This is where cudf links to it as well, so I was following [their lead](https://github.com/rapidsai/cudf/blob/branch-23.12/cpp/benchmarks/CMakeLists.txt#L25).

src/main/cpp/src/parse_uri.cu
hyperbolic2346: First blush on this says yes. I will investigate.

src/main/cpp/benchmarks/CMakeLists.txt
jlowe: I still don't think that's the correct choice.  This is test code, and test code shouldn't be in the main artifact if possible,.  Doesn't seem difficult to fix this.

hyperbolic2346: I didn't mean to imply resistance, I was showing the reasoning for the change here. I have moved it.

hyperbolic2346:
build

hyperbolic2346:
> Are the JNI bindings for this new kernel being postponed to a followup PR?

JNI bindings have been added.

hyperbolic2346:
build

jlowe (commented):
Normally when adding the JNI we also add the Java bindings to leverage that JNI so we can test it via a Java unit test.

src/main/cpp/src/parse_uri.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <rmm/cuda_stream_view.hpp>
+
+#include <memory>
+
+namespace spark_rapids_jni {
+
+/**
+ * @brief Convert a string column into an integer column.
+ *
+ * @param dtype Type of column to return.
```
ttnghia, line 31: Outdated docs.

src/main/cpp/src/parse_uri.hpp
```diff
@@ -0,0 +1,45 @@
...
+ * @param string_col Incoming string column to convert to integers.
+ * @param ansi_mode If true, strict conversion and throws on erorr.
+ *                  If false, null invalid entries.
+ * @param strip if true leading and trailing white space is ignored.
+ * @param stream Stream on which to operate.
+ * @param mr Memory resource for returned column
+ * @return std::unique_ptr<column> Integer column that was created from string_col.
```
ttnghia, line 38: Outdated docs.

src/main/cpp/src/parse_uri.hpp
```diff
@@ -0,0 +1,45 @@
...
+ */
+std::unique_ptr<cudf::column> parse_uri_to_protocol(
+  cudf::strings_column_view const& string_col,
```
ttnghia, line 41: Let's use the conventional name `input`.

hyperbolic2346:
build

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Parse protocol for each URI from the incoming column.
+   *
+   * @param URIColumn The input strings column in which each row contains a URI.
+   * @return A string column with protocol data extracted.
+   */
+  public static ColumnVector parseURIProtocol(ColumnView URIColumn) {
+    assert URIColumn.getType().equals(DType.STRING) : "Input type must be String";
+    return new ColumnVector(parseProtocol(URIColumn.getNativeView()));
```
jlowe, line 38: Nit: Variable names start with a lower case letter, otherwise they look like type names.
```suggestion
  public static ColumnVector parseURIProtocol(ColumnView uriColumn) {
    assert uriColumn.getType().equals(DType.STRING) : "Input type must be String";
    return new ColumnVector(parseProtocol(uriColumn.getNativeView()));
```

src/test/java/com/nvidia/spark/rapids/jni/ParseURITest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import ai.rapids.cudf.AssertUtils;
+import ai.rapids.cudf.ColumnVector;
+
+public class ParseURITest {
+  @Test
+  void parseURIToProtocolTest() {
+    String[] testData = {"https://nvidia.com/https&#://nvidia.com",
```
jlowe, line 30: Would be nice to test null and empty strings.

src/test/java/com/nvidia/spark/rapids/jni/ParseURITest.java
```diff
@@ -0,0 +1,70 @@
...
+        scheme = uri.getScheme();
+      } catch (URISyntaxException ex) {
+        // leave the scheme null if URI is invalid
+      }
+      expectedStrings[i] = scheme;
+    }
+    try (ColumnVector v0 = ColumnVector.fromStrings(testData);
+      ColumnVector expected = ColumnVector.fromStrings(expectedStrings);
+      ColumnVector result = ParseURI.parseURIProtocol(v0);
+    ) {
```
jlowe, line 66:
```suggestion
      ColumnVector result = ParseURI.parseURIProtocol(v0)) {
```

ttnghia (approved):

jlowe (approved):

PR #1506, opened 2023-10-17
Patch parcentile evaluation to produce similar output as Spark CPU

ttnghia:
This is a minor modification to the final evaluation expression during percentile computation, preventing some sort of compiler optimization such that our GPU code produces the output that is as close as possible to Spark CPU.

### DIFF
--- a/src/main/cpp/src/histogram.cu
+++ b/src/main/cpp/src/histogram.cu
@@ -97,6 +97,8 @@ struct fill_percentile_fn {
       return;
     }
 
+    // Using `volatile` qualifier to prevent the compiler from combining `lower_part` and
+    // `upper_part` which may lead to output with different round-off error.
     volatile double const lower_part =
         (static_cast<double>(higher) - position) * static_cast<double>(lower_element);
     volatile double const higher_part =
--- a/src/main/cpp/src/histogram.cu
+++ b/src/main/cpp/src/histogram.cu
@@ -97,7 +97,11 @@ struct fill_percentile_fn {
       return;
     }
 
-    output[idx] = (higher - position) * lower_element + (position - lower) * higher_element;
+    volatile double const lower_part =
+        (static_cast<double>(higher) - position) * static_cast<double>(lower_element);
+    volatile double const higher_part =
+        (position - static_cast<double>(lower)) * static_cast<double>(higher_element);
+    output[idx] = lower_part + higher_part;
   }
 
   fill_percentile_fn(cudf::size_type const *const offsets_, ElementIterator const sorted_input_,
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -173,8 +173,6 @@ add_library(
   src/utilities.cu
   src/xxhash64.cu
   src/zorder.cu
-  ../../../thirdparty/cudf/cpp/tests/utilities/debug_utilities.cu
-  ../../../thirdparty/cudf/cpp/tests/utilities/column_utilities.cu
 )
 
 set_target_properties(
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -173,6 +173,8 @@ add_library(
   src/utilities.cu
   src/xxhash64.cu
   src/zorder.cu
+  ../../../thirdparty/cudf/cpp/tests/utilities/debug_utilities.cu
+  ../../../thirdparty/cudf/cpp/tests/utilities/column_utilities.cu
 )
 
 set_target_properties(
--- a/src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java
@@ -17,12 +17,8 @@
 package com.nvidia.spark.rapids.jni;
 
 import ai.rapids.cudf.ColumnVector;
-import ai.rapids.cudf.ColumnView;
 import ai.rapids.cudf.DType;
-import ai.rapids.cudf.ReductionAggregation;
-import ai.rapids.cudf.Scalar;
 import ai.rapids.cudf.Table;
-import ai.rapids.cudf.TableDebug;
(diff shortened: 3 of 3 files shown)


ttnghia:
build

ttnghia:
build

src/main/cpp/src/histogram.cu
```diff
@@ -97,7 +97,13 @@ struct fill_percentile_fn {
...
     }
 
-    output[idx] = (higher - position) * lower_element + (position - lower) * higher_element;
+    // Using `volatile` qualifier to prevent the compiler from combining `lower_part` and
+    // `upper_part` which may lead to output with different round-off error.
+    volatile double const lower_part =
+        (static_cast<double>(higher) - position) * static_cast<double>(lower_element);
+    volatile double const higher_part =
+        (position - static_cast<double>(lower)) * static_cast<double>(higher_element);
+    output[idx] = lower_part + higher_part;
```
mythrocks, line 106: :+1: It is interesting to know that this sort of thing is still happening.

There are still `TODO`s in the original `libcudf` device code that allude to a similar problem, purported to be eventually resolved in `CUDA 10.2`. :]

mythrocks (approved):

revans2:
I found at least one other issue. If I run all of the tests with the following patch I see a number of test failures still. I added in a new test for double that I also found with a different length that is concerning because even though the values are small the percentage difference is rather large.

```diff
diff --git a/integration_tests/src/main/python/hash_aggregate_test.py b/integration_tests/src/main/python/hash_aggregate_test.py
index aec1f5393..aa606463d 100644
--- a/integration_tests/src/main/python/hash_aggregate_test.py
+++ b/integration_tests/src/main/python/hash_aggregate_test.py
@@ -926,13 +926,21 @@ def exact_percentile_reduction(df):
     )
 
 @incompat
-@approximate_float
 @pytest.mark.parametrize('data_gen', exact_percentile_reduction_data_gen, ids=idfn)
 def test_exact_percentile_reduction(data_gen):
     assert_gpu_and_cpu_are_equal_collect(
         lambda spark: exact_percentile_reduction(gen_df(spark, data_gen))
     )
 
+@incompat
+def test_exact_percentile_reduction_bobby():
+  data_gen = [('val', RepeatSeqGen(DoubleGen(), length=100)),
+     ('freq', LongGen(min_val=0, max_val=1000000, nullable=False)
+                     .with_special_case(0, weight=100))]
+
+  assert_gpu_and_cpu_are_equal_collect(
+    lambda spark: exact_percentile_reduction(gen_df(spark, data_gen, length=373)))
+
 exact_percentile_reduction_cpu_fallback_data_gen = [
     [('val', data_gen),
      ('freq', LongGen(min_val=0, max_val=1000000, nullable=False)
@@ -940,7 +948,6 @@ exact_percentile_reduction_cpu_fallback_data_gen = [
     for data_gen in [IntegerGen(), DoubleGen()]]
 
 @incompat
-@approximate_float
 @allow_non_gpu('ObjectHashAggregateExec', 'SortAggregateExec', 'ShuffleExchangeExec', 'HashPartitioning',
                'AggregateExpression', 'Alias', 'Cast', 'Literal', 'ProjectExec',
                'Percentile')
@@ -967,7 +974,8 @@ def test_exact_percentile_reduction_partial_fallback_to_cpu(data_gen,  replace_m
             'percentile(val, 0.1)',
             'percentile(val, array(0, 0.0001, 0.5, 0.9999, 1))',
             'percentile(val, 0.1, abs(freq))',
-            'percentile(val, array(0, 0.0001, 0.5, 0.9999, 1), abs(freq))'),
+            'percentile(val, array(0, 0.0001, 0.5, 0.9999, 1), abs(freq))'
+            ),
         exist_classes=','.join(exist_clz),
         non_exist_classes=','.join(non_exist_clz),
         conf={'spark.rapids.sql.hashAgg.replaceMode': replace_mode,
@@ -1003,7 +1011,6 @@ def exact_percentile_groupby(df):
     )
 
 @incompat
-@approximate_float
 @ignore_order
 @pytest.mark.parametrize('data_gen', exact_percentile_groupby_data_gen, ids=idfn)
 def test_exact_percentile_groupby(data_gen):
@@ -1019,7 +1026,6 @@ exact_percentile_groupby_cpu_fallback_data_gen = [
     for data_gen in [IntegerGen(), DoubleGen()]]
 
 @incompat
-@approximate_float
 @ignore_order
 @allow_non_gpu('ObjectHashAggregateExec', 'SortAggregateExec', 'ShuffleExchangeExec', 'HashPartitioning',
                'AggregateExpression', 'Alias', 'Cast', 'Literal', 'ProjectExec',
```

To be clear, this didn't cause the issue, but it also didn't fix it. I am fine with filing a separate issue for this.

ttnghia:
The test is showing another kind of bug. There seems to be something else wrong. I'm investigating it.

ttnghia:
@revans2 Here is what I found.

I modified your test as follow:
```
data_gen = [('val', RepeatSeqGen(DoubleGen(), length=100))]
@approximate_float
@incompat
def test_exact_percentile_reduction_bobby():
    def doit(spark):
        df1 = gen_df(spark, data_gen, length=373)
        df1.coalesce(1).write.mode("overwrite").parquet("/home/nghiat/TMP/df.parquet")
        df = spark.read.parquet("/home/nghiat/TMP/df.parquet")
       
       # df = df1
        return df.selectExpr(
            'percentile(val, 0.1)')
    assert_gpu_and_cpu_are_equal_collect(
        doit
    )
```

The test always passed, with data read from parquet. When uncommenting `# df = df1`, the test failed.

Similarly, when testing with input read from just one parquet file, the test always passed:
```
def doit(spark):
        df = spark.read.parquet("/home/nghiat/TMP/df.parquet")
        return df.selectExpr(
            'percentile(val, 0.1)')
```

I put a `printf` in my JNI code to print the final percentile value. In both cases, my code always outputs the same value. 

So there seems to be a bug with Spark CPU?

ttnghia:
Merge this as the failed test seems unrelated anyway.

PR #1511, opened 2023-10-19
Improve quoting of command line arguments for build/run scripts

### LINKED ISSUE #1092: [BUG] build-in-docker and run-in-docker do not quote command line arguments properly
**Describe the bug**
If you want to build with multiple architectures you separate them with a `;`, but this does not work because we quite args incorrectly and turn it all into a single string, so everything after the ; is ignored because it is the end of a bash line.

HERE: https://github.com/NVIDIA/spark-rapids-jni/blob/7b1eb3bb2841e361b8c0ef82fc4f50b94f3bd4ce/build/build-in-docker#L48-L53

and HERE: https://github.com/NVIDIA/spark-rapids-jni/blob/7b1eb3bb2841e361b8c0ef82fc4f50b94f3bd4ce/build/run-in-docker#L53
and HERE: https://github.com/NVIDIA/spark-rapids-jni/blob/7b1eb3bb2841e361b8c0ef82fc4f50b94f3bd4ce/build/run-in-docker#L77

possibly more.

jlowe:
Fixes #1092.  This adds proper handling of arguments, quoting them when necessary to preserve them through double-parsing by the shell.  Tested this by running a standard build-in-docker command along with one that specifies multiple GPU architectures that failed in the past, e.g.: `build-in-docker -DGPU_ARCHS="75-real;80-real;86-real;90-real" clean install`

### DIFF
--- a/build/run-in-docker
+++ b/build/run-in-docker
@@ -50,7 +50,7 @@ if (( $# == 0 )); then
   DOCKER_OPTS="${DOCKER_OPTS} -it"
   RUN_CMD="/bin/bash"
 else
-  RUN_CMD="$*"
+  RUN_CMD="${@@Q}"
 fi
 
 $DOCKER_CMD run $DOCKER_GPU_OPTS $DOCKER_RUN_EXTRA_ARGS -u $(id -u):$(id -g) --rm \
--- a/build/build-in-docker
+++ b/build/build-in-docker
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #
-# Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


build/run-in-docker
```diff
@@ -50,7 +50,7 @@ if (( $# == 0 )); then
   DOCKER_OPTS="${DOCKER_OPTS} -it"
   RUN_CMD="/bin/bash"
 else
-  RUN_CMD="$*"
+  RUN_CMD="${@@Q}"
```
gerashegalov, line 53: nit: we would not need to build a string from an array if we used the alternative syntax https://access.redhat.com/documentation/en-us/red_hat_software_collections/3/html/packaging_guide/sect-enabling_the_software_collection
with `--` on L77

```diff
 if (( $# == 0 )); then
   # no arguments gets an interactive shell
   DOCKER_OPTS="${DOCKER_OPTS} -it"
-  RUN_CMD="/bin/bash"
+  RUN_CMD=("/bin/bash")
 else
-  RUN_CMD="${@@Q}"
+  RUN_CMD=("$@")
 fi
...
-  scl enable devtoolset-11 "$RUN_CMD"
+  scl enable devtoolset-11 -- "${RUN_CMD[@]}"
```

gerashegalov (approved):
LGTM

PR #1537, opened 2023-11-01
Implement timestamp rebase from Julian to Gregorian calendars

ttnghia:
Support rebasing day/microsecond timestamps from the legacy hybrid (Julian + Gregorian) calendar to Proleptic Gregorian calendar.

Merge checklist:
 * [X] Implementation
 * [X] Unit tests
 * [x] Test with spark-rapids (https://github.com/NVIDIA/spark-rapids/pull/9649, all integration tests pass)

### DIFF
--- a/src/main/cpp/src/datetime_rebase.cu
+++ b/src/main/cpp/src/datetime_rebase.cu
@@ -205,7 +205,7 @@ __device__ __inline__ time_components get_time_components(int64_t micros) {
 // Convert the given number of microseconds since the epoch day 1970-01-01T00:00:00Z to a local
 // date-time in Proleptic Gregorian calendar, reinterpreting the result as in Julian calendar, then
 // compute the number of microseconds since the epoch from that Julian local date-time.
-// This is to match with Apache Spark's `rebaseGregorianToJulianMicros` function with timezone
+// This is to match with Apache Spark's `localRebaseGregorianToJulianMicros` function with timezone
 // fixed to UTC.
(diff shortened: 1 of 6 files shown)


ttnghia:
build

src/main/cpp/src/datetime_rebase.cu
```diff
@@ -93,6 +94,59 @@ std::unique_ptr<cudf::column> gregorian_to_julian_days(cudf::column_view const &
   return output;
 }
 
+// Convert a number of Julian days since epoch to local date in Julian calendar.
+// Follow the implementation of `julian_from_days` from
+// https://howardhinnant.github.io/date_algorithms.html
+__device__ __inline__ cuda::std::chrono::year_month_day julian_from_days(int32_t days) {
+  auto const z = days + 719470;
```
gerashegalov, line 101: can you add comments about the hardcoded constants used in the conversion

ttnghia: Sorry there are many constants in the formula (1153`, `1460`, `719470` etc), which can't be explained in some sentences. Even the original implementation (https://howardhinnant.github.io/date_algorithms.html) doesn't have any effective way to explain them. One needs to read through the entire documentation to understand what they are. That's why the `cuda/std` adoption of such formulae (in `detail/libcxx/include/chrono` header) also can't comment anything but has to refer to the original implementation website above.

gerashegalov (approved):
LGTM

PR #1543, opened 2023-11-07
Update state retry state machine for CPU alloc support

revans2:
This updates the state machine to reduce the number of states.
Migrate the shuffle states to be more generic around thread pools instead of just shuffle
And to add in support for reusing the state machine/retry framework for CPU memory allocations too.

This is a breaking change and https://github.com/NVIDIA/spark-rapids/pull/9656 is needed in the plugin to avoid breaking the plugin.


revans2:
build

revans2:
build

parthosa:
nit: Fix typo in PR title `state maching`

src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
```diff
@@ -423,4 +523,79 @@ public static long getAndResetComputeTimeLostToRetryNs(long taskId) {
...
+      local = sra;
+    }
+    if (local != null && local.isOpen()) {
+      local.postCpuAllocSuccess(ptr, amount, blocking, wasRecursive);
+    }
+  }
+
+  /**
+   * The allocation failed, and spilling didn't save it.
+   * @param wasOom wat the failure caused by an OOM or something else.
```
jbrennan333, line 568: typo: wat

abellina (commented):
I am half way through `SparkResourceAdaptorJni`. It looks good so far, and I have mostly minor stuff.

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkTest.java
```diff
@@ -332,36 +335,101 @@ public void testInsertOOMs() {
...
+      assertEquals(RmmSparkThreadState.THREAD_RUNNING, RmmSpark.getStateOf(threadId));
+
+      try {
+        Thread.sleep(1); // Just in case we run on a really fast system in the future where
+        // all of this is sub-nanosecond...
+      } catch (InterruptedException e) {
+        // Ignored
+      }
+      // Force an exception
+      RmmSpark.forceRetryOOM(threadId);
```
jbrennan333, line 401: Doesn't this force a GpuRetryOOM?

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkTest.java
```diff
@@ -332,36 +335,101 @@ public void testInsertOOMs() {
...
+      assertEquals(1, RmmSpark.getAndResetNumRetryThrow(taskid));
+      assertEquals(0, RmmSpark.getAndResetNumSplitRetryThrow(taskid));
+      RmmSpark.blockThreadUntilReady();
+
+      // Allocate something small and verify that it works...
+      LimitingOffHeapAllocForTests.alloc(100).close();
+      assertEquals(RmmSparkThreadState.THREAD_RUNNING, RmmSpark.getStateOf(threadId));
+
+      // Force another exception
+      RmmSpark.forceSplitAndRetryOOM(threadId);
```
jbrennan333, line 419: Same as above - isn't this a GpuSplitAndRetryOOM?

jbrennan333 (commented):
Still reading through the code.  Only minor comments/questions so far.

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
```diff
@@ -46,13 +47,25 @@ public SparkResourceAdaptor(RmmEventHandlerResourceAdaptor<RmmDeviceMemoryResour
   public SparkResourceAdaptor(RmmEventHandlerResourceAdaptor<RmmDeviceMemoryResource> wrapped,
       String logLoc) {
     super(wrapped);
+    watchDog = new Thread(() -> {
+      try {
+        while (handle > 0) {
+          checkAndBreakDeadlocks();
+          Thread.sleep(100);
```
gerashegalov, line 54: I would make it a System property with a default

jbrennan333 (commented):
Overall looks good to me.  Only minor comments.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -572,78 +808,54 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+  }
+
+  bool cpu_postalloc_failed(bool was_oom, bool blocking, bool was_recursive) {
+    std::unique_lock<std::mutex> lock(state_mutex);
+    auto thread_id = static_cast<long>(pthread_self());
+    return post_alloc_failed_core(thread_id, true, was_oom, blocking, was_recursive, lock);
+  }
+
+  void cpu_dealloc(void* addr, size_t amount) {
+    // addr is not used yet, but is here in case we want it in the future.
```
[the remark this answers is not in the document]
revans2: Probably not. It is there so we can do memory leak detection/tracking. But we don't have a need for it now. Probably later.  Also it matches the RMM APIs so I wanted to keep it the same.

revans2:
@jbrennan333 @gerashegalov @abellina I think I have addressed all of the review comments.

revans2:
build

revans2:
build

gerashegalov (dismissed):
LGTM,  [my comment ](https://github.com/NVIDIA/spark-rapids-jni/pull/1543#discussion_r1397923142) does not appear addressed but it does not rise to the level of blocking this PR

revans2:
> LGTM, [my comment ](https://github.com/NVIDIA/spark-rapids-jni/pull/1543#discussion_r1397923142) does not appear addressed but it does not rise to the level of blocking this PR

Sorry should be fixed now

revans2:
build

gerashegalov (approved):

jbrennan333 (approved):
Looks good.  Thanks for the updates!

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -169,6 +197,47 @@ class thread_priority {
   long task_priority() const { return std::numeric_limits<long>::max() - (task_id + 1); }
 };
 
+/**
+ * Holds metrics for a given task/thread about retry counts and times. It is here
+ * because the mapping between tasks and threads can be complicated and can span
+ * different time ranges too.
+ */
+class task_metrics {
+  public:
```
ttnghia, line 206: If no private then just use `struct`:
```suggestion
struct task_metrics {
```

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -169,6 +197,47 @@ class thread_priority {
...
+
+  void take_from(task_metrics& other) {
+    this->num_times_retry_throw += other.num_times_retry_throw;
+    other.num_times_retry_throw = 0;
+    this->num_times_split_retry_throw += other.num_times_split_retry_throw;
+    other.num_times_split_retry_throw = 0;
+    this->time_blocked_nanos += other.time_blocked_nanos;
+    other.time_blocked_nanos = 0;
+    this->time_lost_nanos += other.time_lost_nanos;
+    other.time_lost_nanos = 0;
```
ttnghia, line 223: That is very cumbersome. Let use this:
```suggestion
  void take_from(task_metrics& other) {
    *this = other;
    other = task_metrics{};
```

ttnghia: Or we can just ignore `other` to let it keep its original values.

src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
```diff
@@ -114,47 +114,105 @@ public static long getCurrentThreadId() {
...
+   * @param taskId the ID of the task that this is starting work on.
+   */
+  public static void poolThreadWorkingOnTask(long taskId) {
+    long threadId = getCurrentThreadId();
+    Thread thread = Thread.currentThread();
+    long[] taskIds = new long[]{taskId};
+    synchronized (Rmm.class) {
+      if (sra != null && sra.isOpen()) {
+        ThreadStateRegistry.addThread(threadId, thread);
+        sra.poolThreadWorkingOnTasks(false, threadId, taskIds);
```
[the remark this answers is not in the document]
revans2: java does not support named parameters and I really dislike /* comments */ embedded in the code. If this is a deal breaker for you, then I will change it to an enum. Just let me know.

jbrennan333 (dismissed):

revans2:
build

revans2:
@ttnghia and @gerashegalov I think all of the review comments have been addressed. Please take another look.

gerashegalov (dismissed):

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -14,11 +14,13 @@
  * limitations under the License.
  */
 
+#include <algorithm>
```
ttnghia, line 17: This is known to be extremely heavy and we should avoid it if possible. What APIs do we need from it here?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -306,28 +374,57 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
     if (shutting_down) { throw std::runtime_error("spark_resource_adaptor is shutting down"); }
+    auto found = threads.find(thread_id);
+    if (found != threads.end()) {
+      if (found->second.task_id >= 0 && found->second.task_id != task_id) {
+        if (is_log_enabled) {
+          std::stringstream ss;
+          ss << "desired task_id " << task_id;
+
+          log_status("FIXUP", thread_id, found->second.task_id,
+                  found->second.state, ss.str().c_str());
```
ttnghia, line 398: The pattern `ss.str().c_str()` is used a lot here and can cause undefined behavior/crash when logging is enabled. That is because `.str().c_str()` is a pointer to a temporary string produced by `.str()` which can be destroyed immediately.
```suggestion
         auto const log_str = ss.str();
          log_status("FIXUP", thread_id, found->second.task_id,
                  found->second.state, log_str.c_str());
```

ttnghia: Another option is, for short string message like this, just use `std::string` directly:
```
auto const log_str = std::string("desired task_id ") + std::string(task_id);
          log_status("FIXUP", thread_id, found->second.task_id,
                  found->second.state, log_str.c_str());

```

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1115,61 +1365,62 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
-                   << thread_id_to_wake << " " << as_str(thread->second.state);
-                throw std::runtime_error(ss.str());
+        // 4. Wake up the BUFN thread if we should
+        if (is_to_wake_set) {
+          long thread_id_to_wake = to_wake.get_thread_id();
+          if (thread_id_to_wake > 0) {
+            // Don't wake up yourself on a free. It is not adding more memory for this thread
+            // to use on a retry and we might need a split instead to break a deadlock
+            auto this_id = static_cast<long>(pthread_self());
+            auto thread  = threads.find(thread_id_to_wake);
```
ttnghia, line 1402: Nit: Try to use `auto const`. All variables should be `const`  if possible.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -45,25 +90,16 @@ constexpr char const* SPLIT_AND_RETRY_OOM_CLASS = "com/nvidia/spark/rapids/jni/S
 enum thread_state {
```
ttnghia, line 90: Please use `enum class`. In C++, pure `enum` can be implicitly converted into `int` and vice versa.

ttnghia: Of course we still can be **explicitly** converted from `enum class` to `int`, but that is always "explicit", i.e., intentional.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -890,15 +1080,11 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     bool are_any_tasks_just_blocked = false;
     for (auto thread = threads.begin(); thread != threads.end(); thread++) {
```
ttnghia, line 1081: This pattern also repeats a lot. We can do better with this:
```suggestion
    for (auto const [thread_id, thread_state]: threads) {
```

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -926,35 +1112,57 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
    * returns true if the thread that ended was a normally running task thread.
    * This should be used to decide if wake_up_threads_after_task_finishes is called or not.
    */
-  bool remove_thread_association(long thread_id, const std::unique_lock<std::mutex>& lock)
+  bool remove_thread_association(long thread_id, long remove_task_id, const std::unique_lock<std::mutex>& lock)
```
ttnghia, line 1115: Nit: Please put `const` after type (west const) to be consistent with the current repository style.

ttnghia (changes requested):
There seems to be a serious problem with pointer to temporary so I block this until it is fixed.

revans2:
@ttnghia please take another look I think I have fixed all of your comments.

revans2:
build

ttnghia:
This is good now. You need to run `pre-commit` (https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.02/CONTRIBUTING.md#c).

revans2:
> This is good now. You need to run `pre-commit run --all-files` (https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.02/CONTRIBUTING.md#c).

The pre-commit scripts do not work out of the box with the docker environment so trying to do formatting locally is way too difficult. I will continue to let CI do my pre-commit until it is fixed.

revans2:
build

ttnghia (approved):

PR #1553, opened 2023-11-14
Custom kernel for converting timestamps b/n UTC and non-UTC non-DST timezones

NVnavkumar:
For https://github.com/NVIDIA/spark-rapids/issues/6831.

This is a custom kernel implementation of transition timezones to and from UTC. This caches the timezone transition database from Java for use on the GPU to be compatible with the Spark implementation.

This passes the test suite used for the CPU POC  (`TimeZoneSuite.scala`) in the spark-rapids codebase to be compatible with Apache Spark (See https://github.com/NVIDIA/spark-rapids/pull/9739 for updates to that test suite)


NVnavkumar:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,250 @@
...
+  Map<String, Integer> getZoneIDMap() {
+    try {
+      return zoneIdToTableFuture.get();
+    } catch (InterruptedException | ExecutionException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  Table getTransitions() {
+    ColumnVector fixedTransitions = getFixedTransitions();
```
revans2, line 225: nit:

```
try ( ColumnVector fixedTransitions = getFixedTransitions()) {
  return new Table(fixedTransitions);
}
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,250 @@
...
+
+import ai.rapids.cudf.ColumnVector;
+import ai.rapids.cudf.DType;
+import ai.rapids.cudf.HostColumnVector;
+import ai.rapids.cudf.Table;
+
+public class GpuTimeZoneDB {
+  
+  private CompletableFuture<Map<String, Integer>> zoneIdToTableFuture;
+  private CompletableFuture<HostColumnVector> fixedTransitionsFuture;
```
revans2, line 39: So what exactly is the data type stored here? It looks to be a `LIST<STRUCT<startSeconds: int64, endSecond: Int64, offsetSeconds: Int64>>`?

src/test/java/com/nvidia/spark/rapids/jni/TimeZoneTest.java
```diff
@@ -0,0 +1,206 @@
...
+    assertNotNull(transitions);
+    assertEquals(1, transitions.size());
+    transitions = instance.getHostFixedTransitions("Asia/Shanghai");
+    assertNotNull(transitions);
+    ZoneId shanghai = ZoneId.of("Asia/Shanghai").normalized();
+    assertEquals(shanghai.getRules().getTransitions().size() + 1, transitions.size());
+  }
+  
+  @Test
+  void convertToUtcSecondsTest() {
```
revans2, line 53: Why is there no test for Days? The code says that it supports it.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,223 @@
...
+  // get the fixed transitions
+  auto const ft_cdv_ptr = column_device_view::create(transitions.column(0), stream);
+  lists_column_device_view fixed_transitions = cudf::detail::lists_column_device_view{*ft_cdv_ptr};
+
+  auto results = cudf::make_timestamp_column(input.type(), input.size(),
+                                            cudf::detail::copy_bitmask(input, stream, mr),
+                                            input.null_count(), stream, mr);
+
+  switch (type) {
+    case cudf::type_id::TIMESTAMP_DAYS:
```
revans2, line 173: Do we really need anything here except microseconds?  After all that is all that Spark supports. If we think there are more that we could/should support, then fine, but DAYS in particular I don't think should be supported. There are no tests for it, and I don't think Spark will do any conversion on Date/DAYS to different timezones.

NVnavkumar: Actually I was planning to remove DAYS, so I will remove it. I do think that  Spark can cast timestamp to a date and vice-versa (timezone aware), but I'm not sure that this code would handle it anyways.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,250 @@
...
+    synchronized (this) {
+      try {
+        Map<String, Integer> zoneIdToTable = new HashMap<>();
+        List<List<HostColumnVector.StructData>> masterTransitions = new ArrayList<>();
+        for (String tzId : TimeZone.getAvailableIDs()) {
+          ZoneId zoneId;
+          try {
+            zoneId = ZoneId.of(tzId).normalized(); // we use the normalized form to dedupe
+          } catch (ZoneRulesException e) {
+            continue;
```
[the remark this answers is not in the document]
NVnavkumar: It actually would never happen in this case.  This try/catch might have been added by the IDE, but it's not necessary here. This is an exception that occurs when you pass in an invalid Timezone Id to this method that can't be found in the IANA database.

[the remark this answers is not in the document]
NVnavkumar: Should add, that the source of this is Java itself (`TimeZone.getAvailableIds()`), in which the data is coming from the same place.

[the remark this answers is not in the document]
NVnavkumar: Never mind, it seems that this data is somewhat inconsistent (probably because there is some ambiguity to be resolved, ie the 3-letter abbreviations which are available but deprecated). Maybe we should file a followup issue to handle that case?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,250 @@
...
+  
+  /**
+   * Start to cache the database. This should be called on startup of an executor. It should start
+   * to cache the data on the CPU in a background thread. It should return immediately and allow the
+   * other APIs to be called. Depending on what we want to * do we can have the other APIs block
+   * until this is done caching, or we can have private APIs that would let us load and use specific
+   * parts of the database. I prefer the former solution at least until we see a performance hit
+   * where we are waiting on the database to finish loading.
+   */
+  public static void cacheDatabase() {
```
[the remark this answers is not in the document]
revans2: Note that this is fine to do as a follow on PR. We just need it for host memory limits at some point.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,250 @@
...
+            zoneIdToTable.put(zoneId.getId(), idx);
+          }
+        }
+        HostColumnVector.DataType childType = new HostColumnVector.StructType(false,
+            new HostColumnVector.BasicType(false, DType.INT64),
+            new HostColumnVector.BasicType(false, DType.INT64),
+            new HostColumnVector.BasicType(false, DType.INT32));
+        HostColumnVector.DataType resultType =
+            new HostColumnVector.ListType(false, childType);
+        HostColumnVector fixedTransitions = HostColumnVector.fromLists(resultType,
```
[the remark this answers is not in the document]
revans2: This also is fine to do as a follow on issue.

NVnavkumar:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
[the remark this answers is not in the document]
revans2: This all depends on how many time zones we are going to be able to support, and if we end up supporting them dynamically or not. A comment and a follow on issue should be fine. I am happy to have them skipped for now, but eventually we will need a full fix.

revans2 (commented):
Looking good.

ttnghia (commented):
Please format code using clang-format v16. The style file should be `./src/main/cpp/.clang-format`.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,201 @@
...
+using struct_view = cudf::struct_view;
+
+namespace {
+
+/**
+ * @brief adjust timestamp value by offset
+ * 
+ * @tparam typestamp_type type of the input and output timestamp
+ * @param timestamp input timestamp
+ * @param transitions the transitions
```
ttnghia, line 46: Missing two more `@param`.

src/main/cpp/src/GpuTimeZoneDBJni.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_GpuTimeZoneDB_convertTimestampColumnToUTC(
+    JNIEnv *env, jclass, jlong input_handle, jlong transitions_handle, jint tz_index) {
+    JNI_NULL_CHECK(env, input_handle, "column is null", 0);
+    JNI_NULL_CHECK(env, transitions_handle, "column is null", 0);
+    try {
+        cudf::jni::auto_set_device(env);
+        auto input = reinterpret_cast<cudf::column_view const*>(input_handle);
+        auto transitions = reinterpret_cast<cudf::table_view const*>(transitions_handle);
+        auto index = static_cast<cudf::size_type>(tz_index);
```
ttnghia, line 29: It is recommended to use `auto const`.

src/main/cpp/src/GpuTimeZoneDBJni.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_GpuTimeZoneDB_convertUTCTimestampColumnToTimeZone(
+    JNIEnv *env, jclass, jlong input_handle, jlong transitions_handle, jint tz_index) {
+    JNI_NULL_CHECK(env, input_handle, "column is null", 0);
+    JNI_NULL_CHECK(env, transitions_handle, "column is null", 0);
+    try {
+        cudf::jni::auto_set_device(env);
+        auto input = reinterpret_cast<cudf::column_view const*>(input_handle);
+        auto transitions = reinterpret_cast<cudf::table_view const*>(transitions_handle);
+        auto index = static_cast<cudf::size_type>(tz_index);
```
ttnghia, line 44: Similarly, `auto const`.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * @brief adjust timestamp value by offset
+ * 
+ * @tparam typestamp_type type of the input and output timestamp
+ * @param timestamp input timestamp
+ * @param transitions the transitions 
+ */
+template <typename timestamp_type>
+__device__ timestamp_type convert_timestamp_timezone(timestamp_type const& timestamp,
+    lists_column_device_view const& transitions,
```
nvdbaranec, line 50: If you convert this to a struct/functor, you could eliminate the wrapping lambdas at the calls to thrust::transform and the counting iterator. eg:

```
template<typename timestamp_type>
struct convert_timestamp_timezone {
  list_column_device_view & fixed_transitions;
 size_type tz_index;
 

  __device__ timestamp_type operator()(timestamp_type t){
     ...
  }
};

thrust::transform(rmm::exec_policy(stream),
     input_data.begin<T>(),
     input_data.end<T>(),
     results->mutable_view().begin<cudf::timestamp_s>(),
     convert_timestamp_timezone{fixed_transitions, tz_index});
```

NVnavkumar: I actually had this way before, but for some reason it was not the easiest code to debug. I couldn't step through the code or set breakpoints properly with (cuda-)gdb.

nvdbaranec: Yeah, cuda-gdb falls apart pretty quickly.

NVnavkumar:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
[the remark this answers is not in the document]
NVnavkumar: Filed https://github.com/NVIDIA/spark-rapids-jni/issues/1571

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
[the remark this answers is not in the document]
NVnavkumar: Filed #1570

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,278 @@
...
+        throw e;
+      }
+    }
+  }
+
+  private void close() {
+    try (HostColumnVector hcv = getHostFixedTransitions()) {
+      // automatically closed
+    }
+  }
```
gerashegalov, line 225: is not this try-with-resources equivalent to 
```suggestion
  private void close() {
      getHostFixedTransitions().close()
  }
```

either  
- getHostFixedTransitions() throws and then there is nothing to call close on
- or it does not then it will be called in the implicit "finally"

neither the exception from getHostFixedTransitions nor from getHostFixedTransitions().close() are caught

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @tparam typestamp_type type of the input and output timestamp
+ * @param timestamp input timestamp
+ * @param transitions the list column of transitions to figure out the correct offset
+ *                    to adjust the timestamp. The type of the values in this column is
+ *                    LIST<STRUCT<utcInstant: int64, tzInstant: int64, utcOffset: int32>>
+ * @param tz_index the index of the specified zone id in the transitions table
+ * @param to_utc whether we are converting to UTC or converting to the timezone
+ */
+template <typename timestamp_type>
```
mythrocks, line 59: I suspect `to_utc` might be worth considering for promotion to a template parameter. Its value is constant for the entire column/operation. Either we're converting `TO_UTC` (i.e. `true`), or `FROM_UTC` (i.e. `false`).

We might consider:
```c++
template <typename timestamp_type,  bool to_utc>
```
If we then had a couple of constants:
```c++
auto constexpr TO_UTC = true;
auto constexpr FROM_UTC = false;
```
Then the bools used at the call-site would look less opaque:
```c++
thrust::transform( ..., [&] __device__( auto const& i ) {
  convert_timezone<cudf::timestamp_s, TO_UTC>( timestamp, fixed_transitions, tz_index )
});
```
More importantly, all the `to_utc` checks could be `constexpr`, and not checked once per row.

Just a thought. (@nvdbaranec?)

mythrocks: I toyed with using an enum instead:
```c++
enum class timestamp_direction : bool { FROM_UTC, TO_UTC };
```
But it seemed more verbose:
```c++
convert_timezone<cudf::timestamp_s, timestamp_direction::TO_UTC>( ... )
```
And we can't use `using` declarations with `enum` until `c++20`.  Maybe we should stick with the `bool` instead of `enum`.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+
+  using duration_type = typename timestamp_type::duration;
+  using cuda::std::chrono::duration_cast;
+
+  auto epoch_seconds =
+      static_cast<int64_t>(duration_cast<cudf::duration_s>(timestamp.time_since_epoch()).count());
+
+  auto const tz_transitions = cudf::list_device_view{transitions, tz_index};
+
+  auto size = tz_transitions.size();
```
mythrocks, line 72: Nit:
```suggestion
  auto const size = tz_transitions.size();
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+
+  cudf::device_span<int64_t const> transition_times(
+      &(transitions.child()
+            .child(to_utc ? 1 : 0)
+            .data<int64_t>()[tz_transitions.element_offset(0)]),
+      static_cast<size_t>(size));
+
+  auto idx = thrust::upper_bound(thrust::seq, transition_times.begin(), transition_times.end(),
```
mythrocks, line 80: Nit:
```suggestion
  auto const idx = thrust::upper_bound(thrust::seq, transition_times.begin(), transition_times.end(),
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+                                 epoch_seconds) -
+             transition_times.begin();
+
+  auto const list_offset = tz_transitions.element_offset(size_type(idx - 1));
+  auto const utc_offset = duration_cast<duration_type>(cudf::duration_s{
+      static_cast<int64_t>(transitions.child().child(2).element<int32_t>(list_offset))});
+  return to_utc ? timestamp - utc_offset : timestamp + utc_offset;
```
mythrocks, line 87: If `to_utc` were a template parameter, this would read as:
```c++
if constexpr (to_utc) {
  return timestamp - utc_offset;
}
else {
  return timestamp + utc_offset;
}
```
Verbose, but one branch would get eliminated.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+    case cudf::type_id::TIMESTAMP_SECONDS:
+      thrust::transform(rmm::exec_policy(stream), thrust::make_counting_iterator(0),
+                        thrust::make_counting_iterator(num_rows),
+                        results->mutable_view().begin<cudf::timestamp_s>(),
+                        [input_data = input.begin<cudf::timestamp_s>(), fixed_transitions,
+                         tz_index] __device__(auto const i) {
+                          auto const timestamp = input_data[i];
+                          return convert_timestamp_timezone<cudf::timestamp_s>(
+                              timestamp, fixed_transitions, tz_index, true);
+                        });
```
mythrocks, line 120: Since this `transform` is boilerplate, have we considered moving this to a separate function template, parametrized on `timestamp_type`?

src/main/cpp/tests/timezones.cpp
```diff
@@ -0,0 +1,280 @@
...
+        {LONG_MIN, LONG_MIN, -1585904400L, -933667200L, -922093200L, -908870400L, -888829200L,
+         -650019600L, 515527200L, 558464400L, 684867600L});
+    auto instants_to_utc_col = cudf::test::fixed_width_column_wrapper<int64_t>(
+        {LONG_MIN, LONG_MIN, -1585904400L, -933634800L, -922064400L, -908838000L, -888796800L,
+         -649990800L, 515559600L, 558493200L, 684896400L});
+    auto utc_offsets_col = cudf::test::fixed_width_column_wrapper<int32_t>(
+        {18000, 29143, 28800, 32400, 28800, 32400, 28800, 28800, 32400, 28800, 28800});
+    auto struct_column = cudf::test::structs_column_wrapper{
+        {instants_from_utc_col, instants_to_utc_col, utc_offsets_col},
+        {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1}};
```
mythrocks, line 59: There's a shortcut for the "all valid" case, in [`cudf_test/iterator_utilities.hpp`](https://github.com/rapidsai/cudf/blob/branch-23.12/cpp/include/cudf_test/iterator_utilities.hpp#L116):
```c++
    auto struct_column = cudf::test::structs_column_wrapper{
￼        {instants_from_utc_col, instants_to_utc_col, utc_offsets_col},
￼        cudf::test::iterators::no_nulls()};
```
Please feel free to ignore, if the current version reads better for you.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,278 @@
...
+            thread.setName("gpu-timezone-database-0");
+            thread.setDaemon(true);
+            return thread;
+          }
+        });
+      instance.loadData(executor);
+    }
+  }
+
+
```
mythrocks, line 83: Was the extra whitespace intentional?

mythrocks (commented):
Some very minor nitpicks.  No substantive changes were suggested.

src/main/cpp/src/timezones.cu
ttnghia: Using enum is more difficult to pass the parameter from Java. We have to reimplement the enum again in Java, then do a conversion from Java enum to C++ enum.

ttnghia: In addition, if using `to_utc` as template parameter, almost entire code (except for the `constexpr if` branches) of this function will be recompiled and duplicated. That is even worse.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+ * @param tz_index the index of the specified zone id in the transitions table
+ * @param to_utc whether we are converting to UTC or converting to the timezone
+ */
+template <typename timestamp_type>
+__device__ timestamp_type convert_timestamp_timezone(timestamp_type const &timestamp,
+                                                     lists_column_device_view const &transitions,
+                                                     size_type tz_index, bool to_utc) {
+
+  using duration_type = typename timestamp_type::duration;
+  using cuda::std::chrono::duration_cast;
```
ttnghia, line 65: Please do not type alias any name having `chrono::` prefix. It is very error-prone.

src/main/cpp/src/timezones.cu
ttnghia: Also remove the empty line before this line.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+
+  cudf::device_span<int64_t const> transition_times(
+      &(transitions.child()
+            .child(to_utc ? 1 : 0)
+            .data<int64_t>()[tz_transitions.element_offset(0)]),
+      static_cast<size_t>(size));
+
+  auto idx = thrust::upper_bound(thrust::seq, transition_times.begin(), transition_times.end(),
+                                 epoch_seconds) -
+             transition_times.begin();
```
ttnghia, line 82: Do not subtract iterator from binary search like this. Instead, use like this:
```
auto const it = thrust::upper_bound(thrust::seq, begin, end, val);
auto const idx = static_cast<cudf::size_type>(thrust::distance(begin, it));
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+
+  auto const list_offset = tz_transitions.element_offset(size_type(idx - 1));
```
ttnghia, line 84: If we have to subtract one from `idx`, how about using `thrust::lower_bound` in the code above.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,206 @@
...
+
+namespace spark_rapids_jni {
+
+std::unique_ptr<column> convert_timestamp_to_utc(cudf::column_view const &input,
+                                                 cudf::table_view const &transitions,
+                                                 size_type tz_index, rmm::cuda_stream_view stream,
+                                                 rmm::mr::device_memory_resource *mr) {
+
+  auto type = input.type().id();
+  auto num_rows = input.size();
```
ttnghia, line 100: `auto const`.

src/main/cpp/src/timezones.cu
ttnghia, line 120: Recommend to follow what is suggested here: https://github.com/NVIDIA/spark-rapids-jni/pull/1553#discussion_r1396481946

src/main/cpp/src/timezones.cu
NVnavkumar: It doesn't actually work for this algorithm. When the timestamp actually matches the instant that it is being compared to, we need that element in this binary search. `upper::bound` gives us the insertion point after this element, so we subtract `1`. When the timestamp doesn't match, then lower bound will still give us an insertion point after the element that we need to use. So we use upper bound, and always subtract `1` to get the element we want every time.

NVnavkumar: Also, to be clear the nature of the data in this list is that it will never have repeating elements (since it's a list of timestamps for each transition, and there is only 1 transition at any given time).

src/main/cpp/src/timezones.cu
mythrocks: > difficult to pass the parameter from Java.

That's an interesting point, but that does not apply here.  This function is private to implementation, and `to_utc` is specified internally.  There are different functions for `from_utc` and `to_utc`.

mythrocks: >  this function will be recompiled and duplicated. 

I suspected that's how we will be leaning.  Given the number of architectures we have to support, a runtime check might be better.

Let's ignore this suggestion.

src/main/cpp/src/timezones.cu
mythrocks: Why is this specifically error prone?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,250 @@
...
+        fixedTransitionsFuture.complete(fixedTransitions);
+      } catch (Exception e) {
+        zoneIdToTableFuture.completeExceptionally(e);
+        fixedTransitionsFuture.completeExceptionally(e);
+        throw e;
+      }
+    }
+  }
+
+  HostColumnVector getHostFixedTransitions() {
```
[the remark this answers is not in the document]
NVnavkumar: Made most of these private anyways, since there is no reason for them to be default visibility

src/main/cpp/src/timezones.cu
ttnghia: >  whereas lower::bound gives us the insertion point before this element

No not that. Indeed, `lower_bound` gives the position "no less than" the search element. In our use case, we don't have duplicate elements thus it will give the exact position of the element. And that is also the same as `upper_bound - 1`. So I think we can actually use `lower_bound` here without `-1`.

ttnghia: Found this: 
image

NVnavkumar:
premerge CI is blocked by https://github.com/NVIDIA/spark-rapids-jni/issues/1567

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,184 @@
...
+ * @brief Convert the timestamp value to either UTC or a specified timezone
+ *
+ * This device functor uses a binary search to find the instant of the transition
+ * to find the right offset to do the transition.
+ *
+ * To transition to UTC: do a binary search on the tzInstant child column and subtract
+ * the offset
+ *
+ * To transition from UTC: do a binary search on the utcInstant child column and add
+ * the offset
```
ttnghia, line 49:
```suggestion
 * To transition to UTC: do a binary search on the tzInstant child column and subtract
 * the offset.
 *
 * To transition from UTC: do a binary search on the utcInstant child column and add
 * the offset.
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,184 @@
...
+ * 
+ * @param timestamp input timestamp
+ */
+template <typename timestamp_type>
+struct convert_timestamp_tz_functor {
+  using duration_type = typename timestamp_type::duration;
+
+  lists_column_device_view transitions;
+  size_type tz_index;
+  bool to_utc;
```
ttnghia, line 66:
```suggestion
  lists_column_device_view const transitions;
  size_type const tz_index;
  bool const to_utc;
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,184 @@
...
+        &(transitions.child()
+              .child(to_utc ? 1 : 0)
+              .data<int64_t>()[tz_transitions.element_offset(0)]),
+        static_cast<size_t>(list_size));
+
+    auto const it = thrust::upper_bound(thrust::seq, transition_times.begin(), transition_times.end(),
+                                  epoch_seconds);
+    auto const idx = static_cast<size_type>(thrust::distance(transition_times.begin(), it));
+
+    auto const list_offset = tz_transitions.element_offset(size_type(idx - 1));
```
ttnghia, line 85: `idx` is already `size_type`.
```suggestion
    auto const list_offset = tz_transitions.element_offset(idx - 1);
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,184 @@
...
+    auto const epoch_seconds =
+        static_cast<int64_t>(cuda::std::chrono::duration_cast<cudf::duration_s>(
+          timestamp.time_since_epoch()).count());
+    auto const tz_transitions = cudf::list_device_view{transitions, tz_index};
+    auto const list_size = tz_transitions.size();
+
+    cudf::device_span<int64_t const> transition_times(
+        &(transitions.child()
+              .child(to_utc ? 1 : 0)
+              .data<int64_t>()[tz_transitions.element_offset(0)]),
```
ttnghia, line 78:
```suggestion
    auto const transition_times = cudf::device_span<int64_t const>(
        transitions.child().child(to_utc ? 1 : 0).data<int64_t>() + tz_transitions.element_offset(0),
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,184 @@
...
+        static_cast<size_t>(list_size));
+
+    auto const it = thrust::upper_bound(thrust::seq, transition_times.begin(), transition_times.end(),
+                                  epoch_seconds);
+    auto const idx = static_cast<size_type>(thrust::distance(transition_times.begin(), it));
+
+    auto const list_offset = tz_transitions.element_offset(size_type(idx - 1));
+    auto const utc_offset = cuda::std::chrono::duration_cast<duration_type>(cudf::duration_s{
+        static_cast<int64_t>(transitions.child().child(2).element<int32_t>(list_offset))});
```
ttnghia, line 87: I recommend to type alias these `child().child(..)` into variables at the beginning of the function:
```
auto const utcInstant = transitions.child().child(0);
auto const tzInstant = transitions.child().child(1);
auto const utcOffsets = transitions.child().child(2);
```

So later on:
```
static_cast<int64_t>(utcOffsets.element<size_type>(list_offset))});
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,184 @@
...
+                                                 cudf::table_view const &transitions,
+                                                 size_type tz_index, rmm::cuda_stream_view stream,
+                                                 rmm::mr::device_memory_resource *mr) {
+
+  auto const type = input.type().id();
+  auto const num_rows = input.size();
+
+  // get the fixed transitions
+  auto const ft_cdv_ptr = column_device_view::create(transitions.column(0), stream);
+  lists_column_device_view fixed_transitions = cudf::detail::lists_column_device_view{*ft_cdv_ptr};
```
ttnghia, line 106:
```suggestion
  auto const fixed_transitions = cudf::detail::lists_column_device_view{*ft_cdv_ptr};
```

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,184 @@
...
+                                             cudf::detail::copy_bitmask(input, stream, mr),
+                                             input.null_count(), stream, mr);
+
+  switch (type) {
+    case cudf::type_id::TIMESTAMP_SECONDS:
+      thrust::transform(rmm::exec_policy(stream), 
+                        input.begin<cudf::timestamp_s>(),
+                        input.end<cudf::timestamp_s>(),
+                        results->mutable_view().begin<cudf::timestamp_s>(),
+                        convert_timestamp_tz_functor<cudf::timestamp_s>{fixed_transitions, tz_index, true});
```
ttnghia, line 118: This snippet is called 6 times. So we can extract it as a template function:
```
template<typename Ts>
auto convert_ts(input, transitions, tz_index, to_utc, stream) {
  auto const fixed_transitions = ....;
  auto result = ....;
  
  thrust::transform(rmm::exec_policy(stream), 
                        input.begin<Ts>(), input.end<Ts>(),
                        results->mutable_view().begin<Ts>(),
                        convert_timestamp_tz_functor<Ts>{fixed_transitions, tz_index, to_utc});
}
```

Then later on, call this 6 times like:
```
switch(type) {
 case TIMESTAMP_SECOND: return convert_ts<cudf::timestamp_s>(input, transitions, tz_index, to_utc, stream);
 case TIMESTAMP_MILLISECOND: return convert_ts<...>();
...
}
```

src/main/cpp/src/timezones.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+namespace spark_rapids_jni {
+
+/**
+ * @brief Convert input column timestamps in current timezone to UTC
+ *
+ * The transition rules are in enclosed in a table, and the index corresponding to the
+ * current timezone is given.
+ * 
+ * This method is the converse of convert_utc_timestamp_to_timezone
```
ttnghia, line 32:
```suggestion
 * This method is the converse of convert_utc_timestamp_to_timezone.
```

src/main/cpp/src/timezones.hpp
```diff
@@ -0,0 +1,64 @@
...
+    rmm::cuda_stream_view stream = cudf::get_default_stream(),
+    rmm::mr::device_memory_resource *mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Convert input column timestamps in UTC to specified timezone
+ *
+ * The transition rules are in enclosed in a table, and the index corresponding to the
+ * specific timezone is given.
+ * 
+ * This method is the converse of convert_timestamp_to_utc
```
ttnghia, line 51:
```suggestion
 * This method is the converse of convert_timestamp_to_utc.
```

src/main/cpp/src/timezones.hpp
ttnghia: Should `converse` be `reverse`?

src/main/cpp/src/timezones.hpp
ttnghia: Should `converse` be `reverse`?

src/main/cpp/tests/timezones.cpp
ttnghia: Please mark all variables as `const` whenever possible.

src/main/cpp/tests/timezones.cpp
```diff
@@ -0,0 +1,280 @@
...
+  void SetUp() override
+  {
+    transitions = make_transitions_table();
+  }
+  std::unique_ptr<cudf::table> transitions;
+
+private:
+  std::unique_ptr<cudf::table> make_transitions_table()
+  {
+    auto instants_from_utc_col = cudf::test::fixed_width_column_wrapper<int64_t>(
```
ttnghia, line 49: Since we are using type aliasing like `seconds_col`, please do similarly for `int64_t` such as `int64s_col`.

src/main/cpp/tests/timezones.cpp
```diff
@@ -0,0 +1,280 @@
...
+        {LONG_MIN, LONG_MIN, -1585904400L, -933667200L, -922093200L, -908870400L, -888829200L,
```
ttnghia, line 50: Do not use the old style macro `***_MIN/MAX`. Use this instead:
```
#include <limits>
....
auto constexpr int64_min = std::numeric_limits<int64_t>::min();
auto constexpr int64_max = ...
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,313 @@
...
+  
+  private static GpuTimeZoneDB instance = new GpuTimeZoneDB();
+  static GpuTimeZoneDB getInstance() {
+    return instance;
+  }
+  
+  /**
+   * Start to cache the database. This should be called on startup of an executor. It should start
+   * to cache the data on the CPU in a background thread. It should return immediately and allow the
+   * other APIs to be called. Depending on what we want to * do we can have the other APIs block
```
ttnghia, line 62:
```suggestion
   * other APIs to be called. Depending on what we want to do we can have the other APIs block
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,313 @@
...
+
+  private boolean closed = false;
+
+  GpuTimeZoneDB() {
+    zoneIdToTableFuture = new CompletableFuture<>();
+    fixedTransitionsFuture = new CompletableFuture<>();
+  }
+  
+  private static GpuTimeZoneDB instance = new GpuTimeZoneDB();
+  static GpuTimeZoneDB getInstance() {
```
ttnghia, line 55: Should be `public static`?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,313 @@
...
+   * Start to cache the database. This should be called on startup of an executor. It should start
+   * to cache the data on the CPU in a background thread. It should return immediately and allow the
+   * other APIs to be called. Depending on what we want to * do we can have the other APIs block
+   * until this is done caching, or we can have private APIs that would let us load and use specific
+   * parts of the database. I prefer the former solution at least until we see a performance hit
+   * where we are waiting on the database to finish loading.
+   */
+  public static void cacheDatabase() {
+    if (!instance.isLoaded()) {
+      Executor executor = Executors.newSingleThreadExecutor(
```
ttnghia, line 69: Will this threadsafe? Should we add to the function the `synchronized` qualifier?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,313 @@
...
+          currentTimeZone.toString()));
+    }
+    if (!instance.isLoaded()) {
+      cacheDatabase(); // lazy load the database
+    }
+    Integer tzIndex = instance.getZoneIDMap().get(currentTimeZone.normalized().toString());
+    Table transitions = instance.getTransitions();
+    ColumnVector result = new ColumnVector(convertTimestampColumnToUTC(input.getNativeView(),
+        transitions.getNativeView(), tzIndex));
+    transitions.close();
```
ttnghia, line 106: I feel this is very expensive since we upload data to GPU and create a new transition table every time we call this function. Can we cache the transition table inside `instant`?

NVnavkumar: Yeah. I plan to update this in a future PR since right now we just need to see if we are computing the right thing on the GPU. I think it's a still an open question as to how to cache it and what makes sense. I will file a follow up issue on optimizing this.

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,184 @@
...
+#include <cudf/detail/null_mask.hpp>
+#include <cudf/lists/list_device_view.cuh>
+#include <cudf/lists/lists_column_device_view.cuh>
+#include <cudf/table/table.hpp>
+#include <cudf/types.hpp>
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/exec_policy.hpp>
+#include <thrust/binary_search.h>
+
+#include "timezones.hpp"
```
ttnghia, line 29: The headers should be grouped by a "near to far" order: local headers first, then `cudf_test`, then `cudf/`, then `thrust`, then `rmm`, finally C++ built-in. This applies for all C++ files.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
NVnavkumar: Also, right now the functionality will be hidden behind a configuration flag, so there is time to optimize before fully exposing.

NVnavkumar:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
NVnavkumar: Filed this https://github.com/NVIDIA/spark-rapids-jni/issues/1588

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
NVnavkumar: No this is default visiblity for testing purposes only. Added a comment.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -0,0 +1,313 @@
...
+            thread.setDaemon(true);
+            return thread;
+          }
+        });
+      instance.loadData(executor);
+    }
+  }
+
+
+  public static void shutdown() {
```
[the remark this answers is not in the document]
NVnavkumar: It's synchronized in the close method.

src/main/cpp/src/timezones.hpp
NVnavkumar: Actually changed to `inverse`

NVnavkumar:
build

src/main/cpp/src/timezones.cu
```diff
@@ -0,0 +1,154 @@
...
+}
+
+} // namespace
+
+namespace spark_rapids_jni {
+
+std::unique_ptr<column> convert_timestamp_to_utc(column_view const &input,
+                                                 table_view const &transitions, size_type tz_index,
+                                                 rmm::cuda_stream_view stream,
+                                                 rmm::mr::device_memory_resource *mr) {
```
ttnghia, line 114: Sorry I have another idea to further minimize code duplicate:
```suggestion
namespace {
....
std::unique_ptr<column> convert_timestamp(column_view const &input,
                                                 table_view const &transitions, size_type tz_index,
                                                 bool to_utc,
                                                 rmm::cuda_stream_view stream,
                                                 rmm::mr::device_memory_resource *mr) {
....
//replace `true`/`false` by `to_utc`
}

} // namespace

std::unique_ptr<column> convert_timestamp_to_utc(...) {
  return convert_timestamp(input, transition, tz_index, true, stream, mr);
}

std::unique_ptr<column> convert_utc_timestamp_to_timezone(...) {
  return convert_timestamp(input, transition, tz_index, false, stream, mr);
}
...

````

**Comment by NVnavkumar** [comment]
build

File: src/main/cpp/src/timezones.hpp [hunk 43]
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+#include <cudf/column/column_view.hpp>
+#include <cudf/table/table.hpp>
+#include <cudf/utilities/default_stream.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+
+#include <cstddef>
+
```
ttnghia, line 24:
```suggestion
```

ttnghia (dismissed):

NVnavkumar:
build

ttnghia (approved):

PR #1554, opened 2023-11-14
Updating `parse_uri` to properly validate the URI before returning results

### LINKED ISSUE #1501: Implement PROTOCOL url parsing as part of parse_url support

hyperbolic2346:
These are changes to incorporate validation to match Spark of URIs for `parse_uri`. This results in some major changes to the kernel to switch from warp-per-row to thread-per-row. This is due to the complexity of the validation. The plan is to produce a working kernel and then focus on performance if necessary. This method will be slower than the warp-per-row method, but the complexity of such a kernel leads us to the simpler kernel for now.

closes #1501


hyperbolic2346:
build

thirtiseven (commented):
Tested with plugin, it is much closer! Only a few mismatches found in tests:
```
"http://?",
"http://??",
"http://??/",
"http://#",
"http://user:pass@host/file;param?query;p2",
"http://[1:2:3:4:5:6:7::]",
"http://[::2:3:4:5:6:7:8]",
"http://[fe80::7:8%eth0]",
"http://[fe80::7:8%1]",
```

Performance test:
```
spark.time(df.selectExpr("COUNT(parse_url(url, 'PROTOCOL')) as pr1", "COUNT(parse_url(url, 'PROTOCOL')) as pr2", "COUNT(parse_url(url, 'PROTOCOL')) as pr3", "COUNT(parse_url(url, 'PROTOCOL')) as pr4", "COUNT(parse_url(url, 'PROTOCOL')) as pr5", "COUNT(parse_url(url, 'PROTOCOL')) as pr6", "COUNT(parse_url(url, 'PROTOCOL')) as pr7", "COUNT(parse_url(url, 'PROTOCOL')) as pr8", "COUNT(parse_url(url, 'PROTOCOL')) as pr9", "COUNT(parse_url(url, 'PROTOCOL')) as pr0").show())
```
against 651191 urls' dataset, repeated 100 times.

CPU: 10752 ms
GPU: 1032 ms

It's about 10x faster


Will test with the large dataset later to see if there are more cases.

thirtiseven:
3772 cases out of 651191 urls failed from [kaggle](https://www.kaggle.com/datasets/sid321axn/malicious-urls-dataset) dataset, details in following csv:

[part-00000-f74abea5-dbc6-4a2c-9fbe-9fb9790e48d9-c000.csv](https://github.com/NVIDIA/spark-rapids-jni/files/13374378/part-00000-f74abea5-dbc6-4a2c-9fbe-9fb9790e48d9-c000.csv)

hyperbolic2346:
I added fixes for these cases, but the java test is still failing on `http://#` and `http://?` and I don't know why yet. The cpp tests with the same strings pass, but something is different coming in from Java. Will fix the last two tomorrow.

thirtiseven:
Here are new results from the Kaggle dataset, only 467/651191 cases left.

[part-00000-b118b9bb-468e-4b95-b08a-beb43acf7b17-c000.csv](https://github.com/NVIDIA/spark-rapids-jni/files/13413917/part-00000-b118b9bb-468e-4b95-b08a-beb43acf7b17-c000.csv)

Some patterns I have seen from tests:
- If an url has no protocol and has a ':' in it, the protocol will be the first part split by ':' and the validation for the left part is simpler (just `[^\\#]*`, I think)
- Many urls containing `\[0-9]{3}` passed the validation, but they shouldn't.
- Double `#` in REF. I remember that REF has it's own allowed character set.
- urls like `http://foo.bar/abc/\\\http://foo.bar/abc.gif\\\`

hyperbolic2346:
build

hyperbolic2346:
build

hyperbolic2346:
build

src/main/cpp/src/parse_uri.cu
```diff
@@ -23,279 +23,743 @@
...
+
+enum URI_chunks { PROTOCOL, HOST, AUTHORITY, PATH, QUERY, USERINFO };
+
+enum chunk_validity { VALID, INVALID, FATAL };
+
 namespace {
 
-// utility to validate a character is valid in a URI
-constexpr bool is_valid_character(char ch, bool alphanum_only)
+// some parsing errors are fatal and some parsing errors simply mean this
```
ttnghia, line 59:
```suggestion
// Some parsing errors are fatal and some parsing errors simply mean this
```

src/main/cpp/src/parse_uri.cu
```diff
@@ -23,279 +23,743 @@
...
-    if (ch >= 'a' && ch <= 'z') return true;               // a-z
-  } else {
-    if (ch >= '!' && ch <= ';' && ch != '"') return true;  // 0-9 and !#%&'()*+,-./
-    if (ch >= '=' && ch <= 'Z' && ch != '>') return true;  // A-Z and =?@
-    if (ch >= '_' && ch <= 'z' && ch != '`') return true;  // a-z and _
+  return is_numeric(c) || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F');
+}
+
+__device__ bool skip_and_validate_special(string_view::const_iterator& iter,
+                                          string_view::const_iterator end,
```
ttnghia, line 78:
```suggestion
__device__ bool skip_and_validate_special(string_view::const_iterator iter,
                                          string_view::const_iterator end,
```

src/main/cpp/src/parse_uri.cu
```diff
@@ -23,279 +23,743 @@
...
-          if (warp_lane == cudf::detail::warp_size - 1) {
-            clear_bit(out_validity, row_idx);
-            out_counts[row_idx] = 0;
-          }
+bool __device__ validate_port(string_view port)
+{
+  // port is positive numeric >=0 according to spark...shrug
+  return validate_chunk(port, [] __device__(string_view::const_iterator iter) {
+    auto const c = *iter;
+    if (c < '0' && c > '9') return false;
```
ttnghia, line 418:
```suggestion
    if (c < '0' && c > '9') { return false; }
```

src/main/cpp/src/parse_uri.cu
ttnghia: Also recommend to capitalize the first letter of sentences in all comments.

src/main/cpp/src/parse_uri.cu
hyperbolic2346: This is actually being modified inside this function as we may skip characters such as hex encoded characters or unicode. I also need to return validity information. I originally had this returning the iterator after it was modified, but switched to this when I also needed validity. Do you think this is easier to understand returning a tuple?

hyperbolic2346: Implemented this. Your question tells me that it isn't clear what is happening like if it is a returned value.

src/main/cpp/src/parse_uri.cu
```diff
@@ -23,279 +23,749 @@
 #include <cudf/scalar/scalar_factories.hpp>
 #include <cudf/strings/detail/strings_children.cuh>
 #include <cudf/strings/detail/utilities.cuh>
+#include <cudf/strings/string_view.cuh>
 
 #include <rmm/cuda_stream_view.hpp>
 #include <rmm/exec_policy.hpp>
 
 #include <memory>
+#include <tuple>
```
ttnghia, line 32: I don't see it is used anywhere?

src/main/cpp/src/parse_uri.cu
```diff
@@ -23,279 +23,749 @@
...
-    if (ch >= 'A' && ch <= 'Z') return true;               // A-Z
-    if (ch >= 'a' && ch <= 'z') return true;               // a-z
-  } else {
-    if (ch >= '!' && ch <= ';' && ch != '"') return true;  // 0-9 and !#%&'()*+,-./
-    if (ch >= '=' && ch <= 'Z' && ch != '>') return true;  // A-Z and =?@
-    if (ch >= '_' && ch <= 'z' && ch != '`') return true;  // a-z and _
+  return is_numeric(c) || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F');
+}
+
+__device__ std::pair<bool, string_view::const_iterator> skip_and_validate_special(
```
ttnghia, line 77: It is recommended to use `thrust::pair` in device code.

src/main/cpp/src/parse_uri.cu
```diff
@@ -23,279 +23,749 @@
...
-  auto const nwarps     = static_cast<size_type>(gridDim.x * blockDim.x / cudf::detail::warp_size);
-  char* in_chars_shared = temporary_buffer[local_warp_id];
-
-  // Loop through strings, and assign each string to a warp
-  for (thread_index_type tidx = global_warp_id; tidx < in_strings.size(); tidx += nwarps) {
+  // thread per row
+  auto const tid      = threadIdx.x + blockIdx.x * blockDim.x;
+  auto const base_ptr = in_strings.child(strings_column_view::chars_column_index).data<char>();
+
+  for (thread_index_type tidx = tid; tidx < in_strings.size(); tidx += blockDim.x * gridDim.x) {
```
ttnghia, line 677: I vaguely remember that there is some access function for `blockDim.x * gridDim.x`?

src/main/cpp/src/parse_uri.cu
```diff
@@ -306,6 +776,9 @@ std::unique_ptr<column> parse_uri_to_protocol(strings_column_view const& input,
   auto offsets_column = make_numeric_column(
     data_type{type_to_id<size_type>()}, offset_count, mask_state::UNALLOCATED, stream, mr);
 
+  // build src offsets buffer
+  auto src_offsets = rmm::device_buffer{strings_count * sizeof(size_type), stream};
```
ttnghia, line 780: Why not `device_uvector<size_type>`?

hyperbolic2346: This is simply a temporary buffer of data and is used as such. It could be a `device_uvector`, but we only ever use '.data()`. I'm not against making it a `device_uvector`, but it doesn't buy us much outside of not using `sizeof` directly in our code.

ttnghia:
You probably have some accidental changes in the cudf folder?

ttnghia (approved):

hyperbolic2346:
build

ttnghia (approved):

PR #1573, opened 2023-11-18
Enable Debug build for native code in spark-rapids-jni

NVnavkumar:
This adds the ability to build the native code with debug symbols that can be used in gdb and cuda-gdb. We separate out the build types for libcudf and libsparkrapidsjni because building libcudf with debug symbols is incredibly slow and should only be used by an advanced developer. 

In order to build the debug version of libsparkrapidsjni, you can use:

```shell script
./build/build-in-docker install -DGPU_ARCHS='<compute_archs>' -DCPP_PARALLEL_LEVEL=64 -DBUILD_TESTS=ON -DBUILD_TYPE=Debug -DCUDA_FLAGS_DEBUG'=-G -g'
```

Note, the `CUDA_FLAGS_DEBUG` setting here. By default this value is only `-g` which only enables debug symbols in host code. You need to add `-G` to enable debug symbols for device code.  You need to set the `BUILD_TYPE` to `Debug` to enable the usage of these flags (by default, this is `Release`).

Debug builds can be quite slow so this should only be used if you really need to debug native code on device.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,6 +81,9 @@
     <RMM_LOGGING_LEVEL>OFF</RMM_LOGGING_LEVEL>
     <SPARK_RAPIDS_JNI_CXX_FLAGS/>
     <USE_GDS>OFF</USE_GDS>
+    <LIBCUDF_BUILD_TYPE>Release</LIBCUDF_BUILD_TYPE>
+    <BUILD_TYPE>${LIBCUDF_BUILD_TYPE}</BUILD_TYPE>
+    <CUDA_FLAGS_DEBUG>-g</CUDA_FLAGS_DEBUG>
     <BUILD_TESTS>OFF</BUILD_TESTS>
     <BUILD_BENCHMARKS>OFF</BUILD_BENCHMARKS>
     <BUILD_FAULTINJ>ON</BUILD_FAULTINJ>
@@ -393,6 +396,7 @@
                 <property name="RMM_LOGGING_LEVEL" value="${RMM_LOGGING_LEVEL}"/>
                 <property name="USE_GDS" value="${USE_GDS}"/>
                 <property name="BUILD_BENCHMARKS" value="${BUILD_BENCHMARKS}"/>
+                <property name="LIBCUDF_BUILD_TYPE" value="${LIBCUDF_BUILD_TYPE}"/>
                 <ant antfile="build-libcudf.xml" target="build-libcudf"/>
               </target>
             </configuration>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -81,6 +81,9 @@
     <RMM_LOGGING_LEVEL>OFF</RMM_LOGGING_LEVEL>
     <SPARK_RAPIDS_JNI_CXX_FLAGS/>
     <USE_GDS>OFF</USE_GDS>
+    <LIBCUDF_BUILD_TYPE>Release</LIBCUDF_BUILD_TYPE>
+    <BUILD_TYPE>${LIBCUDF_BUILD_TYPE}</BUILD_TYPE>
+    <CUDA_FLAGS_DEBUG>-g</CUDA_FLAGS_DEBUG>
```
jlowe, line 86: Is this intended to be overridden by the user for some use cases?  I assume this is only used for Debug builds, and I'm a bit surprised this needs to be specified at all.

jlowe (commented):
Changes look OK but it would be nice to update CONTRIBUTING.md.

pom.xml
```diff
@@ -81,6 +81,9 @@
     <RMM_LOGGING_LEVEL>OFF</RMM_LOGGING_LEVEL>
     <SPARK_RAPIDS_JNI_CXX_FLAGS/>
     <USE_GDS>OFF</USE_GDS>
+    <LIBCUDF_BUILD_TYPE>Release</LIBCUDF_BUILD_TYPE>
```
gerashegalov, line 84: In my previous experience, libcudf as a whole fails to build with `BUILD_TYPE=Debug`. I have been relying on selective Debug compiling https://github.com/rapidsai/cudf/blob/branch-23.12/CONTRIBUTING.md#device-debug-symbols. After a certain threshold depending on how many files are compiled with debugging symbols, the plugin slow to start or hangs for me.

pom.xml
NVnavkumar: Yeah, it's a little weird, but the default flags only adds `-g` (which enables debugging in host code), you need `-G` to enable debugging in device code as well.

pom.xml
NVnavkumar: I'm wondering now, if I should try the CUDF approach here with spark-rapids-jni code and then maybe file another PR with a similar section in spark-rapids-jni CONTRIBUTING.md if that works with the way we build.

Might make a bit more sense then.

NVnavkumar: Closing this for now.

PR #1583, opened 2023-11-21
Fix default values for stream parameter

ttnghia:
This changes the default values for stream parameter in various APIs. We should switch to use `cudf::get_default_stream()` which is consistent throughout the library (it will be either `rmm::cuda_stream_default` or `rmm::cuda_stream_per_thread`). Using only `rmm::cuda_stream_default` may lead to mixed use of `rmm::cuda_stream_default` and `rmm::cuda_stream_per_thread` in the same place.

### DIFF
--- a/src/main/cpp/src/decimal_utils.hpp
+++ b/src/main/cpp/src/decimal_utils.hpp
@@ -26,29 +26,39 @@
 
 namespace cudf::jni {
 
-std::unique_ptr<cudf::table>
-multiply_decimal128(cudf::column_view const &a, cudf::column_view const &b, int32_t product_scale,
-                    rmm::cuda_stream_view stream = cudf::get_default_stream());
+std::unique_ptr<cudf::table> multiply_decimal128(
+  cudf::column_view const& a,
+  cudf::column_view const& b,
+  int32_t product_scale,
+  rmm::cuda_stream_view stream = cudf::get_default_stream());
 
-std::unique_ptr<cudf::table>
-divide_decimal128(cudf::column_view const &a, cudf::column_view const &b, int32_t quotient_scale,
-                  rmm::cuda_stream_view stream = cudf::get_default_stream());
+std::unique_ptr<cudf::table> divide_decimal128(
+  cudf::column_view const& a,
+  cudf::column_view const& b,
+  int32_t quotient_scale,
+  rmm::cuda_stream_view stream = cudf::get_default_stream());
 
-std::unique_ptr<cudf::table>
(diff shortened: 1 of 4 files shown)


ttnghia:
build

src/main/cpp/src/decimal_utils.hpp
```diff
@@ -16,45 +16,39 @@
 
 #include <cudf/column/column_view.hpp>
 #include <cudf/table/table.hpp>
+#include <cudf/utilities/default_stream.hpp>
+
+//
```
jlowe, line 21: Comment with no text?  Should either remove these or add a comment.  I'm assuming this is to delineate sections of the includes (IMO not necessary, pretty obvious from include paths or lack thereof), but without text it's not clear what the intent is.

Applies to everywhere this was added in this PR.

ttnghia: This is due to the issue with clang-format that messes up the header order. I have to use an empty comment line to prevent header reordering.

jbrennan333 (approved):
+1 this looks good to me.

src/main/cpp/src/decimal_utils.hpp
jlowe: Does it need to be empty, or can the comment say "// this is here to avoid clang-format from rearranging the header files".  Also wondering why we're explicitly avoiding the recommended header order?

ttnghia: The current clang-format config combines headers together and doesn't respect the conventional "near to far"  order. That means we should include our header first, then library headers, then builtin headers. By using "near to far" order, we can ensure that most headers have their own dependencies included themselves.

We probably should change the config a little bit to account for that.

ttnghia: Ideally if we switch to use `.clang-format` style from cudf then we don't have to worry about this as cudf config can avoid this mixing header issue.

ttnghia: There is another option: fix the style to preserve header grouping: https://github.com/NVIDIA/spark-rapids-jni/pull/1584. With that, the header organization is totally up to the code author.

jlowe (dismissed):

jlowe (approved):

jlowe:
This last commit was so much easer to read! 😸

ttnghia:
build

hyperbolic2346 (approved):

PR #1584, opened 2023-11-21
Fix C++ `clang-format` style to preserve grouping in headers [skip ci]

ttnghia:
Currently, grouping in headers is not allowed:

Before:
```
#include <cudf/lists/lists_column_view.hpp>
#include <cudf/table/table_view.hpp>
#include <cudf/utilities/default_stream.hpp>

#include <thrust/transform.h>

#include <rmm/cuda_stream_view.hpp>

#include <memory>
```

After formatting:
```
#include <memory>

#include <cudf/lists/lists_column_view.hpp>
#include <cudf/table/table_view.hpp>
#include <cudf/utilities/default_stream.hpp>
#include <rmm/cuda_stream_view.hpp>
#include <thrust/transform.h>
```

The formatter should not change the header grouping in the input to respect to "near to far" conventional order. Only headers within each group are sorted alphabetically.

### DIFF
--- a/src/main/cpp/.clang-format
+++ b/src/main/cpp/.clang-format
@@ -133,7 +133,7 @@ ForEachMacros:
   - foreach
   - Q_FOREACH
   - BOOST_FOREACH
-IncludeBlocks:   Regroup
+IncludeBlocks:   Preserve
 IncludeCategories:
   - Regex:           '<[[:alnum:]]+>'
     Priority:        0
--- a/src/main/cpp/.clang-format
+++ b/src/main/cpp/.clang-format
@@ -1,3 +1,17 @@
+# =============================================================================
+# Copyright (c) 2019-2023, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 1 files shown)


src/main/cpp/.clang-format
jlowe, line 1: This file is missing copyrights and ASLv2 license.  Surprised it's not being flagged by the build checks.

ttnghia: Oh wow, I never realized this. The cudf's style also doesn't have copyright. I'll notify them.

gerashegalov: why does spark-rapids-jni needs our own style file? 

The pre-commit hook uses [the one from cudf](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-23.12/.pre-commit-config.yaml#L10). Can we not get all the necessary improvements via cudf?

ttnghia: Good point. I asked that before and vaguely remember that @jlowe said we have a reason to keep a different format config but can't remember.

ttnghia: @jlowe @gerashegalov Do you think this `.clang-format` version is redundant and should be removed completely?  We are now using commit hook style from cudf C++ thus we should start usinig it everywhere in this repository.

jlowe: I'm fine with using the cudf version if that solves more problems than it creates.  Agree that whatever precommit hook for spark-rapids-jni repo uses for formatting (note that thirdparty/cudf will potentially have its own, separate precommit hooks) should match what a manual clang format produces.

hyperbolic2346 (approved):

ttnghia:
Since we are using cudf's C++ style for precommit hook, I'm going to close this and will remove the current `.clang-format` file completely.

PR #1591, opened 2023-11-22
Fixing potential overflow in array read in string to float cast

### LINKED ISSUE #1574: [BUG] Invalid memory access bug in `string_to_float_kernel`
I ran `compute_sanitizer` on `CAST_STRING` test and found this:
```
[ RUN      ] StringToFloatTests/0.Simple
========= Invalid __shared__ read of size 8 bytes
=========     at 0x1f70 in void spark_rapids_jni::detail::string_to_float_kernel<float, (int)256>(T1 *, unsigned int *, int *, int *, const char *, const int *, const unsigned int *, int)
=========     by thread (128,0,0) in block (1,0,0)
=========     Address 0x98 is out of bounds
```

hyperbolic2346:
String to float cast would read into undefined memory if there were >= 19 digits in the string. This ultimately wasn't an issue in our test because the number of digits read to that point was 0, so the math was `(0 * undef) + ...`. Fixing it by creating a constant where the outer code can read it as well to ensure that it is large enough to handle the values.

Fixes #1574

### DIFF
--- a/src/main/cpp/src/cast_string_to_float.cu
+++ b/src/main/cpp/src/cast_string_to_float.cu
@@ -194,6 +194,8 @@ class string_to_float {
     compute_validity(_valid, _except);
   }
 
+  static constexpr int max_safe_digits = 19;
+
  private:
   // shuffle down to remove whitespace
   __device__ void remove_leading_whitespace()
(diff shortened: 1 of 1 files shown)


hyperbolic2346:
build

ttnghia (approved):
Please make sure to run `comupter-sanitizer` on the test.

PR #1607, opened 2023-11-30
[submodule-sync] bot-submodule-sync-branch-24.02 to branch-24.02 [skip ci] [bot]

nvauto:
submodule-sync to create a PR keeping thirdparty/cudf up-to-date.
HEAD commit SHA: 2b0d693138a568fc7870770006452500d6dc2b13, cudf commit SHA: https://github.com/rapidsai/cudf/commit/ac35d1950a8e34866706db10e627ef1a5a471e5d

This PR will be auto-merged if test passed. If failed, it will remain open until test pass or manually fix.

PR #1608, opened 2023-12-01
Fix faultinj build error after spdlog/fmt upgrade

jlowe:
Resolves #1607.  Recently cudf upgraded spdlog/fmt to a new version, and it now requires explicit formatting of enums or a compile error is generated.  This updates the faultinj logging to allow the necessary enums to be logged and also updates the cudf submodule to the latest on branch-24.02 that picks up the spdlog/fmt dependency upgrade.


jlowe:
build

gerashegalov (approved):
LGTM

PR #1623, opened 2023-12-07
Handle Decimal-128 Multiplication For Newer Spark Versions

razajafri:
This PR adds a new method for multiplying two Decimal 128-bit numbers without casting the interim result before casting to the scale the final answer is expected to be in. 

When multiplying two decimal numbers with a precision of 38, earlier versions of Spark are capped by the precision of 38 which causes the answer to be casted twice essentially, once right after multiplication and once before reporting the final answer which leads to the answer being inaccurate. 

For more details please look at the Spark [issue](https://issues.apache.org/jira/browse/SPARK-40129) which explains the issue in greater detail. 

**Changes Made:**
* Added a boolean param to control whether to apply the interim cast.
* Added a new `multiply128` method that takes in interimCast boolean. 
* Added a warning to the old method to highlight that it knowingly adds a bug to the multiplication to match Spark versions

**Tests:**
* Added a unit test

### DIFF
--- a/src/main/cpp/src/DecimalUtilsJni.cpp
+++ b/src/main/cpp/src/DecimalUtilsJni.cpp
@@ -19,8 +19,13 @@
 
 extern "C" {
 
-JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_DecimalUtils_multiply128(
-  JNIEnv* env, jclass, jlong j_view_a, jlong j_view_b, jint j_product_scale)
+JNIEXPORT jlongArray JNICALL
+Java_com_nvidia_spark_rapids_jni_DecimalUtils_multiply128(JNIEnv* env,
+                                                          jclass,
+                                                          jlong j_view_a,
+                                                          jlong j_view_b,
+                                                          jint j_product_scale,
+                                                          bool cast_interim_result)
 {
   JNI_NULL_CHECK(env, j_view_a, "column is null", 0);
   JNI_NULL_CHECK(env, j_view_b, "column is null", 0);
@@ -30,7 +35,7 @@ JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_DecimalUtils_multi
     auto view_b = reinterpret_cast<cudf::column_view const*>(j_view_b);
     auto scale  = static_cast<int>(j_product_scale);
     return cudf::jni::convert_table_for_return(
-      env, cudf::jni::multiply_decimal128(*view_a, *view_b, scale));
+      env, cudf::jni::multiply_decimal128(*view_a, *view_b, scale, cast_interim_result));
   }
   CATCH_STD(env, 0);
 }
--- a/src/main/cpp/src/decimal_utils.cu
+++ b/src/main/cpp/src/decimal_utils.cu
@@ -681,11 +681,11 @@ struct dec128_multiplier {
 
     int mult_scale = a_scale + b_scale;
 
-    // Spark does some really odd things that I personally think are a bug
-    // https://issues.apache.org/jira/browse/SPARK-40129
-    // But to match Spark we need to first round the result to a precision of 38
-    // and this is specific to the value in the result of the multiply.
-    // Then we need to round the result to the final scale that we care about.
+    // According to https://issues.apache.org/jira/browse/SPARK-40129
+    // and https://issues.apache.org/jira/browse/SPARK-45786, Spark has a bug in versions 3.2.4, 3.3.3, 3.4.1, 3.5.0
+    // and 4.0.0
+    // The bug is fixed for later versions but to match the legacy behavior we need to first round the result to a
+    // precision of 38 then we need to round the result to the final scale that we care about.
     if (cast_interim_result) {
       int first_div_precision = dec_precision - 38;
       if (first_div_precision > 0) {
--- a/src/main/cpp/src/decimal_utils.cu
+++ b/src/main/cpp/src/decimal_utils.cu
@@ -682,10 +682,10 @@ struct dec128_multiplier {
     int mult_scale = a_scale + b_scale;
 
     // According to https://issues.apache.org/jira/browse/SPARK-40129
-    // and https://issues.apache.org/jira/browse/SPARK-45786, Spark has a bug in versions 3.2.4, 3.3.3, 3.4.1, 3.5.0
-    // and 4.0.0
-    // The bug is fixed for later versions but to match the legacy behavior we need to first round the result to a
-    // precision of 38 then we need to round the result to the final scale that we care about.
+    // and https://issues.apache.org/jira/browse/SPARK-45786, Spark has a bug in
+    // versions 3.2.4, 3.3.3, 3.4.1, 3.5.0 and 4.0.0 The bug is fixed for later versions but to
+    // match the legacy behavior we need to first round the result to a precision of 38 then we need
+    // to round the result to the final scale that we care about.
     if (cast_interim_result) {
       int first_div_precision = dec_precision - 38;
       if (first_div_precision > 0) {
--- a/src/main/cpp/src/decimal_utils.cu
+++ b/src/main/cpp/src/decimal_utils.cu
@@ -975,7 +975,7 @@ namespace cudf::jni {
 std::unique_ptr<cudf::table> multiply_decimal128(cudf::column_view const& a,
                                                  cudf::column_view const& b,
                                                  int32_t product_scale,
-                                                 bool const& cast_interim_result,
+                                                 bool const cast_interim_result,
                                                  rmm::cuda_stream_view stream)
 {
   CUDF_EXPECTS(a.type().id() == cudf::type_id::DECIMAL128, "not a DECIMAL128 column");
--- a/src/main/cpp/src/decimal_utils.cu
+++ b/src/main/cpp/src/decimal_utils.cu
@@ -658,7 +658,7 @@ struct dec128_multiplier {
                     cudf::mutable_column_view const& product_view,
                     cudf::column_view const& a_col,
                     cudf::column_view const& b_col,
-                    bool const& cast_interim_result)
+                    bool const cast_interim_result)
     : overflows(overflows),
       a_data(a_col.data<__int128_t>()),
       b_data(b_col.data<__int128_t>()),
@@ -679,23 +679,24 @@ struct dec128_multiplier {
 
     int dec_precision = precision10(product);
 
-    int mult_scale = a_scale + b_scale;
+    int const mult_scale = [&]() {
+        // According to https://issues.apache.org/jira/browse/SPARK-40129
+        // and https://issues.apache.org/jira/browse/SPARK-45786, Spark has a bug in
+        // versions 3.2.4, 3.3.3, 3.4.1, 3.5.0 and 4.0.0 The bug is fixed for later versions but to
+        // match the legacy behavior we need to first round the result to a precision of 38 then we need
+        // to round the result to the final scale that we care about.
+        if (cast_interim_result) {
+          int first_div_precision = dec_precision - 38;
+          if (first_div_precision > 0) {
+            auto const first_div_scale_divisor = pow_ten(first_div_precision).as_128_bits();
+            product                            = divide_and_round(product, first_div_scale_divisor);
 
-    // According to https://issues.apache.org/jira/browse/SPARK-40129
-    // and https://issues.apache.org/jira/browse/SPARK-45786, Spark has a bug in
-    // versions 3.2.4, 3.3.3, 3.4.1, 3.5.0 and 4.0.0 The bug is fixed for later versions but to
-    // match the legacy behavior we need to first round the result to a precision of 38 then we need
-    // to round the result to the final scale that we care about.
-    if (cast_interim_result) {
-      int first_div_precision = dec_precision - 38;
-      if (first_div_precision > 0) {
-        auto const first_div_scale_divisor = pow_ten(first_div_precision).as_128_bits();
-        product                            = divide_and_round(product, first_div_scale_divisor);
-
-        // a_scale and b_scale are negative. first_div_precision is not
-        mult_scale = a_scale + b_scale + first_div_precision;
-      }
-    }
+            // a_scale and b_scale are negative. first_div_precision is not
+            return a_scale + b_scale + first_div_precision;
+          }
+        }
+        return a_scale + b_scale;
+    }();
 
     int exponent = prod_scale - mult_scale;
     if (exponent < 0) {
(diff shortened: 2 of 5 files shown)


tgravescs:
related to https://github.com/NVIDIA/spark-rapids/issues/9859

razajafri:
@tgravescs I think I have answered your previous comment. Do you have any other comments?

src/main/cpp/src/decimal_utils.cu
```diff
@@ -675,21 +677,24 @@ struct dec128_multiplier {
 
     chunked256 product = multiply(a, b);
 
+    int dec_precision = precision10(product);
+
+    int mult_scale = a_scale + b_scale;
+
     // Spark does some really odd things that I personally think are a bug
```
revans2, line 684: nit: Could we update this comment to explain what some versions of Spark do this and it has been fixed.

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -25,11 +25,32 @@ public class DecimalUtils {
...
+   * even if the number still fits in a 128-bit representation.
+   * @param a factor input, must match row count of the other factor input
+   * @param b factor input, must match row count of the other factor input
+   * @param productScale scale to use for the product type
+   * @return table containing a boolean column and a DECIMAL128 product column of the specified
+   *         scale. The boolean value will be true if an overflow was detected for that row's
+   *         DECIMAL128 product value. A null input row will result in a corresponding null output
+   *         row.
+   */
+  public static Table mul128(ColumnView a, ColumnView b, int productScale) {
```
revans2, line 40: Can we please just expose the boolean instead of having it hidden with `mul128` vs `multiply128`. It is too subtle of a name change to stand out if there is a bug. Or name them `multiply128Corrected` and then `multiply128SparkLegacy`, but keep the original `multiply128` until the plugin is updated.

razajafri:
build

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
razajafri: Originally I wanted to pass in a boolean but talked myself out of it because I thought passing a boolean was not very pretty and ruled in favor of having a separate method. But coming up with a name was challenging and considering that you are raising this, I think I will just pass a boolean parameter.

revans2 (approved):

src/main/cpp/src/decimal_utils.cu
```diff
@@ -675,21 +677,24 @@ struct dec128_multiplier {
...
+    // to round the result to the final scale that we care about.
+    if (cast_interim_result) {
+      int first_div_precision = dec_precision - 38;
+      if (first_div_precision > 0) {
+        auto const first_div_scale_divisor = pow_ten(first_div_precision).as_128_bits();
+        product                            = divide_and_round(product, first_div_scale_divisor);
+
+        // a_scale and b_scale are negative. first_div_precision is not
+        mult_scale = a_scale + b_scale + first_div_precision;
+      }
```
hyperbolic2346, line 697:
```suggestion
    int const mult_scale = [&]() {
      // According to https://issues.apache.org/jira/browse/SPARK-40129
      // and https://issues.apache.org/jira/browse/SPARK-45786, Spark has a bug in
      // versions 3.2.4, 3.3.3, 3.4.1, 3.5.0 and 4.0.0 The bug is fixed for later versions but to
      // match the legacy behavior we need to first round the result to a precision of 38 then we need
      // to round the result to the final scale that we care about.
      if (cast_interim_result) {
        int first_div_precision = dec_precision - 38;
        if (first_div_precision > 0) {
          auto const first_div_scale_divisor = pow_ten(first_div_precision).as_128_bits();
          product                            = divide_and_round(product, first_div_scale_divisor);

          // a_scale and b_scale are negative. first_div_precision is not
          return a_scale + b_scale + first_div_precision;
      }
      return a_scale + b_scale;
      })();
```
This isn't 100% correct due to the diff not picking up everything, but the general idea here is that we can make this const if we use an immediately evaluated lambda.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -968,6 +974,7 @@ namespace cudf::jni {
 std::unique_ptr<cudf::table> multiply_decimal128(cudf::column_view const& a,
                                                  cudf::column_view const& b,
                                                  int32_t product_scale,
+                                                 bool const& cast_interim_result,
```
hyperbolic2346, line 977:
```suggestion
                                                 bool const cast_interim_result,
```
No real advantage to passing it as a pointer if it is const and smaller than a pointer.

src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java
```diff
@@ -86,6 +86,22 @@ void simplePosMultiplyZeroByNegOne() {
...
+  void multiply128WithoutInterimCast() {
+    try (ColumnVector lhs =
+                 makeDec128Column("-8533444864753048107770677711.1312637916");
+         ColumnVector rhs =
+                 makeDec128Column("-12.0000000000");
+         ColumnVector expectedBasic =
+                 makeDec128Column("102401338377036577293248132533.575165");
+         ColumnVector expectedValid =
+                 ColumnVector.fromBooleans(false);
+         Table found = DecimalUtils.multiply128(lhs, rhs, -6, false)) {
```
hyperbolic2346, line 99: Should there be an explicit test for the interim cast as well?

hyperbolic2346 (changes requested):
Some minor nits, looks good overall and I like the comments added.

src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java
razajafri: Yeah, there are. All the multiply128 tests call into the one with `interimCast=true`. Here is one example https://github.com/NVIDIA/spark-rapids-jni/blob/f6bb2052af9d3bce60d087c758c76b6d77b9936f/src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java#L74

src/main/cpp/src/decimal_utils.cu
razajafri: It's always evaluating `cast_interim_result` to false which always results in test failures.

razajafri:
build

razajafri:
build

hyperbolic2346 (dismissed):

razajafri:
CI failing because of 
```
undefined reference to `cudf::jni::multiply_decimal128(cudf::column_view const&, cudf::column_view const&, int, bool const&, rmm::cuda_stream_view)'
```
but passes locally even after doing a clean build

razajafri:
build

src/main/cpp/src/DecimalUtilsJni.cpp
```diff
@@ -19,8 +19,13 @@
...
 
-JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_DecimalUtils_multiply128(
-  JNIEnv* env, jclass, jlong j_view_a, jlong j_view_b, jint j_product_scale)
+JNIEXPORT jlongArray JNICALL
+Java_com_nvidia_spark_rapids_jni_DecimalUtils_multiply128(JNIEnv* env,
+                                                          jclass,
+                                                          jlong j_view_a,
+                                                          jlong j_view_b,
+                                                          jint j_product_scale,
+                                                          bool cast_interim_result)
```
ttnghia, line 28: The question now is, since this new parameter is added to avoid a known bug, will it be turned on all the time to fix the bug, or will it be turned on/off depending on the situations? If so then when it is turned on?

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -25,21 +25,50 @@ public class DecimalUtils {
...
+   *                    precision
    * @return table containing a boolean column and a DECIMAL128 product column of the specified
    *         scale. The boolean value will be true if an overflow was detected for that row's
    *         DECIMAL128 product value. A null input row will result in a corresponding null output
    *         row.
    */
-  public static Table multiply128(ColumnView a, ColumnView b, int productScale) {
-    return new Table(multiply128(a.getNativeView(), b.getNativeView(), productScale));
+  public static Table multiply128(ColumnView a, ColumnView b, int productScale, boolean interimCast) {
+    return new Table(multiply128(a.getNativeView(), b.getNativeView(), productScale, interimCast));
```
ttnghia, line 71: If the `interimCast` is applied to Spark versions 3.2.4, 3.3.3, 3.4.1, 3.5.0 and 4.0.0 then please add such clarification into the docs of this function.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -675,22 +677,26 @@ struct dec128_multiplier {
...
-
-    int mult_scale = a_scale + b_scale;
-    if (first_div_precision > 0) {
-      auto const first_div_scale_divisor = pow_ten(first_div_precision).as_128_bits();
-      product                            = divide_and_round(product, first_div_scale_divisor);
-
-      // a_scale and b_scale are negative. first_div_precision is not
-      mult_scale = a_scale + b_scale + first_div_precision;
-    }
+    int dec_precision = precision10(product);
```
ttnghia, line 680: `dec_precision` is used only once below. Please move it right above the line using it.

ttnghia: Also please add `const` qualifier to all the variables.

src/main/cpp/src/decimal_utils.cu
```diff
@@ -675,22 +677,26 @@ struct dec128_multiplier {
...
+
+    int const mult_scale = [&]() {
+      // According to https://issues.apache.org/jira/browse/SPARK-40129
+      // and https://issues.apache.org/jira/browse/SPARK-45786, Spark has a bug in
+      // versions 3.2.4, 3.3.3, 3.4.1, 3.5.0 and 4.0.0 The bug is fixed for later versions but to
+      // match the legacy behavior we need to first round the result to a precision of 38 then we
+      // need to round the result to the final scale that we care about.
+      if (cast_interim_result) {
+        int first_div_precision = dec_precision - 38;
+        if (first_div_precision > 0) {
```
ttnghia, line 690:
```suggestion
        if (precision10(product) > 38) {
```

src/main/cpp/src/decimal_utils.cu
```diff
@@ -675,22 +677,26 @@ struct dec128_multiplier {
...
+    int dec_precision = precision10(product);
+
+    int const mult_scale = [&]() {
+      // According to https://issues.apache.org/jira/browse/SPARK-40129
+      // and https://issues.apache.org/jira/browse/SPARK-45786, Spark has a bug in
+      // versions 3.2.4, 3.3.3, 3.4.1, 3.5.0 and 4.0.0 The bug is fixed for later versions but to
+      // match the legacy behavior we need to first round the result to a precision of 38 then we
+      // need to round the result to the final scale that we care about.
+      if (cast_interim_result) {
+        int first_div_precision = dec_precision - 38;
```
ttnghia, line 689:
```suggestion
        int first_div_precision = precision10(product) - 38;
```

src/main/cpp/src/decimal_utils.cu
```diff
@@ -718,6 +724,7 @@ struct dec128_multiplier {
  private:
   // output column for overflow detected
   bool* const overflows;
+  bool cast_interim_result;
```
ttnghia, line 727:
```suggestion
  bool const cast_interim_result;
```

src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java
```diff
@@ -86,6 +86,22 @@ void simplePosMultiplyZeroByNegOne() {
...
+  @Test
+  void multiply128WithoutInterimCast() {
+    try (ColumnVector lhs =
+                 makeDec128Column("-8533444864753048107770677711.1312637916");
+         ColumnVector rhs =
+                 makeDec128Column("-12.0000000000");
+         ColumnVector expectedBasic =
+                 makeDec128Column("102401338377036577293248132533.575165");
+         ColumnVector expectedValid =
+                 ColumnVector.fromBooleans(false);
```
ttnghia, line 98: I guess the code was not formated?

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
razajafri: I thought the docs were pretty clear, if there is anything else you want me to add please share and I will add.

src/main/cpp/src/DecimalUtilsJni.cpp
razajafri: The new parameter is only set to false for Spark versions for which the multiplication bug was fixed. Namely Spark versions 3.4.2, 4.0.0, 3.5.1

src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java
razajafri: It is formatted, but it could be improved by shifting things on the same line :) 

Thanks for pointing this out, it looked ugly

razajafri:
build

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -25,21 +25,50 @@ public class DecimalUtils {
...
+   * WARNING: With interimCast set to  true, this method has a bug which we match with Spark versions before 3.4.2,
+   * 4.0.0, 3.5.1. Consider the following example using Decimal with a precision of 38 and scale of 10:
+   * -8533444864753048107770677711.1312637916 * -12.0000000000 = 102401338377036577293248132533.575166
+   * while the actual answer based on Java BigDecimal is 102401338377036577293248132533.575165
+   *
    * @param a factor input, must match row count of the other factor input
    * @param b factor input, must match row count of the other factor input
    * @param productScale scale to use for the product type
+   * @param interimCast whether to cast the result of the division to 38 precision before casting it again to the final
+   *                    precision
```
ttnghia, line 64: Nit: Sorry, I think these lines are longer than usual thus we may need to rewrite them a little bit.

ttnghia (approved):

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
razajafri: Yeah, good point. I see the file has other instances where we are doing this. To keep this PR focused, I don't want to make the changes in other places and since the CI has passed I would really appreciate if we can do that as a follow-on.

razajafri:
Thanks @ttnghia I will file a follow-on for the line length! appreciate your help

PR #1637, opened 2023-12-11
Make OOM injections separable by host & device

gerashegalov:
This PR makes it possible to inject 
- host OOM only
- device OOM only
- in addition to the current mixed OOM injections

It also enables deferring OOM to the N+1st allocation by skipping first N


src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -226,6 +226,24 @@ struct task_metrics {
...
+
+struct oom_state_type {
+  int hit_count   = 0;
+  int skip_count  = 0;
+  oom_type filter = oom_type::CPU_OR_GPU;
+
+  bool matches(bool is_for_cpu)
+  {
+    return (is_for_cpu && (filter == oom_type::CPU || filter == oom_type::CPU_OR_GPU)) ||
+           (!is_for_cpu && (filter == oom_type::GPU || filter == oom_type::CPU_OR_GPU));
```
abellina, line 243: nit, slightly cleaner? 
```suggestion
    return filter == oom_type::CPU_OR_GPU || 
           is_for_cpu && filter == oom_type::CPU ||
           !is_for_cpu && filter == oom_type::GPU;
```

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -226,6 +226,24 @@ struct task_metrics {
   }
 };
 
+enum class oom_type {
+  CPU_OR_GPU = 0,
+  CPU,
+  GPU,
+};
+
+struct oom_state_type {
```
abellina, line 235: should this just be `oom_state`?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -663,12 +683,23 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
    */
-  void force_retry_oom(long const thread_id, int const num_ooms)
-  {
+  void force_retry_oom(long const thread_id,
+                       int const num_ooms,
+                       int const oom_filter,
+                       int const skip_count)
+  {
+    if (num_ooms < 0) { throw std::invalid_argument("num_ooms cannot be negative"); }
+    if (skip_count < 0) { throw std::invalid_argument("skipCount cannot be negative"); }
```
abellina, line 692:
```suggestion
    if (skip_count < 0) { throw std::invalid_argument("skip_count cannot be negative"); }
```

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1228,15 +1262,21 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+          logger->info("SKIPPING INJECTED_RETRY_OOM({}) thread_id={},task_id={},skip_count={}",
+                       is_for_cpu ? "CPU" : "GPU",
+                       thread_id,
+                       thread->second.task_id,
+                       thread->second.retry_oom.skip_count);
+          --thread->second.retry_oom.skip_count;
+        } else if (thread->second.retry_oom.hit_count > 0) {
+          thread->second.retry_oom.hit_count--;
+          thread->second.metrics.num_times_retry_throw++;
+          log_status("INJECTED_RETRY_OOM", thread_id, thread->second.task_id, thread->second.state);
```
abellina, line 1276: should this be `INJECTED_RETRY_OOM([CPU|GPU])`?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1248,16 +1288,28 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+            "SKIPPING INJECTED_SPLIT_AND_RETRY_OOM({}) thread_id={},task_id={},skip_count={}",
+            is_for_cpu ? "CPU" : "GPU",
+            thread_id,
+            thread->second.task_id,
+            thread->second.retry_oom.skip_count);
+          thread->second.split_and_retry_oom.skip_count--;
+        } else if (thread->second.split_and_retry_oom.hit_count > 0) {
+          thread->second.split_and_retry_oom.hit_count--;
+          thread->second.metrics.num_times_split_retry_throw++;
+          log_status("INJECTED_SPLIT_AND_RETRY_OOM",
```
abellina, line 1303: same comment here, do we want to update the status to include CPU|GPU?

src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
```diff
@@ -23,11 +23,59 @@
...
+            Integer.parseInt(injectionConf.getOrDefault("skip", "0")),
+            Integer.parseInt(injectionConf.getOrDefault("num_ooms", "1")));
+      }
+    }
+
+    private static OomInjection defaultOomInjection() {
+      return new OomInjection(OomInjectionType.CPU_OR_GPU, 0, 1);
+    }
+
+    private static OomInjection zerInjection() {
```
abellina, line 69: `noopInjection`?

abellina (commented):
Looks pretty good so far. Mostly minor comments.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1228,15 +1262,21 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
-        thread->second.metrics.num_times_retry_throw++;
-        log_status("INJECTED_RETRY_OOM", thread_id, thread->second.task_id, thread->second.state);
-        thread->second.record_failed_retry_time();
-        if (is_for_cpu) {
-          throw_java_exception(CPU_RETRY_OOM_CLASS, "injected RetryOOM");
-        } else {
-          throw_java_exception(GPU_RETRY_OOM_CLASS, "injected RetryOOM");
+      if (thread->second.retry_oom.matches(is_for_cpu)) {
+        if (thread->second.retry_oom.skip_count > 0) {
+          logger->info("SKIPPING INJECTED_RETRY_OOM({}) thread_id={},task_id={},skip_count={}",
```
revans2, line 1267: nit: The logger was setup with a very specific pattern to make it so we could parse it. This is not that hard to filter out, but why not just have it do a log_status? or change it to debug like the other one.

src/main/java/com/nvidia/spark/rapids/jni/RmmSpark.java
```diff
@@ -23,11 +23,59 @@
...
+    public final int skipCount;
+    public final int oomCount;
+
+    private OomInjection(OomInjectionType injectionType, int skipCount, int oomCount) {
+      this.injectionType = injectionType;
+      this.skipCount = skipCount;
+      this.oomCount = oomCount;
+    }
+
+    public static OomInjection fromString(String s) {
```
revans2, line 51: At a minimum can we have documentation on what this does? "true" and "false" I guess makes since, but why make it all lower case in the plugin instead of making this more lenient on the inputs? Also it would be really nice to document the map config pattern, and what happens if you provide values that are not skip or num_ooms. and what skip/num_ooms is intended to do.

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
```diff
@@ -182,13 +184,21 @@ public void doneWaitingOnPool(long threadId) {
...
+
   /**
    * Force the thread with the given ID to throw a GpuRetryOOM on their next allocation attempt.
    * @param threadId the ID of the thread to throw the exception (not java thread id).
    * @param numOOMs the number of times the GpuRetryOOM should be thrown
    */
-  public void forceRetryOOM(long threadId, int numOOMs) {
-    forceRetryOOM(getHandle(), threadId, numOOMs);
+  public void forceRetryOOM(long threadId, int numOOMs, int oomMode, int skipCount) {
+    assert numOOMs >= 0 : "non-negative numOoms expected: actual=" + numOOMs;
```
revans2, line 197: nit: Would it be better to have these checks in the C++ code? If bad things can happen if we set these values wrong (like an infinite loop), then C++ would be better to avoid the possibility of ever setting them wrong even when assertions are disabled. But if nothing too bad is going to happen, then here is fine as this really is just test code anyways.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -683,7 +714,10 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     std::unique_lock<std::mutex> lock(state_mutex);
     auto const threads_at = threads.find(thread_id);
     if (threads_at != threads.end()) {
-      threads_at->second.split_and_retry_oom_injected = num_ooms;
+      auto& oom_state      = threads_at->second.split_and_retry_oom;
+      oom_state.hit_count  = num_ooms;
+      oom_state.filter     = oom_type::CPU_OR_GPU;
+      oom_state.skip_count = 0;
```
jbrennan333, line 720: Looks like we've added the some of the code to handle this below.  Maybe we should just do it all in this PR?

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
```diff
@@ -182,13 +184,21 @@ public void doneWaitingOnPool(long threadId) {
...
 
+  public void forceRetryOOM(long threadId, int numOOMs) {
+    forceRetryOOM(getHandle(), threadId, numOOMs, RmmSpark.OomInjectionType.CPU_OR_GPU.ordinal(), 0);
+  }
+
   /**
    * Force the thread with the given ID to throw a GpuRetryOOM on their next allocation attempt.
    * @param threadId the ID of the thread to throw the exception (not java thread id).
    * @param numOOMs the number of times the GpuRetryOOM should be thrown
    */
```
jbrennan333, line 195: Add new parameters.

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
gerashegalov: I am doing it in C++ as well to catch stale `native` method signatures in Java.

jbrennan333 (commented):
I'm not sure I understand why the OomInjection class lives in RmmSpark in the JNI?
It seems like it is dealing with parsing the spark-rapids plugin configuration strings to convert them into the parameters for calls to RmmSpark.forceRetryOOM, etc...   Wouldn't it be better if that lived in the plugin?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
gerashegalov: IMO in snake case it becomes cumbersome when you want to have a local variable

camel case is fine `OomState oomState` but `oom_state oom_state` is not

revans2 (approved):

jbrennan333 (approved):
Thanks for the updates!   This looks good to me.

revans2 (approved):

gerashegalov:
build

Issue #1579, opened 2023-11-21
[BUG] Investigate `inclusive_scan` issue in column-row conversion code


ttnghia:
In https://github.com/NVIDIA/spark-rapids-jni/issues/1567, we have the issue with invalid memory access:
```
C++ exception with description "inclusive_scan failed to synchronize: cudaErrorIllegalAddress: an illegal memory access was encountered" thrown in the test body.
```

https://github.com/NVIDIA/spark-rapids-jni/pull/1577 is a temporary fix, but that does not fix the root cause of the problem and we should spend more time investigating what is really wrong here.

sameerz:
We are not calling the code referenced in #1577 so I am moving this to 24.02 for investigation.

ttnghia:
After investigation, we found that the issue is due to kernel visibility problem. It has been fixed by https://github.com/rapidsai/rapids-cmake/pull/523 and https://github.com/NVIDIA/cuCollections/pull/422.

Close this as resolved.

Issue #1567, opened 2023-11-15
[BUG] nightly test failed ROW_CONVERSION constantly


pxLi:
**Describe the bug**
spark-rapids-jni_nightly-pre_release, build ID: 204

~~currently we only saw this in cuda12 once~~, jni ref: ff59e68cf91b4f3154355a67b5c4ce45bd9b1417, cudf ref: https://github.com/rapidsai/cudf/commit/330d389b26a05676d9f079503a3d96b571762337

we started seeing this more frequent in also submodule sync up pipeline (cuda11 ENV)
spark-rapids-jni_submodule-sync-pre_release, build ID **593**, 591, 592 (constantly failing now)

failed 
```
[2023-11-15T06:06:25.878Z] [INFO]      [exec] The following tests FAILED:
[2023-11-15T06:06:25.878Z] [INFO]      [exec] 	  4 - ROW_CONVERSION (Failed)
[2023-11-15T06:06:25.878Z] [INFO]      [exec] Errors while running CTest
```

from target/cmake-build/Testing/Temporary/LastTest.log (full log [204.log](https://github.com/NVIDIA/spark-rapids-jni/files/13360996/204.log))
```
4/8 Testing: ROW_CONVERSION
4/8 Test: ROW_CONVERSION
Command: "/home/jenkins/agent/workspace/jenkins-spark-rapids-jni_nightly-pre_release-204-cuda12/target/cmake-build/gtests/ROW_CONVERSION"
Directory: /home/jenkins/agent/workspace/jenkins-spark-rapids-jni_nightly-pre_release-204-cuda12/target/cmake-build/tests
"ROW_CONVERSION" start time: Nov 15 06:06 UTC
Output:
----------------------------------------------------------
Running main() from /home/jenkins/agent/workspace/jenkins-spark-rapids-jni_nightly-pre_release-204-cuda12/thirdparty/cudf/cpp/build/_deps/gtest-src/googletest/src/gtest_main.cc
[==========] Running 28 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 13 tests from ColumnToRowTests
[ RUN      ] ColumnToRowTests.Single
[       OK ] ColumnToRowTests.Single (587 ms)
[ RUN      ] ColumnToRowTests.SimpleString
[       OK ] ColumnToRowTests.SimpleString (14 ms)
[ RUN      ] ColumnToRowTests.DoubleString
[...]

PR #1662, opened 2023-12-19
Use cuda::proclaim_return_type on device lambdas

### LINKED ISSUE #1639: Audit code for proclaim return type changes
rapidsai/cudf#14576 is updating to CCCL 2.2.0, and that requires using cuda::proclaim_return_type.  We need to audit our spark-rapids-jni kernels to see if any need to be updated before this breaking change is merged into cudf.

This PR shows the kind of changes that will be needed to make CCCL 2.2.0 work:  https://github.com/rapidsai/cudf/pull/14577

hyperbolic2346:
This PR makes spark-rapids-jni compatible with Thrust 2 by adding `cuda::proclaim_return_type`. This is following suit with cudf, who recently did this work with https://github.com/rapidsai/cudf/pull/14577

closes #1639


hyperbolic2346:
build

ttnghia:
Although I was requested to review, I've never received any email notification?

src/main/cpp/src/parse_uri.cu
```diff
@@ -164,11 +166,11 @@ bool __device__ validate_ipv6(string_view s)
   int address_char_count{0};
   bool address_has_hex{false};
 
-  auto const leading_double_colon = [&]() {
+  auto const leading_double_colon = cuda::proclaim_return_type<bool>([&]() {
```
ttnghia, line 169: This is not device lambda so don't need to change here.

ttnghia: Wait, I just realized that this lambda is unused. The only place that refers to it is https://github.com/NVIDIA/spark-rapids-jni/blob/67845706969fe7c9495e2578db7de278729841fe/src/main/cpp/src/parse_uri.cu#L212 which is commented out.

hyperbolic2346: Yeah. I wrote it all to match the RFC, but Spark/Java don't follow it exactly and I need to match them. I documented the differences by commenting out the original path. I don't know the take on this and if I should remove it or not.

ttnghia: If you commented out the caller then you should also comment out the function too, since it is now unused.

mythrocks:
For the record, we'll also need to change `CMakeLists.txt`, to fetch CCCL 2:
```diff
diff --git a/src/main/cpp/CMakeLists.txt b/src/main/cpp/CMakeLists.txt
index 18c0cd1..a905297 100644
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -94,11 +94,8 @@ include(cmake/Modules/ConfigureCUDA.cmake) # set other CUDA compilation flags
 # ##################################################################################################
 # * dependencies ----------------------------------------------------------------------------------
 
-# find libcu++
-include(${rapids-cmake-dir}/cpm/libcudacxx.cmake)
-
-# find thrust/cub
-include(${CUDF_DIR}/cpp/cmake/thirdparty/get_thrust.cmake)
+# find CCCL
+include(${CUDF_DIR}/cpp/cmake/thirdparty/get_cccl.cmake)
 
 # JNI
 find_package(JNI REQUIRED)
```

mythrocks:
I think the changes in this PR should be good to go.  

They now include the changes I had done independently. (i.e. changes to `CMakeLists.txt`, `map_utils.cu`, and `row_conversion.cu`.)

I do see that we have trouble running `tests/ROW_CONVERSION` successfully with CCCL2. But that might well be a pre-existing problem.

ttnghia:
Yes the issue I was trying to solve (https://github.com/NVIDIA/spark-rapids-jni/issues/1567) now shows up---my previous fix doesn't work anymore. We now need to fix it some other way.

mythrocks:
My conservative vote would be not to change the `thirdparty/cudf` submodule commit hash to go to CCCL-2, for now.

But we need to, there might be value in commenting out `RowToColumnTests.ManyStrings` to allow the build to run through with CCCL-2.  We can then chase down the `inclusive_scan` problem that `row_conversion.cu` runs into (#1579).

abellina:
> My conservative vote would be not to change the thirdparty/cudf submodule commit hash to go to CCCL-2, for now.

I am confused @mythrocks, why do you want to hold?

mythrocks:
I'm :+1: on this change. We might need to rebase it for the latest in `parse_uri.cu`.

mythrocks:
> I am confused @mythrocks, why do you want to hold?

@abellina: Don't mind me. I'm irrationally apprehensive about disruptive changes near the holidays.

hyperbolic2346:
I feel the same way. I don't see the value in checking in something and then going on holiday. So many are out until the new year already. If we run across some major issue, we may not be around to fix or triage it. I'd prefer to leave things running until January than check this in and have unknown issues crop up during that reduced productivity phase. I'm all for fast-following cudf, but I don't see that making this change to CCCL 2.2 now vs in 2 weeks makes a huge difference and it comes with a potential for extended disruptions.

This ship may have already sailed though since even with a pinned cudf, we are pulling in the latest rmm and our builds are broken.

hyperbolic2346:
> I'm 👍 on this change. We might need to rebase it for the latest in `parse_uri.cu`.

Done

hyperbolic2346:
build

ttnghia:
We need to wait until tomorrow so the changes in cudf can propagate here to build.

src/main/cpp/src/row_conversion.cu
```diff
@@ -2456,9 +2467,10 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const& input,
     std::vector<char*> string_data_col_ptrs;
     for (auto& col_string_lengths : string_lengths) {
       device_uvector<size_type> output_string_offsets(num_rows + 1, stream, mr);
-      auto tmp = [num_rows, col_string_lengths] __device__(auto const& i) {
-        return i < num_rows ? col_string_lengths[i] : 0;
-      };
+      auto tmp = cuda::proclaim_return_type<int32_t>(
```
bdice, line 2470: Match the type of `output_string_offsets`.
```suggestion
      auto tmp = cuda::proclaim_return_type<size_type>(
```

bdice (commented):
All the changes thus far look good to me. One comment on `size_type` vs. `int32_t`.

mythrocks (dismissed):
LGTM, save for what @bdice suggested regarding `size_type` vs `int32_t`.

src/main/cpp/CMakeLists.txt
```diff
@@ -94,11 +94,8 @@ include(cmake/Modules/ConfigureCUDA.cmake) # set other CUDA compilation flags
 # ##################################################################################################
 # * dependencies ----------------------------------------------------------------------------------
 
-# find libcu++
-include(${rapids-cmake-dir}/cpm/libcudacxx.cmake)
-
-# find thrust/cub
-include(${CUDF_DIR}/cpp/cmake/thirdparty/get_thrust.cmake)
+# find CCCL
+include(${CUDF_DIR}/cpp/cmake/thirdparty/get_cccl.cmake)
```
bdice, line 98: This can be reverted. We only need the C++ changes for `cuda::proclaim_return_type` in this PR. Then we can update everything (including telling CMake to use CCCL instead of Thrust/libcudacxx) in a follow-up PR.

hyperbolic2346:
build

ttnghia:
Depends on https://github.com/NVIDIA/spark-rapids-jni/pull/1668.

src/main/cpp/CMakeLists.txt
ttnghia: Now with the latest cudf commit we can't compile if don't change this.

ttnghia:
build

src/main/cpp/CMakeLists.txt
bdice: Correct. If you want to use a newer cudf build, you need the CMake changes. I was hoping we could merge the C++ changes separately but it's fine to combine (nonbreaking) C++ changes with (breaking) CMake changes and submodule updates in this PR for expediency.

mythrocks (approved):

bdice (approved):
The CCCL 2.2.0 C++ and CMake changes look good to me.

PR #1667, opened 2023-12-20
Test with latest CCCL.

bdice:
**This is a draft.**

This PR adds CMake logic to override versions of CPM dependencies used by cuDF. This allows us to use a newer CCCL build to see if a bug in CUB is fixed.


bdice:
Closing this PR - it is no longer needed.

PR #1670, opened 2023-12-30
Make the GpuTimeZoneDB class idempotent

NVnavkumar:
Fixes https://github.com/NVIDIA/spark-rapids/issues/10129.

The GpuTimeZoneDB class needs to be able to recover from `shutdown()` being called and potentially being revived again. This can happen when the executor plugin context is shutdown but the JVM is reused with a new executor plugin context.


NVnavkumar:
build

jlowe (approved):

PR #1676, opened 2024-01-04
Fix a bug in format_float kernel

thirtiseven:
`format_number(0.485362439659, 13)` will get 1.8536243965900 in the plugin. It's a kernel bug.

Current `round_half_even` will handle trailing zeros, but it's unnecessary because it will be handled later anyway, and it will confuse the following logic which checks the length of the rounded number.

Also add some minor refactoring and testing.

### DIFF
--- a/src/main/cpp/src/ftos_converter.cuh
+++ b/src/main/cpp/src/ftos_converter.cuh
@@ -1202,11 +1202,8 @@ __device__ inline T round_half_even(T const input, int const olength, int const 
 {
   // "round" a integer to digits digits, with the half-even rounding mode.
   if (digits > olength) {
-    T num = input;
-    for (int i = 0; i < digits - olength; i++) {
-      num *= 10;
-    }
-    return num;
+    // trailing zeros will be handled later
+    return input;
   }
   T div = POW10_TABLE[olength - digits];
   T mod = input % div;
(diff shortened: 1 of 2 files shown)


thirtiseven:
build

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1215,10 +1212,10 @@ __device__ inline T round_half_even(T const input, int const olength, int const
   return num;
 }
 
-__device__ inline int to_formated_chars(floating_decimal_64 const v,
-                                        bool const sign,
-                                        char* const result,
-                                        int digits)
+__device__ inline int to_formated_double_chars(floating_decimal_64 const v,
```
mythrocks, line 1215: Nit: Minor typo.
```suggestion
__device__ inline int to_formatted_double_chars(floating_decimal_64 const v,
```

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1353,10 +1351,10 @@ __device__ inline int format_float_size(floating_decimal_64 const v, bool const
   return index;
 }
 
-__device__ inline int to_formated_chars(floating_decimal_32 const v,
-                                        bool const sign,
-                                        char* const result,
-                                        int digits)
+__device__ inline int to_formated_float_chars(floating_decimal_32 const v,
```
mythrocks, line 1354: Nit:
```suggestion
__device__ inline int to_formatted_float_chars(floating_decimal_32 const v,
```

mythrocks (approved):
I'm ok with this change, but we're going to have to address some minor stuff in this source file, at some point.
1. It would be good to add some function-level docs to `to_formatted_float_chars`, `to_formatted_double_chars`.
2. Is there commonality between `to_formatted_float_chars` and `to_formatted_double_chars`? If we discount the difference of calling `decimalLength17` and `decimalLength9`, they look like this could be a single function template. (I haven't looked closely enough, maybe?)
3. Is `actural_round` supposed to be `actual_round`?

We can address this later, in a follow-on.

thirtiseven:
Thanks for the review @mythrocks ! will address those later.

PR #1679, opened 2024-01-05
Fix compile warnings and refactor ftos_converter.cuh

### LINKED ISSUE #1678: [BUG] Compile warnings in `ftos_converter.cuh`
We have the following compile warnings that need to be addressed:
```
[INFO]      [exec] [21/56] Building CUDA object CMakeFiles/spark_rapids_jni.dir/src/format_float.cu.o
[INFO]      [exec] src/main/cpp/src/ftos_converter.cuh(496): warning #68-D: integer conversion resulted in a change of sign
[INFO]      [exec] 
[INFO]      [exec] src/main/cpp/src/ftos_converter.cuh(672): warning #68-D: integer conversion resulted in a change of sign
[INFO]      [exec] 
[INFO]      [exec] src/main/cpp/src/ftos_converter.cuh(496): warning #68-D: integer conversion resulted in a change of sign
[INFO]      [exec] 
[INFO]      [exec] src/main/cpp/src/ftos_converter.cuh(672): warning #68-D: integer conversion resulted in a change of sign
[INFO]      [exec] 
```

thirtiseven:
Fixes #1678 

Fix compile warnings in ftos_converter.cuh
Also do some refactoring according to comments from #1676

### DIFF
--- a/src/main/cpp/src/ftos_converter.cuh
+++ b/src/main/cpp/src/ftos_converter.cuh
@@ -1216,10 +1216,10 @@ __device__ inline T round_half_even(T const input, int const olength, int const 
  * Convert a floating_decimal_32/64 to a formatted string as the default format (#,###,###.##)
  * of format_number in Spark.
  *
- * @param v: floating_decimal_32/64 value
- * @param sign: sign of the number
- * @param result: output string
- * @param digits: number of digits after decimal point
+ * @param v The input floating_decimal_32/64 value
+ * @param sign Sign of the number
+ * @param result Output string
+ * @param digits Number of digits after decimal point
  */
 template <typename T, typename U>
 __device__ inline int to_formatted_chars(T const v, bool const sign, char* const result, int digits)
@@ -1340,7 +1340,7 @@ __device__ inline int format_size(T const v, bool const sign, int digits)
   int index = 0;
   if (sign) { index++; }
   U output               = v.mantissa;
-  const uint32_t olength = decimalLength(output);
+  uint32_t const  olength = decimalLength(output);
   int32_t exp            = v.exponent + (int32_t)olength - 1;
   if (exp < 0) {
     index += 2 + digits;
--- a/src/main/cpp/src/ftos_converter.cuh
+++ b/src/main/cpp/src/ftos_converter.cuh
@@ -147,9 +147,25 @@ __constant__ uint64_t const DOUBLE_POW5_TABLE[POW5_TABLE_SIZE] = {1ull,
 
 //===== common.h from ryu =====
 
-// Returns the number of decimal digits in v, which must not contain more than 9 digits.
-__device__ inline uint32_t decimalLength(uint32_t const v)
+// Returns the number of decimal digits in v, which must not contain more than 9/17 digits.
+template <typename T>
+__device__ inline uint32_t decimal_length(T const v)
 {
+  static_assert(std::is_integral_v<T> && !std::is_signed_v<T>);
+  if constexpr (sizeof(T) == sizeof(int64_t)) {
+    // The average output length is 16.38 digits, so we check high-to-low.
+    // Function precondition: v is not an 18, 19, or 20-digit number.
+    // (17 digits are sufficient for round-tripping.)
+    assert(v < 100000000000000000L);
+    if (v >= 10000000000000000L) { return 17; }
+    if (v >= 1000000000000000L) { return 16; }
+    if (v >= 100000000000000L) { return 15; }
+    if (v >= 10000000000000L) { return 14; }
+    if (v >= 1000000000000L) { return 13; }
+    if (v >= 100000000000L) { return 12; }
+    if (v >= 10000000000L) { return 11; }
+    if (v >= 1000000000L) { return 10; }
+  }
   // Function precondition: v is not a 10-digit number.
   // (f2s: 9 digits are sufficient for round-tripping.)
   // (d2fixed: We print 9-digit blocks.)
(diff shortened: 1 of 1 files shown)


src/main/cpp/src/ftos_converter.cuh
```diff
@@ -461,7 +461,7 @@ __device__ inline uint32_t mulPow5divPow2(uint32_t const m, uint32_t const i, in
 
 //===== d2s.c and f2s.c from ryu =====
 
-__device__ inline uint32_t decimalLength17(uint64_t const v)
+__device__ inline uint32_t decimalLength(uint64_t const v)
```
ttnghia, line 464: How about this?
```
template<typename T>
__device inline uint32_t decimal_length(T const v) {
if constexpr(sizeof(T) == sizeof(int64_t)) {
  // The average output length is 16.38 digits, so we check high-to-low.
  // Function precondition: v is not an 18, 19, or 20-digit number.
  // (17 digits are sufficient for round-tripping.)
 assert(v < 100000000000000000L);
 if (v >= 10000000000000000L) { return 17; }
  if (v >= 1000000000000000L) { return 16; }
  if (v >= 100000000000000L) { return 15; }
  if (v >= 10000000000000L) { return 14; }
  if (v >= 1000000000000L) { return 13; }
  if (v >= 100000000000L) { return 12; }
  if (v >= 10000000000L) { return 11; }
  if (v >= 1000000000L) { return 10; }
} else {
// Function precondition: v is not a 10-digit number.
  // (f2s: 9 digits are sufficient for round-tripping.)
  // (d2fixed: We print 9-digit blocks.)
  assert(v < 1000000000);
 }

  if (v >= 100000000) { return 9; }
  if (v >= 10000000) { return 8; }
  if (v >= 1000000) { return 7; }
  if (v >= 100000) { return 6; }
  if (v >= 10000) { return 5; }
  if (v >= 1000) { return 4; }
  if (v >= 100) { return 3; }
  if (v >= 10) { return 2; }
  return 1;
}
```

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -493,7 +493,7 @@ __device__ inline floating_decimal_64 d2d(uint64_t const ieeeMantissa, uint32_t
   uint64_t m2;
   if (ieeeExponent == 0) {
     // We subtract 2 so that the bounds computation has 2 additional bits.
-    e2 = 1 - DOUBLE_BIAS - DOUBLE_MANTISSA_BITS - 2;
+    e2 = static_cast<int32_t>(1 - DOUBLE_BIAS - DOUBLE_MANTISSA_BITS - 2);
```
ttnghia, line 496: Please also modify line 499 below, using `static_cast` instead of C-style cast.

ttnghia: OMG, there are too many instances of such C-style cast :(

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1465,12 +1334,13 @@ __device__ inline int to_formated_float_chars(floating_decimal_32 const v,
...
-__device__ inline int format_float_size(floating_decimal_32 const v, bool const sign, int digits)
+template <typename T, typename U>
+__device__ inline int format_size(T const v, bool const sign, int digits)
 {
   int index = 0;
   if (sign) { index++; }
-  uint64_t output        = v.mantissa;
-  uint32_t const olength = decimalLength9(output);
+  U output               = v.mantissa;
+  const uint32_t olength = decimalLength(output);
```
ttnghia, line 1343:
```suggestion
  uint32_t const  olength = decimalLength(output);
```

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1212,48 +1212,55 @@ __device__ inline T round_half_even(T const input, int const olength, int const
...
-                                               char* const result,
-                                               int digits)
+/*
+ * Convert a floating_decimal_32/64 to a formatted string as the default format (#,###,###.##)
+ * of format_number in Spark.
+ *
+ * @param v: floating_decimal_32/64 value
+ * @param sign: sign of the number
+ * @param result: output string
+ * @param digits: number of digits after decimal point
```
ttnghia, line 1222:
```suggestion
 * @param v floating_decimal_32/64 value
 * @param sign sign of the number
 * @param result output string
 * @param digits number of digits after decimal point
```

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1212,48 +1212,55 @@ __device__ inline T round_half_even(T const input, int const olength, int const
...
+ */
+template <typename T, typename U>
```
ttnghia, line 1224: Since `U` depends on `T`, we can just remove it here, then:
```
__device__ to_formatted_chars(...) {
static_assert(std::is_same_v<T, floating_decimal_32> || std::is_same_v<T, floating_decimal_64>);
using U = std::conditional_t<std::is_same_v<T, floating_decimal_32>, uint32_t, uint64_t>;

ttnghia: The same can be applied to the function `format_size` below.

src/main/cpp/src/ftos_converter.cuh
thirtiseven: 👍  done

src/main/cpp/src/ftos_converter.cuh
thirtiseven: Modified every C-style cast I found

src/main/cpp/src/ftos_converter.cuh
thirtiseven: done.

thirtiseven:
build

PR #1686, opened 2024-01-09
WIP: Update GpuTimeZoneDB to support DST

NVnavkumar:
Fixes https://github.com/NVIDIA/spark-rapids/issues/6840.

This adds rules the the timezone database for DST support. 

Eventually this will need to be updated with C++ code to take the rule information and update timestamps based on DST. This means that the rules will need to be applied on the fly to determine what offset to use. It will probably require using extraction to figure out which time of the year the right transition would apply as opposed to the binary search from earlier.


res-life:
Target to 24.04, close it, will put another PR.

Issue #1571, opened 2023-11-17
[FEA] See if we can avoid the warning about leaking HCV used to store GPU Timezone Database


NVnavkumar:
See https://github.com/NVIDIA/spark-rapids-jni/pull/1553#discussion_r1395910122

Since the HCV is kept around for the entire duration of the executor process, when the executor stops (either normally or by failure), we are not always in a situation whether `GpuTimeZoneDB.shutdown()` is called at the end. Therefore, we need to suppress the warnings that will come from this part of the code when the host column vector is leaked.

PR #1689, opened 2024-01-10
Fix memory leak in time zone DB

res-life:
closes https://github.com/NVIDIA/spark-rapids/issues/10169
closes https://github.com/NVIDIA/spark-rapids-jni/issues/1571

### root cases
My debug logs show there are two times running for `doLoadCache`
The implementation causes [doLoadCache](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.02/src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java#L174) running two times, so [here](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.02/src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java#L244) leaks.
```
HostColumnVector.fromLists
```

[here](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.02/src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java#L123) is not synced:
```
    if (!instance.isLoaded()) {
      cacheDatabase(); // lazy load the database
    }
```
When running case `test_cast_timestamp_to_date`, there are multiple threads running the `fromUtcTimestampToTimestamp` concurrently, so `doLoadCache` runs multiple times.

[link](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.02/src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java#L167)
```
executor.execute(this::doLoadData);
```
Above code starts a new thread, and the sync scope breaks.

### changes
- Make a final `GpuTimeZoneDB`, because it's a synced object. final makes it more safe.
- Add `cacheDatabaseAsync` to let plugin invokes.
- Make the whole cacheDatabase function synced.
- Add `cacheEmpty` bool to avoid multiple caching the transition rules.

Now there are no leaks, because `shutdown` will close the singleton memory.


res-life:
build

res-life:
```

```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
jlowe, line 1: 2024 copyrights

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -34,64 +34,78 @@
...
+
+  /**
+   * This should be called on startup of an executor.
+   */
+  public static void cacheDatabaseAsync() {
+    Runnable runnable = () -> {
+      try {
+        cacheDatabase();
+      } catch (Exception e) {
+        // ignore
```
jlowe, line 63: Do we really want to ignore any exceptions thrown here?  Why not let Spark's uncaught exception handler at least log it, or log it here?  Silently eating exceptions is very frustrating to debug.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -34,64 +34,78 @@
...
   }
 
-
-  public static void shutdown() {
-    if (instance.isLoaded()) {
-      instance.close();
-      // Recreate a new instance to reload the database if necessary
-      instance = new GpuTimeZoneDB();
+  public synchronized void shutdownImpl() {
+    if (!cacheEmpty) {
```
jlowe, line 102: Do we have a potential race condition here where a shutdown occurs very quickly after a startup, so quickly (or database load is so slow) that we don't think the cache is there but it will be once the async thread completes?  Seems like we need to also check if there's a pending async load and wait on it if that's the case.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -157,128 +165,89 @@ public static ZoneId getZoneId(String timeZoneId) {
...
+      HostColumnVector.DataType childType = new HostColumnVector.StructType(false,
+          new HostColumnVector.BasicType(false, DType.INT64),
+          new HostColumnVector.BasicType(false, DType.INT64),
+          new HostColumnVector.BasicType(false, DType.INT32));
+      HostColumnVector.DataType resultType =
+          new HostColumnVector.ListType(false, childType);
+      fixedTransitions = HostColumnVector.fromLists(resultType,
+          masterTransitions.toArray(new List[0]));
+    } catch (Exception e) {
+      throw new IllegalArgumentException("load time zone DB cache failed!", e);
```
jlowe, line 241: Curious why IllegalArgumentException?  RuntimeException (or a custom exception derived from RuntimeException) seems more appropriate since there isn't necessarily an illegal argument to this method that triggered the error.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -34,64 +34,78 @@
...
-              return thread;
-            }
-          });
-        instance.loadData(executor);
+  public static void shutdown() {
+    instance.shutdownImpl();
+  }
+
+  private synchronized void cacheDatabaseImpl() {
+    if (cacheEmpty) {
```
jlowe, line 84: Same race as described on shutdown.  What if async load on startup has not completed yet before a method is called that blindly calls cacheDatabase?  Seems like we could have to separate threads trying to load the database at that point (caller thread and async thread).  I think we need to check if there's a pending async load in-flight and wait on the thread if that's the case.

jlowe: Ah, synchronized method solves the issue here, but the race on shutdown is still there.  shutdown could "win" the race to the lock followed by async thread, leaving a cached database in memory despite being shutdown.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: The caches APIs are synchronized, so multiple callings will be serial and wait other to be done.
```scala
private synchronized void cacheDatabaseImpl() {
private synchronized void shutdownImpl() {
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Sorry, my intention is `IllegalStateException`

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: The caches APIs are synchronized, so multiple callings will be serial and wait other to be done.
```scala
private synchronized void cacheDatabaseImpl() {
private synchronized void shutdownImpl() {
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Done.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Done. Now log the error.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Done.

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
jlowe: `synchronized` doesn't solve the race I'm talking about.  All `synchronized` does is guarantee two threads aren't in that method at the same time, but it does nothing to ensure what _order_ those threads call it.  If two or more threads are trying to do the same thing (e.g.; initialize the cache), we're fine, since we don't really care about the order of the threads since the operation they are performing is identical and idempotent.  However if the threads are doing _different_ operations, then we care very much about the order of operations.

If one thread is trying to start the cache and the other is trying to shutdown, we could leave the database in a cached state despite trying to shut it down.  Here's a simplified example:
- On startup, we launch the background thread to initialize the cache
- Before the background thread is able to boot up and call the cache init method, the shutdown method is called (e.g.: very quick test that doesn't involve time DB and tears down on test completion, error on startup, whatever)
- shutdown is called _first_, notices the cache is not initialized, does nothing.
- background thread finally gets around to calling cache init (note there is no guarantee how quick this happens) and proceeds to initialize the cache
- Now we're in a state where the application thinks the time DB cache has been shutdown and freed, but in reality it has not and we've "leaked" it in a sense.

One way to solve this is to set a boolean flag before we start the background thread.  The background caching thread would atomically clear this flag and call the cache init method when it runs.  The shutdown method would check for this flag being set, and if it is, release the lock and wait on the background thread to complete before proceeding to retry the shutdown.  We can optimize it a bit by having shutdown set a flag after the cache clear has occurred indicating a shutdown has happened.  The background init thread can check for this flag being set and skip cache init when that happens.

jlowe: A simpler approach is to have a boolean flag indicating a shutdown has occurred.  Caching the database will clear this flag.  The background thread avoids initializing the cache if the shutdown flag has been set.  The shutdown method always wait on the background init Thread (outside of the lock) if it's not null, setting it to null afterwards.

res-life: The cache is a global singleton, and we do not want to shutdown it unless the Plugin is shutdown.
It’s almost not a scenario that one thread is caching cache and other is shutdown the cache concurrently.
 
Currently test cases do not call cacheDatabase/shutdown concurrently, see the following:
```scala
public class TimeZoneTest {
  @BeforeAll
  static void cacheTimezoneDatabase() {
    GpuTimeZoneDB.cacheDatabase();
  }
  
  @AfterAll
  static void cleanup() {
    GpuTimeZoneDB.shutdown();
  }
```
One potential corner case is:
Another time zone test like `TimeZoneTest2` doing the same like `TimeZoneTest`.
The following sequence will cause stale cache as you mentioned.
- TimeZoneTest.cache
- do TimeZoneTest.test1
- TimeZoneTest2.cache
- TimeZoneTest.shutdown
- do TimeZoneTest2.test2:  cache is shutdown, will cause a problem.
- TimeZoneTest2.shutdown

Typically production usage is:
Call `cacheDatabaseAsync` start the Plugin
Call `shutdown` when  stop the Plugin.

And another solution:
Remove the `shutdown` and mark the host memory column as [noWarnLeakExpected](https://github.com/rapidsai/cudf/blob/85acdc640701940e47b3969b14a811f33e7faf5b/java/src/main/java/ai/rapids/cudf/MemoryCleaner.java#L139), this will disable the cache leak warning for the global singleton host column.When the JVM exits the host memory column will be cleaned.

Do we need to do the implementation you mentioned for the rare use case?

jlowe: I'm not worried about the production use-case but rather the potential corner cases with tests that may shutdown and restart the database.  It's really not that hard to fix this race condition (I posted the details for two approaches above, happy to provide the patch for either), so why not close that race?  Even if we can prove to ourselves it's never going to be a race today, it doesn't guarantee someone won't change the way this is used to expose the race in the future.

res-life: got it, I'll fix it.

res-life: >One way to solve this is to set a boolean flag before we start the background thread. The background caching thread would atomically clear this flag and call the cache init method when it runs. The shutdown method would check for this flag being set, and if it is, release the lock and wait on the background thread to complete before proceeding to retry the shutdown. We can optimize it a bit by having shutdown set a flag after the cache clear has occurred indicating a shutdown has happened. The background init thread can check for this flag being set and skip cache init when that happens.

Now use this approach you mentioned.
Please review.
Changes are:
- Move all sync code into static methods.
- You mentioned changes

res-life:
build

res-life:
build

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -26,72 +33,171 @@
...
   }
-  
+
+  static class LoadingLock {
+    Boolean isLoading = false;
+
+    // record whether a shutdown is called ever.
+    // if `isCloseCalledEver` is true, then the following loading should be skipped.
+    Boolean isShutdownCalledEver = false;
+  }
```
jlowe, line 79: I'm not sure we need a separate lock class for this rather than locking the whole database via synchronized methods as was done previously.  We just need to lock the GpuTimeZoneDB instance when manipulating the boolean flags which can be direct members of the database. cacheDatabaseAsync, cacheDatabaseImpl and shutdown can just be synchronized methods, which removes the need for most of them to wait on threads -- the wait will be implicit in obtaining the database lock for the synchronized method. The only wait we need is in the shutdown method, and only when isLoading is true.

By leaving the methods synchronized rather than having a separate lock class, it's a little easier to reason about the synchronization safety because we don't have to keep track of whether we're holding LoadingLock at the right times.

PR #1539, opened 2023-11-06
Add timestamp parser to parse timestamp string with time zone

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/6846
contributes to https://github.com/NVIDIA/spark-rapids-jni/issues/1654
closes [#1721](https://github.com/NVIDIA/spark-rapids-jni/issues/1721)

Add kernel code for [SparkDateTimeUtils](https://github.com/apache/spark/blob/v3.5.0/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala) functions:
- SparkDateTimeUtils.stringToTimestamp
- SparkDateTimeUtils.stringToTimestampWithoutTimeZone
- SparkDateTimeUtils.stringToTimestampAnsi

Main logic is in: [parseTimestampString](https://github.com/apache/spark/blob/v3.5.0/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala#L394): 

This PR contains 3 parts:
- parser
- time rebase part:    
rebase local time construction (year,month,day,hour,minute,second) in a time zone to UTC time.
- JNI part. 

Cast(string as timestamp) for special strings: now, today, ...
```
from pyspark.sql.types import *
schema = StructType([
    StructField("c1", StringType()),
    StructField("c2", IntegerType()),
])
data = [
    ("today",1),
    ("now",2)
]
df = spark.createDataFrame(
        SparkContext.getOrCreate().parallelize(data, numSlices=2),
        schema).createOrReplaceTempView("tab")
spark.sql("select cast(c1 as timestamp) from tab").show()
```
- Spark 311:  return non-null values. Supports special strings.
- Spark 320 and 320+: return null values. Do not support special strings.

Note:
- This Kernel does not supports special strings: now, today...
- Will create another PR to handle: Cast to date
- Do not support parse just time, refer to [follow-up issue](https://github.com/NVIDIA/spark-rapids/issues/10221)

Signed-off-by: sperlingxx <lovedreamf@gmail.com>
Signed-off-by: Chong Gao <res_life@163.com>

PR #1700, opened 2024-01-16
Adding timeAdd with timezone kernel

thirtiseven:
This PR adds a kernel for timeAdd with a non-utc timezone (non-DST).

In Spark, timeAdd will try to keep the original offset after adding, so we need a special handling if the results are in an overlap. Details in the code comments.

The integration tests from the Spark side are passed, I will add some more tests in jni later.


src/main/cpp/src/timezones.cu
```diff
@@ -163,4 +342,50 @@ std::unique_ptr<column> convert_utc_timestamp_to_timezone(column_view const& inp
...
+                                 scalar_i64 const& duration,
+                                 table_view const& transitions,
+                                 size_type tz_index,
+                                 rmm::cuda_stream_view stream,
+                                 rmm::mr::device_memory_resource* mr)
+{
+  auto const type = input.type().id();
+
+  switch (type) {
+    case cudf::type_id::TIMESTAMP_SECONDS:
```
revans2, line 355: Here too. Spark only supports microseconds for timestamps. Why are we supporting seconds and mulliseconds at all?

src/main/cpp/src/timezones.cu
```diff
@@ -163,4 +342,50 @@ std::unique_ptr<column> convert_utc_timestamp_to_timezone(column_view const& inp
...
+        input, duration, transitions, tz_index, stream, mr);
+    case cudf::type_id::TIMESTAMP_MICROSECONDS:
+      return time_add_with_tz<cudf::timestamp_us>(
+        input, duration, transitions, tz_index, stream, mr);
+    default: CUDF_FAIL("Unsupported timestamp unit for timezone conversion");
+  }
+}
+
+std::unique_ptr<column> time_add(column_view const& input,
+                                 column_view const& duration,
```
revans2, line 369: What is the supported duration type(s)? CUDF supports multiple ones.

https://github.com/rapidsai/cudf/blob/726a7f30757d1a06d74d86bb82cf311cb159f7fd/cpp/include/cudf/types.hpp#L205-L209

Spark also supports multiple ones.

https://github.com/apache/spark/blob/ebd2b78f87fa6086c41d5e6bcade5efeefac75d0/common/unsafe/src/main/java/org/apache/spark/unsafe/types/CalendarInterval.java#L50-L52

https://github.com/apache/spark/blob/ebd2b78f87fa6086c41d5e6bcade5efeefac75d0/sql/api/src/main/scala/org/apache/spark/sql/types/DayTimeIntervalType.scala#L27-L30

https://github.com/apache/spark/blob/ebd2b78f87fa6086c41d5e6bcade5efeefac75d0/sql/api/src/main/scala/org/apache/spark/sql/types/YearMonthIntervalType.scala#L27-L28

The code appears to assume that the duration must be in microseconds. That might be fine in the short term, but I don't see anywhere that we check for this condition and enforce it. We just assume that the input is 64 bits and treat it like it was in microseconds.

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +120,183 @@ auto convert_timestamp_tz(column_view const& input,
...
+  auto [null_mask, null_count] =
+    cudf::detail::bitmask_and(cudf::table_view{{input, duration}}, stream, mr);
+
+  auto results = cudf::make_timestamp_column(
+    input.type(), input.size(), rmm::device_buffer(null_mask, stream), null_count, stream, mr);
+
+  thrust::transform(rmm::exec_policy(stream),
+                    input.begin<timestamp_type>(),
+                    input.end<timestamp_type>(),
+                    duration.begin<int64_t>(),
```
revans2, line 293: This would not work if the duration type was TIMESTAMP_DAYS.

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +120,183 @@ auto convert_timestamp_tz(column_view const& input,
...
+  __device__ inline timestamp_type plus_with_tz(timestamp_type const& timestamp,
+                                                int64_t const& duration) const
+  {
+    if (duration == 0L) { return timestamp; }
+
+    auto const utc_instants = transitions.child().child(0);
+    auto const tz_instants  = transitions.child().child(1);
+    auto const utc_offsets  = transitions.child().child(2);
+
+    auto const epoch_seconds_utc = static_cast<int64_t>(
```
revans2, line 142: Why do we need to cast the result to int64_t? What is the type that this returns? if duration for seconds in not an `int64_t` or larger already we risk overflowing the result.

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +120,183 @@ auto convert_timestamp_tz(column_view const& input,
...
+    auto const transition_times_utc = cudf::device_span<int64_t const>(
+      utc_instants.data<int64_t>() + tz_transitions.element_offset(0),
+      static_cast<size_t>(list_size));
+
+    auto const transition_times_tz = cudf::device_span<int64_t const>(
+      tz_instants.data<int64_t>() + tz_transitions.element_offset(0),
+      static_cast<size_t>(list_size));
+
+    // step 1: Binary search on utc to find the correct offset to convert utc to local
+    auto const utc_it = thrust::upper_bound(
```
revans2, line 158: Don't we do this some place else? Shouldn't transitioning between time zones like this be a static inline function? And if not can you add comments as to why this is different from `GpuTimeZoneDB.fromUtcTimestampToTimestamp`?

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +120,183 @@ auto convert_timestamp_tz(column_view const& input,
...
+      static_cast<size_type>(thrust::distance(transition_times_utc.begin(), utc_it));
+    auto const utc_list_offset = tz_transitions.element_offset(utc_idx - 1);
+    auto const to_local_offset =
+      static_cast<int64_t>(utc_offsets.element<int32_t>(utc_list_offset));
+
+    auto const to_local_offset_duration = cuda::std::chrono::duration_cast<duration_type>(
+      cudf::duration_s{static_cast<int64_t>(to_local_offset)});
+
+    auto const duration_typed = cuda::std::chrono::duration_cast<duration_type>(
+      cudf::duration_us{static_cast<int64_t>(duration)});
```
revans2, line 170: Is the curation always microseconds then?

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +120,183 @@ auto convert_timestamp_tz(column_view const& input,
...
+
+    // step 2: add the duration to the local timestamp
+    auto const local_timestamp_res = timestamp + to_local_offset_duration + duration_typed;
+
+    auto const result_epoch_seconds = static_cast<int64_t>(
+      cuda::std::chrono::duration_cast<cudf::duration_s>(local_timestamp_res.time_since_epoch())
+        .count());
+
+    // step 3: Binary search on local to find the correct offset to convert local to utc
+    auto const local_it = thrust::upper_bound(
```
revans2, line 180: Here too can we refactor the existing code so we can share most/all of it? If not can we list why and try to refactor the code to share as much as possible? If there isn't much that would overlap it is fine. But I still would like the code to be located near the other transition code with a name that is clear what it is doing differently.

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +120,183 @@ auto convert_timestamp_tz(column_view const& input,
...
+    // In GpuTimeZoneDB, we build the transition list with
+    // (utcInstant, utcInstant + OffsetBefore, OffsetAfter) if it is a overlap.
+    // But in the step 3, we actually need to binary search on the utcInstant + OffsetBefore
+    // to find the correct offset to convert local to utc. To reuse the code, we need to
+    // add a special post processing here, to make sure we get the correct id that
+    // utcInstant + OffsetBefore is larger than the result_epoch_seconds.
+    auto const temp_list_offset = tz_transitions.element_offset(local_idx);
+    auto const temp_offset = static_cast<int64_t>(utc_offsets.element<int32_t>(temp_list_offset));
+
+    // We don't want to check this if the idx is the first or last
```
revans2, line 194: Last I understand, but first I am a but confused about.

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +120,183 @@ auto convert_timestamp_tz(column_view const& input,
...
+    if (local_idx != 0 && transition_times_utc[local_idx] != INT64_MAX &&
```
revans2, line 195: On a separate note. Why are the offsets stored as a long? This would be a follow on issue just to investigate if we could get away with an int for the offset. We shouldn't have any offsets of more than a day really.The offsets are stored in seconds. Which would need just larger than a short to store 24 hours.  Even java returns the offset in seconds using an int.

https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html#getTotalSeconds--

I think we probably could get away with changing it and using INT32_MAX here instead.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -130,6 +133,42 @@ public static ColumnVector fromUtcTimestampToTimestamp(ColumnVector input, ZoneI
     transitions.close();
     return result;
   }
+
+  public static ColumnVector timeAdd(ColumnVector input, Scalar duration, ZoneId currentTimeZone) {
+    // TODO: Remove this check when all timezones are supported
```
res-life, line 138: Nit: Consider extract a method like `checkTimeZone` from:
```
    // TODO: Remove this check when all timezones are supported
    // (See https://github.com/NVIDIA/spark-rapids/issues/6840)
    if (!isSupportedTimeZone(currentTimeZone)) {
      throw new IllegalArgumentException(String.format("Unsupported timezone: %s",
          currentTimeZone.toString()));
    }
```

src/main/cpp/src/timezones.hpp
```diff
@@ -66,4 +66,33 @@ std::unique_ptr<cudf::column> convert_utc_timestamp_to_timezone(
...
+ * specific timezone is given.
+ *
+ * @param input the column of input timestamps in UTC
+ * @param intervals the column of intervals to add to the input timestamps
+ * @param transitions the table of transitions for all timezones
+ * @param tz_index the index of the row in `transitions` corresponding to the specific timezone
+ * @param stream CUDA stream used for device memory operations and kernel launches.
+ * @param mr Device memory resource used to allocate the returned timestamp column's memory
+ */
+std::unique_ptr<cudf::column> time_add(
```
res-life, line 82: Nit: duration parameter has 2 types: `numeric_scalar` and `column_view`.
Extract a template for duration parameter to combine the two `time_add` functions to one function.

src/main/cpp/src/timezones.cu
res-life: Nit: use std::numeric_limits<Type>::max() instead of INT64_MAX.

res-life:
Please check 2024 Copyrights.

src/main/cpp/src/timezones.cu
```diff
@@ -163,4 +342,50 @@ std::unique_ptr<column> convert_utc_timestamp_to_timezone(column_view const& inp
...
+                                 column_view const& duration,
+                                 table_view const& transitions,
+                                 size_type tz_index,
+                                 rmm::cuda_stream_view stream,
+                                 rmm::mr::device_memory_resource* mr)
+{
+  auto const type = input.type().id();
+
+  switch (type) {
+    case cudf::type_id::TIMESTAMP_SECONDS:
```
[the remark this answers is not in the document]
thirtiseven: Updated to support microseconds only. I guess convert_timestamp supports all timestamp types to make it generally usable, but this kernel is just for timeAdd in Spark, so we don't need to do that.

src/main/cpp/src/timezones.cu
thirtiseven: ditto

src/main/cpp/src/GpuTimeZoneDBJni.cpp
```diff
@@ -50,4 +50,49 @@ Java_com_nvidia_spark_rapids_jni_GpuTimeZoneDB_convertUTCTimestampColumnToTimeZo
   }
   CATCH_STD(env, 0);
 }
+
```
HaoYang670, line 53: update copyright?

src/main/cpp/src/timezones.cu
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, NVIDIA CORPORATION.
+ * Copyright (c) 2024, NVIDIA CORPORATION.
```
HaoYang670, line 2: 2023-2024?

res-life: should be 2023-2024, not 2024.

thirtiseven:
There is a bug in the current code:
In spark, TimeAdd is a plusDays followed by a plus(microseconds):

```
def timestampAddDayTime(micros: Long, dayTime: Long, zoneId: ZoneId): Long = {
  val days = dayTime / MICROS_PER_DAY
  val microseconds = dayTime - days * MICROS_PER_DAY
  val resultTimestamp = microsToInstant(micros)
    .atZone(zoneId)
    .plusDays(days)
    .plus(microseconds, ChronoUnit.MICROS)
  instantToMicros(resultTimestamp.toInstant)
}
```
which results in:
- plusDays happens in LocalTime, and `plusDays` is not equivalent to `plus(MICROS_PER_DAY * days)`, so this PR is wrong here. I'm implementing this.
- plusDays in ZonedDateTime will also `resolve` (choose a preferred offset from two possible when in a overlap), so there will be two `resolve`s in total, which might cause some unexpected behavior, I will investigate it more.

src/main/cpp/src/GpuTimeZoneDBJni.cpp
```diff
@@ -50,4 +50,50 @@ Java_com_nvidia_spark_rapids_jni_GpuTimeZoneDB_convertUTCTimestampColumnToTimeZo
...
+                                                         jlong duration_handle,
+                                                         jlong transitions_handle,
+                                                         jint tz_index)
+{
+  JNI_NULL_CHECK(env, input_handle, "column is null", 0);
+  JNI_NULL_CHECK(env, transitions_handle, "column is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
+    auto const input = reinterpret_cast<cudf::column_view const*>(input_handle);
+    auto const duration =
```
revans2, line 67: This is problematic because c++ is going to say that `duration_handle` is a pointer to `cudf::duration_scalar<cudf::duration_us>`, even if it is not. The only thing that is guaranteed is.

```
    auto duration = reinterpret_cast<cudf::scalar *>(duration_handle);
```

Beyond that you will need to use dynamic_cast where C++ will verify that it is the type you expect. Note that dynamic_cast returns nullptr if it is not the right type.

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +119,221 @@ auto convert_timestamp_tz(column_view const& input,
...
+
+    // In Spark, timeAdd will add the days of the duration to the timestamp first with `plusDays`,
+    // resolve the offset if the result is in the overlap, and then add the remainder microseconds
+    // to the results with `plus`, then resolve again. In java, the `plusDays` adds the days to the
+    // localDateTime in day's field, so we need take transitions into account. While the `plus` step
+    // adds the remainder microseconds to the localDateTime in microsecond's field, so it is
+    // equivalent to add the remainder to the utc result without resolving the offset.
+    auto const duration_value =
+      static_cast<int64_t>(cuda::std::chrono::duration_cast<cudf::duration_us>(duration).count());
+    auto const microseconds_per_day = 86400000000ll;
```
revans2, line 234: Have we tested with more microseconds than in a day, along with a days value? I know that this is an odd corner case, but if we have to apply days separately from microseconds wouldn't this result in the same kind of errors if we assume that we can always `convert microseconds_per_day` into days? It looks like for DayTimeIntervalType microsecond values outside of a day are not technically allowed, so we are probably okay there, but CalendarIntervalType does not indicate any limitations like that, and reading the code there are none.

We might be able to work around this by restricting the usage of CalendarInterval, but I'd prefer not to need to do that.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -217,19 +215,43 @@ public static ColumnVector fromTimestampToUtcTimestamp(ColumnVector input, ZoneI
...
+    warnUnsupportedTimeZone(currentTimeZone);
+    cacheDatabase();
+    Integer tzIndex = instance.getZoneIDMap().get(currentTimeZone.normalized().toString());
+    try (Table transitions = instance.getTransitions()) {
+      return new ColumnVector(timeAddCC(input.getNativeView(), duration.getNativeView(),
+          transitions.getNativeView(), tzIndex));
+    }
+  }
+
+  private static void warnUnsupportedTimeZone(ZoneId zoneId) {
```
revans2, line 247: nit: This is not a warning, could we call it something like `assertTimeZoneSupported`?

revans2 (commented):
It looks a lot better. The JNI casts for scalar values need to be fixed. I also still think that there is a lot of code that could be made more common, but I am not going to keep pushing for that. It is a not now.

I also want to understand a long term plan for how we want to make the gap processing common. I think there is a lot of overlap between parsing timestamps and adding time intervals. If you think about it.

Add Interval looks kind of like.

Take timestamp and convert to local timeZone.
extract year, month, day, hour, min, second, sub-second
Add Intervals
Convert from TZ to UTC timestamp minding any gap.

Parsing a timestamp ends up being:
Parse out year, month, day, hour, min, second, sub-second and TZ from string
Convert from TZ to UTC timestamp minding any gap.

But that is all hidden details and we don't have to do it all now. My biggest concern is that each of the PRs are adding in different ways to find the gap, and I want to be sure that we don't have too many conflicts when merging them both.

src/main/cpp/src/timezones.cu
thirtiseven: Following test can pass with this pr:
```
@pytest.mark.parametrize('data_gen', [(0, 86400000001), (1, 86400000001), (1, 86400000000*100)], ids=idfn)
@allow_non_gpu(*non_supported_tz_allow)
def test_timeadd_debug(data_gen):
    days, microseconds = data_gen
    assert_gpu_and_cpu_are_equal_collect(
        lambda spark: unary_op_df(spark, TimestampGen(), length=200000)
            .selectExpr("a + (interval {} days {} microseconds)".format(days, microseconds)))
```

It's an interesting case I hadn't thought of, I'll check the code to see if there is some validation in spark/pyspark to pass it. 

If it is possible to pass an invalid CalendarInterval, we can have a simple workaround in plugin side to pass only the days in this kernel and add the microseconds directly to the result.

src/main/cpp/src/GpuTimeZoneDBJni.cpp
```diff
@@ -50,4 +50,50 @@ Java_com_nvidia_spark_rapids_jni_GpuTimeZoneDB_convertUTCTimestampColumnToTimeZo
...
+JNIEXPORT jlong JNICALL
+Java_com_nvidia_spark_rapids_jni_GpuTimeZoneDB_timeAddCS(JNIEnv* env,
+                                                         jclass,
+                                                         jlong input_handle,
+                                                         jlong duration_handle,
+                                                         jlong transitions_handle,
+                                                         jint tz_index)
+{
+  JNI_NULL_CHECK(env, input_handle, "column is null", 0);
+  JNI_NULL_CHECK(env, transitions_handle, "column is null", 0);
```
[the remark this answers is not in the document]
thirtiseven: done.

src/main/cpp/src/GpuTimeZoneDBJni.cpp
thirtiseven: Thanks. I moved the `dynamic_cast` and type checking to timezones.cu L392-L386 because scalar is forward declaration here. I'm not quite sure I did things right there, please help check again.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
thirtiseven: done.

thirtiseven:
@revans2 Thanks for the review. I will try to make the code more common and investigate the corner case, and I think other comments are addressed.
The common parts between this pr and #1539 is  the GpuTimeZoneDB.java, there will be some conflicts but we can handle that because 1539 mainly works on the new column it added.

src/main/cpp/src/timezones.cu
thirtiseven: Ok, this case failed on 311. Spark seems to automatically use DayTimeIntervalType for interval literals in higher versions. Will match this behavior in plugin PR.

thirtiseven: Matched this in plugin datetimeExpressionsUtils.scala

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -230,6 +269,11 @@ private void doLoadData() {
                   );
                 }
               });
+              ZoneOffsetTransition last = transitions.get(transitions.size() - 1);
```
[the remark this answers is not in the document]
res-life: Seems it's OK to add a max item in the end.
Not sure if add a max item fixed a bug.

If the searched item is the last item, then `idx-1` refer to the wrong index.
Refer to the following example:
Search `8` get end(), then `idx -1` is not want we want.

[upper_bound](https://thrust.github.io/doc/group__binary__search_gac00fd24a457a398e91c3d3822e02b95a.html#gac00fd24a457a398e91c3d3822e02b95a):
```cpp
thrust::device_vector<int> input(5);
input[0] = 0;
input[1] = 2;
input[2] = 5;
input[3] = 7;
input[4] = 8;
thrust::upper_bound(thrust::device, input.begin(), input.end(), 0); // returns input.begin() + 1
thrust::upper_bound(thrust::device, input.begin(), input.end(), 1); // returns input.begin() + 1
thrust::upper_bound(thrust::device, input.begin(), input.end(), 2); // returns input.begin() + 2
thrust::upper_bound(thrust::device, input.begin(), input.end(), 3); // returns input.begin() + 2
thrust::upper_bound(thrust::device, input.begin(), input.end(), 8); // returns input.end()
thrust::upper_bound(thrust::device, input.begin(), input.end(), 9); // returns input.end()
```

Another option: use `lower_bound` instead.
then:  `it != end && *it = searched_item` to find an item.
Then we can remove the min and max items.

I propose to remove the min/max items and use `lower_bound`. Should be a follow-up.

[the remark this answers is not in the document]
res-life: Be careful when there is no item in the vector,  `end - 1` will point to a wrong address.

res-life:
> I also want to understand a long term plan for how we want to make the gap processing common. I think there is a lot of overlap between parsing timestamps and adding time intervals. If you think about it.
> 
> Add Interval looks kind of like.
> 
> Take timestamp and convert to local timeZone. extract year, month, day, hour, min, second, sub-second Add Intervals Convert from TZ to UTC timestamp minding any gap.
> 
> Parsing a timestamp ends up being: Parse out year, month, day, hour, min, second, sub-second and TZ from string Convert from TZ to UTC timestamp minding any gap.
> 
> But that is all hidden details and we don't have to do it all now. My biggest concern is that each of the PRs are adding in different ways to find the gap, and I want to be sure that we don't have too many conflicts when merging them both.

Refer to Java `ZoneRules`, there are mainly 3 columns:
```
    private final long[] savingsInstantTransitions;   // save n epoch seconds
    private final LocalDateTime[] savingsLocalTransitions; // save 2*n local times
    private final ZoneOffset[] wallOffsets;  // offsets

    for (ZoneOffsetTransition trans : transitionList) {
            if (trans.isGap()) {
                localTransitionList.add(trans.getDateTimeBefore());
                localTransitionList.add(trans.getDateTimeAfter());
            } else {
                localTransitionList.add(trans.getDateTimeAfter());
                localTransitionList.add(trans.getDateTimeBefore());
            }
            localTransitionOffsetList.add(trans.getOffsetAfter());
        }
```
In https://github.com/NVIDIA/spark-rapids-jni/pull/1539, there are already 4 columns.
I propose to use the method of Java, then implements:
```
private Object getOffsetInfo(LocalDateTime dt) // used for local to UTC
public ZoneOffset getOffset(Instant instant) // used for UTC to local
```
`savingsLocalTransitions` savles 2*n items(n is fixed transition number), then binary search will add one extra match compared to our currently solution.

For `savingsLocalTransitions`, we can map them and save as epoch seconds. 

Then, we will update `timeAdd`, `timezones.cu` and `parser ts`
other:
  Remove the first min and the last max transitions to save GPU memory, and use `lower_bound` instead of `upper_bound` to do binary search.

For DST, mainly about:
- Generate a new Transition rule if `year` is larger than the last transition.
- Find the time on specified week day.

We can reuse the code for DST.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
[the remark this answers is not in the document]
thirtiseven: > Be careful when there is no item in the vector, `end - 1` will point to a wrong address.

Yes, since there is a start point and an end point, I think this case will not happen.

thirtiseven:
build

thirtiseven:
Mark as draft because we need a long term plan for timezone issues.

thirtiseven:
Close for not planned recently

PR #1706, opened 2024-01-17
[submodule-sync] bot-submodule-sync-branch-24.02 to branch-24.02 [skip ci] [bot]

nvauto:
submodule-sync to create a PR keeping thirdparty/cudf up-to-date.
HEAD commit SHA: 63f0ed923f8280c751d7c4df08e46e584839c671, cudf commit SHA: https://github.com/rapidsai/cudf/commit/c7acdaa231fb0ffe7751611590f9b85ba7508d4d

This PR will be auto-merged if test passed. If failed, it will remain open until test pass or manually fix.

PR #1708, opened 2024-01-17
Update to new cudf strings where character data is no longer a child column

jlowe:
Fixes the submodule sync error reported in #1706.  This updates to the changes in rapidsai/cudf#14202 where the strings column has the character data stored in a device buffer rather than a child column.


jlowe:
build

jlowe:
build

hyperbolic2346 (approved):

Issue #1735, opened 2024-01-25
[BUG] match parse_url QUERY behavior when url contains no protocol


thirtiseven:
**Describe the bug**
When test parse_url with current jni, the results different from gpu and cpu:

```
|bab.bg/en_page.php?id=33 |null |cpu     |
|bab.bg/en_page.php?id=33 |33   |gpu     |
```

seems cpu will always return null for urls without protocol, we should match this in parse_uri kernel.

**Steps/Code to reproduce bug**
Run:
```
df.selectExpr("*", "parse_url(url, 'QUERY', 'query') as pr")
```
with plugin and try url with no protocol.

thirtiseven:
The pattern is: if the last value of key in query part has length of **exactly strlen(query)-3**, and query with a key that can't match, gpu will return last value.

PR #1740, opened 2024-01-26
Fix a parse_uri query bug

thirtiseven:
Fixes https://github.com/NVIDIA/spark-rapids-jni/issues/1735

This pr fixes s small bug in the parse_uri query kernel: If the last value of key in query part has length of **exactly strlen(query)-3** or **0**, and query with a key that can't match, gpu will returns last value.

Tested the kernel with kaggle dataset and got passed, I think now it is correct enough.


hyperbolic2346 (approved):

thirtiseven:
build

PR #1843, opened 2024-03-07
Use whoami instead of $USER

gerashegalov:
Docker does not create the envrironment variable USER per https://docs.docker.com/engine/reference/run/#environment-variables

The user property in the build-info is missing as seen in #1840 

### DIFF
--- a/build/build-info
+++ b/build/build-info
@@ -27,7 +27,7 @@ echo_build_properties() {
   git_path=$2
   shift 2
   echo version=$version
-  echo user=$USER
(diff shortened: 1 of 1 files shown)


gerashegalov:
build

build/build-info
jlowe, line 1: 2024 copyrights

gerashegalov: will wait until #1840 is merged to save CI resources

jlowe (approved):

PR #1858, opened 2024-03-11
Improve build-info script

gerashegalov:
When running the script outside Docker on the host without CUDA runtime I noticed that it may succeed while generating a corrupt properties file.

```
version=24.04.0-SNAPSHOT
user=user1
revision=d9ec683518b78fb3a52272b324f2184f19381318
branch=branchXY
date=2024-03-11T18:01:51Z
url=git@github.com:NVIDIA/spark-rapids-jni
/home/user1/gits/NVIDIA/spark-rapids-jni/build/build-info: line 53: cuobjdump: command not found
```

This PR improves the script as follows:

- Enable pipefail option to make sure that errors in the pipe are detected by bash
- Separate assignment gpu_architectures to avoid masking the subshell failure by `echo`
- Since awk is symlinked to gawk in the Docker image, we can implement the entire logic for gpu archs in gawk
- Removing the trailing ';' from the list of archs

**Testing**

```
~/gits/NVIDIA/spark-rapids-jni$ $PWD/build/build-info 24.04.0-SNAPSHOT $PWD $PWD/target/libcudf-install/lib64/libcudf.a 
version=24.04.0-SNAPSHOT
user=user1
revision=d9ec683518b78fb3a52272b324f2184f19381318
branch=fixTrailingSemicolon
date=2024-03-11T21:18:48Z
url=git@github.com:NVIDIA/spark-rapids-jni
/home/user1/gits/NVIDIA/spark-rapids-jni/build/build-info: line 53: cuobjdump: command not found
~/gits/NVIDIA/spark-rapids-jni$ echo $?
127
```

```
./build/run-in-docker $PWD/build/build-info 24.04.0-SNAPSHOT $PWD $PWD/target/libcudf-install/lib64/libcudf.a 
version=24.04.0-SNAPSHOT
user=user1
revision=d9ec683518b78fb3a52272b324f2184f19381318
branch=fixTrailingSemicolon
date=2024-03-11T18:27:31Z
url=git@github.com:NVIDIA/spark-rapids-jni
gpu_architectures=75;86
```

### DIFF
--- a/build/build-info
+++ b/build/build-info
@@ -35,10 +35,15 @@ echo_build_properties() {
   echo date=$(date -u +%Y-%m-%dT%H:%M:%SZ)
   echo url=$(cd "$git_path" && git config --get remote.origin.url)
   echo gpu_architectures=$(
-    cuobjdump libcudf.a | \
-    awk '
-      match($0, /arch = sm_([0-9]+)/, tmp) { arch[tmp[1]]=1 } 
-      END { for (a in arch) s = s ";" a; printf substr(s, 2) }
+    cuobjdump "$libcudf_path" | gawk '
+      match($0, /arch = sm_([0-9]+)/, tmp) { arch[tmp[1]]=tmp[1] }
+      END {
+     	asort(arch) 
+      	for (i in arch) {
+	  s = s ";" arch[i]
+        }
+	printf substr(s, 2) 
+      }
     '
   )
   for arg in "$@"; do
--- a/build/build-info
+++ b/build/build-info
@@ -44,7 +44,7 @@ echo_build_properties() {
       END {
         n = asorti(arch)
         if (n == 0) {
-          print "ERROR: No fatbin ELF / PTX code sections found"
+          print "ERROR: No fatbin ELF / PTX code sections found" > "/dev/stderr"
           exit(1)
         }
         s = arch[1]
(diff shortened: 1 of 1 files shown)


gerashegalov:
build

build/build-info
```diff
@@ -34,7 +35,27 @@ echo_build_properties() {
...
+        n = asorti(arch)
+        if (n == 0) {
+          print "ERROR: No fatbin ELF / PTX code sections found" > "/dev/stderr"
+          exit(1)
+        }
+        s = arch[1]
+        for (i = 2; i <= n; i++) {
+          s = s ";" arch[i]
+        }
+        printf s
```
jlowe, line 54: We don't want to try to interpret formatting characters here
```suggestion
        print s
```

gerashegalov: Originally, I was not using the concatenation, and used `printf` in the loop to print without a newline suffix.

Here the newline is also unnecessary but it gets stripped away by the output capture `$( ...)` anyways, so print is fine too.

There is no format interpretation unless there are further arguments to printf. If due to a bug the first argument contains a format-control letter it will error out

```
bash-4.2$ gawk 'BEGIN { str="75;%s;86"; printf str }'
gawk: cmd. line:1: fatal: not enough arguments to satisfy format string
        `75;%s;86'
            ^ ran out for this one
```

jlowe (approved):

gerashegalov:
build

gerashegalov:
build

parthosa (approved):
Interesting catch @gerashegalov. Thanks

PR #1861, opened 2024-03-12
Add path parser for get_json_object

### LINKED ISSUE #1830: [FEA] GetJsonObject:  Implement a Path parser like Spark does, can refer to cuDF current Path parser
Implement a Path parser like Spark does, can refer to cuDF current Path parser

**Additional context**
Epic issue: https://github.com/NVIDIA/spark-rapids-jni/issues/1823

thirtiseven:
Closes #1830 

This PR adds a path parser for get_json_object to match spark's behavior.

Spark code [here](https://github.com/apache/spark/blob/72a95bcad7f1906c97fb0971ed6338374ec3009d/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala#L59-L106). We do not have RegexParsers in cpp, so the implementation is more long and complicated than in scala.

Here is a draft diagram describing the parsing rules:

path_parser

Local tests against those cases I added got 100% line coverage and branch coverage.


src/main/cpp/src/get_json_object.hpp
```diff
@@ -19,20 +19,270 @@
...
+  path_instruction_type type;
+
+  // used when type is named type
+  cudf::string_view name{"", 0};
+
+  // used when type is index
+  long index{-1};
+};
+
+class JsonPathParser {
```
revans2, line 50: Would it be better to just pass down the parsed version from Spark? This feels like it is providing very little value over the Spark version.

src/main/cpp/src/get_json_object.hpp
```diff
@@ -19,20 +19,270 @@
...
+    size_t pos    = 0;
+    if (str.size() == 2) {
+      // the while loop below will not run if str.size() == 2
+      // and there is no valid path
+      return thrust::nullopt;
+    }
+    if (str[pos++] != '$') { return thrust::nullopt; }
+    bool in_brackets = false;
+    bool in_quotes   = false;
+    if (str[pos] == '[') { in_brackets = true; }
```
revans2, line 75: This does not protect against an overflow. `$` is a valid path and this is not checking for that.

src/main/cpp/src/get_json_object.hpp
```diff
@@ -19,20 +19,270 @@
...
+    if (str[pos] == '\'') {
+      // .'name' is invalid
+      return thrust::nullopt;
+    }
+    while (pos < str.size() - 1) {
+      char from = str[pos];
+      char to   = str[pos + 1];
+      // std::cout << "from: " << from << " to: " << to << " " << in_brackets << " " << in_quotes <<
```
revans2, line 83: Can we remove the debug logging, not just comment it out. Or put it under an `#ifdef`.

src/main/cpp/src/get_json_object.hpp
```diff
@@ -19,20 +19,270 @@
...
+        } else {
+          // subscript: [*] or [123]
+          in_brackets = true;
+          instructions.push_back(path_instruction{path_instruction_type::SUBSCRIPT});
+          if (to == '*') {
+            // subscript wildcard
+            current_instruction = path_instruction{path_instruction_type::WILDCARD};
+          } else if (is_digit(to)) {
+            // subscript index
+            // TODO long overflow
```
revans2, line 119: Are we planning on addressing this?

revans2 (commented):
I looked at the code, and found a few issues, but even if I didn't find anything I would feel much better about keeping the parsing in Spark and just pass the tokens down though JNI, especially if we are on a tight time scale.

src/main/cpp/tests/CMakeLists.txt
```diff
@@ -60,12 +60,15 @@ ConfigureTest(CAST_FLOAT_TO_STRING
 ConfigureTest(DATETIME_REBASE
     datetime_rebase.cpp)
 
-ConfigureTest(ROW_CONVERSION
-    row_conversion.cpp)
+ConfigureTest(JSON_PATH_PARSER
+    json_path_parser.cpp)
```
res-life, line 64: Nit, we can put this into GET_JSON_OBJECT binary.
```
ConfigureTest(GET_JSON_OBJECT
    json_parser_tests.cpp,
    json_path_parser.cpp
    )
```

thirtiseven:
Moved to plugin and jni pr, closing

Issue #1848, opened 2024-03-08
[FEA] GetJsonObject: Implement JNI interface for get-json-object kernel


res-life:
Epic issue: https://github.com/NVIDIA/spark-rapids-jni/issues/1823

res-life:
PR merged to feature branch: https://github.com/NVIDIA/spark-rapids-jni/pull/1862

PR #1862, opened 2024-03-12
Add JNI for GetJsonObject

thirtiseven:
Closes https://github.com/NVIDIA/spark-rapids-jni/issues/1848

This PR adds JNI for getJsonObject.

I haven't tested it because the feature isn't ready yet, but we can merge it to the dev branch first.


src/main/cpp/src/get_json_object.cu
```diff
@@ -305,6 +304,8 @@ std::unique_ptr<cudf::column> get_json_object(cudf::strings_column_view const& c
   // parse the json_path into a command buffer
   auto path_commands_optional = parse_path(json_path);
 
+  auto options = json_parser_options{};
```
revans2, line 307: As a follow on for cleanup. If we don't have a need to pass down a json_parser_options class, then do we even have a need for the class?  Would it be better to just remove it because there is only one set of options we ever run with?

revans2 (approved):

src/main/cpp/src/getJsonObjectJni.cpp
```diff
@@ -0,0 +1,37 @@
+/*
```
ttnghia, line 1: File name for JNI should have all first letter Capitalized.

In addition, we may have other JSON functionalities in the future, beyond `getJsonObject`. Thus, it's better to call `JsonUtilsJni.cpp` and `JsonUtils.java` or similar.

src/main/cpp/src/getJsonObjectJni.cpp
```diff
@@ -0,0 +1,37 @@
...
+  JNIEnv* env, jclass, jlong input_column, jlong json_path)
+{
+  JNI_NULL_CHECK(env, input_column, "input column is null", 0);
+  JNI_NULL_CHECK(env, json_path, "json path is null", 0);
+
+  try {
+    cudf::jni::auto_set_device(env);
+    auto const input = reinterpret_cast<cudf::column_view const*>(input_column);
+    auto const n_strings_col_view(*n_column_view);
+    auto const* n_scalar_path = reinterpret_cast<cudf::string_scalar*>(j_scalar_handle);
```
ttnghia, line 31: Where does `j_scalar_handle` come frome?

src/main/cpp/src/getJsonObjectJni.cpp
thirtiseven: Thanks, done.

src/main/cpp/src/getJsonObjectJni.cpp
thirtiseven: I forgot to update `CmakeLists` so this cpp file did not work at all 😅, updated.

src/main/cpp/src/JSONUtilsJni.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+  JNI_NULL_CHECK(env, input_column, "input column is null", 0);
+  JNI_NULL_CHECK(env, path_ins_types, "path_ins_types is null", 0);
+  JNI_NULL_CHECK(env, path_ins_names, "path_ins_names is null", 0);
+  JNI_NULL_CHECK(env, path_ins_indexes, "path_ins_indexes is null", 0);
+
+  try {
+    cudf::jni::auto_set_device(env);
+    cudf::column_view* n_column_view = reinterpret_cast<cudf::column_view*>(input_column);
+    cudf::strings_column_view n_strings_col_view(*n_column_view);
```
ttnghia, line 41:
```suggestion
    auto const n_column_view = reinterpret_cast<cudf::column_view const*>(input_column);
    auto const n_strings_col_view = cudf::strings_column_view{*n_column_view};
```

src/main/cpp/src/JSONUtilsJni.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    cudf::jni::native_jintArray path_ins_types_n(env, path_ins_types);
+    cudf::jni::native_jstringArray path_ins_names_n(env, path_ins_names);
+    cudf::jni::native_jlongArray path_ins_indexes_n(env, path_ins_indexes);
+
+    std::vector<int32_t> path_ins_types_v = std::vector<int32_t>(path_ins_types_n.size());
+    env->GetIntArrayRegion(path_ins_types, 0, path_ins_types_n.size(), path_ins_types_v.data());
```
ttnghia, line 48:
```suggestion
    auto const path_ins_types_v = std::vector<int32_t>(path_ins_types_n.begin(), path_ins_types_n.end());
```

src/main/cpp/src/JSONUtilsJni.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    std::vector<int64_t> path_ins_indexes_v = std::vector<int64_t>(path_ins_indexes_n.size());
+    env->GetLongArrayRegion(
+      path_ins_indexes, 0, path_ins_indexes_n.size(), path_ins_indexes_v.data());
```
ttnghia, line 52:
```suggestion
    auto const path_ins_indexes_v = std::vector<int64_t>(path_ins_indexes_n.begin(), path_ins_indexes_n.end());
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -74,10 +74,37 @@ struct path_instruction {
 };
 
 // TODO parse JSON path
```
ttnghia, line 76: Still have `TODO`?

src/main/cpp/src/get_json_object.cu
```diff
@@ -74,10 +74,37 @@ struct path_instruction {
...
+      case 4:  // named
+        path_instructions.push_back(path_instruction(path_instruction_type::named));
+        path_instructions.back().name =
+          cudf::string_view(path_names[i].c_str(), path_names[i].size());
+        break;
+      default: throw std::runtime_error("Invalid path type");
+    }
+  }
+  return cudf::detail::make_device_uvector_sync(
+    path_instructions, stream, rmm::mr::get_current_device_resource());
```
ttnghia, line 107: Oh this will crash. The `cudf::string_view` objects are viewing host strings. We cannot process in device using view of the host strings.

In general, we only create `string_view` for local processing and should never pass it across host<=>device. Depending on the size of `path_names`, we may want to create an `std::vector` of `string_scalar` or a strings column and use it instead.

src/main/cpp/src/get_json_object.cu
```diff
@@ -325,16 +355,15 @@ std::unique_ptr<cudf::column> get_json_object(cudf::strings_column_view const& c
...
-      path_commands_optional.value().size(),
-      sizes.data(),
-      d_offsets,
-      thrust::nullopt,
-      thrust::nullopt,
-      thrust::nullopt,
-      options);
+    <<<grid.num_blocks, grid.num_threads_per_block, 0, stream.value()>>>(*cdv,
+                                                                         path_commands.data(),
+                                                                         path_commands.size(),
```
ttnghia, line 360: `data()` and `size()` parameters can always combined into `device_span`.
```suggestion
device_span{path_commands.data(), path_commands.size()},
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -355,8 +384,8 @@ std::unique_ptr<cudf::column> get_json_object(cudf::strings_column_view const& c
   get_json_object_kernel<block_size>
     <<<grid.num_blocks, grid.num_threads_per_block, 0, stream.value()>>>(
       *cdv,
-      path_commands_optional.value().data(),
-      path_commands_optional.value().size(),
+      path_commands.data(),
+      path_commands.size(),
```
ttnghia, line 388: `device_span`.

src/main/cpp/src/get_json_object.hpp
```diff
@@ -30,8 +31,9 @@ namespace spark_rapids_jni {
  */
 std::unique_ptr<cudf::column> get_json_object(
   cudf::strings_column_view const& col,
-  cudf::string_scalar const& json_path,
-  spark_rapids_jni::json_parser_options options,
+  std::vector<int32_t> const& path_types,
+  std::vector<std::string> const& path_names,
+  std::vector<int64_t> const& path_indexes,
```
ttnghia, line 36: If this set of values always come together, let make a separate struct for them:
```
struct path_info { // this name is just a suggestion, it can be something else if that's better
  int32_t type;
  std::string name;
  int64_t indices;
};
....
get_json_object(..., 
                             std::vector<path_info> paths,
...);
```

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public static ColumnVector getJsonObject(ColumnVector input, int[] path_ins_types, String[] path_ins_names,
+  long[] path_ins_indexes) {
+    assert(input.getType().equals(DType.STRING)) : "column must be a String";
+    assert(path_ins_types.length == path_ins_names.length) : "path_ins_types and path_ins_names must have the same size";
+    assert(path_ins_types.length == path_ins_indexes.length) : "path_ins_types and path_ins_indexes must have the same size";
+    return new ColumnVector(getJsonObject(input.getNativeView(), path_ins_types, path_ins_names, path_ins_indexes));
+  }
+
+  private static native Long getJsonObject(long input, int[] path_ins_types, String[] path_ins_names, long[] path_ins_indexes);
```
ttnghia, line 34: Don't need boxed value here.

```suggestion
  private static native long getJsonObject(long input, int[] path_ins_types, String[] path_ins_names, long[] path_ins_indexes);
```

src/main/cpp/src/get_json_object.hpp
thirtiseven: Updated to `Table` now

src/main/cpp/src/get_json_object.cu
```diff
@@ -70,14 +73,52 @@ struct path_instruction {
...
 
   // used when type is index
-  int index{-1};
+  int64_t index{-1};
 };
 
-// TODO parse JSON path
-thrust::optional<rmm::device_uvector<path_instruction>> parse_path(
-  cudf::string_scalar const& json_path)
+rmm::device_uvector<path_instruction> construct_path_commands(cudf::table_view instructions,
```
ttnghia, line 79:
```suggestion
rmm::device_uvector<path_instruction> construct_path_commands(cudf::table_view const& instructions,
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -70,14 +73,52 @@ struct path_instruction {
...
-  return thrust::nullopt;
+  auto const ins_types   = instructions.column(0);
+  auto const ins_names   = instructions.column(1);
+  auto const ins_indexes = instructions.column(2);
+
+  auto const s_ins_types = cudf::strings_column_view(ins_types);
+  auto const s_ins_names = cudf::strings_column_view(ins_names);
+
+  auto const d_ins_types   = cudf::column_device_view::create(s_ins_types.parent(), stream);
+  auto const d_ins_names   = cudf::column_device_view::create(s_ins_names.parent(), stream);
```
ttnghia, line 91:
```suggestion
  auto const d_ins_types   = cudf::column_device_view::create(ins_types, stream);
  auto const d_ins_names   = cudf::column_device_view::create(ins_names, stream);
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -70,14 +73,52 @@ struct path_instruction {
...
+
+  thrust::transform(rmm::exec_policy(stream),
+                    thrust::make_counting_iterator(0),
+                    thrust::make_counting_iterator(instructions.num_rows()),
+                    path_commands.begin(),
+                    [d_types   = *d_ins_types,
+                     d_names   = *d_ins_names,
+                     d_indexes = *d_ins_indexes] __device__(auto idx) {
+                      path_instruction instruction(path_instruction_type::named);
+                      auto const type_str = d_types.element<cudf::string_view>(idx);
```
ttnghia, line 104: Why don't use `path_instruction_type` but string type here?

res-life: The path_instructions are transferred from JAVA to C++.
We can use int to represent `path_instruction_type `, like 1-> wildcard, 2-> key, ...
But these integers will be magic number in JNI repo.
Becasue the size of path_instructions is probably small, so the perf impact will be small.
There is a tradoff between readability and performance.
@ttnghia  which one do you like?

src/main/cpp/src/get_json_object.cu
```diff
@@ -70,14 +73,49 @@ struct path_instruction {
...
 
   // used when type is index
-  int index{-1};
+  int64_t index{-1};
 };
 
-// TODO parse JSON path
-thrust::optional<rmm::device_uvector<path_instruction>> parse_path(
-  cudf::string_scalar const& json_path)
+rmm::device_uvector<path_instruction> construct_path_commands(cudf::table_view const& instructions,
```
revans2, line 79: Why is the input to this a table? We only support a scalar value as input for the path. The data is small, so we are going to launch a kernel to parse a list of strings on the GPU, when the CPU could do it and send the results to the GPU faster than this. Please change it so that we have a list/vector of C++ objects that represent the parsed data. This cannot be a string because I may want to use the word `star`, `key` or really any other of the special case words you have here in my path, and I don't want it to fail. The magic numbers should not pass through into the cuda kernel, they should only be between the java and the JNI implementation. Java should have an enum for the type, and there should be a corresponding enum in C++. The java code should translate the enum to an int, or byte in this case because there are not many values.

src/main/cpp/src/get_json_object.cu
ttnghia: If we have just a small number of path instructions to process then it may even be better like @revans2 said, using CPU to process them could be better. We should use the enum value only on a large dataset.

src/main/cpp/src/get_json_object.cu
ttnghia: If the input is a small number of instructions (?), let just do all processing on the CPU using host code, or even call the Spark's code to do such preprocessing. Then, we can create a `device_uvector` from the result using `make_device_uvector_sync` for further GPU processing.

thirtiseven: Thanks all! updated to matched enum and struct in cpu.

src/main/cpp/src/get_json_object.cu
res-life: This [commit](https://github.com/NVIDIA/spark-rapids-jni/pull/1868/commits/d45422cb9cf2028a1496f5b15d03737e21c5c9dd) removed the `json_parser_options`.
When merging, will remove it also.

src/main/cpp/src/get_json_object.cu
```diff
@@ -70,14 +68,49 @@ struct path_instruction {
...
+        path_commands.emplace_back(path_instruction{path_instruction_type::named});
+        path_commands.back().name =
+          cudf::string_view(all_names_scalar.data() + name_pos, name.size());
+        name_pos += name.size();
+        break;
+      default: CUDF_FAIL("Invalid path instruction type");
+    }
+  }
+  // convert to uvector
+  return cudf::detail::make_device_uvector_sync(path_commands, stream, mr);
```
res-life, line 113: I doubt if this can work for std::string type.
This method is copying a memory block, but string has a pointer.
I suspect it only copy a pointer and does not copy the data that pointer refers to.
Need to test in kernel code.

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -0,0 +1,53 @@
...
+      this.index = index;
+    }
+  }
+
+  public static ColumnVector getJsonObject(ColumnVector input, int size, PathInstructionJni[] path_instructions) {
+    assert(input.getType().equals(DType.STRING)) : "column must be a String";
+    return new ColumnVector(getJsonObject(input.getNativeView(), size, path_instructions));
+  }
+
+  private static native long getJsonObject(long input, int size, PathInstructionJni[] path_instructions);
```
res-life, line 52: Another option:
Pass in a GPU column view which is a structure:
```
{
    int type,
    int index,
    string name
}
```

or pass in three column views: type cv, index cv, name cv.

src/main/cpp/src/get_json_object.cu
```diff
@@ -294,26 +319,24 @@ __launch_bounds__(block_size) CUDF_KERNEL
...
-  // if the json path is empty, return a string column containing all nulls
-  if (!path_commands_optional.has_value()) {
-    return std::make_unique<cudf::column>(
-      cudf::data_type{cudf::type_id::STRING},
-      col.size(),
-      rmm::device_buffer{0, stream, mr},  // no data
-      cudf::detail::create_null_mask(col.size(), cudf::mask_state::ALL_NULL, stream, mr),
-      col.size());                        // null count
+  // get a string buffer to store all the names and convert to device
+  std::string all_names;
```
res-life, line 331: Rename it to `name`?
Is it a name in a instruction? or names for all instructions?

src/main/cpp/src/get_json_object.cu
```diff
@@ -294,26 +319,24 @@ __launch_bounds__(block_size) CUDF_KERNEL
...
+  for (auto const& inst : instructions) {
+    all_names += std::get<1>(inst);
   }
+  cudf::string_scalar all_names_scalar(all_names);
```
res-life, line 335: Please be careful the life cycle of this class.
It's local variable, and the device memory in the scalar will be de deallocated soon.
Please double check.

res-life: Sorry for this, it's not a problem.

thirtiseven:
Hi @ttnghia , I'm merging this into the feature branch for integration testing with the main flow. Please take another look and I will create follow up pr to fix your comments if needed, thanks.

Issue #1832, opened 2024-03-05
[FEA] GetJsonObject: Implement get-json-object main function


res-life:
**Is your feature request related to a problem? Please describe.**
Implement get-json-object main function

**Additional context**
Epic issue: https://github.com/NVIDIA/spark-rapids-jni/issues/1823

res-life:
Draft PR: https://github.com/NVIDIA/spark-rapids-jni/pull/1868

PR #1868, opened 2024-03-13
get-json-object: main flow

res-life:
closes [#1832](https://github.com/NVIDIA/spark-rapids-jni/issues/1832)

- Implementation of get-json-object main flow
- integration test with JSON generator

### DIFF
--- a/src/main/cpp/src/get_json_object.cu
+++ b/src/main/cpp/src/get_json_object.cu
@@ -51,26 +51,6 @@ namespace spark_rapids_jni {
 namespace detail {
 // namespace {
 
-/**
- * write JSON style
- */
-enum class write_style { raw_style, quoted_style, flatten_style };
-
-/**
- * path instruction
- */
-struct path_instruction {
-  CUDF_HOST_DEVICE inline path_instruction(path_instruction_type _type) : type(_type) {}
-
-  path_instruction_type type;
-
-  // used when type is named type
-  cudf::string_view name;
-
-  // used when type is index
-  int64_t index{-1};
-};
-
 rmm::device_uvector<path_instruction> construct_path_commands(
   std::vector<std::tuple<path_instruction_type, std::string, int64_t>> const& instructions,
   cudf::string_scalar const& all_names_scalar,
@@ -113,105 +93,34 @@ rmm::device_uvector<path_instruction> construct_path_commands(
   return cudf::detail::make_device_uvector_sync(path_commands, stream, mr);
 }
 
-/**
- * TODO: JSON generator
- *
- */
-template <int max_json_nesting_depth = curr_max_json_nesting_depth>
-class json_generator {
- public:
-  CUDF_HOST_DEVICE json_generator(char* _output, size_t _output_len)
-    : output(_output), output_len(_output_len)
-  {
-  }
-  CUDF_HOST_DEVICE json_generator() : output(nullptr), output_len(0) {}
-
-  // create a nested child generator based on this parent generator
-  // child generator is a view
-  CUDF_HOST_DEVICE json_generator new_child_generator()
-  {
-    if (nullptr == output) {
-      return json_generator();
-    } else {
-      return json_generator(output + output_len, 0);
-    }
-  }
-
-  CUDF_HOST_DEVICE json_generator finish_child_generator(json_generator const& child_generator)
-  {
-    // logically delete child generator
-    output_len += child_generator.get_output_len();
-  }
-
-  CUDF_HOST_DEVICE void write_start_array()
-  {
-    // TODO
-  }
-
-  CUDF_HOST_DEVICE void write_end_array()
-  {
-    // TODO
-  }
-
-  CUDF_HOST_DEVICE void copy_current_structure(json_parser<max_json_nesting_depth>& parser)
-  {
-    // TODO
-  }
-
-  /**
-   * Get current text from JSON parser and then write the text
-   * Note: Because JSON strings contains '\' to do escape,
-   * JSON parser should do unescape to remove '\' and JSON parser
-   * then can not return a pointer and length pair (char *, len),
-   * For number token, JSON parser can return a pair (char *, len)
-   */
-  CUDF_HOST_DEVICE void write_raw(json_parser<max_json_nesting_depth>& parser)
-  {
-    if (output) {
-      auto copied = parser.write_unescaped_text(output + output_len);
-      output_len += copied;
-    }
-  }
-
-  CUDF_HOST_DEVICE inline size_t get_output_len() const { return output_len; }
-
- private:
-  char const* const output;
-  size_t output_len;
-};
-
-/**
- * @brief Result of calling a parse function.
- *
- * The primary use of this is to distinguish between "success" and
- * "success but no data" return cases.  For example, if you are reading the
- * values of an array you might call a parse function in a while loop. You
- * would want to continue doing this until you either encounter an error
- * (parse_result::ERROR) or you get nothing back (parse_result::EMPTY)
- */
-enum class parse_result {
-  ERROR,          // failure
-  SUCCESS,        // success
-  MISSING_FIELD,  // success, but the field is missing
-  EMPTY,          // success, but no data
-};
+__device__ bool evaluate_path(json_parser<>& p,
+                              json_generator<>& g,
+                              write_style style,
+                              path_instruction const* path_ptr,
+                              int path_size)
+{
+  return path_evaluator::evaluate_path(p, g, style, path_ptr, path_size);
+}
 
 /**
  * @brief Parse a single json string using the provided command buffer
  *
- * @param j_state The incoming json string and associated parser
- * @param commands The command buffer to be applied to the string. Always ends
- * with a path_operator_type::END
- * @param output Buffer user to store the results of the query
+ * @param j_parser The incoming json string and associated parser
+ * @param path_ptr The command buffer to be applied to the string.
+ * @param path_size Command buffer size
+ * @param output Buffer used to store the results of the query
  * @returns A result code indicating success/fail/empty.
  */
-template <int max_json_nesting_depth = curr_max_json_nesting_depth>
-__device__ parse_result parse_json_path(json_parser<max_json_nesting_depth>& j_parser,
-                                        cudf::device_span<path_instruction const> path_commands,
-                                        json_generator<max_json_nesting_depth>& output)
+__device__ inline bool parse_json_path(json_parser<>& j_parser,
+                                       path_instruction const* path_ptr,
+                                       size_t path_size,
+                                       json_generator<>& output)
 {
-  // TODO
-  return parse_result::SUCCESS;
+  j_parser.next_token();
+  // JSON validation check
+  if (json_token::ERROR == j_parser.get_current_token()) { return false; }
+
+  return evaluate_path(j_parser, output, write_style::raw_style, path_ptr, path_size);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/cpp/src/get_json_object.cu
```diff
@@ -104,25 +104,29 @@ class json_generator {
...
 
-  CUDF_HOST_DEVICE void write_end_array()
+  CUDF_HOST_DEVICE void copy_current_unescaped_text(json_parser<max_json_nesting_depth>& parser)
   {
-    // TODO
+    if (nullptr != output) {
+      int len = parser.write_unescaped_text(output + output_len);
+      output_len += len;
+    } else {
+      int len = parser.compute_unescaped_text(output + output_len);
```
revans2, line 117: Why exactly does the `compute_unescaped_text` take an output at all?

src/main/cpp/src/get_json_object.cu
```diff
@@ -199,13 +542,15 @@ __device__ parse_result parse_json_path(json_parser<max_json_nesting_depth>& j_p
...
-                       size_t out_buf_size,
-                       json_parser_options options)
+get_json_object_single(
+  char const* input,
+  cudf::size_type input_len,
+  path_instruction const* path_commands_ptr,
+  int path_commands_size,
+  char* out_buf,
+  size_t out_buf_size,
+  json_parser_options options)  // TODO make this a reference? use a global singleton options?
```
revans2, line 552: Or just don't have a json_parser_options at all. We don't have a way to change them, so why not just hard code it?

revans2 (commented):
This generally looks like it follows the spark code, but it is kind of hard to tell until we actually have something working.

src/main/cpp/src/get_json_object.cu
res-life: Done.

Now: 
compute_unescaped_len()

src/main/cpp/src/get_json_object.cu
res-life: OK, I'll hard code it.

res-life:
@ttnghia  Help review.

Ready to do integration test with JSON generator.
Will fix more bugs when doing integration test.

TODO: Add more integration cases

res-life:
Added more UT cases from [link](https://github.com/apache/spark/blob/v3.5.0/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/JsonExpressionsSuite.scala)

src/main/cpp/src/get_json_object.cu
```diff
@@ -38,40 +38,20 @@
...
 
 /**
- * path instruction
+ * @brief Result of calling a parse function.
+ *
+ * The primary use of this is to distinguish between "success" and
+ * "success but no data" return cases.  For example, if you are reading the
+ * values of an array you might call a parse function in a while loop. You
+ * would want to continue doing this until you either encounter an error
+ * (parse_result::ERROR) or you get nothing back (parse_result::EMPTY)
```
revans2, line 52: But there is no parse_result::EMPTY in the enum?

src/main/cpp/src/get_json_object.cu
```diff
@@ -38,40 +38,20 @@
...
- * path instruction type
- */
-enum class path_instruction_type { subscript, wildcard, key, index, named };
 
 /**
- * path instruction
+ * @brief Result of calling a parse function.
+ *
+ * The primary use of this is to distinguish between "success" and
+ * "success but no data" return cases.  For example, if you are reading the
```
ttnghia, line 49:
```suggestion
 * "success but no data" return cases. For example, if you are reading the
```

src/main/cpp/src/get_json_object.cu
ttnghia: Same question.

src/main/cpp/src/get_json_object.cu
```diff
@@ -80,106 +60,39 @@ thrust::optional<rmm::device_uvector<path_instruction>> parse_path(
...
- * (parse_result::ERROR) or you get nothing back (parse_result::EMPTY)
- */
-enum class parse_result {
-  ERROR,          // failure
-  SUCCESS,        // success
-  MISSING_FIELD,  // success, but the field is missing
-  EMPTY,          // success, but no data
-};
+template <int max_json_nesting_depth>
+CUDF_HOST_DEVICE bool evaluate_path(json_parser<max_json_nesting_depth>& p,
```
ttnghia, line 64: Nit: Typically we should not use host+device qualifier too much. Only use it for short computation functions. If it does not need to be on device then just declare as a normal function.

src/main/cpp/src/get_json_object.hpp
```diff
@@ -19,10 +19,596 @@
...
+
+/**
+ * write JSON style
+ */
+enum class write_style { raw_style, quoted_style, flatten_style };
+
+/**
+ * path instruction type
+ */
+enum class path_instruction_type { subscript, wildcard, key, index, named };
```
ttnghia, line 41: According to our current style, enum values are `ALL_CAP`.
```suggestion
enum class write_style { RAW_STYLE, ... };
/**
 * path instruction type
 */
enum class path_instruction_type { subscript, wildcard, key, index, named };
```

src/main/cpp/src/json_parser.hpp
```diff
@@ -16,35 +16,40 @@
...
+#include <thrust/tuple.h>
 
 namespace spark_rapids_jni {
 
 /**
  * write style when writing out JSON string
  */
 enum class write_style {
-  // e.g.: '\\r' is a string with 2 chars '\' '\r', writes 1 char '\r'
+  // e.g.: '\\r' is a string with 2 chars '\' 'r', writes 1 char '\r'
```
ttnghia, line 31: Nit: Sorry, again, following style guide, enum values are `ALL_CAP`.

src/main/cpp/src/json_parser.hpp
```diff
@@ -524,23 +536,26 @@ class json_parser {
...
   {
-    return try_parse_quoted_string(
-      str_pos,
-      '\'',
-      to_match_str_pos,  // match str pos, nullptr means do not match
-      to_match_str_end,  // match str end
-      copy_destination,  // copy destination while parsing, nullptr means do not copy
-      w_style);
+    return try_parse_quoted_string(str_pos,
+                                   '\'',
```
ttnghia, line 540: I'm not sure where this function will run (host or device?), as a lot of functions here are marked as `CUDF_HOST_DEVICE`. Generally we should not do that, only `__device__` or without marking (host functions).

Anyway, this may crash, since it is passing a `char*` pointer that may point to memory and that pointer may be passed into device function. If this is a host function then it's fine. If it is a device function, pass a `string_scalar` instead.

src/main/cpp/src/json_parser.hpp
```diff
@@ -550,13 +565,13 @@ class json_parser {
...
   {
-    return try_parse_quoted_string(
-      str_pos,
-      '\"',
-      to_match_str_pos,  // match str pos, nullptr means do not match
-      to_match_str_end,  // match str end
-      copy_destination,  // copy destination while parsing, nullptr means do not copy
-      w_style);
+    return try_parse_quoted_string(str_pos,
+                                   '\"',
```
ttnghia, line 569: Similarly, do not pass a string literal. Pass either `std::string` or `string_scalar`.

ttnghia:
I just did another review iteration. What's most confusing is `CUDA_HOST_DEVICE` is used everywhere. If possible, please just use `__device__` or none if the function does not run on GPU. That will reduce confusion when we want to see where the function is running to check the relevant context.

src/main/cpp/src/get_json_object.cu
```diff
@@ -80,106 +60,39 @@ thrust::optional<rmm::device_uvector<path_instruction>> parse_path(
...
- * with a path_operator_type::END
- * @param output Buffer user to store the results of the query
+ * @param j_parser The incoming json string and associated parser
+ * @param path_ptr The command buffer to be applied to the string.
+ * @param path_size Command buffer size
+ * @param output Buffer used to store the results of the query
  * @returns A result code indicating success/fail/empty.
  */
 template <int max_json_nesting_depth = curr_max_json_nesting_depth>
 __device__ parse_result parse_json_path(json_parser<max_json_nesting_depth>& j_parser,
```
ttnghia, line 84: I see that this function is called only once? If so, we can just eliminate it completely and put the code at the caller.

src/main/cpp/src/get_json_object.hpp
```diff
@@ -19,10 +19,596 @@
...
+#include <thrust/optional.h>
+#include <thrust/pair.h>
+#include <thrust/scan.h>
+#include <thrust/tuple.h>
+
 #include <memory>
 
 namespace spark_rapids_jni {
 
+namespace detail {
```
ttnghia, line 31: So everything in this file and `json_parser.hpp` are running on host CPU?

res-life: In production, `json_parser.hpp`  runs on GPU.
But for the unit test cases in `json_parser_tests.cpp`, they are running on CPU.

`json_parser` `json_start_pos` scans the JSON string view on GPU and writes output to cuDF string column.

`json_parser` is straightforward, it scans char array(parameter is `char *`) and writes to char buffer (parameter is `char *`). So it's convenient to use `CUDF_HOST_DEVICE` to test the logic.
Sorry for introduced extra `HOST` annotation.

src/main/cpp/src/get_json_object.cu
res-life: Updated, added annotation: `inline`

src/main/cpp/src/json_parser.hpp
res-life: The `\"` char will be on GPU stack, IMO, it's safe to put it in the parameter.
I'll test this PR from end to end after this [PR](https://github.com/NVIDIA/spark-rapids-jni/pull/1862) is done

src/main/cpp/src/json_parser.hpp
res-life: > I'm not sure where this function will run (host or device?)

Refer to this [comment](https://github.com/NVIDIA/spark-rapids-jni/pull/1868#discussion_r1530015452)

src/main/cpp/src/get_json_object.hpp
res-life: I'll update the enum to upper case after integration test is done.

src/main/cpp/src/get_json_object.cu
res-life: Refer to this [comment](https://github.com/NVIDIA/spark-rapids-jni/pull/1868#discussion_r1530015452)

src/main/cpp/src/get_json_object.cu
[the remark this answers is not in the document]
res-life: Removed `parse_result`

src/main/cpp/src/get_json_object.cu
res-life: Removed `parse_result`

src/main/cpp/src/get_json_object.cu
res-life: Removed `parse_result`
We do not need `parse_result::EMPTY`.
If `evaluate_path` success and result is empty, sets `output_size` as 0 and `evaluate_path` returns true.

res-life:
Note: the integration test (between JSON generator, JSON parser, Main flow) is done.
The end to end integration test is not done.

res-life:
@ttnghia  Merge this to feature branch to unblock the End to End testing.
We can call another review when merging to Branch-24.04.

src/main/cpp/src/get_json_object.hpp
ttnghia: Although the code can run on both CPU and GPU, I suspect that there may be some parts that cannot run exactly the same way on both places. We should better to implement it only as GPU code, and run unit tests on the GPU data.

res-life: OK, will update

Issue #1666, opened 2023-12-20
[BUG] The build can be broken due to dependency changes of cudf


hyperbolic2346:
**Describe the bug**
The cudf build always grabs the [latest RMM](https://github.com/rapidsai/rapids-cmake/blob/c9a3eeb4b291806b9aa940ed7e3ff4a5a30fe907/rapids-cmake/cpm/versions.json#L86) of a specific branch to build. This can and has caused build breakages as the RMM may not be compatible with the cudf version. Most recently, this was the CCCL 2.2 change that hit RMM and cudf at the same time. We pin a specific cudf commit, so our cudf didn't move forward, but RMM was pulled from latest and so it was using CCCL 2.2. This resulted in builds failing even though we had a pinned cudf commit.

**Steps/Code to reproduce bug**
Sync to a commit of spark-rapids-jni before the CCCL 2.2 changes were merged and build.

**Expected behavior**
We run CI and put a specific cudf commit into the submodule to prevent breakages like this. We should be able to continue using the toolchain from the time of the CI build.

**Additional context**
Bradley proposed a fix by creating a `versions.json` file that can be used with [rapids cpm package override](https://docs.rapids.ai/api/rapids-cmake/stable/command/rapids_cpm_package_override/) to pin to the versions that we actually used to build and test in CI. This leaves cmake itself as the only wildcard. This [can also be pinned](https://github.com/rapidsai/rapids-cmake/blob/branch-24.02/README.md#overriding-rapidscmake) to a specific git SHA.

sameerz:
Related issue: https://github.com/rapidsai/cudf/issues/11533

bdice:
[...]

PR #1877, opened 2024-03-18
Support cudf pin with rapids-cmake

GaryShen2008:
- Set cuDF build with pinned rapids-cmakes in POM
- Use git download the rapids-cmake in cudf build so that we can get the rapids-cmake's SHA
- Update ci/submodule-sync.sh to get the latest cuDF commit and with the latest rapids-cmake

Close https://github.com/NVIDIA/spark-rapids-jni/issues/1666

Main code is copied from https://github.com/robertmaynard/spark-rapids-jni/commit/85f5a2901afe8ae28c945b9d4e58033edef0c85d


GaryShen2008:
build

ci/submodule-sync.sh
```diff
@@ -89,6 +90,18 @@ else
...
+# Extract the rapids-cmake sha1 that we need to pin too
+rapids_cmake_sha=$(git -C thirdparty/cudf/cpp/build/_deps/rapids-cmake-src/ rev-parse HEAD)
+echo "Update rapids-cmake pinned SHA1 to ${rapids_cmake_sha}"
+echo "${rapids_cmake_sha}" > thirdparty/cudf-pins/rapids-cmake.sha
+
+# Do the git add after the build so that we get
+# the updated versions.json generated by the build
+echo "Update cudf submodule to ${cudf_sha} with updated pinned versions"
+git add .
+git diff-index --quiet HEAD || git commit -s -m "Update submodule cudf to ${cudf_sha}"
```
jlowe, line 102: There was already a git add and commit performed above.  That block above was removed in the reference version of this change.

jlowe (commented):
For proper attribution and to make this a bit easier to review vs. the original source, I think this PR should have been done as a cherry pick of the original reference commit to preserve the original author and then commits added to that (e.g.: fixing copyrights, adding the pinned json from the build, etc.)

ci/submodule-sync.sh
GaryShen2008: oh, let me remove above one.
I'll use cherry-pick way to force push the branch and update this PR.

GaryShen2008:
build

robertmaynard (approved):

GaryShen2008:
build

GaryShen2008:
build

ci/submodule-sync.sh
```diff
@@ -63,6 +63,10 @@ if [[ "${cudf_sha}" == "${cudf_prev_sha}" ]]; then
   exit 0
 fi
 
+echo "Try update cudf submodule to ${cudf_sha}..."
+git add .
+git diff-index --quiet HEAD || git commit -s -m "Update submodule cudf to ${cudf_sha}"
```
jlowe, line 68: It looks like this is going to commit twice, since we're committing here and also on line 101 below.  Do we need to commit twice?  Should the second commit be via `--amend`?

ci/submodule-sync.sh
jlowe: build/submodule-check will fail if you don't `git add` the submodule, it shouldn't have to be committed.

ci/submodule-sync.sh
GaryShen2008: Good idea. I forgot the amend way to reduce the twice commits. Let me try.

ci/submodule-sync.sh
GaryShen2008: Let me run a test against it, if so, we just need 2 git add and one git commit at last.

GaryShen2008: Thanks, @jlowe. It works. I remove the first commit line.

ci/submodule-sync.sh
GaryShen2008: Since we remove the first commit, it doesn't require the --amend way because it's a single commit.

jlowe (approved):

jlowe:
build

Issue #1831, opened 2024-03-05
[FEA] GetJsonObject: Implement JSON generator to print JSON items


res-life:
**Is your feature request related to a problem? Please describe.**
Implement JSON generator to print JSON items

**Additional context**
Epic issue: https://github.com/NVIDIA/spark-rapids-jni/issues/1823

SurajAralihalli:
In the cudf's `get_json_object` , writing the output to the result column happens in two phases. In the first phase the size of the output for each row (per thread) in computed. `if (!out_buf.has_value()) { d_sizes[tid] = output_size; }` [json_path.cu#L952](https://github.com/rapidsai/cudf/blob/bd68b1c897741d97684c8555487de759c7576758/cpp/src/json/json_path.cu#L952)



Consider a JSON Column with 2 rows: `{ "ab" : "pqr" },  { "ab" : "lmn" }`

get_json_object( col, "$.ab") returns a cudf column which is made from `make_strings_column`.
 [json_path.cu#L1046](https://github.com/rapidsai/cudf/blob/e612a8aee5ba54f397b8d5be14201776bac9dd2d/cpp/src/json/json_path.cu#L1046)

```
col.size() -> 2
offsets -> 0,3
chars.release() -> pqrlmn
nullcount -> 0
validity -> 1,1
```

In the second phase, the offsets are used in each thread to write to the correct location in the device memory [json_path.cu#L938C7](https://github.com/rapidsai/cudf/blob/e612a8aee5ba54f397b8d5be14201776bac9dd2d/cpp/src/json/json_path.cu#L938C7-L938C89)

 1.   Memory before : xxxxxx
       thread 1 writes at 0 -> pqrxxx
       thread 2 writes at 3 -> xxxlmn

 2.  Memory after: pqrlmn
 
 
However Spark uses temporary memory while writing JSON. [jsonExpressions.scala#L278](https://github.com/apache/spark/blob/ce5ddad990373636e94071e7cef2f31021add07b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala#L278).

```
// temporarily buffer child matches, the emitted json will need to be
        // modified slightly if there is only a single element written
        val buffer = new StringWriter()

        var dirty = 0
        Utils.tryWithResource(jsonFactory.createGenerator(buffer)) { flattenGenerator =>
          flattenGenerator.writeStartArray()

          while (p.nextToken() != END_ARRAY) {
[...]

PR #1897, opened 2024-03-26
getJsonObject number normalization

thirtiseven:
Closes https://github.com/NVIDIA/spark-rapids-jni/issues/1831

This PR supports number normalization in getJsonObject.

In getJsonObject in Spark, a float number is converted to a double when parsing and then converted to a string when returning. This PR uses `stod` in cudf and `ftos_converter` in jni to simulate this behavior. There may be some compatibility issues in `ftos_converter` because it is based on ryu, but it is acceptable because the difference is very minor. We use this solution to support double to string in spark-rapids as well.

For int number, the only special case I know is "-0" is normalized to "0". Note that "-0000000" is invalid as data.

We need to merge it into the feature branch after removing the cpu tests, because this pr calls device-only functions.


src/main/cpp/src/get_json_object.hpp
```diff
@@ -65,22 +65,20 @@ struct path_instruction {
...
     : output(_output), output_len(0), hide_outer_array_tokens(_hide_outer_array_tokens)
   {
   }
-  CUDF_HOST_DEVICE json_generator(bool _hide_outer_array_tokens)
+  __device__ json_generator(bool _hide_outer_array_tokens)
     : output(nullptr), output_len(0), hide_outer_array_tokens(_hide_outer_array_tokens)
   {
   }
-  CUDF_HOST_DEVICE CUDF_HOST_DEVICE json_generator<>& operator=(const json_generator<>& other)
+  __device__ __device__ json_generator<>& operator=(const json_generator<>& other)
```
res-life, line 81: two __device__

res-life:
Add several JNI cases will be good.

res-life:
Yes, -0000000 is not a valid JSON number.

PR #1901, opened 2024-03-26
Pin cudf dependencies during build

### LINKED ISSUE #1666: [BUG] The build can be broken due to dependency changes of cudf

jlowe:
Closes #1666.  This restores the changes from #1877 and updates the rapids-cmake version to avoid the issue reported in NVIDIA/spark-rapids#10627.


jlowe:
build

jlowe:
Verified RAPIDS Accelerator zstd tests pass with this build.

gerashegalov (dismissed):
LGTM

ttnghia (commented):
Does this mean we always pin, and we have to manually upgrade after a while?

jlowe:
> Does this mean we always pin, and we have to manually upgrade after a while?

No.  There's two modes to the build.  The default is to use the pinned versions, but the submodule sync build, which automatically runs every few hours, will build in a mode that updates the pinned versions file to whatever versions it finds when it moves to the updated cudf version.  See https://github.com/NVIDIA/spark-rapids-jni/pull/1901/files#diff-a4b52f1881e34609ce5cd5e0722da2f2f708d1eef03334723a5628082fe94d31

jlowe:
I just realized I should update this PR to refresh the pin versions, since it's been almost a week since those were updated in this PR.  Should have that updated shortly.

jlowe:
build

ttnghia (approved):

gerashegalov (approved):

bdice:
@jlowe This is awesome! I'm really glad this approach worked out.

jlowe:
Thank @bdice, but all the credit goes to @robertmaynard for the idea and implementation.

Issue #1917, opened 2024-04-01
[BUG] spark-rapids-jni JNI does not set RMM_LOGGING_LEVEL


abellina:
This is the mirror image of this issue: https://github.com/rapidsai/cudf/issues/15417 but for spark-rapids-jni.

spark-rapids-jni should have this flag consistently passed. At this stage, I think this is more of a defensive move in case we move pool creation to spark-rapids-jni in the future.

PR #1918, opened 2024-04-01
Set LIBCUDF_LOGGING_LEVEL to allow us to turn off rmm log

abellina:
This is related to https://github.com/NVIDIA/spark-rapids-jni/issues/1917

@jlowe suggested a patch to build-libcudf.xml to ensure we can turn off rmm log. This is necessary due to the new cuDF pinned pool added in cuDF jni, where a pinned pool OOM is treated as a fallback to cudaMallocHost/cudaFreeHost: we don't need a log to be generated here.

This PR removes the generation of such a log.

### DIFF
--- a/build-libcudf.xml
+++ b/build-libcudf.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <!--
-  Copyright (c) 2022, NVIDIA CORPORATION.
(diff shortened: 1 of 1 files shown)


jlowe (approved):

abellina:
build

Issue #1923, opened 2024-04-03
[BUG] getJsonObject gets infinity instead of zero for a very small float number


res-life:
**Describe the bug**
CPU gets zero for number: 9.299999257686047e-0005603333574677677
But GPU gets infinity.
The number is a very small value nearly zero, so zero is the correct answer.

**Steps/Code to reproduce bug**
get_json_object("9.299999257686047e-0005603333574677677", "$")
CPU: 0.0
GPU: infinity

**Additional context**
JNI uses cuDF `string to double` and `double to string`.
cuDF can not hanlde very small float value, need to discuss how to fix:
- fix in JNI
- fix in cuDF

thirtiseven:
For number normalization, first use `stod` to convert string to double, then convert back to string with `ftos_converter`:
```
double d_value = cudf::strings::detail::stod(cudf::string_view(current_token_start_pos, number_token_len))
return spark_rapids_jni::ftos_converter::double_normalization(d_value, destination)
```

It seems that `cudf::strings::detail::stod` will return inf for this case, because the precision is very large for a normal use of string to double. We could try to have a workaround in jni to fix this issue.

I tested that in jni's string to double kernel, this case will return null. maybe investigating on how jackson handles this issue can be helpful.

res-life:
Haoyang filed one follow-up issue in cuDF repo:  https://github.com/rapidsai/cudf/issues/15508

thirtiseven:
Filed https://github.com/rapidsai/cudf/issues/15508 in cuDF. 

I use cuDF’s implementation because our string to double does not have a single kernel to call easily, It needs some refactoring to be used for getJsonObject or other usage. And it is not a bit to bit match to Spark also, although it should be much closer than one in cuDF, we might want to fix it in a long term.

revans2:
> we might want to fix it in a long term.

Then please file an issue to do so.

res-life:
> > we might want to fix it in a long term.
> 
> Then please file an issue to do so.

Filed [an issue](https://github.com/NVIDIA/spark-rapids-jni/issues/1963) to track

res-life:
Close this issue as PR #1944 merged.

PR #1944, opened 2024-04-10
Fix getJsonObject number normalization 0.0 <=> inf bug

thirtiseven:
Fixes https://github.com/NVIDIA/spark-rapids-jni/issues/1923

Floating number normalization in getJsonObject kernel using `cudf::strings::detail::stod` to convert string to double, then convert back to string with `ftos_converter`.

For numbers with very large exp, like `9.299999257686047e-0005603333574677677` or `9.299999257686047e0005603333574677677`, following code will overflow and produce wrong sign:

`9.299999257686047e-0005603333574677677` should be 0.0, but cuDF gets Infinity
`9.299999257686047e0005603333574677677` should be infinity, but cuDF gets 0.0

cuDF is using `int` to save exponent value.
For the above 1st double string, it will overflow from negative value to positive value.
For the above 2nd double string, it will overflow from positive value to negative value.
So cuDF gets totally wrong result.

```diff
  // check for exponent char
  int exp_ten  = 0;
  int exp_sign = 1;
  if (in_ptr < end) {
    char ch = *in_ptr++;
    if (ch == 'e' || ch == 'E') {
      if (in_ptr < end) {
        ch = *in_ptr;
        if (ch == '-' || ch == '+') {
          exp_sign = (ch == '-' ? -1 : 1);
          ++in_ptr;
        }
        while (in_ptr < end) {
          ch = *in_ptr++;
          if (ch < '0' || ch > '9') break;
          exp_ten = (exp_ten * 10) + (int)(ch - '0');
+          if (exp_ten >= 1e8) break;
        }
      }
    }
  }
```

Then with a wrong sign, it will turn to `0.0` or `inf` in a wrong way.

Adding `if (exp_ten >= 1e8) break;` will prevent it from overflowing, and since the exp here is big enough to be an inf, it doesn't matter what the real value is. I choose `1e8` here because it's the largest power of 10 that can be multiplied by 10 one more time without overflowing, so it's the largest possible limit here. The reason not to choose `std::numeric_limits<double>::max_exponent10` (which is 308) here is that there is an `exp_off`, which could be negative, to calculate the final exp later.

I think it's a cudf bug, but let's hot fix it in a copy of stod in jni, to make getJsonObject happy first.

Will file a cudf issue soon.


res-life:
Add some cases in GetJsonObjectTest.java

res-life:
There is only one line change:
```
+          if (exp_ten >= 1e8) break;
```
But the following code will adjust `exp_ten` again:
```
 exp_ten *= exp_sign;
  exp_ten += exp_off;
  exp_ten += num_digits - 1;
```
So if `exp_off` or `num_digits` have very large values, then the change is equivalent to the original logic.

Fortunately, JSON parser defined the max digits for JSON numbers, refer to `json_parser.cuh`
```
constexpr int max_num_len = 1000;
```

So this change is OK.
Please add comment like:
Because of max digits for number in JSON is 1000 (consistent with Spark),  so the adjust will not impact result.

src/main/cpp/src/string_to_float_cudf.cuh
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * @brief This function converts the given string into a
+ * floating point double value.
+ *
+ * This will also map strings containing "NaN", "Inf", etc.
+ * to the appropriate float values.
+ *
+ * This function will also handle scientific notation format.
+ */
+__device__ inline double stod(cudf::string_view const& d_str)
```
res-life, line 36: Please describe this is copied from cuDF to do a temporary fix for branch-24.04.
Long term fix will be in cuDF.

src/main/cpp/src/json_parser.cuh
```diff
@@ -1516,8 +1517,8 @@ class json_parser {
         // 200.000 => 200.0, 351.980 => 351.98, 12345678900000000000.0
         // => 1.23456789E19 0.0000000000003 => 3.0E-13; 0.003 => 0.003; 0.0003
         // => 3.0E-4 leverage function: `get_current_float_parts`
-        double d_value =
-          cudf::strings::detail::stod(cudf::string_view(current_token_start_pos, number_token_len));
+        double d_value = spark_rapids_jni::detail::stod(
```
res-life, line 1520: BTW, please update the comments:
```suggestion
        // number normalization:
        // 0.03E-2 => 0.3E-4; large value => infinity; small value => 0.0
        // 200.000 => 200.0, 351.980 => 351.98, 12345678900000000000.0
        // => 1.23456789E19 0.0000000000003 => 3.0E-13; 0.003 => 0.003; 0.0003
        // => 3.0E-4
```

thirtiseven: done.

src/main/cpp/src/string_to_float_cudf.cuh
thirtiseven: done.

thirtiseven:
build

res-life (approved):

src/main/cpp/src/string_to_float_cudf.cuh
```diff
@@ -0,0 +1,141 @@
...
+        ch = *in_ptr;
+        if (ch == '-' || ch == '+') {
+          exp_sign = (ch == '-' ? -1 : 1);
+          ++in_ptr;
+        }
+        while (in_ptr < end) {
+          ch = *in_ptr++;
+          if (ch < '0' || ch > '9') break;
+          exp_ten = (exp_ten * 10) + (int)(ch - '0');
+          if (exp_ten >= 1e8) break;
```
revans2, line 107: Can we have an explanation about why this is here? This is the only change between CUDF and this version so I would like to understand it better. Mainly because it has a magic number of 1e8 in it and I would really like to understand where that came from.

revans2 (commented):
Just one more comment.

PR #1983, opened 2024-04-19
Explicitly pull in GTest dependency when building C++ tests

jlowe:
@robertmaynard let us know that our explicit use of an implicit GTest dependency via libcudf is going to break in an upcoming libcudf update.  This updates our build with his suggested change to make the dependency explicit.


jlowe:
build

robertmaynard (approved):

abellina (approved):

PR #1987, opened 2024-04-22
Remove Path key/subscript to save GPU memory usage to improve perf

res-life:
Remove Path key/subscript to save GPU memory usage to improve perf.

I found less GPU stack memory usage will help on perf.
I decreased the max path depth from 16 to 8, and get perf improvement.
```
constexpr int max_path_depth = 8;
```
And I optimized the main flow logic to reduced query paths:  
- key and named to one path => named
- subscript and index/wildcard to one path => index/wildcard
From the definition of JSON query path,  key and named always ocurr in the same time, and also subscript and index/wildcard are always occur in the same time. So it's safe to reduce (key, named) to (named),  and it's safe to reduce (subscript, index/wildcard) to (index/wildcard)

Note: If query path is a single `wildcard` , e.g.:  $.* , Spark returns null, so the path is equivalent to invalid path.

## performance test
| Customer   | Base line(1st, 2nd, 3rd) | After this PR  | speed |
|------------|-----------|---------|------------|
| customer 1 | 231500 ms, 232760 ms, 232647 ms | 275719 ms, 272898 ms, 274407 ms |   about 84%  |

### DIFF
--- a/src/main/cpp/src/get_json_object.cu
+++ b/src/main/cpp/src/get_json_object.cu
@@ -322,25 +322,12 @@ __device__ inline bool path_match_elements(path_instruction const* path_ptr,
   return path_ptr[0].type == path_type0 && path_ptr[1].type == path_type1;
 }
 
-__device__ inline bool path_match_elements(path_instruction const* path_ptr,
-                                           size_t path_size,
-                                           path_instruction_type path_type0,
-                                           path_instruction_type path_type1,
-                                           path_instruction_type path_type2,
-                                           path_instruction_type path_type3)
+__device__ inline thrust::tuple<bool, int> path_match_index(path_instruction const* path_ptr,
+                                                            size_t path_size)
 {
-  if (path_size < 4) { return false; }
-  return path_ptr[0].type == path_type0 && path_ptr[1].type == path_type1 &&
-         path_ptr[2].type == path_type2 && path_ptr[3].type == path_type3;
-}
-
-__device__ inline thrust::tuple<bool, int> path_match_subscript_index(
-  path_instruction const* path_ptr, size_t path_size)
-{
-  auto match = path_match_elements(
-    path_ptr, path_size, path_instruction_type::SUBSCRIPT, path_instruction_type::INDEX);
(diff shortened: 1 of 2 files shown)


res-life:
Some IT cases are not passed:

```
========= 9 failed, 58 passed, 27086 deselected, 9 warnings in 18.72s ==========
```

res-life:
All IT cases passed after applying two PRs:

- https://github.com/NVIDIA/spark-rapids/pull/10734
- This PR

src/main/cpp/src/get_json_object.cu
```diff
@@ -696,7 +400,7 @@ __device__ bool evaluate_path(json_parser& p,
   // There is a same constant in JSONUtil.java, keep them consistent when changing
   // Note: Spark-Rapids should guarantee the path depth is less or equal to this limit,
   // or GPU reports cudaErrorIllegalAddress
-  constexpr int max_path_depth = 16;
+  constexpr int max_path_depth = 8;
```
revans2, line 403: Can we please split up these two changes so that we can evaluate the performance of each of them separately? I would really like to see max_path_depth be turned into a template parameter for the kernel. That way we can have a few different kernels and select the right one depending on the path done. That or we can do like @nvdbaranec did on the original version and set the amount of shared memory that would be needed as part of launching the kernel.  But that can be follow on work.

revans2 (approved):

res-life:
build

src/main/cpp/src/get_json_object.cu
res-life: > Can we please split up these two changes so that we can evaluate the performance of each of them separately?

Now, we combine [Key, Named] 2 paths to a single [Named] path, so the change from 16 to 8 is not a tuning.
And: 
[Subscribe, Index] => [Index]
[Subscribe, Wildcard] => [Wildcard]
No need to separate to two PRs.

> That or we can do like @nvdbaranec did on the original version and set the amount of shared memory that would be needed as part of launching the kernel. But that can be follow on work.

OK, will add in a follow-up.

PR #1993, opened 2024-04-23
[submodule-sync] bot-submodule-sync-branch-24.06 to branch-24.06 [skip ci] [bot]

nvauto:
submodule-sync to create a PR keeping thirdparty/cudf up-to-date.
HEAD commit SHA: 40264d34c9acc77c30c55e774d2f11aa8c2140b4, cudf commit SHA: https://github.com/rapidsai/cudf/commit/7341866495b03bdf3f01f8f4e57953741c77e7aa

This PR will be auto-merged if test passed. If failed, it will remain open until test pass or manually fix.

PR #1994, opened 2024-04-23
Use RAPIDS cmake to locate GTest

jlowe:
Fixes submodule sync failure reported at #1993.  We need to use RAPIDS cmake to locate GTest, as a normal find_package will not find it after libcudf dropped it as an exposed dependency.


jlowe:
build

sameerz (approved):

Issue #1963, opened 2024-04-12
[FEA] [follow-up] Revert the PR for number normalization 0.0 <=> inf bug and use cuDF fix


res-life:
**Is your feature request related to a problem? Please describe.**
PR https://github.com/NVIDIA/spark-rapids-jni/pull/1944 is using a temparary fix.

After cuDF fix [this issue](https://github.com/rapidsai/cudf/issues/15508), should revert PR #1944

PR #2000, opened 2024-04-26
Switch string to double in getJsonObject back to cudf 

thirtiseven:
Closes https://github.com/NVIDIA/spark-rapids-jni/issues/1963

To support number normalization in getJsonObject, we need to call string to double function in cudf. But there were an overflow issue, so we ported their code to spark-rapids-jni and fixed it.

After https://github.com/rapidsai/cudf/pull/15517 merged, it is safe to switch it to cudf.


thirtiseven:
build

res-life (approved):
LGTM

Issue #1991, opened 2024-04-23
[BUG] Drop CentOS7 support in 24.06


NvTimLiu:
**Describe the bug**
We plan to drop CentOS 7 (which uses glibc 2.17) & Ubuntu18.04(glibc 2.27) in RAPIDS 24.06. This issue documents some of the work items needed to complete the drop in platform support.  Refor to :

https://docs.rapids.ai/notices/rsn0037/

https://github.com/rapidsai/build-planning/issues/23

https://github.com/NVIDIA/spark-rapids/issues/10496

The new minimum glibc supported by RAPIDS will become 2.28 (used by Rocky 8), because that is the oldest glibc of any operating system we currently support.
- CentOS 7 : 2.17
- Ubuntu18 : 2.27
- Rocky8 :  2.28
- Debian10: 2.28
- Ubuntu20: 2.31
- Ubuntu22: 2.35

Will need to:
  - [ ] Update scripts/Docker of 24.06 for JNI/spark-rapids-JNI to support Rocky8 
  - [ ] Build 24.06 cuDF JNI/spark-rapids-JNI on Rocky8
  - [ ] Aggregate spark-rapids-jni into 24.06 spark-rapids
  - [ ] Drop spark-rapids Integration tests against Ubuntu18 or CentOS7
  - [ ] Check IT on CSP to makes sure no CentOS7 and Ubuntu18 or below OS

NvTimLiu:
Built out amd64/arm64 spark-rapids-jni jars on Rocky8, running integration tests with spark-rapids to see if any GLIBC issues.

Issue #1958, opened 2024-04-11
Support for arm64 in build scripts


jlowe:
The scripts under build/ do not work for ARM builds.  There is a Dockerfile.multi which appears to be unreferenced (at least in this repo) that has some support for ARM, and it's unclear why that file is separate from Dockerfile if it supports building images across multiple architectures.  Ideally build/build-in-docker and build/run-in-docker would work on both x64 and arm64 machines.

NvTimLiu:
I'll take care to this issue later.

NvTimLiu:
Fixed by https://github.com/NVIDIA/spark-rapids-jni/pull/2010

PR #2010, opened 2024-05-02
Drop Centos7 support

NvTimLiu:
To fix: https://github.com/NVIDIA/spark-rapids-jni/issues/1991

Drop Centos7 support, switch to build in a Rocky 8 Docker image

Update the script to support both amd64 and arm64 CPUs


build/build-in-docker
```diff
@@ -24,30 +24,44 @@ set -e
 SCRIPTDIR=$(cd $(dirname $0); pwd)
 
 LOCAL_MAVEN_REPO=${LOCAL_MAVEN_REPO:-"$HOME/.m2/repository"}
-CUDF_USE_PER_THREAD_DEFAULT_STREAM=${CUDF_USE_PER_THREAD_DEFAULT_STREAM:-ON}
 USE_GDS=${USE_GDS:-ON}
 export CMAKE_GENERATOR=${CMAKE_GENERATOR:-"Ninja"}
+export DOCKER_BUILD_EXTRA_ARGS="--platform=linux/amd64 --build-arg CMAKE_ARCH=x86_64"
```
NvTimLiu, line 29: Align with the nightly build script : https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.06/ci/nightly-build.sh

build/build-in-docker
```diff
@@ -24,30 +24,44 @@ set -e
...
+git submodule update --init --recursive
+
 $SCRIPTDIR/run-in-docker mvn \
+    -P${profiles} \
+    -Dlibcudf.build.configure=true \
     -Dmaven.repo.local=$LOCAL_MAVEN_REPO \
-    -DCUDF_USE_PER_THREAD_DEFAULT_STREAM=$CUDF_USE_PER_THREAD_DEFAULT_STREAM \
-    -DUSE_GDS=$USE_GDS \
-    $_CUDF_CLEAN_SKIP \
+    -DUSE_GDS=$USE_GDS -Dtest=*,!CuFileTest,!CudaFatalTest,!ColumnViewNonEmptyNullsTest \
```
jlowe, line 64: This should not be passing test flags, that's a surprising behavior for a build script used by devs.  Arguably something is broken with the setup if we cannot normally build without overriding the test flags, and hacking the build script is not the ideal way to do it.  Unrelated to Centos7 changes.

ci/Dockerfile
jlowe, line 1: Curious why we're not just putting the changes from Dockerfile.multi in to Dockerfile?  Dockerfile is the standard name, not sure why we need to deviate here.

NvTimLiu: Sounds good to me

NvTimLiu: > Curious why we're not just putting the changes from Dockerfile.multi in to Dockerfile? Dockerfile is the standard name, not sure why we need to deviate here.

Sounds good to me

NvTimLiu:
build

NvTimLiu:
build

build/build-in-docker
```diff
@@ -24,30 +24,39 @@ set -e
...
+    arch=amd64;;
+  aarch64|arm64)
+    arch=arm64;;
+  *)
+    echo "Unsupported CPU architecture"; exit 1;;
+esac
+
+# Set env for arm64 build
+if [ "$arch" == "arm64" ]; then
+  profiles="${profiles},arm64"
```
jlowe, line 50: `profiles` appears to be unused.

ci/Dockerfile
```diff
@@ -17,31 +17,28 @@
...
-RUN yum install -y zlib-devel maven tar wget patch ninja-build
-# require git 2.18+ to keep consistent submodule operations
-RUN yum -y install https://packages.endpointdev.com/rhel/7/os/x86_64/endpoint-repo.x86_64.rpm && yum install -y git
-# pin urllib3<2.0 for https://github.com/psf/requests/issues/6432
-RUN scl enable rh-python38 "pip install requests 'urllib3<2.0'"
-
+RUN dnf --enablerepo=powertools install -y scl-utils gcc-toolset-${TOOLSET_VERSION} python39 zlib-devel maven tar wget patch ninja-build git
 ## pre-create the CMAKE_INSTALL_PREFIX folder, set writable by any user for Jenkins
-RUN mkdir /usr/local/rapids && mkdir /rapids && chmod 777 /usr/local/rapids && chmod 777 /rapids
+RUN mkdir /usr/local/rapids /rapids && chmod 777 /usr/local/rapids /rapids
```
jlowe, line 32:
```suggestion
RUN mkdir -m 777 /usr/local/rapids /rapids
```

build/build-in-docker
```diff
@@ -24,30 +24,39 @@ set -e
 SCRIPTDIR=$(cd $(dirname $0); pwd)
 
 LOCAL_MAVEN_REPO=${LOCAL_MAVEN_REPO:-"$HOME/.m2/repository"}
-CUDF_USE_PER_THREAD_DEFAULT_STREAM=${CUDF_USE_PER_THREAD_DEFAULT_STREAM:-ON}
 USE_GDS=${USE_GDS:-ON}
 export CMAKE_GENERATOR=${CMAKE_GENERATOR:-"Ninja"}
+export DOCKER_BUILD_EXTRA_ARGS="--platform=linux/amd64 --build-arg CMAKE_ARCH=x86_64"
+CUDA_VER=${CUDA_VER:-cuda11}
```
jlowe, line 30: run-in-docker is using CUDA_VERSION, this needs to be consistent with that, or we could end up using the wrong Dockerfile for what cuda version Maven thinks it is building for.

build/build-in-docker
```diff
@@ -24,30 +24,39 @@ set -e
...
+  *)
+    echo "Unsupported CPU architecture"; exit 1;;
+esac
+
+# Set env for arm64 build
+if [ "$arch" == "arm64" ]; then
+  profiles="${profiles},arm64"
+  USE_GDS="OFF"
+  USE_SANITIZER="ON"
+  BUILD_FAULTINJ="OFF"
```
jlowe, line 53: Curious why we don't build faultinj on arm64?  Would be good to have a comment here explaining the deviation.

build/build-in-docker
```diff
@@ -24,30 +24,39 @@ set -e
...
+  aarch64|arm64)
+    arch=arm64;;
+  *)
+    echo "Unsupported CPU architecture"; exit 1;;
+esac
+
+# Set env for arm64 build
+if [ "$arch" == "arm64" ]; then
+  profiles="${profiles},arm64"
+  USE_GDS="OFF"
```
jlowe, line 51: Curious why we don't build GDS on arm64?  It used to be separate but now is part of the CUDA toolkit.  Is it not part of the arm64 CUDA toolkit?

build/build-in-docker
```diff
@@ -24,30 +24,39 @@ set -e
...
+  export DOCKER_BUILD_EXTRA_ARGS="--platform=linux/arm64 --build-arg CMAKE_ARCH=aarch64"
 fi
 
 $SCRIPTDIR/run-in-docker mvn \
     -Dmaven.repo.local=$LOCAL_MAVEN_REPO \
-    -DCUDF_USE_PER_THREAD_DEFAULT_STREAM=$CUDF_USE_PER_THREAD_DEFAULT_STREAM \
     -DUSE_GDS=$USE_GDS \
-    $_CUDF_CLEAN_SKIP \
+    -DBUILD_TESTS=ON -DBUILD_FAULTINJ=${BUILD_FAULTINJ} -Dcuda.version=$CUDA_VER \
+    -DUSE_SANITIZER=${USE_SANITIZER} \
```
jlowe, line 61: As mentioned before, this script is used by developers, shouldn't be doing surprising things like adding cmdline options to Maven that may conflict with options the developer is specifying.  Not sure why these aren't specified in the CI jobs using this script rather than putting them here.  If we want these to always be on, they shouldn't be specified here, they should be specified in the pom.xml.  Specifically talking about `BUILD_TESTS`,  and `USE_SANITIZER`.  I'm OK with specifying cuda.version since that's closely tied to the Docker image used, and run-in-docker is already deciding that based on CUDA_VERSION.

build/build-in-docker
jlowe: Actually, a better way to handle that profile is to update pom.xml so it automatically triggers based on the CPU architecture, that way users and scripts never need to specify it to get the right one, e.g:
```
  <profile>
    <id>arm64</id>
    <activation>
      <os>
        <arch>arm64</arch>
      </os>
    </activation>
    ....
```

NvTimLiu:
build

build/build-in-docker
NvTimLiu: > Curious why we don't build GDS on arm64? It used to be separate but now is part of the CUDA toolkit. Is it not part of the arm64 CUDA toolkit?

Yes, GDS cuFiles lib is not in the arm64 CUDA toolkit

[INFO]      [exec]   Could NOT find cuFile (missing: cuFile_LIBRARY cuFileRDMA_LIBRARY
[INFO]      [exec]   cuFile_INCLUDE_DIR)

NvTimLiu:
build

build/build-in-docker
NvTimLiu: Sounds good to me, updated!

build/build-in-docker
NvTimLiu: Added explanations for why we set these vars

build/build-in-docker
```diff
@@ -24,30 +24,29 @@ set -e
 SCRIPTDIR=$(cd $(dirname $0); pwd)
 
 LOCAL_MAVEN_REPO=${LOCAL_MAVEN_REPO:-"$HOME/.m2/repository"}
-CUDF_USE_PER_THREAD_DEFAULT_STREAM=${CUDF_USE_PER_THREAD_DEFAULT_STREAM:-ON}
 USE_GDS=${USE_GDS:-ON}
 export CMAKE_GENERATOR=${CMAKE_GENERATOR:-"Ninja"}
+export DOCKER_BUILD_EXTRA_ARGS="--build-arg TARGETPLATFORM=linux/amd64 --build-arg CMAKE_ARCH=x86_64"
+CUDA_VERSION=${CUDA_VERSION:-cuda11}
```
jlowe, line 30: Argh, run-in-docker expects CUDA_VERSION to be 11.8.0 not cuda11, but this script needs the classifier version.  To make it consistent, I think it needs to look like this:
```suggestion
export CUDA_VERSION=${CUDA_VERSION:-11.8.0}
CUDA_CLASSIFIER=cuda${CUDA_VERSION%%.*}
```
and use $CUDA_CLASSIFIER where CUDA_VERSION is used below.

build/build-in-docker
jlowe: We need to fix the tests on arm64, not hack the build script to run the sanitizer to hide the problem.  We will not be running the sanitizer in production, so letting this slip through by hacking the build script is not what we want.

build/build-in-docker
NvTimLiu: Good catch, fixed

NvTimLiu:
@jlowe  I've filed an issue for the `rmm OOM UT failed on arm64 if USE_SANITIZER=OFF`

Should we leave it unfixed and follow it up in the issue below?  
We can merge this PR first to drop centOS7 support for 24.06 release since there're only 2-3 weeks remaining before v24.06 release, thanks!

https://github.com/NVIDIA/spark-rapids-jni/issues/2022

NvTimLiu:
build

jlowe:
If tests are failing it's a blocker for this PR.  Hacking the script to pass the test is not the right answer.  The goal is to produce a good build on rocky8, but the failing tests indicate there may be issues with the build.  The tests are there for a reason, to indicate whether the code is working properly.  If the tests used to pass when building on centos7 but fail when building on rocky8 then there may be an issue with the code produced by the rocky8 build.  We need to investigate the test failures and fix them.

build/build-in-docker
```diff
@@ -24,30 +24,30 @@ set -e
...
-    # Don't skip clean if ccache is enabled
-    # unless the user overrides
-    _CUDF_CLEAN_SKIP="-Dlibcudf.clean.skip=false"
-  fi
+# Set env for arm64 build, The possible values of 'uname -m' : [x86_64/i386/aarch64/mips/...]
+if [ "$(uname -m)" == "aarch64" ]; then
+  USE_GDS="OFF" # The GDS cuFiles RDMA libraries are not included in the arm64 CUDA toolkit.
+  USE_SANITIZER="ON" # OOM failures in rmm occur during arm64 unit tests when USE_SANITIZER=OFF
+  BUILD_FAULTINJ="OFF" # libcupti_static.a linked by cufaultinj, does not exist in the arm64 CUDA toolkit.
+  export DOCKER_BUILD_EXTRA_ARGS="--build-arg TARGETPLATFORM=linux/arm64 --build-arg CMAKE_ARCH=aarch64"
```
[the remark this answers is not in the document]
NvTimLiu: "Besides DOCKER_BUILD_EXTRA_ARGS, all other parameters/variables defined in build-in-docke are for the use of run-in-docker."

"Do you mean that we should merge these two files, build-in-docke and run-in-docker, into one file called run-in-docker?"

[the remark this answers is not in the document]
jlowe: No, definitely do not merge these two scripts.  build-in-docker is designed for building, it knows about maven, etc.  run-in-docker is a lower-level script used by build-in-docker that developers can use to run something _other_ than maven.  What I'm saying here is that this line looks like it's trying to force a certain image to use based on the platform architecture.  That's something run-in-docker should do, not this script, since it has nothing to do with just building with maven -- anything we want to run needs to pick the proper image, thus it's a run-in-docker change.

To be honest, I was surprised to see so many changes to build-in-docker since run-in-docker is the script that's supposed to know the intimate details of what image to pick.  Given cupti is not available on arm64, it makes some sense, but a lot of other changes don't.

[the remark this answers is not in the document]
NvTimLiu: agree, let me update it!

NvTimLiu:
build

ci/Dockerfile
```diff
@@ -17,31 +17,29 @@
...
 #
 ###
 ARG CUDA_VERSION=11.8.0
-FROM nvidia/cuda:$CUDA_VERSION-devel-centos7
-ARG DEVTOOLSET_VERSION=11
+ARG OS_RELEASE=8
+ARG TARGETPLATFORM=linux/amd64
+# multi-platform build with: docker buildx build --platform linux/arm64,linux/amd64 <ARGS> on either amd64 or arm64 host
+# check available official arm-based docker images at https://hub.docker.com/r/nvidia/cuda/tags (OS/ARCH)
+FROM --platform=$TARGETPLATFORM nvidia/cuda:$CUDA_VERSION-devel-rockylinux$OS_RELEASE
```
jlowe, line 28: Do we need to specify a platform here?  I would expect Docker to automatically select a base image compatible with the current processor architecture (e.g.: we didn't have to specify this before when using other OS base images available on multiple architectures).  If we're always going to build on the platform we expect to run on, I'm not sure why we need to explicitly state the platform here and in the build command.

jlowe:
Converted this to a draft PR, as we don't want this merged until we get to the bottom of the arm64 test failures after building on rocky8, tracked by #2022.

build/build-in-docker
```diff
@@ -24,30 +24,26 @@ set -e
 SCRIPTDIR=$(cd $(dirname $0); pwd)
 
 LOCAL_MAVEN_REPO=${LOCAL_MAVEN_REPO:-"$HOME/.m2/repository"}
-CUDF_USE_PER_THREAD_DEFAULT_STREAM=${CUDF_USE_PER_THREAD_DEFAULT_STREAM:-ON}
 USE_GDS=${USE_GDS:-ON}
-export CMAKE_GENERATOR=${CMAKE_GENERATOR:-"Ninja"}
```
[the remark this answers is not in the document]
NvTimLiu: CMAKE_GENERATOR is similar to DOCKER_BUILD_EXTRA_ARGS, also a var for docker env, not for maven build.

ci/Dockerfile
NvTimLiu: We have the CI jobs to cross-build the arm64 JNI images on the x86_64 Blossom machines.

`--platform=$TARGETPLATFORM` is set to be compatible with docker cross-building docker cases.

build/build-in-docker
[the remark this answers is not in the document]
jlowe: CMAKE_GENERATOR is not a variable for the Docker environment.  It's only used when invoking cmake to configure a projecft, not building cmake itself. Developers have used run-in-docker many times in the past, it's not needed here, hence wondering why this is being moved here.  It is only relevant when building the project, and that's what build-in-docker is for.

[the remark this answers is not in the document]
NvTimLiu: Sounds good to me, moved it back

NvTimLiu:
Here's the relevant fix for the issue2022 : https://github.com/rapidsai/cudf/pull/15706

jlowe (approved):

NvTimLiu:
build

ttnghia:
Unfortunately this breaks my local build:
```
[INFO]      [exec] /usr/local/cmake-3.26.4-linux-x86_64/bin/cmake --regenerate-during-build -S/home/nghiat/Devel/jni/1/thirdparty/cudf/cpp -B/home/nghiat/Devel/jni/1/thirdparty/cudf/cpp/build
[INFO]      [exec] CMake Error at CMakeLists.txt:27 (project):
[INFO]      [exec]   The CMAKE_C_COMPILER:
[INFO]      [exec] 
[INFO]      [exec]     /opt/rh/devtoolset-11/root/usr/bin/cc
[INFO]      [exec] 
[INFO]      [exec]   is not a full path to an existing compiler tool.
[INFO]      [exec] 
[INFO]      [exec]   Tell CMake where to find the compiler by setting either the environment
[INFO]      [exec]   variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
[INFO]      [exec]   the compiler, or to the compiler name if it is in the PATH.
[INFO]      [exec] 
[INFO]      [exec] 
[INFO]      [exec] CMake Error at CMakeLists.txt:27 (project):
[INFO]      [exec]   The CMAKE_CXX_COMPILER:
[INFO]      [exec] 
[INFO]      [exec]     /opt/rh/devtoolset-11/root/usr/bin/c++
[INFO]      [exec] 
[INFO]      [exec]   is not a full path to an existing compiler tool.
[INFO]      [exec] 
[INFO]      [exec]   Tell CMake where to find the compiler by setting either the environment
[INFO]      [exec]   variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
[INFO]      [exec]   to the compiler, or to the compiler name if it is in the PATH.
[INFO]      [exec] 
[INFO]      [exec] 
[INFO]      [exec] ninja: error: rebuilding 'build.ninja': subcommand failed

```

jlowe:
I suspect if you clean out thirdparty/cudf/cpp/build it will fix that issue.  The compiler changed locations between the old and new image, so anything caching the compiler location (such as an existing cmake cache file) will break.

ttnghia:
Oh yes, that fixes it. Thanks.

NvTimLiu:
Also fixed https://github.com/NVIDIA/spark-rapids-jni/issues/1958

Issue #2029, opened 2024-05-09
[BUG] Incorrect percentiles when `frequency == 0`


mythrocks:
This has to do with https://github.com/NVIDIA/spark-rapids/issues/10719 and https://github.com/NVIDIA/spark-rapids/issues/10738.

The bugs above are a result of percentile values being reported incorrectly for rows where the frequency is set to `0`. Here is an illustration, in SparkSQL:
```sql
SELECT val, PERCENTILE( val, 1, freq ) 
FROM VALUES (10, 0L), (30, 1) AS TAB(val, freq) GROUP BY val
```
Apache Spark SQL returns:
```
+---+------------------------+
|val|percentile(val, 1, freq)|
+---+------------------------+
| 10|                    NULL|
| 30|                    30.0|
+---+------------------------+
```
The plugin returns:
```
+---+------------------------+
|val|percentile(val, 1, freq)|
+---+------------------------+
| 10|                     0.0|
| 30|                    30.0|
+---+------------------------+
```

This can be simulated with the following Java test in `spark-rapids-jni`:
```java
package com.nvidia.spark.rapids.jni;

import ai.rapids.cudf.AssertUtils;
import ai.rapids.cudf.ColumnVector;

import ai.rapids.cudf.TableDebug;
import org.junit.jupiter.api.Test;

public class PercentileTest {

    @Test
    void testZeroFrequency() {
      try (ColumnVector values = ColumnVector.fromInts(10, 30);
           ColumnVector freqs = ColumnVector.fromLongs(0, 1);
           ColumnVector histo = Histogram.createHistogramIfValid(values, freqs, true);
           ColumnVector percentiles = Histogram.percentileFromHistogram(histo, new double[]{1}, false);
           ColumnVector expected = ColumnVector.fromBoxedInts(null, 30)) {

        TableDebug.get().debug("Histograms: ", histo);
        System.out.println("\n");
        TableDebug.get().debug("Percentiles as non-list: ", percentiles);
        AssertUtils.assertColumnsAreEqual(percentiles, expected);
      }
    }
}
```

It appears that the problem is not in the construction of the histogram, but in the evaluation of the percentile values. From my tracing, the histograms look about right:
```
GPU COLUMN Histograms:  - NC: 0 DATA: null VAL: null
GPU COLUMN Histograms: :DATA - NC: 0 DATA: null VAL: null
[...]

PR #2030, opened 2024-05-10
Fix null handling in `percentile` evaluation when the input histogram is empty

ttnghia:
Previously, `percentile` outputs a `null` when the input histogram contains all nulls. It missed the case when the input is empty (which also need to output a `null`).

This PR fixes that issue.

Closes https://github.com/NVIDIA/spark-rapids-jni/issues/2029.


ttnghia:
build

mythrocks (approved):
LGTM! Thank you for the fix.

PR #2042, opened 2024-05-15
Add Python requests module to Dockerfile

jlowe:
After #2010 the Python requests module was dropped from the Docker image.  This adds it back.


jlowe:
build

sameerz (approved):

ttnghia (approved):

ttnghia:
This seems not enough. At least we need to add the module `request`:
```
13:57:43      import requests
13:57:43  ModuleNotFoundError: No module named 'requests'
```

jlowe:
I manually verified the requests module can be imported after this change, so my guess is that the CI ran in a stale Docker image.

PR #2047, opened 2024-05-16
Make the max jsonpath depth 16 to match other values in the code

revans2:
I will do some more testing, but my initial benchmarks didn't show any regression in performance by making this change.

### DIFF
--- a/src/main/cpp/src/get_json_object.cu
+++ b/src/main/cpp/src/get_json_object.cu
@@ -387,7 +387,7 @@ __device__ bool evaluate_path(json_parser& p,
   // There is a same constant in JSONUtil.java, keep them consistent when changing
   // Note: Spark-Rapids should guarantee the path depth is less or equal to this limit,
   // or GPU reports cudaErrorIllegalAddress
-  constexpr int max_path_depth = 8;
+  constexpr int max_path_depth = 16;
 
   // define stack; plus 1 indicates root context task needs an extra memory
   context stack[max_path_depth + 1];


revans2:
build

revans2:
I ran the get_json_object benchmarks.  There is a small performance regression . It appears to be up to 2.63%. I think that is an acceptable price to pay so that a customer can actually execute the queries that they have, but I am open to discussions.  Either way the JAVA and native values for this were out of sync with each other and at least one of them needs to be updated.

```
['./16.json', './8.json']
# get_json_object

## [0] NVIDIA RTX A6000

|  size_bytes  |  max_depth  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |         Diff |   %Diff |  Status  |
|--------------|-------------|------------|-------------|------------|-------------|--------------|---------|----------|
|   1000000    |      2      |   3.928 ms |       0.82% |   3.915 ms |       1.30% |   -13.203 us |  -0.34% |   PASS   |
|   10000000   |      2      |   4.372 ms |       2.33% |   4.341 ms |       2.57% |   -30.243 us |  -0.69% |   PASS   |
|  100000000   |      2      |  18.506 ms |       0.42% |  18.361 ms |       0.48% |  -145.114 us |  -0.78% |   FAIL   |
|  1000000000  |      2      | 154.152 ms |       0.40% | 151.985 ms |       0.15% | -2166.727 us |  -1.41% |   FAIL   |
|   1000000    |      4      |   3.982 ms |       0.18% |   3.928 ms |       0.17% |   -54.017 us |  -1.36% |   FAIL   |
|   10000000   |      4      |   4.421 ms |       1.30% |   4.389 ms |       1.66% |   -31.769 us |  -0.72% |   PASS   |
|  100000000   |      4      |  18.038 ms |       0.24% |  17.641 ms |       0.50% |  -396.275 us |  -2.20% |   FAIL   |
|  1000000000  |      4      | 152.524 ms |       0.10% | 149.448 ms |       0.20% | -3076.387 us |  -2.02% |   FAIL   |
|   1000000    |      6      |   4.027 ms |       0.13% |   4.002 ms |       0.17% |   -25.417 us |  -0.63% |   FAIL   |
|   10000000   |      6      |   4.464 ms |       5.87% |   4.442 ms |       0.42% |   -22.206 us |  -0.50% |   FAIL   |
|  100000000   |      6      |  18.192 ms |       0.42% |  17.713 ms |       0.36% |  -478.902 us |  -2.63% |   FAIL   |
|  1000000000  |      6      | 153.036 ms |       0.21% | 150.535 ms |       0.17% | -2500.682 us |  -1.63% |   FAIL   |
|   1000000    |      8      |   4.059 ms |       0.15% |   4.006 ms |       0.23% |   -53.456 us |  -1.32% |   FAIL   |
|   10000000   |      8      |   4.495 ms |       0.32% |   4.472 ms |       1.67% |   -23.268 us |  -0.52% |   FAIL   |
|  100000000   |      8      |  18.409 ms |       0.50% |  18.018 ms |       0.56% |  -391.007 us |  -2.12% |   FAIL   |
|  1000000000  |      8      | 154.194 ms |       0.19% | 152.705 ms |       0.20% | -1488.704 us |  -0.97% |   FAIL   |

# Summary

- Total Matches: 16
  - Pass    (diff <= min_noise): 3
  - Unknown (infinite noise):    0
  - Failure (diff > min_noise):  13
```

revans2:
I am converting this to draft because it looks like I don't understand the mapping between the two values very well, and we probably need to update the docs to make it all clear.

res-life:
After this [change](https://github.com/NVIDIA/spark-rapids-jni/pull/1987), Spark-Rapids uses length: 
```
MAX_PATH_DEPTH = 16;
```

JNI uses depth 8.

Because JNI reduces path items by dividing 2.
e.g.:  $.key has 2 path items: KEY, NAMED, [change](https://github.com/NVIDIA/spark-rapids-jni/pull/1987) removed the KEY path.

(KEY NAMED) always occurs with pair.
And (SUBSRIPT INDEX) always occurs with pair.

So [change](https://github.com/NVIDIA/spark-rapids-jni/pull/1987) removed the KEY and SUBSRIPT before do actual `get-json-object`

Should also modify `JSONUtils.MAX_PATH_DEPTH`,  this is used by Spark-Rapids jar.

e.g.:

Update `get_json_object.cu`
```
constexpr int max_path_depth = 12;
```
Update `JSONUtils`
```
public static final int MAX_PATH_DEPTH = 24;
```

res-life:
Legacy cuDF counts continuous NAMEDs as 1 level, new version counts continuous NAMEDs as multiple levels.
So cuDF supports `$.k1.k2.k3.k4.k5.k6.k7.k8.k9.k10`, but new version will fail.
Because new version needs to validate JSON while doing main flow, it's hard to combine the multiple NAMED handling.

For example:
JSON is
```
[
    {
        "k1": {
            "k2": "v2",
            "k_other": "xx"
        },
        "k_other": "xx"
    }, 

    {
        "k1": {
            "k2": "v2",
            "k_other": "xx"
        },
        "k_other": "xx"
    }
]
```
Query is $[*].k1.k2
Now new version logic is:
Push 1st element (let's note array[0] ) in the array into stack to handle
Push .k1 into Stack
Push .k2 into Stack
Execute .k2, writes JSON generator; skip/validate the keys after 'k2' (here it's k_other)
Pop .k2
Execute .k1; skip/validate the keys after 'k1' (here it's k_other)
Pop .k1
Pop array[0]
Push 2nd element (let's note array[1]) in the array into stack to handle
// the following are similar
Push .k1 into Stack
Push .k2 into Stack
Execute .k2, writes JSON generator; skip/validate the keys after 'k2' (here it's k_other)
Pop .k2
Execute .k1; skip/validate the keys after 'k1' (here it's k_other)
Pop .k1
Pop array[1]

revans2:
Thanks for the info @res-life I figured that out/remembered it after reading through the old PR again. It was really confusing from an end user standpoint so I put up another PR in the plugin, that along with this one should make it more consistent.

This PR now depends on https://github.com/NVIDIA/spark-rapids/pull/10833 to go in first, or it will break the plugin build.

revans2:
The info from nvcc for the spill/etc for a max depth of 8 is.

```
[INFO]      [exec]     976 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
[INFO]      [exec] ptxas info    : Used 86 registers, 128 bytes smem, 504 bytes cmem[0], 10336 bytes cmem[2]
```

For 16 it is.

```
[INFO]      [exec]     1744 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
[INFO]      [exec] ptxas info    : Used 86 registers, 128 bytes smem, 504 bytes cmem[0], 10336 bytes cmem[2]
```

So it increases the stack frame size by 768 bytes. But it didn't change the amount of spill or other memory used, so it didn't technically change the occupancy of the kernel at all, which is probably why there is no performance difference.

revans2:
I did some more profiling and neither the theoretical nor the actual occupancy changed at all with this change in the current form of the kernel. This is because even though the stack frame size changes a lot. The actual physical memory assigned to it didn't change at all.

revans2:
build

jlowe (approved):

res-life (approved):
LGTM

res-life:
Tested Baidu/Bytedance benchmark data, did not show perf regression compared to `max_depth = 8`

PR #2057, opened 2024-05-21
Fail get_json_object calls that are too long

revans2:
This is a backup just in case the plugin messes up and we need a get_json_object call that is too deep for that the code can support. The will change a possible data corruption/hang into an error.


revans2:
build

revans2:
build

jlowe (approved):

PR #2063, opened 2024-05-22
Fix cast_string_to_float with trailing whitespaces for inf and nan string

Feng-Jiang28:
Fix the issue listed in [#10794](https://github.com/NVIDIA/spark-rapids/issues/10794). 
 
In RapidsCastSuit of Spark UT, "Process Infinity, -Infinity, NaN in case insensitive manner" case would fail, the reason is that in jni's `cast_to_float.cu`, the functions for checking inf and nan were considering "inf"("infinity")  and "nan" strings with additional characters as illegal. I added the logic for checking if the additional characters are whitespaces.


src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -241,15 +241,24 @@ class string_to_float {
...
-        _except = _len != 3;
-      }
-      return true;
+      // if we start with 'nan', then even if we have other garbage character(excluding
+      // whitespaces), this is a null row. but for e.g. : "nan   " cases. spark will treat the as
+      // "nan", when the trailing characters are whitespaces, it is still a valid string. if we're
+      // in ansi mode and this is not -precisely- nan, report that so that we can throw an exception
+      // later.
+
+      // move forward the curren position by 3
```
thirtiseven, line 250: typo: current?

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -241,15 +241,24 @@ class string_to_float {
...
+      _bpos += 3;
+      _c = __shfl_down_sync(0xffffffff, _c, 3);
+
+      // remove the trailing whitespaces, if there exits
+      remove_leading_whitespace();
```
thirtiseven, line 255: How about renaming the function to a more general name like just `remove_whitespace` since we are removing trailing whitespaces but not leading whitespaces here.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -59,6 +59,60 @@ void castToIntegerTest() {
     }
   }
 
+
+  //This is for testing the CastStrings class with the input of "inf" etc string
```
thirtiseven, line 63: nit: This testcase looks self-describing, how about removing the comments? Also for `castToFloatNanTest`.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -59,6 +59,60 @@ void castToIntegerTest() {
...
+  @Test
+  void castToFloatsInfTest(){
+    // The test data: Table.TestBuilder object with a column containing the string "inf"
+    Table.TestBuilder tb2 = new Table.TestBuilder();
+    tb2.column("INFINITY ", "inf", "+inf ", " -INF  ");
+
+    Table.TestBuilder tb = new Table.TestBuilder();
+    tb.column(Float.POSITIVE_INFINITY, Float.POSITIVE_INFINITY, Float.POSITIVE_INFINITY, Float.NEGATIVE_INFINITY);
```
thirtiseven, line 71: Let's test double type too.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -59,6 +59,60 @@ void castToIntegerTest() {
     }
   }
 
+
+  //This is for testing the CastStrings class with the input of "inf" etc string
+  @Test
+  void castToFloatsInfTest(){
```
thirtiseven, line 65: nit: How about placing these cases somewhere near existing castToFloats test cases like after `castToFloatsTrimTest`?

src/main/cpp/src/cast_string_to_float.cu
hyperbolic2346: This isn't what the function does though. I know exactly what will happen if I call `remove_leading_whitespace("  the quick brown fox");`, but what is the expected output of `remove_whitespace("  the quick brown fox");`? It could easily be expected that it means `thequickbrownfox`, which this function isn't doing for us. The internal implementation that we are moving a pointer around is not immediately obvious, but each time we call that, we are removing whitespace from the start of the current string.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -295,13 +304,21 @@ class string_to_float {
                                                  (_warp_lane == 4 && (_c == 'Y' || _c == 'y')));
       if (infinity_mask == 0x1f) {
         _bpos += 5;
+        _c = __shfl_down_sync(0xffffffff, _c, 5);
         // if we're at the end
         if (_bpos == _len) { return true; }
```
hyperbolic2346, line 309:
```suggestion
        // if we're at the end
        if (_bpos == _len) { return true; }
        _c = __shfl_down_sync(0xffffffff, _c, 5);
```
No need to shuffle down if we're leaving

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
hyperbolic2346: Some negative tests would be nice. Some strings like `INFINITY AND BEYOND`.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -59,6 +59,60 @@ void castToIntegerTest() {
...
+      }
+    }
+  }
+
+  //This is for testing the CastStrings class with the input of "nan " etc string
+  @Test
+  void castToFloatNanTest(){
+    // The test data: Table.TestBuilder object with a column containing the string "inf"
+    Table.TestBuilder tb2 = new Table.TestBuilder();
+    tb2.column("nan", "nan ", " nan ", "NAN", "nAn ", " NAn ");
```
hyperbolic2346, line 94: Negative tests? `nan o`

hyperbolic2346 (changes requested):
Looks good. Thank you for doing this work!

src/main/cpp/src/cast_string_to_float.cu
Feng-Jiang28: Yes, this remove_leading_whitespace() removes the leading whitespace of the remaining string to process. The name make sense this way.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
Feng-Jiang28: Done

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
Feng-Jiang28: `INFINITY AND BEYOND`  added.
with aother one test string `INF`.

src/main/cpp/src/cast_string_to_float.cu
Feng-Jiang28: done thank you

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
Feng-Jiang28: "Nan 0", "nan  nan" test string added.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -297,11 +306,19 @@ class string_to_float {
         _bpos += 5;
         // if we're at the end
         if (_bpos == _len) { return true; }
+        _c = __shfl_down_sync(0xffffffff, _c, 5);
       }
 
+      // remove the remaining whitespace if exits
```
hyperbolic2346, line 312:
```suggestion
      // remove the remaining whitespace if exists
```

hyperbolic2346 (dismissed):
Small typo fix, but I'm happy with this. Thanks for your work here!

src/main/cpp/src/cast_string_to_float.cu
Feng-Jiang28: typo fixed. :)

Feng-Jiang28:
> Small typo fix, but I'm happy with this. Thanks for your work here!

Thank you for reviewing, appreciate it.

thirtiseven (approved):
Looks good to me 👍

thirtiseven:
build

hyperbolic2346 (approved):

thirtiseven:
build

thirtiseven (approved):

hyperbolic2346 (approved):

PR #2066, opened 2024-05-22
Profiler class and native code to support self-profiling

jlowe:
Contributes to NVIDIA/spark-rapids#10632.  Adds a new Profiler class and native code that can be used to collect CUDA and NVTX profiling data for the current process via CUPTI.  It supports 

The profiler is intended to be started before the process calls any CUDA events so it can capture everything.  The profiler is shipped as a separate shared library in the spark-rapids-jni jar that can be loaded before the spark-rapids-jni shared native library that contains the cudf and spark-rapids-jni CUDA kernels.

The profiling data is serialized as a series of size-prefixed flatbuffers.  See profiler.fbs for the schema.  This also includes a converter tool, built in target/cmake-build/profiler/spark_rapids_profile_converter which can be used to convert the flatbuffer profiling data into various formats.  It currently supports converting to JSON (useful for performing custom analysis on the raw data) or the Nsight NVTXT format.  NVTXT can in turn be converted into an nsys-rep file that the Nsight Systems viewer can load via the ImportNvtxt tool supplied with Nsight Systems.


jlowe:
build

src/main/cpp/profiler/profiler_debug.cpp
```diff
@@ -0,0 +1,185 @@
...
+std::string activity_object_kind_to_string(CUpti_ActivityObjectKind kind)
+{
+  switch (kind) {
+    case CUPTI_ACTIVITY_OBJECT_PROCESS: return "PROCESS";
+    case CUPTI_ACTIVITY_OBJECT_THREAD: return "THREAD";
+    case CUPTI_ACTIVITY_OBJECT_DEVICE: return "DEVICE";
+    case CUPTI_ACTIVITY_OBJECT_CONTEXT: return "CONTEXT";
+    case CUPTI_ACTIVITY_OBJECT_STREAM: return "STREAM";
+    case CUPTI_ACTIVITY_OBJECT_UNKNOWN:
+    default: return "UNKNOWN";
```
revans2, line 47: nit: for all of the Unknown enums do we want to include the ID in the UNKNOWN string? That would at least point us in a direction of what went wrong.

src/main/cpp/profiler/ProfilerJni.cpp
```diff
@@ -0,0 +1,530 @@
...
+    State->writer_thread = std::thread(
+      writer_thread_process, get_jvm(env), writer, write_buffer_size, write_buffer_size);
+    auto rc = cuptiSubscribe(&State->subscriber_handle, callback_handler, nullptr);
+    check_cupti(rc, "Error initializing CUPTI");
+    rc = cuptiEnableCallback(1,
+                             State->subscriber_handle,
+                             CUPTI_CB_DOMAIN_RUNTIME_API,
+                             CUPTI_RUNTIME_TRACE_CBID_cudaDeviceReset_v3020);
+    if (flush_period_msec > 0) {
+      std::cerr << "PROFILER: Flushing activity records every " << flush_period_msec
```
revans2, line 450: nit: I'm not a huge fan of logging to stderr in JNI libraries. Might be good to change this to a logging framework at some point.

src/main/cpp/profiler/ProfilerJni.cpp
```diff
@@ -0,0 +1,530 @@
...
+                << " milliseconds" << std::endl;
+      Flush_period_msec    = static_cast<uint64_t>(flush_period_msec);
+      Last_flush_time_msec = timestamp_now();
+      CUpti_CallbackId const driver_launch_callback_ids[] = {
```
revans2, line 454: Might be nice to have some comments here about what this is doing.

src/main/cpp/profiler/ProfilerJni.cpp
```diff
@@ -0,0 +1,530 @@
...
+  }
+  vm->DetachCurrentThread();
+}
+
+// Enable/disable capture of CUPTI activity events
+void update_activity_enable(bool enable)
+{
+  if (enable) {
+    check_cupti(cuptiActivityEnable(CUPTI_ACTIVITY_KIND_DEVICE), "Error enabling device activity");
+    // check_cupti(cuptiActivityEnable(CUPTI_ACTIVITY_KIND_CONTEXT), "Error enabling context
```
revans2, line 388: Could you remove the commented out code, or add a comment explaining what is happening here.

src/main/cpp/profiler/ProfilerJni.cpp
```diff
@@ -0,0 +1,530 @@
...
+    }
+  }
+  vm->DetachCurrentThread();
+}
+
+// Enable/disable capture of CUPTI activity events
+void update_activity_enable(bool enable)
+{
+  if (enable) {
+    check_cupti(cuptiActivityEnable(CUPTI_ACTIVITY_KIND_DEVICE), "Error enabling device activity");
```
revans2, line 387: Could we have a static list of configs that we can walk both for enable and disable.

src/main/cpp/profiler/ProfilerJni.cpp
```diff
@@ -0,0 +1,530 @@
...
+      has_cupti_callback_errored(false),
+      is_shutdown(false)
+  {
+  }
+};
+
+// Global variables
+subscriber_state* State = nullptr;
+uint64_t Flush_period_msec;
+uint64_t Last_flush_time_msec;
```
revans2, line 209: Do we need some kind of locking for this? if so it would be good to have a comment explaining that here. The flush period should be mostly static once it is set.

src/main/cpp/profiler/ProfilerJni.cpp
```diff
@@ -0,0 +1,530 @@
...
+{
+  if (rc != CUPTI_SUCCESS) { throw std::runtime_error(msg + ": " + get_cupti_error(rc)); }
+}
+
+uint64_t timestamp_now()
+{
+  timespec info;
+  if (clock_gettime(CLOCK_MONOTONIC_RAW, &info) != 0) {
+    std::cerr << "PROFILER: Unable to determine current time, aborting!" << std::endl;
+    abort();
```
revans2, line 250: nit: I'd rather see an exception if possible, but I don't know if it is possible.

src/main/cpp/profiler/ProfilerJni.cpp
```diff
@@ -0,0 +1,530 @@
...
+  }
+  return info.tv_sec * 1e3 + info.tv_nsec / 1e6;
+}
+
+void on_driver_launch_exit()
```
revans2, line 255: Are we guaranteed that this is only every called from a single thread? I am just concerned about accessing global variables with no locking mechanism or explanation about how it is done.

src/main/cpp/profiler/spark_rapids_profile_converter.cpp
```diff
@@ -0,0 +1,756 @@
...
+  uint32_t tid;
+
+  bool operator==(thread_id const& o) const { return pid == o.pid && tid == o.tid; }
+};
+
+template <>
+struct std::hash<thread_id> {
+  std::size_t operator()(thread_id const& t) const
+  {
+    // TODO: use a real hash
```
revans2, line 77: Do we need this? Should there be a follow on issue?

revans2 (approved):
Really just nits, but be aware I do not know CMake so if there are scary thing in there I probably missed them. I also am not an expert on the libraries or formats that you are using, so I am not sure 100% if they are all correct, but things appeared to line up properly.

src/main/cpp/profiler/ProfilerJni.cpp
jlowe: We don't need any locking for Flush_period_msec.  It's set once on startup before any callbacks try to examine it.  The more interesting one is Last_flush_time_msec which theoretically could be set by multiple threads.  This probably ends up being set atomically in practice, but probably prudent to make it an atomic to guarantee it.  We shouldn't care which thread "wins" a race for this, as they're all trying to set it to the current time.  Worst-case, we end up flushing ever so slightly more often than we should because a slightly stale timestamp overwrites a newer one.

src/main/cpp/profiler/ProfilerJni.cpp
jlowe: The problem with throwing an exception here is that it could be theoretically in any thread, not necessarily a Java-registered thread, and thus the exception will likely end up killing the process.  Given this is just used for periodic flush, I could make this fail in a safer but degraded way, where it simply returns the last flush timestamp and thus effectively disables flushing.

src/main/cpp/profiler/spark_rapids_profile_converter.cpp
jlowe: I suspect we don't really need to do this in practice, as the hash is probably "good enough" for the limited data we're using with it.

src/main/cpp/profiler/ProfilerJni.cpp
```diff
@@ -0,0 +1,527 @@
...
+                                             CUPTI_ACTIVITY_KIND_MARKER,
+                                             CUPTI_ACTIVITY_KIND_CONCURRENT_KERNEL,
+                                             CUPTI_ACTIVITY_KIND_OVERHEAD};
+  if (enable) {
+    for (CUpti_ActivityKind const id : activity_ids) {
+      check_cupti(cuptiActivityEnable(id), "Error enabling device activity");
+    }
+  } else {
+    for (CUpti_ActivityKind const id : activity_ids) {
+      check_cupti(cuptiActivityEnable(id), "Error disabling device activity");
```
revans2: I think this is a copy/paste bug and needs to be cuptiActivityDisable

revans2 (dismissed):

jlowe:
build

jlowe:
build

PR #2079, opened 2024-05-28
case when improvement: avoid copy_if_else

res-life:
Case when improvement: avoid copy_if_else

For the following `case when`:
```
select
  case
    when bool_1_expr then "value_1"
    when bool_2_expr then "value_2"
    when bool_3_expr then "value_3"
    ......
    else "value_else"
  end
from tab
```
### Current logic, [link](https://github.com/NVIDIA/spark-rapids/blob/branch-24.06/sql-plugin/src/main/scala/com/nvidia/spark/rapids/conditionalExpressions.scala#L362-L370):
Iteratively invoke `copy_if_else` to merge the tail 2 branches.
This incurr lots of memory operations, here intruduced 3 `copy_if_else`
```
      val elseRet = elseValue
        .map(_.columnarEvalAny(batch))
        .getOrElse(GpuScalar(null, branches.last._2.dataType))
      val any = branches.foldRight[Any](elseRet) {
        case ((predicateExpr, trueExpr), falseRet) =>
          computeIfElse(batch, predicateExpr, trueExpr, falseRet)
      }
```

### Improvement:
First evaluate all the `when` exprs and get bool columns.
Then select the first true in the bool columns and return the column index
Then select salars according to the select column.

implement 2 kernels to handle:

```
/**
 * select the first column index with true value.
 * e.g.:
 * column 0 in table: true,  false, false
 * column 1 in table: false, true,  false
 * column 2 in table: false, false, true
 * 
 * return column: 0, 1, 2
*/
std::unique_ptr<cudf::column> select_first_true_index(
  cudf::table_view const& when_bool_columns,
  rmm::cuda_stream_view stream        = cudf::get_default_stream(),

/**
 * Select strings int scalar column according to index column
 * scalar column: s0, s1, s2
 * index  column: 0,  1,  2,  2,  1,  0,  3
 * output column: s0, s1, s2, s2, s1, s0, null
 * 
*/
std::unique_ptr<cudf::column> select_from_index(
  cudf::strings_column_view const& then_and_else_scalar_column,
  cudf::column_view const& select_index_column,
  rmm::cuda_stream_view stream        = cudf::get_default_stream(),
```

### DIFF
--- a/src/main/cpp/src/CaseWhenJni.cpp
+++ b/src/main/cpp/src/CaseWhenJni.cpp
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2024, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "case_when.hpp"
+#include "cudf_jni_apis.hpp"
+
+#include <vector>
+
+extern "C" {
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CaseWhen_selectFirstTrueIndex(
+  JNIEnv* env, jclass, jlongArray bool_cols)
+{
+  JNI_NULL_CHECK(env, bool_cols, "array of column handles is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
+    cudf::jni::native_jpointerArray<cudf::column_view> n_cudf_bool_columns(env, bool_cols);
+    auto bool_column_views = n_cudf_bool_columns.get_dereferenced();
+    return cudf::jni::release_as_jlong(
+      spark_rapids_jni::select_first_true_index(cudf::table_view(bool_column_views)));
+  }
+  CATCH_STD(env, 0);
+}
+
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CaseWhen_selectFromIndex(JNIEnv* env,
+                                                                                  jclass,
+                                                                                  jlong scalar_cols,
+                                                                                  jlong index_col)
+{
+  JNI_NULL_CHECK(env, scalar_cols, "Column handles is null", 0);
+  JNI_NULL_CHECK(env, index_col, "Column handles is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
+    auto const scalar_column_view      = reinterpret_cast<cudf::column_view const*>(scalar_cols);
+    auto const scalar_strings_col_view = cudf::strings_column_view{*scalar_column_view};
+    auto const index_column_view       = reinterpret_cast<cudf::column_view const*>(index_col);
+    return cudf::jni::release_as_jlong(
+      spark_rapids_jni::select_from_index(scalar_strings_col_view, *index_column_view));
+  }
+  CATCH_STD(env, 0);
+}
+}
--- a/src/main/cpp/src/case_when.cu
+++ b/src/main/cpp/src/case_when.cu
@@ -1,0 +1,183 @@
+/*
+ * Copyright (c) 2024, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "case_when.hpp"
+
+#include <cudf/column/column_factories.hpp>
+#include <cudf/strings/detail/strings_column_factories.cuh>
+#include <cudf/table/table_device_view.cuh>
+#include <cudf/types.hpp>
+
+#include <thrust/transform.h>
+
+// #include <cudf/detail/iterator.cuh>
+// #include <cudf/detail/utilities/vector_factories.hpp>
+
+// #include <rmm/device_uvector.hpp>
+
+namespace spark_rapids_jni {
+namespace detail {
+
+/**
+ * Find the first column index with true value from bool columns for a specified row
+ */
+struct select_first_true_fn {
+  cudf::table_device_view const d_table;
+
+  /**
+   * bool columns number should be the size of case when branches.
+   * For case when that does not have else branch, max returned index is column number which means
+   * use default null path.
+   *
+   * e.g.:
+   * CASE WHEN 'a' THEN 'A' END , max index is 1 which means use null
+   * CASE WHEN 'a' THEN 'A' ELSE '_' END, max index is also 1 which means use else value '_'
+   */
+  __device__ cudf::size_type operator()(std::size_t row_idx)
+  {
+    auto col_num                     = d_table.num_columns();
+    bool found_true                  = false;
+    cudf::size_type first_true_index = col_num;
+    for (auto col_idx = 0; !found_true && col_idx < col_num; col_idx++) {
+      if (d_table.column(col_idx).element<bool>(row_idx)) {
+        found_true       = true;
+        first_true_index = col_idx;
+      }
+    }
+    return first_true_index;
+  }
+};
+
+std::unique_ptr<cudf::column> select_first_true_index(cudf::table_view const& when_bool_columns,
+                                                      rmm::cuda_stream_view stream,
+                                                      rmm::mr::device_memory_resource* mr)
+{
+  // checks
+  auto const num_columns = when_bool_columns.num_columns();
+  CUDF_EXPECTS(num_columns > 0, "At least one column must be specified");
+  auto const row_count = when_bool_columns.num_rows();
+  if (row_count == 0)  // empty begets empty
+    return cudf::make_empty_column(cudf::type_id::INT32);
+
+  auto ret = cudf::make_numeric_column(
+    cudf::data_type{cudf::type_id::INT32}, row_count, cudf::mask_state::ALL_VALID, stream, mr);
+
+  auto d_table = cudf::table_device_view::create(when_bool_columns, stream);
+
+  // select first true index
+  thrust::transform(rmm::exec_policy(stream),
+                    thrust::make_counting_iterator<cudf::size_type>(0),
+                    thrust::make_counting_iterator<cudf::size_type>(row_count),
+                    ret->mutable_view().begin<cudf::size_type>(),
+                    select_first_true_fn{*d_table});
+  return ret;
+}
+
+/**
+ * Exedute SQL `case when` semantic.
+ * If there are multiple branches and each branch uses scalar to generator value,
+ * then it's fast to use this function because it does not generate temp string columns.
+ * Regular logic in RAPIDs Acceselorator invokes multiple cudf::copy_if_else to generate
+ * multiple temp string columns which are time-consuming.
+ *
+ * E.g.:
+ *   SQL is:
+ *     select
+ *        case
+ *          when bool_1_expr then "value_1"
+ *          when bool_2_expr then "value_2"
+ *          when bool_3_expr then "value_3"
+ *          else "value_else"
+ *        end
+ *      from tab
+ *
+ *   E.g.: the input data is:
+ *     bool column 1: [true,  false, false, false]  // the result of
+ bool_1_expr.executeColumnar
+ *     bool column 2: [false, true,  false, flase]  // the result of
+ bool_2_expr.executeColumnar
+ *     bool column 3: [false, false, true,  flase]  // the result of
+ bool_3_expr.executeColumnar
+ *     scalars in SQL are "value_1", "value_1", "value_1", "value_else"
+ *
+ *   Output will be:
+ *     ["value_1", "value_2", "value_3", "value_else"]
+ *
+ * Note: the number of bool columns is greater than the number of scalars by 1.
+ * size(when_bool_columns) == (row of then_and_else_scalars) + 1
+ *
+ */
+
+std::unique_ptr<cudf::column> select_from_index(
+  cudf::strings_column_view const& then_and_else_scalar_column,
+  cudf::column_view const& select_index_column,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr)
+{
+  cudf::size_type num_of_rows   = select_index_column.size();
+  cudf::size_type num_of_scalar = then_and_else_scalar_column.size();
+
+  // create device views
+  auto d_scalars =
+    *(cudf::column_device_view::create(then_and_else_scalar_column.parent(), stream));
+  auto d_select_index = *(cudf::column_device_view::create(select_index_column, stream));
+
+  // Select <str_ptr, size> pairs from multiple scalars according to select index
+  using str_view = thrust::pair<char const*, cudf::size_type>;
+  rmm::device_uvector<str_view> indices(num_of_rows, stream);
+  thrust::transform(rmm::exec_policy(stream),
+                    thrust::make_counting_iterator<cudf::size_type>(0),
+                    thrust::make_counting_iterator<cudf::size_type>(num_of_rows),
+                    indices.begin(),
+                    [d_scalars, num_of_scalar, d_select_index] __device__(cudf::size_type row_idx) {
+                      // select scalar according to index
+                      cudf::size_type scalar_idx = d_select_index.element<cudf::size_type>(row_idx);
+
+                      // NOTE a case: scalar_idx == num_of_scalar, e.g.:
+                      // CASE WHEN 'a' THEN 'A' END , max index is 1 which means use null
+                      // CASE WHEN 'a' THEN 'A' ELSE '_' END, max index is also 1 which means use
+                      // else value '_'
+                      if (scalar_idx < num_of_scalar) {
+                        auto const d_str = d_scalars.element<cudf::string_view>(scalar_idx);
+                        return str_view{d_str.data(), d_str.size_bytes()};
+                      } else {
+                        return str_view{nullptr, 0};
+                      }
+                    });
+
+  // create final string column from string index pairs
+  return cudf::strings::detail::make_strings_column(indices.begin(), indices.end(), stream, mr);
+}
+
+}  // namespace detail
+
+std::unique_ptr<cudf::column> select_first_true_index(cudf::table_view const& when_bool_columns,
+                                                      rmm::cuda_stream_view stream,
+                                                      rmm::mr::device_memory_resource* mr)
+{
+  return detail::select_first_true_index(when_bool_columns, stream, mr);
+}
+
+std::unique_ptr<cudf::column> select_from_index(
+  cudf::strings_column_view const& then_and_else_scalars_column,
+  cudf::column_view const& select_index_column,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr)
+{
+  return detail::select_from_index(then_and_else_scalars_column, select_index_column, stream, mr);
+}
+
+}  // namespace spark_rapids_jni
--- a/src/main/cpp/src/case_when.hpp
+++ b/src/main/cpp/src/case_when.hpp
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2024, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cudf/column/column.hpp>
+#include <cudf/column/column_view.hpp>
+#include <cudf/strings/strings_column_view.hpp>
+#include <cudf/table/table_view.hpp>
+#include <cudf/types.hpp>
+
+#include <rmm/mr/device/per_device_resource.hpp>
+#include <rmm/resource_ref.hpp>
+
+#include <map>
+#include <memory>
+#include <string>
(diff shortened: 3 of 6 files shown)


revans2:
My biggest concern is with side effects. And to be clear that is not an issue with this code. It is a more general problem with case/when I want you to be aware of. Case/when and if/else in Spark are lazy. This means that if an expression in the when part has a side effect, like can throw an exception, then we cannot evaluate it for any row that would cause an exception to be triggered. This appears to be specific to scalars in the case/when so it should be fine.

winningsix:
@res-life created an issue for this. https://github.com/NVIDIA/spark-rapids-jni/issues/2084

src/test/java/com/nvidia/spark/rapids/jni/CaseWhenTest.java
```diff
@@ -42,6 +44,47 @@ void selectIndexTest() {
     }
   }
 
+  public static ColumnVector fromBooleansWithNulls(Boolean... values) {
```
revans2, line 47: Use https://github.com/rapidsai/cudf/blob/ff981a4048a389b0e2582e94d3397a83096d16c9/java/src/main/java/ai/rapids/cudf/ColumnVector.java#L1457 instead

res-life: Done.

res-life:
build

ttnghia:
Please provide benchmarks to show off better how much benefit this can provide?

res-life:
> Please provide benchmarks to show off better how much benefit this can provide?

Will do it.

res-life:
For end to end perf result, refer to:
https://github.com/NVIDIA/spark-rapids/pull/10951#issuecomment-2144673907

@ttnghia  is it needed to add benchmark tests? Above is end to end result.

src/main/cpp/src/CaseWhenJni.cpp
```diff
@@ -0,0 +1,55 @@
...
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CaseWhen_selectFromIndex(JNIEnv* env,
+                                                                                  jclass,
+                                                                                  jlong scalar_cols,
+                                                                                  jlong index_col)
+{
+  JNI_NULL_CHECK(env, scalar_cols, "Column handles is null", 0);
+  JNI_NULL_CHECK(env, index_col, "Column handles is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
+    auto const scalar_column_view      = reinterpret_cast<cudf::column_view const*>(scalar_cols);
```
ttnghia, line 47: Since the scalars are assembled into rows of a column, they are no longer scalars. Probably a better name should be `options_col`.

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+   * CASE WHEN 'a' THEN 'A' END
+   *   The number of bool columns is 1
+   *   The number of scalars is 1
+   *   Max index is 1 which means using NULL(all when exprs are false).
+   * CASE WHEN 'a' THEN 'A' ELSE '_' END
+   *   The number of bool columns is 1
+   *   The number of scalars is 2
+   *   Max index is also 1 which means using else value '_'
+   */
+  __device__ cudf::size_type operator()(std::size_t row_idx)
```
ttnghia, line 51:
```suggestion
  __device__ cudf::size_type operator()(std::size_t row_idx) const
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+std::unique_ptr<cudf::column> select_first_true_index(cudf::table_view const& when_bool_columns,
+                                                      rmm::cuda_stream_view stream,
+                                                      rmm::mr::device_memory_resource* mr)
+{
+  // checks
+  auto const num_columns = when_bool_columns.num_columns();
+  CUDF_EXPECTS(num_columns > 0, "At least one column must be specified");
+  auto const row_count = when_bool_columns.num_rows();
+  if (row_count == 0)  // empty begets empty
+    return cudf::make_empty_column(cudf::type_id::INT32);
```
ttnghia, line 77:
```suggestion
  if (row_count == 0)  { // empty begets empty
    return cudf::make_empty_column(cudf::type_id::INT32);
  }
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+
+  // make output column
+  auto ret = cudf::make_numeric_column(
+    cudf::data_type{cudf::type_id::INT32}, row_count, cudf::mask_state::ALL_VALID, stream, mr);
+
+  // select first true index
+  auto d_table = cudf::table_device_view::create(when_bool_columns, stream);
```
ttnghia, line 84:
```suggestion
  auto const d_table_ptr = cudf::table_device_view::create(when_bool_columns, stream);
```

src/main/cpp/src/case_when.hpp
```diff
@@ -0,0 +1,76 @@
...
+ *   scalar column: s0, s1, s2
+ *   index  column: 0,  1,  2,  2,  1,  0,  3
+ *   output column: s0, s1, s2, s2, s1, s0, NULL
+ *
+ */
+std::unique_ptr<cudf::column> select_from_index(
+  cudf::strings_column_view const& then_and_else_scalar_column,
+  cudf::column_view const& select_index_column,
+  rmm::cuda_stream_view stream        = cudf::get_default_stream(),
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
```
ttnghia, line 74: This is a very good optimization strategy and it should not be limited to string type. Indeed, we can apply this to any data type. Thus, the function signature should take the input as a regular `column_view`, not `string_column_view`. Supporting other data types is not required to be implemented in this PR but the function signature should be prepared for future changes.
```suggestion
 * Select strings in scalar column according to index column.
 * If index is out of bound, use NULL value
 * e.g.:
 *   scalar column: s0, s1, s2
 *   index  column: 0,  1,  2,  2,  1,  0,  3
 *   output column: s0, s1, s2, s2, s1, s0, NULL
 *
 */
std::unique_ptr<cudf::column> select_from_index(
  cudf::column_view const& case_when_options,
  cudf::column_view const& select_index_column,
  rmm::cuda_stream_view stream        = cudf::get_default_stream(),
  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+                    select_first_true_fn{*d_table});
+  return ret;
+}
+
+std::unique_ptr<cudf::column> select_from_index(
+  cudf::strings_column_view const& then_and_else_scalar_column,
+  cudf::column_view const& select_index_column,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr)
+{
```
ttnghia, line 98:
```suggestion
std::unique_ptr<cudf::column> select_from_index(
  cudf::column_view const& case_when_options,
  cudf::column_view const& select_index_column,
  rmm::cuda_stream_view stream,
  rmm::mr::device_memory_resource* mr)
{
  CUDF_EXPECTS(case_when_options.type().id() == cudf::type_id::STRING, "Currently only STRING type is supported in multiple options case when");
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+  cudf::size_type num_of_rows   = select_index_column.size();
+  cudf::size_type num_of_scalar = then_and_else_scalar_column.size();
+
+  // create device views
+  auto d_scalars =
+    *(cudf::column_device_view::create(then_and_else_scalar_column.parent(), stream));
```
ttnghia, line 104: Oh never use this again. This is dereference of a temporary object which is destroyed immediately.

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+}
+
+std::unique_ptr<cudf::column> select_from_index(
+  cudf::strings_column_view const& then_and_else_scalar_column,
+  cudf::column_view const& select_index_column,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr)
+{
+  cudf::size_type num_of_rows   = select_index_column.size();
+  cudf::size_type num_of_scalar = then_and_else_scalar_column.size();
```
ttnghia, line 100:
```suggestion
  cudf::size_type num_options = case_when_options.size();
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+    *(cudf::column_device_view::create(then_and_else_scalar_column.parent(), stream));
+  auto d_select_index = *(cudf::column_device_view::create(select_index_column, stream));
+
+  // Select <str_ptr, str_size> pairs from multiple scalars according to select index
+  using str_view = thrust::pair<char const*, cudf::size_type>;
+  rmm::device_uvector<str_view> indices(num_of_rows, stream);
+  thrust::transform(rmm::exec_policy(stream),
+                    thrust::make_counting_iterator<cudf::size_type>(0),
+                    thrust::make_counting_iterator<cudf::size_type>(num_of_rows),
+                    indices.begin(),
```
ttnghia, line 113:
```suggestion
  thrust::tabulate(rmm::exec_policy(stream),
                    indices.begin(),
                    indices.end(),
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+                    [d_scalars, num_of_scalar, d_select_index] __device__(cudf::size_type row_idx) {
+                      // select scalar according to index
+                      cudf::size_type scalar_idx = d_select_index.element<cudf::size_type>(row_idx);
```
ttnghia, line 116: As mentioned, "scalar" is probably not a good name. Instead, "options" seems to be better.

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+
+                      // return <str_ptr, str_size> pair
+                      if (scalar_idx < num_of_scalar && !d_scalars.is_null(scalar_idx)) {
+                        auto const d_str = d_scalars.element<cudf::string_view>(scalar_idx);
+                        return str_view{d_str.data(), d_str.size_bytes()};
```
ttnghia, line 121:
```suggestion
                        return d_options.element<cudf::string_view>(scalar_idx);
```

src/main/java/com/nvidia/spark/rapids/jni/CaseWhen.java
```diff
@@ -0,0 +1,105 @@
...
+ *          else "value_else"
+ *        end
+ *      from tab
+ *
+ * Execution steps:
+ *   Execute bool exprs to get bool columns, e.g., gets:
+ *     bool column 1: [true,  false, false, false]  // bool_1_expr result
+ *     bool column 2: [false, true,  false, flase]  // bool_2_expr result
+ *     bool column 3: [false, false, true,  flase]  // bool_3_expr result
+ *   Execute `selectFirstTrueIndex` to get the column index for the first true in bool columns.
```
ttnghia, line 43: Here we have 4 options instead of 3 (the `value_else` should be considered). As such, we can generate the 4th bool column containing all-true values. Doing so, we won't need to launch another kernel to set `value_else` after gathering values from the first 3 columns.

src/main/cpp/src/case_when.hpp
ttnghia: Oh wait. I just realize that this is just a gather. So we don't need this function at all. Just call `cudf::gather`, which already supports all data types.

res-life:
@ttnghia  Thanks a lot, I will fix.

src/main/cpp/src/CaseWhenJni.cpp
res-life: This `selectFromIndex` can be replaced by `gather`. Will remove `selectFromIndex`.

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,151 @@
...
+    cudf::size_type first_true_index = col_num;
+    for (auto col_idx = 0; !found_true && col_idx < col_num; col_idx++) {
+      auto const& col = d_table.column(col_idx);
+      if (!col.is_null(row_idx) && col.element<bool>(row_idx)) {
+        // Predicate is true and not null
+        found_true       = true;
+        first_true_index = col_idx;
+      }
+    }
+    return first_true_index;
```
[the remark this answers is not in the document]
res-life:
```
    auto const it = thrust::find_if(thrust::seq, d_table.begin(), d_table.end(), [row_idx]__device__(auto const& col) {
        return !col.is_null(row_idx) && col.element<bool>(row_idx);
    });
    return thrust::distance(d_table.begin(), it); // if not found, this will return `d_table.num_columns()`
```
Got this building error:
[exec] /home/chongg/code/spark-rapids-jni/src/main/cpp/src/case_when.cu(55): error: type name is not allowed.

And I saw thrust::find_if is complicated:
```
template<typename DerivedPolicy, typename InputIterator, typename Predicate>
_CCCL_HOST_DEVICE
InputIterator find_if(thrust::execution_policy<DerivedPolicy> &exec,
                      InputIterator first,
                      InputIterator last,
                      Predicate pred)
{
  typedef typename thrust::iterator_traits<InputIterator>::difference_type difference_type;
  typedef typename thrust::tuple<bool,difference_type> result_type;

  // empty sequence
  if(first == last) return last;

  const difference_type n = thrust::distance(first, last);

  // this implementation breaks up the sequence into separate intervals
  // in an attempt to early-out as soon as a value is found

  // TODO incorporate sizeof(InputType) into interval_threshold and round to multiple of 32
  const difference_type interval_threshold = 1 << 20;
  const difference_type interval_size = (thrust::min)(interval_threshold, n);

  // force transform_iterator output to bool
  typedef thrust::transform_iterator<Predicate, InputIterator, bool> XfrmIterator;
  typedef thrust::tuple<XfrmIterator, thrust::counting_iterator<difference_type> > IteratorTuple;
  typedef thrust::zip_iterator<IteratorTuple> ZipIterator;

  IteratorTuple iter_tuple = thrust::make_tuple(XfrmIterator(first, pred),
                                                thrust::counting_iterator<difference_type>(0));

  ZipIterator begin = thrust::make_zip_iterator(iter_tuple);
  ZipIterator end   = begin + n;

  for(ZipIterator interval_begin = begin; interval_begin < end; interval_begin += interval_size)
  {
    ZipIterator interval_end = interval_begin + interval_size;
    if(end < interval_end)
    {
      interval_end = end;
    } // end if

    result_type result = thrust::reduce(exec,
                                        interval_begin, interval_end,
                                        result_type(false,interval_end - begin),
                                        find_if_functor<result_type>());

    // see if we found something
    if(thrust::get<0>(result))
    {
      return first + thrust::get<1>(result);
    }
  }

  //nothing was found if we reach here...
  return first + n;
}

Let's use current concise impl?

```

res-life:
build

src/main/cpp/src/case_when.cu
res-life: done

src/main/cpp/src/case_when.cu
res-life: done.

src/main/cpp/src/case_when.cu
res-life: done.

src/main/cpp/src/case_when.cu
res-life: It is no longer applicable. Because `select_from_index` was removed.

src/main/cpp/src/case_when.cu
res-life: It is no longer applicable. Because `select_from_index` was removed.

src/main/cpp/src/case_when.cu
res-life: It is no longer applicable. Because `select_from_index` was removed.

src/main/cpp/src/case_when.cu
res-life: It is no longer applicable. Because `select_from_index` was removed.

src/main/cpp/src/case_when.cu
res-life: It is no longer applicable. Because `select_from_index` was removed.

res-life:
@ttnghia  Please review again.

res-life:
Spark-Rapids corresponding change: 
https://github.com/NVIDIA/spark-rapids/pull/10951/commits/7c43e698511b556311165c9e022a8347bd4308af

```
// removed
val finalRet = CaseWhen.selectFromIndex(scalarCol, firstTrueIndex)
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,99 @@
...
+   * CASE WHEN 'a' THEN 'A' ELSE '_' END
+   *   The number of bool columns is 1
+   *   The number of scalars is 2
+   *   Max index is also 1 which means using else value '_'
+   */
+  __device__ cudf::size_type operator()(std::size_t row_idx) const
+  {
+    auto col_num = d_table.num_columns();
+    for (auto col_idx = 0; col_idx < col_num; col_idx++) {
+      auto const& col = d_table.column(col_idx);
```
ttnghia, line 55:
```suggestion
    for (auto const& col : d_table) {
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,99 @@
...
+  // checks
+  auto const num_columns = when_bool_columns.num_columns();
+  CUDF_EXPECTS(num_columns > 0, "At least one column must be specified");
+  auto const row_count = when_bool_columns.num_rows();
+  if (row_count == 0) {  // empty begets empty
+    return cudf::make_empty_column(cudf::type_id::INT32);
+  }
+  // make output column
+  auto ret = cudf::make_numeric_column(
+    cudf::data_type{cudf::type_id::INT32}, row_count, cudf::mask_state::ALL_VALID, stream, mr);
```
ttnghia, line 78:
```suggestion
    cudf::data_type{cudf::type_id::INT32}, row_count, cudf::mask_state::UNALLOCATED, stream, mr);
```

src/main/cpp/src/case_when.hpp
```diff
@@ -0,0 +1,60 @@
...
+ *   4th row is: false, false, false; do not find true, set index to the end index 3
+ *
+ *   output column: 0, 1, 2, 3
+ *   In the `case when` context, here 3 index means using NULL value.
+ *
+ */
+std::unique_ptr<cudf::column> select_first_true_index(
+  cudf::table_view const& when_bool_columns,
+  rmm::cuda_stream_view stream        = cudf::get_default_stream(),
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
```
ttnghia, line 58: We no longer use `rmm::mr::device_memory_resource*`. Instead, use `rmm::device_async_resource_ref`.
```suggestion
  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource());
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,99 @@
...
+                    ret->mutable_view().begin<cudf::size_type>(),
+                    select_first_true_fn{*d_table_ptr});
+  return ret;
+}
+
+}  // namespace detail
+
+std::unique_ptr<cudf::column> select_first_true_index(cudf::table_view const& when_bool_columns,
+                                                      rmm::cuda_stream_view stream,
+                                                      rmm::mr::device_memory_resource* mr)
```
ttnghia, line 94:
```suggestion
                                                      rmm::device_async_resource_ref mr)
```

res-life:
build

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,99 @@
...
+        return col_idx;
+      }
+    }
+    return col_num;
+  }
+};
+
+std::unique_ptr<cudf::column> select_first_true_index(cudf::table_view const& when_bool_columns,
+                                                      rmm::cuda_stream_view stream,
+                                                      rmm::mr::device_memory_resource* mr)
```
ttnghia, line 67:
```suggestion
                                                      rmm::device_async_resource_ref mr)
```

ttnghia:
build

ttnghia (approved):

src/main/cpp/src/CaseWhenJni.cpp
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "case_when.hpp"
+#include "cudf_jni_apis.hpp"
+
+#include <vector>
```
ttnghia, line 20:
```suggestion
```

src/main/cpp/src/case_when.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <cudf/types.hpp>
+
+#include <rmm/mr/device/per_device_resource.hpp>
+#include <rmm/resource_ref.hpp>
+
+#include <map>
+#include <memory>
+#include <string>
+#include <variant>
+#include <vector>
```
ttnghia, line 32:
```suggestion
#include <cudf/table/table_view.hpp>

#include <rmm/mr/device/per_device_resource.hpp>
#include <rmm/resource_ref.hpp>

#include <memory>
```

ttnghia (commented):
Sorry there are redundant headers that need to be removed.

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,99 @@
...
+#include <cudf/table/table_device_view.cuh>
+#include <cudf/types.hpp>
+
+#include <thrust/transform.h>
+
+namespace spark_rapids_jni {
+namespace detail {
+
+/**
+ * Select the column index for the first true in bool columns for the specified row
```
ttnghia, line 30:
```suggestion
 * @brief Select the column index for the first true in bool columns for the specified row
```

src/main/cpp/src/case_when.cu
```diff
@@ -0,0 +1,99 @@
...
+ */
+struct select_first_true_fn {
```
ttnghia, line 32: Nit: We should wrap anything that is locally used in a source file into an anonymous namespace to avoid name clashing in the future with other source files.

res-life:
build

src/main/cpp/src/case_when.cu
res-life: Done.

src/main/cpp/src/case_when.hpp
res-life: Done.

src/main/cpp/src/case_when.cu
res-life: Done.

src/main/cpp/src/CaseWhenJni.cpp
res-life: Done.

src/main/cpp/src/case_when.hpp
res-life: Done.

src/main/cpp/src/case_when.cu
res-life: Done.
Added anonymous namespace

ttnghia (approved):

revans2 (commented):
The code looks good, but what do the performance numbers look like?

wjxiz1992:
The perf numbers
https://github.com/NVIDIA/spark-rapids/pull/10951#issuecomment-2144673907

res-life:
Yes, please refer to the above link.

PR #2105, opened 2024-06-03
Profiler: disable collecting async allocations by default

jlowe:
Disables collecting CUDA async allocation and free events by default.  There are often a large number of these events in the traces and they often are not very interesting in a profile.  Added a config to control whether they are collected, and refactored the profiler configuration to use a builder interface.


jlowe:
build

revans2 (approved):

Issue #2124, opened 2024-06-06
[FEA] Improve `get_json_object` by calling the main kernel only once


ttnghia:
Currently, we have to call the `get_json_object_kernel` twice: one for computing the output size and one for writing the output, and each of these calls is very expensive.

However, the second call does a very similar job to the first call. Essentially, writing the output in the second call is basically just multiple string copies. As such, we can eliminate the second call by storing the copying location right in the first call when computing the output size. By doing so, we can replace the second call with a simple kernel to do string copy using these copying locations.

ttnghia:
An alternative idea for this is to pre-allocate a temporary buffer column for the output without knowing the output string sizes. The temp buffer column can be allocated to have the same size as the input, thus it can store the output string without having size overflow. Then, while calling to the main kernel, we write out both the output strings along with their sizes. A gathering step is then executed to extract the final output strings column.

PR #2129, opened 2024-06-07
Optimize `get_json_object` by calling the main kernel only once

ttnghia:
This PR optimizes the host code for `get_json_object` function, improving performance by calling the main kernel only once.

Currently, the main kernel is called twice, one to compute the output string sizes and another one is to write the output strings. The idea of this work is to pre-allocate a temporary char column to write the output without knowing the output string sizes. Then, while calling to the main kernel, we write out both the output strings along with their sizes, using just one kernel call. A (relatively cheap) gathering step is then executed to extract the final output strings column.

Closes https://github.com/NVIDIA/spark-rapids-jni/issues/2124.


src/main/cpp/src/get_json_object.cu
```diff
@@ -953,64 +918,99 @@ std::unique_ptr<cudf::column> get_json_object(
...
     instructions, all_names_scalar, stream, rmm::mr::get_current_device_resource());
+  auto const d_input_ptr = cudf::column_device_view::create(input.parent(), stream);
+  auto const in_offsets  = cudf::detail::offsetalator_factory::make_input_iterator(input.offsets());
+
+  // A buffer to store the output strings without knowing their sizes.
+  // Since we do not know their sizes, we need to allocate the buffer a bit larger than the input
+  // size so that we will not write output strings into an out-of-bound position.
+  // Checking out-of-bound needs to be performed in the main kernel to make sure we will not have
+  // data corruption.
+  constexpr auto padding_ratio = 1.01;
```
thirtiseven, line 939: Question: How do we get this ratio? I think we have some case that will make the output larger than `input_size*1.01`, like number normalization, adding quotes under some styles, and escaping some chars.

ttnghia: This is the total size, not one row's size. If we have 1M rows, this will be the size of approx. 1M and 10'000 rows. In other word, we have around 10'000 more rows to avoid invalid memory access issue. We don't care if data of row `n` is written cross its boundary to row `n+1` which causes corruption, since we will discard the entire buffer if such overflow is detected. The extra 10'000 rows at the end is just to make sure data of the last row will not be written into out-of-bound of the entire buffer.

ttnghia: The extra `10'000` rows seem to be too much. We can turn up this padding ratio to be a dynamic value. For example: the padding should always be the average size of 100 rows. That should be large enough to avoid spilling of the last row.

ttnghia:
build

ttnghia:
build

src/main/cpp/src/get_json_object.cu
res-life: One extreme case:
```
scala> val d = Seq("['\n']", "['\n\n\n\n\n\n\n\n\n\n']")
scala> val df = d.toDF
scala> df.createOrReplaceTempView("tab")
scala> spark.sql("select length(value) from tab").show()
+-------------+
|length(value)|
+-------------+
|             5|
|           17|
+-------------+

scala> spark.sql("select length(get_json_object(value, '$')) from tab").show()
+---------------------------------+
|length(get_json_object(value, $))|
+---------------------------------+
|                                 6|
|                               30|
+---------------------------------+
```
This case will cause invalid memory access:
round up 2 * 1.01 = 3, avg size is 11, total allocated size is 33, but total write size will be 35 (5 + 30).
It causes overlap writing as expected(is not an issue), but the tailing writing causes invalid memory access.

One option:
Add a method in `json_generator` like:
```
write_byte(char c) {

  if (curr_size >= max_allowed_size) {
     // write nothing
     curr_size++;
  } else {
     *curr_pos = c;
     curr_size++;
  }
}
```

res-life:
It would be nice to a some unit test:
- Causes the 2nd kernel invoking.

ttnghia:
> It would be nice to add a unit test:
> 
> * Causes the 2nd kernel invoking.

Do you have any suggestion for the input? For such tests, the output length need to be much larger than the input.

src/main/cpp/src/get_json_object.cu
ttnghia: In my latest code, I pad the buffer by `100 * average_row_size`. In this case, we have `avg_row_size = 22/2=11`, so buffer size will be `22 + 100*11`, no overflow.

ttnghia: I'll add that code and check the benchmark to see how it will affect the overall performance.

ttnghia:
build

ttnghia:
build

src/main/cpp/src/get_json_object.cu
ttnghia: I updated the padding, which is now computed using max row size. This way can avoid invalid memory access when we have extreme cases such as the column has all null rows except one very long row.

ttnghia:
build

res-life (approved):

thirtiseven (approved):

PR #2153, opened 2024-06-17
Make run-in-docker usable in other repos

gerashegalov:
- Introduces WORKDIR to change the default local path available to Docker to point to another repo

This PR makes it possible to write easy repro steps for cudf or any other path to be executed in the spark-rapids-jni Docker container

### DIFF
--- a/build/run-in-docker
+++ b/build/run-in-docker
@@ -23,6 +23,7 @@ set -e
 # Base paths relative to this script's location
 SCRIPTDIR=$(cd $(dirname $0); pwd)
 REPODIR=$SCRIPTDIR/..
+WORKDIR=${WORKDIR:-$REPODIR}
 
 CUDA_VERSION=${CUDA_VERSION:-11.8.0}
 DOCKER_CMD=${DOCKER_CMD:-docker}
--- a/build/run-in-docker
+++ b/build/run-in-docker
@@ -18,13 +18,11 @@
 
 # Run a command in a Docker container with devtoolset
 
-set -ex
+set -e
 
-SCRIPTPATH=$(readlink -f "$0")
 # Base paths relative to this script's location
-SCRIPTDIR=$(cd $(dirname $SCRIPTPATH); pwd)
+SCRIPTDIR=$(cd $(dirname $0); pwd)
 REPODIR=$SCRIPTDIR/..
-WORKDIR=${WORKDIR:-$REPODIR}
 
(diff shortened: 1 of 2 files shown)


build/run-in-docker
```diff
@@ -18,11 +18,13 @@
 
 # Run a command in a Docker container with devtoolset
 
-set -e
+set -ex
```
jlowe, line 21: We don't want this by default, do we?  run-in-docker is already verbose enough with the Docker build output.

build/run-in-docker
```diff
@@ -55,18 +57,24 @@ if (( $# == 0 )); then
   DOCKER_OPTS="${DOCKER_OPTS} -it"
   RUN_CMD="/bin/bash"
 else
-  RUN_CMD="${@@Q}"
+  THIS_FILE=$(basename $SCRIPTPATH)
+  ARG0_CMD=$(basename "$0")
```
jlowe, line 61: Not sure I'm a fan of this hack, as I could easily see it causing as many problems as it solves once a user forgets they've setup this aliasing.  But as-is, it's inconsistent.  If someone wants to run an argless command this way, it doesn't work because the logic is missing in the `$# == 0` case.

It's also a bit odd that the headline implies this is just about allowing a variable to be overridden, but the majority of the change in this PR are actually for this unrelated change.

gerashegalov: Thanks for pointing out the bug with  `$# == 0`.  I do not really need the symlink based dispatch for this PR

jlowe (approved):

build/run-in-docker
```diff
@@ -63,10 +63,10 @@ $DOCKER_CMD run $DOCKER_GPU_OPTS $DOCKER_RUN_EXTRA_ARGS -u $(id -u):$(id -g) --r
   -v "/etc/passwd:/etc/passwd:ro" \
   -v "/etc/shadow:/etc/shadow:ro" \
   -v "/etc/sudoers.d:/etc/sudoers.d:ro" \
-  -v "$REPODIR:$REPODIR:rw" \
+  -v "$WORKDIR:$WORKDIR:rw" \
```
jlowe, line 66: Oops missed that WORKDIR is not set in this script by default to REPODIR like it was in the previous revision.  Does not work if WORKDIR is not set at all.

jlowe (changes requested):

jlowe (approved):

gerashegalov:
build

PR #2165, opened 2024-06-24
Restore static CUDA runtime linking

### LINKED ISSUE #2162: [BUG] libcudf.so being built with dynamic CUDA runtime dependency
The nightly builds are producing a jar that has a libcudf.so with a dynamic CUDA runtime dependency.  The CUDA runtime dependency is supposed to be static, and this can cause issues with environments that don't provide the CUDA 11 runtime library.
Note the missing `libcudart.so.11.0` dependency on libcudf.so below:
```bash
$ jar xvf ~/.m2/repository/com/nvidia/spark-rapids-jni/24.08.0-SNAPSHOT/spark-rapids-jni-24.08.0-SNAPSHOT-cuda11.jar amd64/Linux/libcudf.so
 inflated: amd64/Linux/libcudf.so
$ ldd amd64/Linux/libcudf.so
	linux-vdso.so.1 (0x00007ffded5e8000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x0000737afc368000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x0000737afc34c000)
	libnvcomp.so => /home/jlowe/src/spark-rapids/amd64/Linux/libnvcomp.so (0x0000737ad2c00000)
	libnvcomp_gdeflate.so => /home/jlowe/src/spark-rapids/amd64/Linux/libnvcomp_gdeflate.so (0x0000737ad1c00000)
	libnvcomp_bitcomp.so => /home/jlowe/src/spark-rapids/amd64/Linux/libnvcomp_bitcomp.so (0x0000737ad0800000)
	libcudart.so.11.0 => not found
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x0000737afc345000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x0000737afc340000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x0000737ad0400000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x0000737afc259000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x0000737afc237000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000737ad0000000)
	/lib64/ld-linux-x86-64.so.2 (0x0000737afc384000)
```

jlowe:
Fixes #2162.  This problem was caused by a [kvikio change](https://github.com/rapidsai/kvikio/pull/369/files#diff-1bba462ab050e89360fd88110a689e85ee037749cea091a1848ab574381d3795R98-R102) that requires downstream projects to set CUDA_STATIC_RUNTIME when they need a static runtime.

In the process of investigating this, I discovered there's an existing rapids_cuda_set_runtime that's supposed to be used for specifying the CUDA runtime settings.  I updated the CMakeLists.txt to use it, although it does not by itself fix the issue.  The CUDA_STATIC_RUNTIME variable is required, but we should update to the RAPIDS method for setting CUDA runtime parameters to match how libcudf is doing it in case the implementation details change in the future.

### DIFF
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -30,6 +30,7 @@ set(CUDF_CPP_BUILD_DIR
   "$ENV{CUDF_CPP_BUILD_DIR}"
   CACHE STRING "path to libcudf build root"
 )
+set(CUDA_STATIC_RUNTIME ON)
 
 include("${CUDF_DIR}/rapids_config.cmake")
 include(rapids-cmake)
@@ -267,6 +268,7 @@ target_link_libraries(
     ${PARQUET_LIB}
     ${THRIFT_LIB}
 )
+rapids_cuda_set_runtime(spark_rapids_jni USE_STATIC ON)
 set_target_properties(spark_rapids_jni PROPERTIES LINK_LANGUAGE "CXX")
 # For backwards-compatibility with the cudf Java bindings and RAPIDS accelerated UDFs,
 # all of the code is built into libcudf.so that is statically linked to the CUDA runtime library.
@@ -303,8 +305,8 @@ if(USE_GDS)
     -Wl,--no-whole-archive
     spark_rapids_jni
     ${cuFile_LIBRARIES}
-    CUDA::cudart_static
   )
+  rapids_cuda_set_runtime(cufilejni USE_STATIC ON)
 endif()
 
 # ##################################################################################################
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -30,6 +30,7 @@ set(CUDF_CPP_BUILD_DIR
   "$ENV{CUDF_CPP_BUILD_DIR}"
   CACHE STRING "path to libcudf build root"
 )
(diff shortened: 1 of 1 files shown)


jlowe:
build

src/main/cpp/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ set(CUDF_CPP_BUILD_DIR
   "$ENV{CUDF_CPP_BUILD_DIR}"
   CACHE STRING "path to libcudf build root"
 )
+set(CUDA_STATIC_RUNTIME ON)
```
revans2, line 33: nit: Can we have some comments about this? It is a little odd that we have to set this two different ways.

revans2 (approved):

jlowe:
build

src/main/cpp/CMakeLists.txt
jlowe: Added a comment.

revans2 (approved):

gerashegalov (approved):

PR #2166, opened 2024-06-24
Avoid re-generating thirdparty/cudf-pins/versions.json after libcudf build

### LINKED ISSUE #2163: [BUG] versions.json being clobbered by subsequent builds
After #2136 versions.json is getting clobbered during submodule sync builds because subsequent builds after the libcudf build are overwriting the file with a small subset of dependencies.  This loss of pinning information can be seen on the [first submodule sync commit after the change](https://github.com/NVIDIA/spark-rapids-jni/commit/97cc5127294eb60785c06f9086801c6534f45b72).

jlowe:
Fixes #2163.  Changes the libcudfjni and spark-rapids-jni builds that occur after the libcudf build to always be pinned.  We want to use the same rapids-cmake and libcudf dependencies that were found during the libcudf build for these followup builds.

This also updates the pinned version files based on this change.


jlowe:
build

revans2 (approved):

Issue #1734, opened 2024-01-25
[FEA] Mark all kernels with hidden visibility


ttnghia:
To avoid potential issues when using a static cuda runtime in the library, we need to hide all kernels from the external world. This can be done by using the static keyword or hidden visibility adding to the kernel declaration, similar to what has been done in cudf: https://github.com/rapidsai/cudf/pull/14726.

For example, we can use this macro instead of `__global__`:
```
#ifndef JNI_KERNEL
# define JNI_KERNEL __attribute__ ((visibility ("hidden"))) __global__
#endif
```

mattahrens:
Additional documentation needed to show how to do this properly with UDFs.

jlowe:
> Additional documentation needed to show how to do this properly with UDFs.

Filed https://github.com/NVIDIA/spark-rapids-examples/issues/356 to track updating the native UDF examples in the spark-rapids-examples repository.

jihoonson:
I'm working on this issue.

PR #2168, opened 2024-06-24
All kernels in the JNI should have hidden visibility

jihoonson:
Fixes https://github.com/NVIDIA/spark-rapids-jni/issues/1734.

This PR addresses a similar issue to https://github.com/rapidsai/cudf/pull/14726. Instead of adding a new macro as suggested in https://github.com/NVIDIA/spark-rapids-jni/issues/1734, it uses the one in libcudf added in https://github.com/rapidsai/cudf/pull/14726 to avoid duplicate macros.

### DIFF
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -157,13 +157,13 @@ process_value(bool first_value, T current_val, T const new_digit, bool adding)
  */
 template <typename T>
 void CUDF_KERNEL string_to_integer_kernel(T* out,
-                                         bitmask_type* validity,
-                                         const char* const chars,
-                                         size_type const* offsets,
-                                         bitmask_type const* incoming_null_mask,
-                                         size_type num_rows,
-                                         bool ansi_mode,
-                                         bool strip)
+                                          bitmask_type* validity,
+                                          const char* const chars,
+                                          size_type const* offsets,
+                                          bitmask_type const* incoming_null_mask,
+                                          size_type num_rows,
+                                          bool ansi_mode,
+                                          bool strip)
 {
   auto const group = cooperative_groups::this_thread_block();
   auto const warp  = cooperative_groups::tiled_partition<cudf::detail::warp_size>(group);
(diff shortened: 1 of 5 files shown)


hyperbolic2346 (approved):
Thanks for doing this work!

ttnghia (dismissed):

ttnghia (approved):

hyperbolic2346 (approved):

hyperbolic2346:
build

src/main/cpp/src/cast_string.cu
```diff
@@ -156,14 +156,14 @@ process_value(bool first_value, T current_val, T const new_digit, bool adding)
...
 template <typename T>
-void __global__ string_to_integer_kernel(T* out,
-                                         bitmask_type* validity,
-                                         const char* const chars,
-                                         size_type const* offsets,
-                                         bitmask_type const* incoming_null_mask,
-                                         size_type num_rows,
-                                         bool ansi_mode,
-                                         bool strip)
+void CUDF_KERNEL string_to_integer_kernel(T* out,
```
sameerz, line 159: Should this be

`void CUDF_KERNEL` or `CUDF_KERNEL void` ?

hyperbolic2346: Validity wise it doesn't matter. Stylistically it can be argued we should be consistent. I think it's outside the scope of this change to enforce it though. This example started with `void __global__` and ended with `void __global__ static`. This looks odd expanded out, but is valid and does the expected thing. I would prefer to see it as `CUDF_KERNEL void` as well.

jihoonson: Oh, I was just replacing all `__global__` with `CUDA_KERNEL` and never noticed this. I agree `CUDA_KERNEL void` is better and consistent. Will fix it.

jihoonson: Ah just noticed it has been fixed already in https://github.com/NVIDIA/spark-rapids-jni/pull/2178.

Issue #2096, opened 2024-05-30
[FEA] Support the hive hashing function


firestarman:
**Is your feature request related to a problem? Please describe.**
 Spark supports [two algorithms to generate the bucket id](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/V1Writes.scala#L123) for the bucketing write. 

Hash partitioning is simple because we already support it. To support the other one we would need to implement the HiveHash expression. That would be a new kernel.

**Describe the solution you'd like**
Here is the code of Hive hash.
https://github.com/apache/spark/blob/25d96f7bacb43a7d5a835454ecc075e40d4f3c93/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/hash.scala#L686-L1035
ints are unchanged.
longs are computed as (int) ((input >>> 32) ^ input);
a byte array uses

    int result = 0;
    for (int i = 0; i < lengthInBytes; i++) {
      result = (result * 31) + (int) Platform.getByte(base, offset + i);
    }
    return result;
for timestamps it is

    val timestampInSeconds = MICROSECONDS.toSeconds(timestamp)
    val nanoSecondsPortion = (timestamp % MICROS_PER_SECOND) * NANOS_PER_MICROS

    var result = timestampInSeconds
    result <<= 30 // the nanosecond part fits in 30 bits
    result |= nanoSecondsPortion
    ((result >>> 32) ^ result).toInt
But for decimal numbers they convert it to a BigDecimal, normalize the value, and then call hashCode on it. I think in the short term we will not be able to support decimal values as the BigDecimal implementation is GPL licensed so we cannot copy it.

The hard parts are with how some data is normalized before it is hashed.

We are also going to have problems with Maps. The way maps are hashed it depends on the order that they are stored. Maps are not ordered so we are not likely to get the same result and it might be best to just not say that we support them.

We also don't really support CalendarInterval very well, so we probably want to say at the beginning that don't support it.

Or we could just fall back to the CPU if it is a hive compatible bucketing and file a follow on for the rest.

firestarman:
This is for https://github.com/NVIDIA/spark-rapids/issues/10366.

firestarman:
Close it as common types are supported. We can file new issue for more types in the future.

PR #2171, opened 2024-06-25
Add support for hive hash

firestarman:
contribute to https://github.com/NVIDIA/spark-rapids-jni/issues/2096

This PR adds the support for Hive hash with the basic types. The supported types list is
`[bool, byte, short, int, long, string, date, timestamp, float, double]`.


firestarman:
build

firestarman:
Hi @mythrocks , could you help review it ? Thx.

firestarman:
build

src/main/cpp/src/hash.cuh
```diff
@@ -134,4 +134,18 @@ std::unique_ptr<cudf::column> xxhash64(
...
+ * @param input The table of columns to hash
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ *
+ * @returns A column where each row is the hash of a column from the input.
+ */
+std::unique_ptr<cudf::column> hive_hash(
+  cudf::table_view const& input,
+  rmm::cuda_stream_view stream      = cudf::get_default_stream(),
+  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource());
```
ttnghia, line 149: There is an issue with function declarations here. These declarations are put in a `cuh` file, then included in `*Jni.cpp` files which are not compiled under `nvcc` but regular `g++`. Thus, in the future we may have compiling issue when `g++` decides to throw compile error when seeing (unused) device code in `.cuh` file.

The solution now should be extracting these function declarations into a separate `hash.hpp` file.

src/main/cpp/src/hive_hash.cu
```diff
@@ -0,0 +1,237 @@
...
+
+constexpr hive_hash_value_t HIVE_HASH_FACTOR = 31;
+constexpr hive_hash_value_t HIVE_INIT_HASH   = 0;
+
+hive_hash_value_t __device__ inline compute_int(int32_t const& key) { return key; }
+
+hive_hash_value_t __device__ inline compute_long(int64_t const& key)
+{
+  return ((uint64_t)key >> 32) ^ key;
+}
```
ttnghia, line 45:
```suggestion
hive_hash_value_t __device__ inline compute_long(int64_t key)
{
  return (static_cast<uint64_t>(key) >> 32) ^ key;
}
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -0,0 +1,237 @@
...
+  return ret;
+}
+
+template <typename Key>
+struct hive_hash_function {
+  using result_t = hive_hash_value_t;
+
+  constexpr hive_hash_function() = delete;
+  // 'seed' is not used in 'hive_hash_function', but required by 'element_hasher'.
+  constexpr hive_hash_function(uint32_t seed) {}
```
ttnghia, line 62: Unused parameters should not have a name, to avoid compiler warning.
```suggestion
  constexpr hive_hash_function(uint32_t) {}
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -0,0 +1,237 @@
...
+   * @return The hash value of the row
+   */
+  __device__ auto operator()(cudf::size_type row_index) const noexcept
+  {
+    return cudf::detail::accumulate(
+      _table.begin(),
+      _table.end(),
+      HIVE_INIT_HASH,
+      cuda::proclaim_return_type<hive_hash_value_t>(
+        [row_index, nulls = this->_check_nulls] __device__(auto hash, auto column) {
```
ttnghia, line 168:
```suggestion
        [row_index, nulls = this->_check_nulls] __device__(auto hash, auto const& column) {
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -0,0 +1,237 @@
...
+   public:
+    __device__ element_hasher_adapter(Nullate check_nulls) noexcept : _check_nulls(check_nulls) {}
+
+    using hash_functor = cudf::experimental::row::hash::element_hasher<hash_fn, Nullate>;
+
+    template <typename T, CUDF_ENABLE_IF(not cudf::is_nested<T>())>
+    __device__ hive_hash_value_t operator()(cudf::column_device_view const& col,
+                                            cudf::size_type row_index) const noexcept
+    {
+      auto const hasher = hash_functor{_check_nulls, HIVE_INIT_HASH, HIVE_INIT_HASH};
```
ttnghia, line 192: `hasher` seems to be constant thus it can be a member variable and initialized upon class construction.

ttnghia: Doing so, we can remove `_check_nulls` from this class.

src/main/cpp/src/hive_hash.cu
```diff
@@ -0,0 +1,237 @@
...
+   */
+  __device__ auto operator()(cudf::size_type row_index) const noexcept
+  {
+    return cudf::detail::accumulate(
+      _table.begin(),
+      _table.end(),
+      HIVE_INIT_HASH,
+      cuda::proclaim_return_type<hive_hash_value_t>(
+        [row_index, nulls = this->_check_nulls] __device__(auto hash, auto column) {
+          auto cur_hash = cudf::type_dispatcher(
```
ttnghia, line 169: Since we don't support nested types, dispatching this way can reduce code size:
```
cudf::type_dispatcher<dispatch_void_if_nested>(...);
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -0,0 +1,237 @@
...
+using hive_hash_value_t = int32_t;
+
+constexpr hive_hash_value_t HIVE_HASH_FACTOR = 31;
+constexpr hive_hash_value_t HIVE_INIT_HASH   = 0;
+
+hive_hash_value_t __device__ inline compute_int(int32_t const& key) { return key; }
+
+hive_hash_value_t __device__ inline compute_long(int64_t const& key)
+{
+  return ((uint64_t)key >> 32) ^ key;
```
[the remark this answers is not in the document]
firestarman: yeah, we have unit tests to cover these cases.

src/main/cpp/src/hive_hash.cu
```diff
@@ -0,0 +1,237 @@
...
+}
+
+hive_hash_value_t __device__ inline compute_bytes(int8_t const* data, cudf::size_type const len)
```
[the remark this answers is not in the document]
firestarman: still good to know

src/main/cpp/src/HashJni.cpp
```diff
@@ -52,4 +52,19 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_Hash_xxhash64(JNIEnv* e
   }
   CATCH_STD(env, 0);
 }
+
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_Hash_hiveHash(JNIEnv* env,
+                                                                       jclass,
+                                                                       jlongArray column_handles)
+{
+  JNI_NULL_CHECK(env, column_handles, "array of column handles is null", 0);
```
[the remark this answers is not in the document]
firestarman: no i think. Just copy this check from the code of other hash functions.

src/main/cpp/src/hash.cuh
firestarman: done

src/main/cpp/src/hive_hash.cu
firestarman: good idea, done

src/main/cpp/src/hive_hash.cu
```diff
@@ -0,0 +1,237 @@
...
+    ret = ret * HIVE_HASH_FACTOR + static_cast<int32_t>(data[i]);
+  }
+  return ret;
+}
+
+template <typename Key>
+struct hive_hash_function {
+  using result_t = hive_hash_value_t;
+
+  constexpr hive_hash_function() = delete;
```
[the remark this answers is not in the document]
firestarman: removed this line

src/main/java/com/nvidia/spark/rapids/jni/Hash.java
```diff
@@ -85,7 +85,26 @@ public static ColumnVector xxhash64(ColumnView columns[]) {
     return xxhash64(DEFAULT_XXHASH64_SEED, columns);
   }
 
+  public static ColumnVector hiveHash(ColumnView columns[]) {
+    if (columns.length < 1) {
+      throw new IllegalArgumentException("Hive hashing requires at least 1 column of input");
+    }
```
ttnghia, line 91: Is this a valid assert? And matched with Spark? I thought that we can call even with 0 column?

src/main/cpp/src/hash.cuh
mythrocks: This is a great catch.  I did notice that we were putting this in a `.cuh`.  But I didn't realize that this was then included in a `.cpp`.

src/main/java/com/nvidia/spark/rapids/jni/Hash.java
mythrocks: Spark allows a table to be unpartitioned, yes.  But I don't think it's valid to call the hive hash without specifying at least one input column.

mythrocks: To clarify:  Spark shouldn't be invoking the hasher, in the case where no partition/bucketing column is specified.

firestarman: Yeah, Spark ensure there is at least one column when calling into hiveHash.
There are similar assertions for murmur3 and xxhash64.

ttnghia:
CI fails due to including the old header (`hash.cuh`). Need to change to `hash.hpp`.

firestarman:
build

firestarman:
>CI fails due to including the old header (hash.cuh) in tests/hash.cpp. Need to change to hash.hpp.

Yeah, made a fix

ttnghia (approved):

firestarman:
Really appreciate your quick reviews. @ttnghia @mythrocks

Issue #2190, opened 2024-07-01
[FEA] large string support


jlowe:
libcudf now supports string columns with more than 2GB of character data.  There are many places in the C++ and Java code that assume offsets are 4 bytes, and large string support violates that assumption, so we disabled libcudf's large string support (see #2189).  We need to find and update all of those places before enabling large-string support.

PR #2191, opened 2024-07-01
Disable large strings support in libcudf build

### LINKED ISSUE #2189: Disable large-string support in cudf configuration
rapidsai/cudf#16037 will default libcudf to enable large string support.  The RAPIDS Accelerator has many places in C++ and Java where it assumes the size of offsets is 4 bytes, and thus large strings would break in various ways.  Until the RAPIDS Accelerator is ready for large string support, we will need to disable this in the cudf configuration.

libcudf build needs to configure with CUDF_LARGE_STRINGS_DISABLED

jlowe:
Fixes #2189. rapidsai/cudf#16037 is enabling large strings support by default in libcudf, but the RAPIDS Accelerator still has many places that assume offsets are always 4 bytes.  Until #2190 is fixed, we will need to explicitly disable large strings support in  the libcudf build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -424,6 +424,7 @@
                   <arg value="-DCUDF_ENABLE_ARROW_PARQUET=ON"/>
                   <arg value="-DCUDF_USE_ARROW_STATIC=ON"/>
                   <arg value="-DCUDF_USE_PER_THREAD_DEFAULT_STREAM=${CUDF_USE_PER_THREAD_DEFAULT_STREAM}" />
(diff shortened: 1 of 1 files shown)


jlowe:
Verified that when rapidsai/cudf#16037 is added to the build, this change adds the appropriate compile-time flag for the libcudf build to prevent large strings support from being enabled by default.  Also verified that this can be added before that cudf change, as it only generates a warning in the cmake build about an unused flag.

jlowe:
build

abellina (approved):

PR #2220, opened 2024-07-11
Fix `get_json_object` producing wrong output with wildcard paths

ttnghia:
The output of `get_json_object` is wrong when the json path has a wildcard following by a name instruction. In such situations, the early return statement in the code path processing name instruction may cause the entire output row to be a null, instead of ignoring the child objects that do not contain the desired field.

This fixes the issue by removing the early return statement mentioned above.

Closes https://github.com/NVIDIA/spark-rapids/issues/11172.

### DIFF
--- a/src/main/cpp/src/get_json_object.cu
+++ b/src/main/cpp/src/get_json_object.cu
@@ -488,10 +488,10 @@ __device__ bool evaluate_path(json_parser& p,
                 return false;
               }
             }
-            ctx.task_is_done = true;
-          } else {
-            return false;
           }
+          // Mark task is done regardless whether the expected child was found.
+          // If it was not found, the root context should eventially still not be dirty.
+          ctx.task_is_done = true;
         } else {
           // below is 1st enter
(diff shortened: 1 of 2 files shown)


ttnghia:
build

src/main/cpp/src/get_json_object.cu
```diff
@@ -488,10 +488,10 @@ __device__ bool evaluate_path(json_parser& p,
                 return false;
               }
             }
-            ctx.task_is_done = true;
-          } else {
-            return false;
           }
+          // Mark task is done regardless whether the expected child was found.
+          // If it was not found, the root context should eventially still not be dirty.
```
ttnghia, line 493:
```suggestion
```

ttnghia:
build

ttnghia:
build

ttnghia:
build

ttnghia:
build

res-life (approved):
LGTM

res-life:
Good catch!

PR #2228, opened 2024-07-16
Fix completion condition for `get_json_object` with wildcard paths but empty input array

ttnghia:
In cases when we have wildcard paths  (`[*]`) but the input array corresponding to it is an empty array, the current code does not report as complete after parsing past the end of the array. Instead, it continues parsing the input string indefinitely until the end of the string. Although that does not produce an incorrect output, it is a wrong behavior.

This fixes such behavior, allowing the code to terminate correctly.


ttnghia:
build

thirtiseven (approved):
LGTM

PR #2232, opened 2024-07-16
Cleanup `get_json_object`

ttnghia:
This does some cleaning up work, including:
 * Change enum type to inherit from `int8_t` and reorganize class member variables to reduce padding. As such, class instances occupy less memory.
 * Add `const` and `static` qualifier to member functions.
 * Remove redundant code.
 * Construct path commands without stream sync, etc.

No new functionality is added in this PR.


ttnghia:
build

ttnghia:
build

src/main/cpp/src/json_parser.cuh
```diff
@@ -1386,20 +1386,22 @@ class json_parser {
...
+      t = next_token();
       if (t == json_token::START_OBJECT || t == json_token::START_ARRAY) {
         ++open;
       } else if (t == json_token::END_OBJECT || t == json_token::END_ARRAY) {
         if (--open == 0) { return true; }
       } else if (t == json_token::ERROR) {
         return false;
       }
-    }
+    } while (t != json_token::SUCCESS);
```
thirtiseven, line 1400: Question: It looks like a logic change to check if `t != json_token::SUCCESS`. Was there a bug?

ttnghia: It should not be a bug. Digging into the code, for normal situations it will return `true` and for situations like `[[[[[[[[.......` I see that this will ultimately result in the `ERROR` token. 

However, a better logic IMO is to have a return at the end of the function so I changed into this. Essentially, the behavior of this function does not change at all.

thirtiseven (approved):

PR #2246, opened 2024-07-20
Replace path-format=absolute with realpath

### LINKED ISSUE #2242: [BUG] run-in-docker can fail with unknown flag: --path-format
When running the run-in-docker script on a GCP Dataproc node running a Debian 11 image, the script fails with:
```
unknown flag: --path-format
See 'docker build --help'
```

The problem stems from improper handling of the `--path-format=absolute` flag by git added in #2238.  This flag apparently isn't recognized or handled properly in older versions of git (in this case git 2.30.2):
```
$ git rev-parse --path-format=absolute --git-common-dir
--path-format=absolute
.git
```

gerashegalov:
Fixes #2242

Accommodate older git versions without path-format option

### DIFF
--- a/build/run-in-docker
+++ b/build/run-in-docker
@@ -20,8 +20,10 @@
 
 set -e
 
-REPODIR=$(git rev-parse --path-format=absolute --show-toplevel)
-GIT_COMMON_DIR=$(git rev-parse --path-format=absolute --git-common-dir)
+REPODIR_REL=$(git rev-parse --show-toplevel)
+REPODIR=$(readlink -e "$REPODIR_REL")
+GIT_COMMON_DIR_REL=$(git rev-parse --git-common-dir)
+GIT_COMMON_DIR=$(readlink -e "$GIT_COMMON_DIR_REL")
 WORKDIR=${WORKDIR:-$REPODIR}
 
 CUDA_VERSION=${CUDA_VERSION:-11.8.0}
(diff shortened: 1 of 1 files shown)


gerashegalov:
build

build/run-in-docker
```diff
@@ -20,8 +20,10 @@
 
 set -e
 
-REPODIR=$(git rev-parse --path-format=absolute --show-toplevel)
-GIT_COMMON_DIR=$(git rev-parse --path-format=absolute --git-common-dir)
+REPODIR_REL=$(git rev-parse --show-toplevel)
+REPODIR=$(readlink -e "$REPODIR_REL")
+GIT_COMMON_DIR_REL=$(git rev-parse --git-common-dir)
+GIT_COMMON_DIR=$(readlink -e "$GIT_COMMON_DIR_REL")
```
jlowe, line 26: Any reason not to use `realpath` here? Not all platforms that provide readlink support the `-e` option, but many have `realpath` which seems to do what we want here without additional arguments.

gerashegalov: I was not familiar with`realpath` for some reason. Thanks for pointing to it. Now I am noticing that the man page for GNU coreutils readlink even states:

> Note realpath(1) is the preferred command to use for canonicalization functionality.

Similarly not all versions of realpath support `-e`. However, we can rely on git that the path exists.

gerashegalov:
build

jlowe (approved):

PR #2255, opened 2024-07-24
Avoid serializing profiler events with invalid timestamps

jlowe:
It is possible for CUPTI to generate records that have timestamps set to 0.  When these appear in the nsys-rep file, Nsight Systems shows a profile with a _huge_ timescale, making it very difficult to navigate the profile in the viewer.  This updates the profile serialization processing to ignore records with invalid timestamps and "uninteresting" marker data records that are essentially all zero except for an ID.


jlowe:
build

revans2 (approved):

PR #2256, opened 2024-07-24
Run `get_json_object_multiple_paths` with thread-parallel kernel based on number of paths

ttnghia:
Currently, either the thread-parallel or warp-parallel kernel is executed based on the input row size. This changes the condition to select which kernel to launch based on the number of JSON paths instead, which can produce better performance when having multiple paths.


ttnghia:
build

ttnghia:
build

ttnghia:
Tested the threshold on switching between thread-parallel vs warp parallel, with a small dataset:
```
thread par, max path size:
2: 40s
4: 17s
6: 15s
8: 11.8s
10: 9.3s
16: 7.6s
32: 5.9s

warp par, max path size:
2: 19s
4: 13s
6: 13s
8: 12s
10: 11.5s
16: 10.6s
32: 10.2s
```

As from the test, warp-parallel kernel is faster when the number of paths is less than `8`. Thus, we will switch to thread-parallel kernel from 8 paths.

ttnghia:
build

ttnghia:
build

ttnghia:
build

revans2 (approved):
The numbers look better for all of the GPUs I have tested it on. I want to do a bit more testing on some other datasets.

Also could we try a hybrid approach where we have the threads in a warp dedicated to a single row up to 32 paths.  So if there are 2 paths 2 of the threads would be active. If there were 32 paths, all of them would be active?  The main reason for this is that I see a huge performance drop off between 7 and 8 paths, but also if I don't use powers of 2 for the number of paths, then I get kind of sporadic performance results.

parallelism scaling

ttnghia:
Close this as it is no longer needed. Instead, it is replaced by https://github.com/NVIDIA/spark-rapids-jni/pull/2258.

PR #2258, opened 2024-07-25
Implement hybrid thread/warp parallel kernel for `get_json_object`

ttnghia:
Instead of selecting either a thread-parallel or warp-parallel kernel depending on the input row size, this approach implements a different hybrid kernel such that:
 * Each warp always processes at max one input row.
 * The number of active threads in a warp depends on the number of JSON paths. The number of warps processing one row is computed as `ceil(num_path / warp_size)`.


ttnghia:
Tested thread-parallel vs warp parallel vs hybrid kernel, with a small (fingerprint) dataset:
```
thread par, max path size:
2: 40s
4: 17s
6: 15s
8: 11.8s
10: 9.3s
16: 7.6s
32: 5.9s

warp par, max path size:
2: 19s
4: 13s
6: 13s
8: 12s
10: 11.5s
16: 10.6s
32: 10.2s

hybrid thread/warp par, max path size:
2: 17.2s
4: 11s
6: 10s
8: 8s
10: 7.6s
16: 6.4s
32: 5.6s
```

ttnghia:
build

revans2 (approved):
The volatility is better, and the low parallelism performance is much better, but it didn't really improve the average performance for the one test case I ran. I think that is okay, because there are less performance low spots that we could hit.


parallelism scaling (1)

revans2:
To be clear the chart is comparing against https://github.com/NVIDIA/spark-rapids-jni/pull/2256#pullrequestreview-2199808677 and I still want to run some more tests, but I think this is good to go.

PR #2265, opened 2024-07-26
Separate device data generation for JSON paths from `get_json_object`

ttnghia:
Instead of re-generating device data for the same set of JSON paths every time calling `get_json_object`, we can generate it beforehand, then pass its device pointer(s) into `get_json_object`, avoiding the overhead of creating the same device data for JSON paths in the long run. Initial tests with this showed a speedup appr. `~5%` on `get_json_object`.

Note that this is a breaking change, thus it needs to be merged with the corresponding changes in the plugin as close as possible.


ttnghia:
build

ttnghia:
build

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -94,25 +100,41 @@ public static ColumnVector[] getJsonObjectMultiplePaths(ColumnVector input,
...
+    long[] ptrs = createGpuJSONPaths(typeNums, names, indexes, pathOffsets);
+    GpuJSONPath[] ret = new GpuJSONPath[ptrs.length];
     for (int i = 0; i < ptrs.length; i++) {
-      ret[i] = new ColumnVector(ptrs[i]);
+      ret[i] = new GpuJSONPath(ptrs[i]);
+    }
+    return ret;
+  }
+
+  public static ColumnVector getJsonObject(ColumnVector input, GpuJSONPath path) {
```
revans2, line 111: nit: To avoid the merge problem, could we keep the old APIs also (for backwards compatibility) and just have something like.

```
GPUJSONPath[] gpuPaths = createGpuJSONPaths(Arrays.asList(path));
try {
  return getJsonObject(input, gpuPaths[0]);
} finally {
  for (GPUJSONPath p: gpuPaths) {
    if (p != nul) {
      p.close();
    }
  }
}
```

Then something similar for the multi case too.

src/test/java/com/nvidia/spark/rapids/jni/GetJsonObjectTest.java
```diff
@@ -707,12 +707,23 @@ void getJsonObjectTest_ExceedMaxNestingDepthInJSONParser() {
...
+         ColumnVector output = getJsonObject(validInput, query)) {
       assertColumnsAreEqual(validInput, output);
-      assertThrows(CudfException.class, () -> JSONUtils.getJsonObject(invalidInput, query));
+      assertThrows(CudfException.class, () -> getJsonObject(invalidInput, query));
     }
   }
 
+  private ColumnVector getJsonObject(ColumnVector jsonCv, JSONUtils.PathInstructionJni[] path) {
+    List<List<JSONUtils.PathInstructionJni>> pathsAsList = Arrays.asList(Arrays.asList(path));
+    JSONUtils.GpuJSONPath[] gpuPaths = JSONUtils.createGpuJSONPaths(pathsAsList);
```
revans2, line 718: Don't these need to be closed or else we will leak GPU memory?

src/test/java/com/nvidia/spark/rapids/jni/GetJsonObjectTest.java
```diff
@@ -707,12 +707,23 @@ void getJsonObjectTest_ExceedMaxNestingDepthInJSONParser() {
...
+    return JSONUtils.getJsonObject(jsonCv, gpuPaths[0]);
+  }
+
+  private ColumnVector[] getJsonObjectMultiplePaths(ColumnVector jsonCv,
+                                                    List<List<JSONUtils.PathInstructionJni>> paths) {
+    JSONUtils.GpuJSONPath[] gpuPaths = JSONUtils.createGpuJSONPaths(paths);
```
revans2, line 724: Don't these need to be closed or else we will leak GPU memory?

revans2:
I spoke with @jlowe and @abellina and I think we have a plan on how to do this for the plugin. I need to talk to @sameerz about priorities/etc.

ttnghia:
Close this as it is not practical.

PR #2285, opened 2024-07-31
Fix out of bound write handling for `get_json_object_multiple_paths`

ttnghia:
This fixes the code path for handling out-of-bound write when there are more than one JSON paths in `get_json_object_multiple_paths`. Currently there are two issues with it:
 * The kernel launcher does not allow running the kernel again with different number of paths, and
 * The indices of the out-of-bound write buffers are not correctly identified.

A unit test is also added to check for this particular code path.

### DIFF
--- a/src/main/cpp/src/get_json_object.cu
+++ b/src/main/cpp/src/get_json_object.cu
@@ -808,6 +808,13 @@ struct json_path_processing_data {
  *
  * The number of warps processing each row is computed as `ceil(num_paths / warp_size)`.
  *
+ * We explicitly set a value for `min_block_per_sm` parameter in the launch bounds to avoid
+ * spilling from the kernel itself. By default NVCC uses a heuristic to find a balance between
+ * the maximum number of registers used by a kernel and the parallelism of the kernel.
+ * If lots of registers are used the parallelism may suffer. But in our case NVCC gets this wrong
+ * and we want to avoid spilling all the time or else the performance is really bad. This
+ * essentially tells NVCC to prefer using lots of registers over spilling.
+ *
  * @param input The input JSON strings stored in a strings column
  * @param path_data Array containing all path data
  * @param num_threads_per_row Number of threads processing each input row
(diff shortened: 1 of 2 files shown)


ttnghia:
build

src/main/cpp/src/get_json_object.cu
```diff
@@ -1123,12 +1104,14 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
...
 
-  for (auto const idx : oob_indices) {
-    output[idx] = cudf::make_strings_column(input.size(),
-                                            std::move(out_offsets_and_sizes[idx].first),
-                                            out_char_buffers[idx].release(),
-                                            out_null_masks_and_null_counts[idx].second,
-                                            std::move(out_null_masks_and_null_counts[idx].first));
+  for (std::size_t idx = 0; idx < oob_indices.size(); ++idx) {
+    auto const out_idx = oob_indices[idx];
+    output[out_idx] =
```
ttnghia, line 1109: Previously it was `output[idx]`. That is incorrect.

src/main/cpp/src/get_json_object.cu
```diff
@@ -1123,12 +1104,14 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
...
+      cudf::make_strings_column(input.size(),
+                                std::move(out_offsets_and_sizes[idx].first),
+                                out_char_buffers[idx].release(),
+                                out_null_masks_and_null_counts[idx].second,
+                                std::move(out_null_masks_and_null_counts[idx].first));
```
ttnghia, line 1114: The indices `out_idx` and `idx` should be different. Previously they were the same. That is incorrect.

src/main/cpp/src/get_json_object.cu
```diff
@@ -1112,7 +1093,7 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
     h_path_data, stream, rmm::mr::get_current_device_resource());
   thrust::uninitialized_fill(
     rmm::exec_policy(stream), d_has_out_of_bound.begin(), d_has_out_of_bound.end(), 0);
-  kernel.exec(*d_input_ptr, d_path_data, stream);
+  kernel_launcher::exec(*d_input_ptr, d_path_data, stream);
```
ttnghia, line 1096: When launching the kernel a second time, the number of paths is different from that number in launching the first time.

revans2 (approved):

revans2:
I ran through all of the JSON tests and redid the performance testing I did before. Everything passed as expected, because this is just here for a very odd corner case that is not going to ever be needed in practice.

PR #2266, opened 2024-07-26
Detect and throw exception in `get_json_object` if the input has nesting depth exceed the maximum allowed value

ttnghia:
Currently, in `get_json_object`, when the maximum nesting depth has been exceeded, the evaluation process may either  return a null or just crash without any clear message. This PR adds the ability to detect such situation and  throw a better error message.

In addition, this also synchronizes between `MAX_JSON_PATH_DEPTH` in C++ JNI and `MAX_PATH_DEPTH` Java code, avoiding the possibility that they are set at different values.

PR #2290, opened 2024-08-01
Fix performance regression of `get_json_object` due to unnecessary stream synchronization

ttnghia:
After a stream sync added in https://github.com/NVIDIA/spark-rapids-jni/pull/2266, the performance of `get_json_object` surprisingly drops by more than 10%. This PR removes that stream sync by combining the error check values together and only synchronizing stream just once.

### DIFF
--- a/src/main/cpp/src/JSONUtilsJni.cpp
+++ b/src/main/cpp/src/JSONUtilsJni.cpp
@@ -25,16 +25,6 @@ using path_instruction_type = spark_rapids_jni::path_instruction_type;
 
 extern "C" {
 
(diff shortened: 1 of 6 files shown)


ttnghia:
build

ttnghia:
build

ttnghia:
build

ttnghia:
build

revans2:
I just ran performance and correctness tests and it looks good. If anything it is slightly faster than before. I want to rerun with the latest merge changes you put in too. But that should only be a few more mins

ttnghia:
build

ttnghia:
build

revans2 (approved):
The performance looks good and all of my tests passed.

PR #2299, opened 2024-08-02
Have multi-get_json_object batch and cluster paths

revans2:
This is to improve performance and try to stay within a memory limit.

This contributes to https://github.com/NVIDIA/spark-rapids/issues/11263

In my performance testing I am seeing an improvement of 9% on an a6000 GPU when there are lots of paths to combine, many of which have common beginning paths.  I will test on some less powerful GPUs too. I think we can do a better job with clustering the paths in a smarter way, but this is super simple and is showing a lot of promise so I thought I would check it in.

Note to fully take advantage of this we will need some changes to the plugin as well https://github.com/NVIDIA/spark-rapids/pull/11289.

### DIFF
--- a/src/main/cpp/src/get_json_object.cu
+++ b/src/main/cpp/src/get_json_object.cu
@@ -17,8 +17,6 @@
 #include "get_json_object.hpp"
 #include "json_parser.cuh"
 
-#include <numeric>
-
 #include <cudf/column/column_device_view.cuh>
 #include <cudf/column/column_factories.hpp>
 #include <cudf/detail/nvtx/ranges.hpp>
@@ -42,6 +40,8 @@
 #include <thrust/pair.h>
 #include <thrust/transform_reduce.h>
 #include <thrust/tuple.h>
+
+#include <numeric>
 
 namespace spark_rapids_jni {
 
@@ -1019,8 +1019,7 @@ bool check_error(cudf::detail::host_vector<int8_t> const& error_check)
 std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
   cudf::strings_column_view const& input,
   cudf::detail::input_offsetalator const& in_offsets,
-  std::vector<cudf::host_span<std::tuple<path_instruction_type, std::string, int32_t>>>&
-    json_paths,
+  std::vector<cudf::host_span<std::tuple<path_instruction_type, std::string, int32_t>>>& json_paths,
   int64_t scratch_size,
   rmm::cuda_stream_view stream,
   rmm::device_async_resource_ref mr)
@@ -1157,8 +1156,7 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
 
 std::vector<std::unique_ptr<cudf::column>> get_json_object(
   cudf::strings_column_view const& input,
-  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>> &
-    json_paths,
+  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>>& json_paths,
   int64_t memory_budget_bytes,
   int32_t parallel_override,
   rmm::cuda_stream_view stream,
@@ -1176,7 +1174,7 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
   }
 
   std::vector<std::size_t> sorted_indices(json_paths.size());
-  std::iota(sorted_indices.begin(), sorted_indices.end(), 0); // Fill with 0, 1, 2, ...
+  std::iota(sorted_indices.begin(), sorted_indices.end(), 0);  // Fill with 0, 1, 2, ...
 
   // Sort indices based on the corresponding paths.
   std::sort(sorted_indices.begin(), sorted_indices.end(), [&json_paths](size_t i, size_t j) {
@@ -1237,14 +1235,14 @@ std::unique_ptr<cudf::column> get_json_object(
   rmm::device_async_resource_ref mr)
 {
   CUDF_FUNC_RANGE();
-  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>> inst{instructions};
+  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>> inst{
+    instructions};
   return std::move(detail::get_json_object(input, inst, -1, -1, stream, mr).front());
 }
 
 std::vector<std::unique_ptr<cudf::column>> get_json_object_multiple_paths(
   cudf::strings_column_view const& input,
-  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>>&
-    json_paths,
+  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>>& json_paths,
   int64_t memory_budget_bytes,
   int32_t parallel_override,
   rmm::cuda_stream_view stream,
--- a/src/main/cpp/src/get_json_object.cu
+++ b/src/main/cpp/src/get_json_object.cu
@@ -1017,7 +1017,7 @@ bool check_error(cudf::detail::host_vector<int8_t> const& error_check)
 }
 
 std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
-  cudf::strings_column_view const& input,
+  cudf::column_device_view const& input,
   cudf::detail::input_offsetalator const& in_offsets,
   std::vector<cudf::host_span<std::tuple<path_instruction_type, std::string, int32_t> const>> const&
     json_paths,
@@ -1030,8 +1030,6 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
 
   auto const num_outputs = json_paths.size();
   std::vector<std::unique_ptr<cudf::column>> output;
-
-  auto const d_input_ptr = cudf::column_device_view::create(input.parent(), stream);
 
   // The error check array contains markers denoting if there is any out-of-bound write occurs
   // (first `num_outputs` elements), or if the nesting depth exceeded its limits (the last element).
@@ -1066,7 +1064,7 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
   thrust::uninitialized_fill(
     rmm::exec_policy(stream), d_error_check.begin(), d_error_check.end(), 0);
 
-  kernel_launcher::exec(*d_input_ptr, d_path_data, d_max_path_depth_exceeded, stream);
+  kernel_launcher::exec(input, d_path_data, d_max_path_depth_exceeded, stream);
   auto h_error_check = cudf::detail::make_host_vector_sync(d_error_check, stream);
   auto has_no_oob    = check_error(h_error_check);
 
@@ -1135,7 +1133,7 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
     h_path_data, stream, rmm::mr::get_current_device_resource());
   thrust::uninitialized_fill(
     rmm::exec_policy(stream), d_error_check.begin(), d_error_check.end(), 0);
-  kernel_launcher::exec(*d_input_ptr, d_path_data, d_max_path_depth_exceeded, stream);
+  kernel_launcher::exec(input, d_path_data, d_max_path_depth_exceeded, stream);
   h_error_check = cudf::detail::make_host_vector_sync(d_error_check, stream);
   has_no_oob    = check_error(h_error_check);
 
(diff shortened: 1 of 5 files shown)


revans2:
build

src/main/cpp/src/get_json_object.hpp
```diff
@@ -53,11 +53,18 @@ std::unique_ptr<cudf::column> get_json_object(
  * This function processes all the JSON paths in parallel, which may be faster than calling
  * to `get_json_object` on the individual JSON paths. However, it may consume much more GPU
  * memory, proportional to the number of JSON paths.
+ * @param input the input string column to parse JSON from
+ * @param json_paths the path operations to read extract
+ * @param memory_budget_bytes a memory budget for temporary memory usage if > 0
+ * @param parallel_override if > 0 then teh number of paths to process in parallel
+ *  this will cause the memory_budget_bytes to be ignored.
```
ttnghia, line 60:
```suggestion
 * @param parallel_override if > 0 then specify the number of paths to process in parallel 
 *        (this will cause the `memory_budget_bytes` to be ignored)
```

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -71,8 +77,36 @@ public static ColumnVector getJsonObject(ColumnVector input, PathInstructionJni[
     return new ColumnVector(getJsonObject(input.getNativeView(), typeNums, names, indexes));
   }
 
+  /**
+   * Extract multiple JSON paths from a JSON column. The paths are processed n a Spark
```
ttnghia, line 81:
```suggestion
   * Extract multiple JSON paths from a JSON column. The paths are processed in a Spark
```

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -55,6 +55,12 @@ public PathInstructionJni(PathInstructionType type, String name, int index) {
     }
   }
 
+  /**
+   * Extract a JSON path from a JSON column. The path is processed in a Spark compatible way.
+   * @param input the string column containing JSON
+   * @param pathInstructions the instructions for the path processing
+   * @return the result of processing the path.
```
ttnghia, line 62: Let's remove all period for consistency (as other lines do not have it).
```suggestion
   * @return the result of processing the path0
```

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -71,8 +77,36 @@ public static ColumnVector getJsonObject(ColumnVector input, PathInstructionJni[
     return new ColumnVector(getJsonObject(input.getNativeView(), typeNums, names, indexes));
   }
 
+  /**
+   * Extract multiple JSON paths from a JSON column. The paths are processed n a Spark
+   * compatible way.
+   * @param input the string column containing JSON
+   * @param paths the instructions for multiple paths.
+   * @return the result of processing each path in the order that they were passed in.
```
ttnghia, line 85:
```suggestion
   * @param paths the instructions for multiple paths
   * @return the result of processing each path in the order that they were passed in
```

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -71,8 +77,36 @@ public static ColumnVector getJsonObject(ColumnVector input, PathInstructionJni[
...
+   */
   public static ColumnVector[] getJsonObjectMultiplePaths(ColumnVector input,
                                                           List<List<PathInstructionJni>> paths) {
+    return getJsonObjectMultiplePaths(input, paths, -1, -1);
+  }
+
+  /**
+   * Extract multiple JSON paths from a JSON column. The paths are processed n a Spark
```
ttnghia, line 93:
```suggestion
   * Extract multiple JSON paths from a JSON column. The paths are processed in a Spark
```

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -71,8 +77,36 @@ public static ColumnVector getJsonObject(ColumnVector input, PathInstructionJni[
...
+   * compatible way.
+   * @param input the string column containing JSON
+   * @param paths the instructions for multiple paths.
```
ttnghia, line 96:
```suggestion
   * @param paths the instructions for multiple paths
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -1154,6 +1154,73 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
...
+    std::tuple<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>, std::size_t>>
+    sorted_paths;
+  for (std::size_t i = 0; i < json_paths.size(); i++) {
+    sorted_paths.emplace_back(json_paths[i], i);
+  }
+  std::sort(sorted_paths.begin(), sorted_paths.end());
+
+  auto const in_offsets =
+    cudf::detail::offsetalator_factory::make_input_iterator(input.offsets(), input.offset());
+  auto const scratch_size = calc_scratch_size(input, in_offsets, stream, mr);
```
ttnghia, line 1187:
```suggestion
  auto const scratch_size = calc_scratch_size(input, in_offsets, stream);
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -1154,6 +1154,73 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
...
+  std::vector<std::unique_ptr<cudf::column>> output(num_outputs);
+  std::size_t starting_path = 0;
+  while (starting_path < num_outputs) {
+    std::size_t at = starting_path;
+    std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>> batch;
+    std::vector<std::size_t> output_ids;
+    if (parallel_override > 0) {
+      int count = 0;
+      while (at < num_outputs && count < parallel_override) {
+        batch.push_back(std::get<0>(sorted_paths[at]));
```
ttnghia, line 1200: This is copying the entire path (a vector of tuple) which is expensive. It is much better to construct a `batch` as `std::vector<cudf::host_span>` so we won't have any copy.

src/main/cpp/src/get_json_object.cu
```diff
@@ -1154,6 +1154,73 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
...
+  auto const scratch_size = calc_scratch_size(input, in_offsets, stream, mr);
+  if (memory_budget_bytes <= 0 && parallel_override <= 0) {
+    parallel_override = static_cast<int>(sorted_paths.size());
+  }
+  std::vector<std::unique_ptr<cudf::column>> output(num_outputs);
+  std::size_t starting_path = 0;
+  while (starting_path < num_outputs) {
+    std::size_t at = starting_path;
+    std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>> batch;
+    std::vector<std::size_t> output_ids;
```
ttnghia, line 1196: Move `batch` and `output_ids` out of the `while` loop, then `resize(0)` inside the loop. By doing so we can avoid the vectors doing memory reallocation when their capacity exceeded.

src/main/cpp/src/get_json_object.cu
ttnghia: This requires changing line 1021-1022 above into
```
std::vector<cudf::host_span<std::tuple<path_instruction_type, std::string, int32_t> const>> const&
    json_paths,
```

Then, `batch` type should be:
```
std::vector<cudf::host_span<std::tuple<path_instruction_type, std::string, int32_t> const>> batch;
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -1154,6 +1154,73 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
...
+      while (at < num_outputs && count < parallel_override) {
+        batch.push_back(std::get<0>(sorted_paths[at]));
+        output_ids.push_back(std::get<1>(sorted_paths[at]));
+        at++;
+        count++;
+      }
+    } else {
+      long budget = 0;
+      while (at < num_outputs && budget < memory_budget_bytes) {
+        batch.push_back(std::get<0>(sorted_paths[at]));
```
ttnghia, line 1208:
```suggestion
        batch.emplace_back(std::get<0>(sorted_paths[at]));
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -1154,6 +1154,73 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object(
...
+    return output;
+  }
+
+  std::vector<
+    std::tuple<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>, std::size_t>>
+    sorted_paths;
+  for (std::size_t i = 0; i < json_paths.size(); i++) {
+    sorted_paths.emplace_back(json_paths[i], i);
+  }
+  std::sort(sorted_paths.begin(), sorted_paths.end());
```
ttnghia, line 1183: IMO this is also expensive, since we have to copy paths from the input `json_paths` into `sorted_paths`. We can do better by just using a sorted order generated from sorting `json_paths`:
```
std::vector<size_t> sorted_indices(json_paths.size());
std::iota(sorted_indices.begin(), sorted_indices.end(), 0); // Fill with 0, 1, 2, ...

// Sort indices based on the corresponding paths
std::sort(sorted_indices.begin(), sorted_indices.end(), [&json_paths](size_t i, size_t j) {
    return paths[i] < paths[j];
});
```

Then later on, this `sorted_indices[at]` will replace `std::get<1>(sorted_paths[at])`.

revans2:
@ttnghia is this what you wanted? I am not sure how to mark a host_span as pointing to const data, so I ended up needing to remove the const from the input paths. If there is a better way please let me know.

ttnghia:
> how to mark a host_span as pointing to const data

That is in my previous suggestion: https://github.com/NVIDIA/spark-rapids-jni/pull/2299#discussion_r1702158302
```
std::vector<cudf::host_span<std::tuple<path_instruction_type, std::string, int32_t> const>> batch;
```

src/main/cpp/src/get_json_object.cu
```diff
@@ -907,7 +909,7 @@ std::tuple<std::vector<rmm::device_uvector<path_instruction>>,
            cudf::string_scalar,
            std::string>
 construct_path_commands(
-  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>> const&
+  std::vector<cudf::host_span<std::tuple<path_instruction_type, std::string, int32_t>>> const&
```
ttnghia, line 912:
```suggestion
  std::vector<cudf::host_span<std::tuple<path_instruction_type, std::string, int32_t> const>> const&
```

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -71,8 +77,36 @@ public static ColumnVector getJsonObject(ColumnVector input, PathInstructionJni[
     return new ColumnVector(getJsonObject(input.getNativeView(), typeNums, names, indexes));
   }
 
+  /**
+   * Extract multiple JSON paths from a JSON column. The paths are processed in a Spark
+   * compatible way.
+   * @param input the string column containing JSON
+   * @param paths the instructions for multiple paths.
```
ttnghia, line 84:
```suggestion
   * @param input the string column containing JSON
   * @param paths the instructions for multiple paths
```

revans2:
build

ttnghia (approved):

src/main/cpp/src/get_json_object.cu
```diff
@@ -988,49 +1016,22 @@ bool check_error(cudf::detail::host_vector<int8_t> const& error_check)
...
-  // Since we do not know their sizes, we need to allocate the buffer a bit larger than the input
-  // size so that we will not write output strings into an out-of-bound position.
-  // Checking out-of-bound needs to be performed in the main kernel to make sure we will not have
-  // data corruption.
-  auto const scratch_size = [&, max_row_size = max_row_size] {
-    // Pad the scratch buffer by an additional size that is a multiple of max row size.
-    auto constexpr padding_rows = 10;
-    return input.chars_size(stream) + max_row_size * padding_rows;
-  }();
+  auto const d_input_ptr = cudf::column_device_view::create(input.parent(), stream);
```
ttnghia, line 1034: Sorry, I missed this. `d_input_ptr` can be at the caller (in `get_json_object`). Then it is passed in this by:
```
std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
  cudf::column_device_view const& input,
)
```
At the caller:
```
auto tmp = get_json_object_batch(*d_input_ptr, in_offsets, batch, scratch_size, stream, mr);
```

This is because each time calling to `column_device_view::create` is also not trivial.

revans2:
build

ttnghia (approved):

PR #2327, opened 2024-08-14
Rename function and class containing `from_json`

ttnghia:
This includes a few changes for `from_json` function:
 * Renames `from_json` into `from_json_to_raw_map`, since the current implementation is not a full-feature `from_json`.
 * Move the corresponding JNI C++ and Java function from `MapUtils` into `JSONUtils`, since JSON is more relevant than map.

Moving away from `Map*` into `JSON*` is necessary for further development of `from_json` to support more output data types other than just map.

To avoid breaking changes in the Spark plugin, the function `MapUtils#extractRawMapFromJsonString` is redirected to  `JSONUtils#extractRawMapFromJsonString`. The entire class `MapUtils` will be removed in the follow up work.


ttnghia:
build

ttnghia:
build

ttnghia:
build

revans2 (commented):
The change looks fine. My main concern is that we are going to break the plugin. Could you please keep MapUtils.java with the `extractRawMapFromJsonString` but have it call into JSONUtils.java's `extractRawMapFromJsonString`? But because of how the plugin works we cannot deprecate the MapUtils API. But we can delete it as soon as the plugin is updated to use the new API.

ttnghia:
build

ttnghia:
> Could you please keep MapUtils.java with the `extractRawMapFromJsonString` but have it call into JSONUtils.java's `extractRawMapFromJsonString`?

That sounds good. Now we have `MapUtils#extractRawMapFromJsonString` calls to `JSONUtils#extractRawMapFromJsonString` so we won't break the plugin build.

revans2 (approved):

PR #2357, opened 2024-08-27
Update for changes to Arrow linkage in libcudf

vyasr:
With https://github.com/rapidsai/cudf/pull/16640, the libcudf library no longer links to arrow. While that dependency has been removed, the Java native code has not been updated to avoid arrow and still needs it for interop with arrow buffers (cudf now uses [nanoarrow](https://github.com/apache/arrow-nanoarrow)). While that may be removed in the future, for now the spark-rapids-jni needs to have its build updated to support the new location of arrow libraries and headers since they are now within the libcudfjni build instead of the libcudf build.

This PR depends on https://github.com/rapidsai/cudf/pull/16640, but will be required for spark-rapids-jni to build once that cudf PR is merged.


mythrocks:
Build

mythrocks (dismissed):
👍 LGTM!
Thanks for working on this, @vyasr. 👏 

I've run a smattering of integration tests in `spark-rapids` based on this change, including the `parquet_pyarrow` tests.  Things appear to be in order.

jlowe (commented):
This PR will need to explicitly update the thirdparty/cudf submodule to a version that contains rapidsai/cudf#16640, e.g.: latest branch-24.10, so it can test with those changes.  We periodically update the submodule to follow cudf changes, but those updates will only automatically occur if it builds and tests pass.  Since it won't build as-is without these changes, we'll need this PR to pull in the cudf changes.

vyasr:
Build

vyasr:
Thanks for confirming Jason. I've updated the branch and attempted to trigger a build although I suspect I won't have the power to do so.

jlowe (approved):

jlowe:
build

src/main/cpp/CMakeLists.txt
```diff
@@ -270,6 +274,7 @@ target_link_libraries(
     cudf::cudf
     nvtx3::nvtx3-cpp
   -Wl,--no-whole-archive
+    ${ARROW_LIB}
```
jlowe, line 277: The fact that we need to add the arrow lib dependency explicitly implies there's a bug in the cudfjni CMakeLists where it should have exported that dependency but did not.

vyasr: The problem is that the cudfjni isn't doing any exporting or installing at all. It is not producing a proper CMake package in the build directory (nor would it if installed, although that's not relevant here since AFAICT it's never installed, just used directly from the build directory). [Note that we're just using `find_library` for the lib instead of `find_package` like we would for something properly set up for export via CMake](https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.10/src/main/cpp/CMakeLists.txt#L152).

PR #2392, opened 2024-09-10
[DRAFT] Device metrics

zpuller:
Addresses [to](https://gitlab-master.nvidia.com/nvspark/spark-rapids-private/-/issues/183)

This is in draft state. Still need to perform additional testing and some cleaning.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -1373,6 +1373,8 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
           transition(thread->second, thread_state::THREAD_RUNNING);
           // TODO why do we not set this to is_for_cpu
           thread->second.is_cpu_alloc = false;
+          // num_bytes is likely not padded, which could cause slight inaccuracies
+          // but for now it shouldn't matter for watermark purposes
           thread->second.memory_allocated_bytes += num_bytes;
           thread->second.metrics.max_memory_allocated = std::max(
             thread->second.metrics.max_memory_allocated,
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -24,6 +24,7 @@
 #include <spdlog/sinks/ostream_sink.h>
 #include <spdlog/spdlog.h>
 
+#include <algorithm>
 #include <chrono>
 #include <exception>
 #include <map>
@@ -203,6 +204,8 @@ struct task_metrics {
   // The amount of time that this thread has lost due to retries (not including blocked time)
   long time_lost_nanos = 0;
 
+  long max_memory_allocated = 0;
+
   void take_from(task_metrics& other)
   {
     add(other);
@@ -215,6 +218,7 @@ struct task_metrics {
     this->num_times_split_retry_throw += other.num_times_split_retry_throw;
     this->time_blocked_nanos += other.time_blocked_nanos;
     this->time_lost_nanos += other.time_lost_nanos;
+    this->max_memory_allocated = std::max(this->max_memory_allocated, other.max_memory_allocated);
   }
 
   void clear()
@@ -295,6 +299,8 @@ class full_thread_state {
   // time)
   long time_retry_running_nanos = 0;
   std::chrono::time_point<std::chrono::steady_clock> block_start;
+  long memory_allocated_bytes = 0;
+
   // metrics for the current thread
   task_metrics metrics;
 
@@ -799,6 +805,10 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     return get_and_reset_metric(task_id, &task_metrics::time_lost_nanos);
   }
 
+  long get_and_reset_max_memory_allocated(long const task_id) {
+    return get_and_reset_metric(task_id, &task_metrics::max_memory_allocated);
+  }
+
   void check_and_break_deadlocks()
   {
     std::unique_lock<std::mutex> lock(state_mutex);
@@ -823,7 +833,7 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     // blocking is not used yet. It could be used for some debugging so we are keeping it.
     std::unique_lock<std::mutex> lock(state_mutex);
     auto const thread_id = static_cast<long>(pthread_self());
-    post_alloc_success_core(thread_id, true, was_recursive, lock);
+    post_alloc_success_core(thread_id, true, was_recursive, amount, lock);
   }
 
   bool cpu_postalloc_failed(bool const was_oom, bool const blocking, bool const was_recursive)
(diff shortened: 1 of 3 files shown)


src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1359,7 +1371,15 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
             throw std::invalid_argument(ss.str());
           }
           transition(thread->second, thread_state::THREAD_RUNNING);
+          // TODO why do we not set this to is_for_cpu
```
revans2, line 1374: I will have to look at the code again to try and remember. I think that this path is only taken for GPU allocations, but then why are we passing in `is_for_cpu`? I think it has something to do with `!was_recursive`, but like I said I am going to read through the code again to know for sure.

revans2 (commented):
Looking good.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2: I remember now. It is just be being a bit defensive and resetting the state when we transitioned to running. It should not need to be set at all, because it only matters when we are in a state that is in the middle of doing an allocation, and THREAD_RUNNING is not in the middle of doing any allocations.

zpuller:
build

zpuller:
build

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1333,18 +1344,22 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
   }
 
   void post_alloc_success_core(long const thread_id,
                                bool const is_for_cpu,
                                bool const was_recursive,
+                               std::size_t const num_bytes,
                                std::unique_lock<std::mutex>& lock)
   {
     // pre allocate checks
+    std::cerr << "post_alloc_success_core" << std::endl;
```
revans2, line 1362: Need to remove this

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1360,6 +1375,11 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
           }
           transition(thread->second, thread_state::THREAD_RUNNING);
           thread->second.is_cpu_alloc = false;
+          // num_bytes is likely not padded, which could cause slight inaccuracies
+          // but for now it shouldn't matter for watermark purposes
```
revans2, line 1379: Please make sure that we document this in the metrics page when it is fully ready

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -799,6 +805,11 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     return get_and_reset_metric(task_id, &task_metrics::time_lost_nanos);
   }
 
+  long get_and_reset_max_memory_allocated(long const task_id)
```
revans2, line 808: Looking at the code this is combined for both CPU and GPU. I was thinking that we would have GPU and CPU memory separate. The main reasons I would like them separate is that, at least in the short term, the GPU value will be accurate, but the CPU one might not be. Also the if there is fragmentation for GPU we might see spilling when the total allocated < the size of the GPUs memory. This would give us a clue that it is happening. But the second one is kind of a long shot.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
zpuller: Which page are you referring to exactly?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
zpuller: Can you be more specific on which code path exactly is currently combined for CPU and GPU? I thought you mentioned in the previous comment that `post_alloc_success_core` should be GPU only, so I'm a bit confused.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2: Sorry it is an external page https://docs.nvidia.com/spark-rapids/user-guide/latest/tuning-guide.html#metrics

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2: Line 837 of this file is calling `post_alloc_success_core` from `cpu_postalloc_success`, which is clearly not gpu for a GPU. If this is intended to be GPU only. Can we please update the name of the metric so it has gpu in it? and only update it when the allocation/free is for a GPU?

As for what I might have said I'm sorry I don't remember the context exactly, but I guess I was wrong.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -203,6 +204,8 @@ struct task_metrics {
   // The amount of time that this thread has lost due to retries (not including blocked time)
   long time_lost_nanos = 0;
 
+  long max_memory_allocated = 0;
```
revans2, line 207: Again because both GPU and CPU memory is processed by this file can we please mark this as gpu_max_memory_allocated?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
revans2, line 808: Here too `get_and_reset_gpu_max_memory_allocated`

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1779,6 +1803,7 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
             if (is_for_cpu == t_state.is_cpu_alloc) {
               transition(t_state, thread_state::THREAD_ALLOC_FREE);
             }
+            t_state.gpu_memory_allocated_bytes -= num_bytes;
```
revans2, line 1806: We also need to do the same thing here and only subtract the bytes allocated if it is for the GPU.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -838,7 +849,7 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     // addr is not used yet, but is here in case we want it in the future.
     // amount is not used yet, but is here in case we want it for debugging/metrics.
```
revans2, line 850: nit: please update the comment here about amount, so it is now correct, or just delete the line.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -823,7 +834,7 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     // blocking is not used yet. It could be used for some debugging so we are keeping it.
```
revans2, line 834: Here too please update the comment about amount, or just delete the line.

revans2 (dismissed):

zpuller:
build

zpuller:
build

PR #2393, opened 2024-09-10
Add HostTable interface to allow wielding of host tables in native code

jlowe:
Depends upon rapidsai/cudf#16784.

Adds a host version of cudf::table_view and cudf::column_view so we can examine and construct tables in host memory, along with APIs to construct a contiguous device table from a host table.  This makes it easier to move some host processing code from Java into native code to avoid excessive object creation for very wide and/or deeply nested table schemas.


src/main/cpp/src/host_table_view.hpp
```diff
@@ -0,0 +1,320 @@
...
+ * some of which may be null as indicated by a bitmask.
+ *
+ * Unless otherwise noted, the memory layout of the `host_column_view`'s data and
+ * bitmask is expected to adhere to the Arrow Physical Memory Layout
+ * Specification: https://arrow.apache.org/docs/memory_layout.html
+ *
+ * Because `host_column_view` is non-owning, no host memory is allocated nor freed
+ * when `host_column_view` objects are created or destroyed.
+ */
+class host_column_view {
```
ttnghia, line 38: I see that this class has almost identical members as `cudf::column_view`. So, given a host data buffer as a column of elements, we should be able to construct an instance of `cudf::column_view` out of it. We just need to keep track that the created instance is viewing host data, and this can be done in Java through a boolean flag.

Did I miss anything here?

jlowe: I thought about using cudf::column_view, but there are member functions that make no sense to call on it (e.g.: the ranged null_count) as it would try to run a device kernel on host data.  Using a separate class makes it type safe, enforced by the compiler, that we aren't passing the wrong type of data to a method.

ttnghia: If we want to hide some member functions, we could implement this class with private inheritance of `cudf::column_view`, with explicit exposure of the member functions we want. For example:
```
class host_column_view : cudf::column_view {
public:
  using cudf::column_view::size();
  using cudf::column_view::data();
...
```

jlowe: Does that work in practice given how children work?  cudf::column_view expects the children to be column_views, but they need to be host_column_views in this case.  Seems the closest to what we want is to derive privately from cudf::detail::column_view_base which has everything except the children, and then host_column_view would add just the children interfaces that expect and return host_column_view.  However that would be using a cudf::detail API directly, and given how trivial this stuff is, I'd rather have the duplication than risk breakage given these aren't ever supposed to be used in the same APIs anyway.

ttnghia: Okay I see what you meant. Then this would be fine, although that seems not a very good approach when having code duplication.

src/main/cpp/src/HostTableJni.cpp
```diff
@@ -0,0 +1,249 @@
...
+    auto buffer          = reinterpret_cast<uint8_t*>(host_address);
+    auto buffer_size     = static_cast<std::size_t>(host_size);
+    auto stream          = reinterpret_cast<cudaStream_t>(jstream);
+    auto host_table_view = to_host_table_async(*table, buffer, buffer_size, stream);
+    return reinterpret_cast<jlong>(host_table_view.release());
+  }
+  CATCH_STD(env, 0);
+}
+
+JNIEXPORT jlongArray JNICALL Java_com_nvidia_spark_rapids_jni_HostTable_toDeviceColumnViews(
```
ttnghia, line 207: Does this mean we are creating `cudf::column_view` from host data pointer? I see that later we call `ColumnVector.fromViewWithContiguousAllocation(columnViewHandles)` on the output of this, but does that  implicitly handle host-device memcpy?

jlowe: We're creating a cudf::column_view from a host view but also applying a constant offset to each pointer that effectively translates the pointers from host memory into device memory,  See https://github.com/NVIDIA/spark-rapids-jni/pull/2393/files#diff-19b375574bb1514d837e2a4fbf847c446512d330b685cf12fa67f33ea9978d1fR116-R118 for details.  The host buffer is copied to device memory and the difference in base pointers of the host and device buffers is used to translate the host views into device views.

ttnghia (approved):

jlowe:
build

PR #2449, opened 2024-09-27
POC: Implement `from_json_to_structs` using `get_json_object` kernel

ttnghia:
TBA


ttnghia:
Close as it is no longer needed.

PR #2457, opened 2024-09-30
Implement `concat_json` to join JSON strings given by strings column

ttnghia:
This adds `concat_json` to join multiple JSON strings given by strings column into one unified JSON string. Along with the concatenated data buffer, the result also contains a vector that indicates whether each row in the input is null or empty, and the delimiter character used for concatenation.


ttnghia:
build

ttnghia:
build

ttnghia:
build

src/main/cpp/src/json_utils.cu
```diff
@@ -0,0 +1,236 @@
...
+#include <thrust/iterator/zip_iterator.h>
+#include <thrust/transform.h>
+#include <thrust/tuple.h>
+#include <thrust/uninitialized_fill.h>
+
+namespace spark_rapids_jni {
+
+namespace detail {
+
+constexpr bool not_whitespace(cudf::char_utf8 ch) { return ch > ' '; }
```
revans2, line 40: white space for JSON is exactly `' ' || '\r' || '\n" || '\t'`

See https://www.json.org/json-en.html

src/main/cpp/src/json_utils.cu
```diff
@@ -0,0 +1,236 @@
...
+    [input = *d_input_ptr] __device__(cudf::size_type idx) -> thrust::tuple<bool, bool> {
+      if (input.is_null(idx)) { return {false, false}; }
+
+      auto const d_str = input.element<cudf::string_view>(idx);
+      auto itr         = d_str.data();
+      auto const end   = itr + d_str.size_bytes();
+
+      while (itr < end) {
+        cudf::char_utf8 ch   = 0;
+        auto const chr_width = cudf::strings::detail::to_char_utf8(itr, ch);
```
revans2, line 92: We don't need to parse it as UTF-8. We don't do it anywhere else in the JSON parsing code.

src/main/cpp/src/json_utils.cu
```diff
@@ -0,0 +1,236 @@
...
+
+      return {not_eol, not_eol};
+    });
+
+  auto constexpr num_levels  = 256;
+  auto constexpr lower_level = std::numeric_limits<char>::min();
+  auto constexpr upper_level = std::numeric_limits<char>::max();
+  auto const num_chars       = input.chars_size(stream);
+
+  rmm::device_uvector<uint32_t> d_histogram(num_levels, stream);
```
revans2, line 134: Have you tested the histogram code with a \n in it? When I did it the delimiter it returned was still \n and I don't know why. I think it would be better to write our own that is just a reduction into a 256-bit value that is a bitmask for which bytes showed up in the data. No need for a count because all we want is a boolean is it there or not value.

src/main/cpp/src/json_utils.cu
```diff
@@ -0,0 +1,236 @@
...
+    auto first_non_existing_char = static_cast<char>(first_zero_count_pos - zero_level);
+    while (!can_be_delimiter(first_non_existing_char)) {
+      first_zero_count_pos = find_first_zero_count_pos(first_zero_count_pos + 1, end);
+      if (first_zero_count_pos == end) { return {'\0', false}; }
+      first_non_existing_char = static_cast<char>(first_zero_count_pos - zero_level);
+    }
+    return {first_non_existing_char, true};
+  };
+
+  // Firstly, search from the `\n` character to the end of the histogram,
```
revans2, line 179: nit: if the issues with not using \n as the delimiter are solved I would rather have us just start at 0 instead of trying really hard to use \n as the delimiter.

ttnghia:
build

ttnghia:
build

ttnghia:
build

ttnghia:
build

revans2 (approved):
The performance got a lot better on my tests 24% better than the current baseline and all JSON related tests passed.

I need to track down what happened with some regexp tests, but they appear unrelated.

PR #2480, opened 2024-10-09
Merge branch-24.10 into main 

nvauto:
Change version to 24.10.0

Note: merge this PR with **Create a merge commit to merge**


NvTimLiu:
Mark the change as drafted

This change is for in advance review, 

I'll up merge the following changes if needed.

jlowe (approved):

NvTimLiu:
build

NvTimLiu (approved):

pxLi (approved):

jlowe:
Why was this merged before CI finished running?

pxLi:
> Why was this merged before CI finished running?

Sorry for the confusion, for merge-to-main change we will start a RC build pipeline internally to verify the change to save time and resources. 

This one should be marked as [skip ci], and then merge

pxLi:
> > This one should be marked as [skip ci], and then merge
> 
> This doesn't make sense to me -- we should not be merging things to main that have never been built. It sounds like we're merging to main and then doing an RC build (because we need to build with the proper commit hash on main, yes?), but then that means we're merging _before_ we know it builds. That's not what we should be doing.

The idea here is all commits to dev branch has already passed all the CIs, for main branch we only updated the release version, we do the RC build (which is a superset of all the build + test cases in pre-merge) to cover the entire CI checks for the release commit. If any issue, there should be a bug fix to the dev branch and go through the pre-merge CI as usual.

And sure, if this doesn't make sense to people. let's stop [skip ci] when merge to main cc @NvTimLiu thanks!

jlowe:
The problem is we're merging code to main without testing it first.  Sure, we test it _eventually_, but not until after it's been merged.  The main branch is supposed to be the most stable branch of all branches, yet we're not treating it that way.

pxLi:
> The problem is we're merging code to main without testing it first. Sure, we test it _eventually_, but not until after it's been merged. The main branch is supposed to be the most stable branch of all branches, yet we're not treating it that way.

Agree, we will always run the CI and make sure it passes before merging into main branch

PR #2483, opened 2024-10-09
Revert nvcomp

revans2:
This is based off of the work done by @gerashegalov, but applies the patch at build time.

I have not tested this with the code that tries to upmerge to a new version of CUDF. I still need to test that.

Also note that this is not intended to be something that coexists with developers trying to work on CUDF or spark-rapids-jni. That said I would like as much feedback on anything I missed as possible.


revans2:
build

revans2:
build

build/apply-patches
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Run a command in a Docker container with devtoolset
+
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
gerashegalov, line 23: nit: since we use git later  we can use  `git rev-parse --show-toplevel` to get the local repo dir,. and build paths from that instead of using ..  `realpath "$SCRIPT_DIR/.."` later

build/apply-patches
```diff
@@ -0,0 +1,39 @@
...
+
+PATCH_DIR=${PATCH_DIR:-$(realpath "$SCRIPT_DIR/../patches/")}
+
+CUDF_DIR=${CUDF_DIR:-$(realpath "$SCRIPT_DIR/../thirdparty/cudf/")}
+
+pushd "$CUDF_DIR"
```
gerashegalov, line 29: nit: if we handle patches via `git apply` we can avoid switching working dirs by using `git -C $CUDF_DIR`

revans2: I could do that with a patch too, but `git status` needs to be in the subdirectory/repo to work properly. `-C` does not work for `git status`.

gerashegalov: Maybe it is a bug in prior git versions 

```
~/gits/NVIDIA/spark-rapids-jni$ git --version
git version 2.34.1

~/gits/NVIDIA/spark-rapids-jni$ git -C thirdparty/cudf status -v --porcelain --untracked-files=no
 M print_env.sh
```

revans2: I would rather stick with what I have tested for now.

revans2:
I have decided not to try and fix the auto-upmerge of CUDF as it is going to be difficult to fix and this is intended to only be a fix for 24.10, at least until the regression in nvcomp can be diagnosed and fixed. Because this is temporary I think it is better to just upmerge manually.

revans2:
@gerashegalov @pxLi please take a look.

build/unapply-patches
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Run a command in a Docker container with devtoolset
+
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
gerashegalov, line 23: nit: the BASE_DIR changes from the apply script apply here as well

ci/submodule-sync.sh
```diff
@@ -70,24 +70,18 @@ echo "Test against ${cudf_sha}..."
 
 MVN="mvn -Dmaven.wagon.http.retryHandler.count=3 -B"
 set +e
-${MVN} verify ${MVN_MIRROR} \
+# Don't do a full build. Just try to update/build CUDF with no patches on top of it.
+${MVN} validate ${MVN_MIRROR} \
```
gerashegalov, line 74: We currently lump together cmake build generation and the actual build step. This is an additional change that would avoid this `~/gits/NVIDIA/spark-rapids-jni.reviews$ git diff pom.xml `

```Diff
diff --git a/pom.xml b/pom.xml
index dc55a89..e9d2efc 100644
--- a/pom.xml
+++ b/pom.xml
@@ -429,7 +429,7 @@
             <id>build-libcudf</id>
             <phase>validate</phase>
             <configuration>
-              <target xmlns:if="ant:if">
+              <target xmlns:if="ant:if" xmlns:unless="ant:unless">
                 <condition property="needConfigure">
                   <or>
                     <istrue value="${libcudf.build.configure}"/>
@@ -465,7 +465,8 @@
                 </exec>
                 <exec dir="${libcudf.build.path}"
                       failonerror="true"
-                      executable="cmake">
+                      executable="cmake"
+                      unless:true="${submodule.patch.skip}">
                   <arg value="--build"/>
                   <arg value="${libcudf.build.path}"/>
                   <arg value="--target"/>
@@ -482,6 +483,7 @@
             <id>build-libcudfjni</id>
             <phase>validate</phase>
             <configuration>
+              <skip>${submodule.patch.skip}</skip>
               <target>
                 <mkdir dir="${libcudfjni.build.path}"/>
                 <exec dir="${libcudfjni.build.path}"
@@ -517,6 +519,7 @@
             <id>build-sparkrapidsjni</id>
             <phase>validate</phase>
             <configuration>
+              <skip>${submodule.patch.skip}</skip>
               <target>
                 <mkdir dir="${native.build.path}"/>
                 <exec dir="${native.build.path}"
@@ -555,6 +558,7 @@
             <id>build-info</id>
             <phase>generate-resources</phase>
             <configuration>
+              <skip>${submodule.patch.skip}</skip>
```

gerashegalov, line 74: I think we care only about one execution in the `validate` phase. So we could

```suggestion
${MVN} antrun:run@build-libcudf ${MVN_MIRROR} \
```

pxLi (approved):
LGTM for the patch parts. 

Let's try address comments for submodule-sync in 24.12, thanks

PR #2502, opened 2024-10-14
Make it so applying and removing patches are repeatable without errors

revans2:
There are a number of issues with dealing with patches for the CUDF repo. This fixes some of them. It does not fix `submodule update --init` as that happens externally. It also does not fix other parts of upmerging a branch unless you have unapplied the patches before hand.  Sadly it also does not provide any good way to do development work on CUDF using the submodule as your development environment.

What this does do it make it so that if you apply patches to CUDF and try to build again without removing those patches it should work.

### DIFF
--- a/build/apply-patches
+++ b/build/apply-patches
@@ -51,9 +51,11 @@ fi
 
 CHANGED_FILES=$(git status --porcelain --untracked-files=no)
 
-if [ \( -s "$FULLY_PATCHED_FILE" \) -a  \( -n "$CHANGED_FILES" \) ] ; then
-  echo "Patches appear to have been applied already"
-  exit 0
+if [ \( -s "$FULLY_PATCHED_FILE" \) -a  \( -n "$CHANGED_FILES" \) ] ; then 
+  if git apply -R --check "$FULLY_PATCHED_FILE" ; then
+    echo "Patches appear to have been applied already"
+    exit 0
+  fi
 fi
 
 if [ -n "$CHANGED_FILES" ] ; then
@@ -64,7 +66,7 @@ fi
 
 find "$PATCH_DIR" -maxdepth 1 -type f -print0 | sort -zV | while IFS= read -r -d '' file; do
   echo "patching with: $file"
-  patch --no-backup-if-mismatch -f -t --reject-file=- -p1 -i "$file"
+  git apply -v "$file"
 done
 
 git diff > "$FULLY_PATCHED_FILE"
--- a/build/apply-patches
+++ b/build/apply-patches
@@ -16,8 +16,6 @@
 # limitations under the License.
 #
 
-# Run a command in a Docker container with devtoolset
-
 set -e
 
 BASE_DIR=$( git rev-parse --show-toplevel )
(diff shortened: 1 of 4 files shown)


revans2:
build

jlowe (commented):
What's here looks OK.

What about the dangerous and surprising behavior of specifying submodule.patch.skip=true as discussed in https://github.com/NVIDIA/spark-rapids-jni/pull/2497#discussion_r1797408136?  If this is an artifact of how the validate build works, maybe we should change the name of the property to something less "inviting" for developers to try, as this won't do what they think it does.

revans2:
> What's here looks OK.
> 
> What about the dangerous and surprising behavior of specifying submodule.patch.skip=true as discussed in [#2497 (comment)](https://github.com/NVIDIA/spark-rapids-jni/pull/2497#discussion_r1797408136)? If this is an artifact of how the validate build works, maybe we should change the name of the property to something less "inviting" for developers to try, as this won't do what they think it does.

That is a good point. At a minimum I need to document it better. Thinking about it I am trying to understand the use case fro it. Originally I put it in as a way to skip the apply/unapply steps. But then I used it as a part of the CI process to auto upmerge CUDF, which made @gerashegalov want to lock it down so we don't accidentally build the plugin without the patches being installed. But I don't see that being the case any more.  Perhaps I can move it back to what it was originally intended to be and document it better to let people know that there is a foot/gun here no matter what it does.

revans2:
build

jlowe (dismissed):
Minor typo, lgtm.  Note that this will make the submodule sync take a bit longer because it will build everything only to rebuild it, but ccache may be able to avoid a lot of the rebuild.

I'm personally OK if we decide to keep the build-skipping as long as we update the name of the config to reflect that.  submodule.patch.skip implies it's only going to skip patching submodules, not also skip all native builds (including those outside of the submodule).

build/unapply-patches
```diff
@@ -16,29 +16,62 @@
...
+if [ -z "$PATCH_FILES" ] ; then
+  echo "No patches to remove"
+  exit 0
 fi
 
-# Check for modifications
-if [ -n "$(git status --porcelain --untracked-files=no)" ] ; then
+CHANGED_FILES=$(git status --porcelain --untracked-files=no)
+
+if [ \( -s "$FULLY_PATCHED_FILE" \) -a  \( -n "$CHANGED_FILES" \) ] ; then
```
gerashegalov, line 54: I would prefer a direct test of patch applicability via `git apply --check` for both forward and `--reverse` . If it is neither of the states we are in, we can ask/warn the developer to commit changes and regenerate the patch if it is intentional

build/apply-patches
```diff
@@ -26,14 +24,49 @@ PATCH_DIR=${PATCH_DIR:-$(realpath "$BASE_DIR/patches/")}
...
+# 2) Rebuild the repo without having to clean and start over
+# 3) upmerge to a new version of the plugin including updating the cudf submodule
+#
+# Building from scratch is simple. We want clean to unapply any patches and 
+# build to apply them. But if we want to rebuild without a clean we need to know what
+# state the CUDF repo is in. Did we apply patches to it or not. The fastest way to do this
+# is to save some state files about what happened. But a user could mess with CUDF directly
+# so we want to have ways to double check that they are indeed correct.
+
+FULLY_PATCHED_FILE="$CUDF_DIR/spark-rapids-jni.patch"
```
gerashegalov, line 41: there may technically be a patch that applies to non-cudf portion of spark-rapids, maybe should go to the top-level dir?

jlowe: Curious what's the use-case for patching outside of the submodule?  I thought we would checkin the change directly rather than a patch of that change.

gerashegalov: We only have a sole use case with the submodule patching right now. No doubt direct modifications of the source tree outside the module is preferrable. In terms of hypotheticals, there could be another submodule or some native dependency that  we want consume bypassing cudf which needs patching.

revans2: Right now all patches apply only to CUDF. I don't want to try and make that more generic right now. If we do hit that situation I would rather have sub-directories under patches for each third party sub directory, and then apply/remove patches for each subdirectory separately.

revans2:
build

ttnghia (approved):

jlowe (approved):

build/apply-patches
gerashegalov: ok, separate subdirs work too

revans2:
@gerashegalov I added in the direct test, and you were also right that git apply is much cleaner than patch.  Please take another look

gerashegalov (approved):
LGTM

Issue #2468, opened 2024-10-04
[FEA] Implement a function to convert a list of strings columns into a structs column with desired schema


ttnghia:
In `GpuJsonToStruct`, after reading the input JSON using `cudf::read_json`, we need to convert the output table (consisting of many strings columns) into the final structs column with the desired schema. Currently, that step is done in Scala code, which can have a lot of overhead of unnecessarily copying the intermediate columns into the final output column. We can do that better in native C++ code, performing the conversion if needed while just moving the intermediate columns into the output column without any copying.

PR #2510, opened 2024-10-17
Implement `from_json_to_structs`

ttnghia:
This implements the function `from_json_to_structs`, which performs parsing a strings column into a structs column with a given data schema. This is the case of running `from_json` SQL function in Spark with struct schema.

Closes https://github.com/NVIDIA/spark-rapids-jni/issues/2468.

Contribute to https://github.com/NVIDIA/spark-rapids/issues/11560.

Depends on:
 * https://github.com/NVIDIA/spark-rapids-jni/pull/2607


src/main/cpp/src/json_utils.cu
```diff
@@ -276,7 +280,7 @@ std::pair<std::unique_ptr<cudf::column>, rmm::device_uvector<bool>> cast_strings
 
   auto output = cudf::make_fixed_width_column(
     cudf::data_type{cudf::type_id::BOOL8}, string_count, cudf::mask_state::UNALLOCATED, stream, mr);
-  auto validity = rmm::device_uvector<bool>(string_count, stream);  // intentionally not use `mr`
+  auto validity = rmm::device_uvector<bool>(string_count, stream);
```
hyperbolic2346, line 283: do we need to pass `mr` now?

ttnghia: No, I just remove the comment as I thought it is not necessary. Maybe will replace it with other comments.

ttnghia:
CC @karthikeyann.

src/main/cpp/src/from_json_to_structs.cu
```diff
@@ -0,0 +1,922 @@
...
+                             /*strip*/ false,
+                             stream,
+                             mr);
+  }
+
+  auto const out_offsets =
+    cudf::detail::offsetalator_factory::make_input_iterator(offsets_column->view());
+  auto chars_data = rmm::device_uvector<char>(bytes, stream, mr);
+
+  // Since the strings store decimal numbers, they should be very short.
```
revans2, line 464: nit: Technically with a DECIMAL_128 it is possible to have 158-bytes per string, but yes, this is generally true.

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -166,80 +166,65 @@ public static ColumnVector extractRawMapFromJsonString(ColumnView input, JSONOpt
...
-   *
-   * @deprecated This method is deprecated since it does not have parameters to control various
-   * JSON reader behaviors.
+   * Parse a JSON string into a struct column following by the given data schema.
    *
    * @param input The input strings column in which each row specifies a json object
-   * @return A map column (i.e., a column of type {@code List<Struct<String,String>>}) in
-   *         which the key-value pairs are extracted directly from the input json strings
+   * @param schema The schema of the output struct column
+   * @param opts The options for parsing JSON strings
```
revans2, line 173: nit: Could we indicate that some of the options will be ignored and are hard coded to what Spark needs?

revans2 (dismissed):

ttnghia:
build

src/main/cpp/src/from_json_to_structs.cu
```diff
@@ -0,0 +1,1100 @@
...
+
+  auto const input_sv = cudf::strings_column_view{input};
+  auto const input_offsets_it =
+    cudf::detail::offsetalator_factory::make_input_iterator(input_sv.offsets());
+  auto const d_input_ptr = cudf::column_device_view::create(input, stream);
+  auto const is_valid_it = cudf::detail::make_validity_iterator<true>(*d_input_ptr);
+
+  // We need to nullify the invalid string rows.
+  // Technically, we should just mask out these rows as invalid and ignore them.
+  // However, `spark_rapids_jni::string_to_integer` cannot handle these non-empty null rows,
```
hyperbolic2346, line 216: Should we make an issue to fix that?

hyperbolic2346 (dismissed):
Question about a possible issue, but otherwise looks good.

src/main/cpp/src/from_json_to_structs.cu
ttnghia: Oh wait, maybe I was wrong. I'll check this again.

ttnghia: Indeed I've seen a bug in `string_to_decimal_kernel` that cannot handle non-empty null rows, not `string_to_integer`. I'll check for `string_to_integer` more carefully tomorrow. The bug in `string_to_decimal_kernel` also needs to be fixed so we can leverage non-empty null handling in this PR.

hyperbolic2346: https://github.com/NVIDIA/spark-rapids-jni/pull/2607

ttnghia:
build

src/main/java/com/nvidia/spark/rapids/jni/JSONUtils.java
```diff
@@ -166,80 +166,78 @@ public static ColumnVector extractRawMapFromJsonString(ColumnView input, JSONOpt
...
+   * @param isUSLocale Whether the current local is US locale, used when converting strings to
+   *        decimal types
+   * @return A struct column in which each row is parsed from the corresponding json string
    */
```
revans2, line 187: I assume that this version of extractRawMapFromJsonString was just not used so it is okay for us to delete it.

revans2 (approved):
Looks god to me, but I am not C++/CUDA expert so I would love it if someone with more experience there could also look at this.

PR #2532, opened 2024-10-24
Introduce kudo serialization format.

liurenjie1024:
## Design

This pr introduces kudo serialization format, which significantly improved serialization/deserialization time compared to jcudf  serialization format. Kudo is optimized for columnar batch serialization used during spark shuffle, and the improvements are based on two observations:

1. During spark shuffle we have a lot of contexts provided by runtime such as table schema, that means we could simplify headers. In kudo's header, we only contains necessary fields such as offset, number of rows, data lens, and one byte for each column to indicate whether it has validity buffer.
2. Gpu's columnar batch is typically much larger than cpu's vectorized execution engine, that means we almost always need to do batch concatanation in shuffle read time. When serializing a part of a larger columnar batch, unlike jcudf, which calculates exact validity buffer and offset buffer, we only record offset and number of rows in header, and copy necessay bytes, since the exact buffer could be restored at read time when do concating. This saves a lot of compuation when doing serialization.

## Performance

We have observed 30%-4000% serialization time improvement, up to 200% deserialization time improvement, and similar concat batching performance.

## Notice

This pr is a draft mainly for skimming through the whole picture of codes, and we could work out a plan for splitting them into smaller pr so that it would be easier to review.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/TableUtils.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/TableUtils.java
@@ -1,7 +1,5 @@
 package com.nvidia.spark.rapids.jni;
 
-import ai.rapids.cudf.*;
-
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


src/main/java/com/nvidia/spark/rapids/jni/kudo/SerializedTableHeader.java
```diff
@@ -0,0 +1,155 @@
...
+ * isInitialized will return true if the metadata was read correctly from the stream.
+ * It will return false if an EOF was encountered at the beginning indicating that
+ * there was no data to be read.
+ */
+public final class SerializedTableHeader {
+    /**
+     * Magic number "KUDO" in ASCII.
+     */
+    private static final int SER_FORMAT_MAGIC_NUMBER = 0x4B55444F;
+    private static final short VERSION_NUMBER = 0x0001;
```
jlowe, line 20: JCudfSerialization was built as a generic serializer, where the software writing and reading could change.  This is specific to shuffle.  I don't think we need a version number, and arguably can embed that concept the magic number if it's ever needed (and I cannot see why it would be).

src/main/java/com/nvidia/spark/rapids/jni/kudo/SerializedTableHeader.java
```diff
@@ -0,0 +1,155 @@
...
+
+    // Useful for reducing calculations in writing.
+    private long offset;
+    private long numRows;
```
jlowe, line 24: It's impossible to have more than a 31-bit row count, this should just be an int.

src/main/java/com/nvidia/spark/rapids/jni/kudo/SerializedTableHeader.java
```diff
@@ -0,0 +1,155 @@
...
+    private long validityBufferLen;
+    private long offsetBufferLen;
+    private long totalDataLen;
```
jlowe, line 27: These should all be ints.  It's not possible to serialize a single object that is > 2GB in Spark.

src/main/java/com/nvidia/spark/rapids/jni/kudo/SerializedTableHeader.java
```diff
@@ -0,0 +1,155 @@
...
+    // This is used to indicate the validity buffer for the columns.
+    // 1 means that this column has validity data, 0 means it does not.
+    private byte[] hasValidityBuffer;
```
jlowe, line 30: This should be a bit mask, not byte-per-column.  We never have to bit shift the entire hasValidityBuffer mask or anything complex like that, just single-bit test this which is trivial to compute, both on the write and read sides, and this saves a lot of space when we're dealing with tons of columns and tons of partitions across tons of tasks.

jlowe (commented):
I added some initial comments from skimming through the code.  One way to break this up into smaller PRs:
- Have a "generic utilities" PR that includes Pair, SlicedTable, TableUtils, etc.
- Have a "schema utilities" PR that includes the visitor framework and related utilities (i.e.: the stuff under the schema package)
- Write-side PR that includes the parts of the Kudo serializer related to writing partitions of a table to a stream (and related tests)
- Read-side PR that includes the remaining parts of the Kudo serializer related to merging partitions (and related tests)

PR #2534, opened 2024-10-24
Add `HiveHash` support for nested types

ustcfy:
This PR is based on https://github.com/wjxiz1992/spark-rapids-jni/pull/9.

This PR adds support for Hive hash with `struct` and `list` types.


firestarman:
build

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,228 @@ class hive_device_row_hasher {
...
+        : column(col), child_idx(0), cur_hash(HIVE_INIT_HASH)
+      {
+      }
+
+      __device__ void update_cur_hash(hive_hash_value_t child_hash)
+      {
+        cur_hash = cur_hash * HIVE_HASH_FACTOR + child_hash;
+      }
+
+      __device__ hive_hash_value_t get_hash() { return cur_hash; }
```
firestarman, line 246: Personally I will keep the `this` pointer when accessing a member field.

firestarman:
Hi @ttnghia, could you help take a look ? Thx in advance.

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,228 @@ class hive_device_row_hasher {
...
+     * next child column to process, and current hash value for the column.
+     *
+     * @param _column The current column being processed.
+     * @param _child_idx The index of the next child column to process.
+     * @param _cur_hash The current hash value for the column.
+     *
+     * @note The default constructor is deleted to prevent uninitialized usage.
+     *
+     * @constructor
+     * @param col The column device view to be processed.
```
ttnghia, line 224:
```suggestion
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -239,15 +483,20 @@ std::unique_ptr<cudf::column> hive_hash(cudf::table_view const& input,
   // Return early if there's nothing to hash
   if (input.num_columns() == 0 || input.num_rows() == 0) { return output; }
 
+  // Nested depth cannot exceed 8
+  constexpr int max_nested_depth = 8;
```
ttnghia, line 487: Is this enough for general purposes? If not, we may want to consider another approach that preprocess the input so we can iterate the nested columns without using a stack.

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,228 @@ class hive_device_row_hasher {
...
     __device__ hive_hash_value_t operator()(cudf::column_device_view const& col,
                                             cudf::size_type row_index) const noexcept
     {
-      CUDF_UNREACHABLE("Nested type is not supported");
+      cudf::column_device_view curr_col = col.slice(row_index, 1);
+      // The default constructor of `col_stack_element` is deleted, so it can not allocate an array
+      // of `col_stack_element` directly.
+      // Instead leverage the byte array to create the col_stack_element array.
+      uint8_t stack_wrapper[MAX_NESTED_DEPTH * sizeof(col_stack_element)];
+      col_stack_element_ptr col_stack = reinterpret_cast<col_stack_element_ptr>(stack_wrapper);
```
ttnghia, line 394: Oh no, never ever do this in general. Any class can have some alignment requirement (such as on a 64 bytes boundaries) while `uint8_t` array may not be aligned. If you want to declare a stack, do this directly:
```
col_stack_element stack[MAX_NESTED_DEPTH];
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,228 @@ class hive_device_row_hasher {
...
+     * @param _cur_hash The current hash value for the column.
+     *
+     * @note The default constructor is deleted to prevent uninitialized usage.
+     *
+     * @constructor
+     * @param col The column device view to be processed.
+     */
+    class col_stack_element {
+     private:
+      cudf::column_device_view _column;  // current column
```
ttnghia, line 228: Instead of storing a column in each stack element, you can first flatten the input nested column into a table (array of columns), then here just store the index of the column in that array. See `cudf::experimental::row::lexicographic::preprocessed_table` for example of such flatten table.

Some preprocessing is needed so we can retrieve the index of the children columns in the array of flatten table.

ttnghia: Check out `cudf/cpp/src/table/row_operators.cu` (`cudf::experimental::row::lexicographic::preprocessed_table::create`) and `cudf/cpp/src/structs/utilities.cpp` (`cudf::structs::detail::flatten_nested_columns`) for example code of flattening table.

src/main/cpp/src/hive_hash.cu
ustcfy: This is because the default constructor of `cudf::column_device_view` is deleted.
I cannot just define `stack[MAX_NESTED_DEPTH]` without initializing it.  :(

src/main/cpp/src/hive_hash.cu
ustcfy: But how should list columns be handled without using a stack? 👀 
At least using `cudf::detail::lists_column_device_view(curr_col).get_sliced_child()` to flatten the list is not feasible. 

As illustrated in the example I provided in the comments, nulls at any nested level affect the result of `HiveHash`, and this information is lost during the flattening process.

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,219 @@ class hive_device_row_hasher {
...
+          } else {
+            // Push the next child column into the stack
+            col_stack[stack_size++] =
+              col_stack_element(cudf::detail::structs_column_device_view(curr_col).get_sliced_child(
+                element.get_and_inc_child_idx()));
+          }
+        } else if (curr_col.type().id() == cudf::type_id::LIST) {
+          // lists_column_device_view has a different interface from structs_column_device_view
+          curr_col = cudf::detail::lists_column_device_view(curr_col).get_sliced_child();
+          if (element.cur_child_idx() == curr_col.size()) {
```
ttnghia, line 409: `curr_col.size()` is the number of rows, not the child index.

ttnghia: Indeed we only need to check if `element.curr_child_idx() == 1` since a lists column has just one child.

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,219 @@ class hive_device_row_hasher {
...
+            if (--stack_size > 0) { col_stack[stack_size - 1].update_cur_hash(element.get_hash()); }
+          } else {
+            col_stack[stack_size++] =
+              col_stack_element(curr_col.slice(element.get_and_inc_child_idx(), 1));
```
ttnghia, line 413: Why `curr_col.slice`? This is to get the slice of rows, not slice of children.
```suggestion
              col_stack_element(curr_col);
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,219 @@ class hive_device_row_hasher {
...
+            if (--stack_size > 0) { col_stack[stack_size - 1].update_cur_hash(element.get_hash()); }
+          } else {
+            // Push the next child column into the stack
+            col_stack[stack_size++] =
+              col_stack_element(cudf::detail::structs_column_device_view(curr_col).get_sliced_child(
+                element.get_and_inc_child_idx()));
+          }
+        } else if (curr_col.type().id() == cudf::type_id::LIST) {
+          // lists_column_device_view has a different interface from structs_column_device_view
+          curr_col = cudf::detail::lists_column_device_view(curr_col).get_sliced_child();
```
ttnghia, line 408:
```suggestion
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,219 @@ class hive_device_row_hasher {
...
+     * @return The computed hive hash value.
+     *
+     * @note This function is only enabled for nested column types.
+     */
     template <typename T, CUDF_ENABLE_IF(cudf::is_nested<T>())>
     __device__ hive_hash_value_t operator()(cudf::column_device_view const& col,
                                             cudf::size_type row_index) const noexcept
     {
-      CUDF_UNREACHABLE("Nested type is not supported");
+      cudf::column_device_view curr_col = col.slice(row_index, 1);
```
ttnghia, line 380: I don't think we will need to slice the input column. When we call the hasher, we just pass `row_index` to it, then it will automatically pull the necessary rows at the specified index.

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,219 @@ class hive_device_row_hasher {
...
+          if (element.cur_child_idx() == curr_col.size()) {
+            if (--stack_size > 0) { col_stack[stack_size - 1].update_cur_hash(element.get_hash()); }
+          } else {
+            col_stack[stack_size++] =
+              col_stack_element(curr_col.slice(element.get_and_inc_child_idx(), 1));
+          }
+        } else {
+          // There is only one element in the column for primitive types
+          auto hash = cudf::type_dispatcher<cudf::experimental::dispatch_void_if_nested>(
+            curr_col.type(), this->hash_functor, curr_col, 0);
```
ttnghia, line 418:
```suggestion
          auto hash = cudf::type_dispatcher<cudf::experimental::dispatch_void_if_nested>(
            curr_col.type(), this->hash_functor, curr_col, row_index);
```

src/main/cpp/src/hive_hash.cu
ustcfy: But the concept of children in a `list_column` is different from that in a `struct_column`. 
For any `list_column`, `num_children()` is always 2: one is `offsets col`, and the other is `child col`.

src/main/cpp/src/hive_hash.cu
ustcfy: I want to get the element from the `child col` at the `child_idx` row.

ustcfy: I got the slice of children at line 408.

src/main/cpp/src/hive_hash.cu
ustcfy: Here, `curr_col` has only one element, so the index should be 0.

src/main/cpp/src/hive_hash.cu
ttnghia: For lists column, the offsets child are not used for hashing thus it is considered to have only one child.

src/main/cpp/src/hive_hash.cu
ustcfy: I need to ensure that all columns in the stack contain only one element (whether nested or non-nested), so I think the slice is necessary.

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,219 @@ class hive_device_row_hasher {
...
+     * This functor produces the same result as "HiveHash" in Spark for nested types.
+     * The pseudocode of Spark's HiveHash function is as follows:
+     *
+     * hive_hash_value_t hive_hash(NestedType element) {
+     *    hive_hash_value_t hash = HIVE_INIT_HASH;
+     *    for (int i = 0; i < element.num_child(); i++) {
+     *        hash = hash * HIVE_HASH_FACTOR + hive_hash(element.get_child(i));
+     *    }
+     *    return hash;
+     * }
```
ttnghia, line 260: This algorithm only applies to struct and non-nested types. When we have lists, which can have many elements, what will we do with them? And in case lists of lists of....of lists, thing will be very complicated.

src/main/cpp/src/hive_hash.cu
ttnghia: Again, `size()` here is the number of rows, nothing related to number of children.

src/main/cpp/src/hive_hash.cu
ustcfy: Here is the [Spark Hive Hash Implementation](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/hash.scala#L920), where the computation logic for `lists` and `structs` is similar.

res-life: > Ref: Here is cudf code for hashing nested elements: https://github.com/rapidsai/cudf/blob/d67d017e8c3176ae214165869d5bb4b02c4207a0/cpp/include/cudf/table/experimental/row_operators.cuh#L1926-L1937

I checked this, seems it's not applicable to hive hash.
e.g.: hive hash for: [[1,2], [3,4,5]] ,
It expects:
`h1 = 31 * hash(1) + hash(2); h2 = 31 * (31 * hash(3) + hash(4)) + hash(5);  final_hash = 31 * h1 + h2`
Above code in `row_operators.cuh` has better performance, but it's logic is:
First update hash according sub list sizes: [2, 3], then combine for each item in each sub list.
For hive hash, [[1,2], [3,4,5]] and [[1,2,3], [4,5]] have diff hash values. We need to use stack to record the intermediate hash of sub list.

@ustcfy  inspired by the approach `row_operators` Nghia pointed,  it's not optimal to put all the ternimal item into stack. E.g.: put 1, 2, 3, 4, 5 to stack, then compute the hash value.
We can detect if current element is list of basic types, if it is, then use a `for` iterator to compute hash directly:
```c++
if (curr_list_element.child() is not nested type) {
  val hash = SEED;
  for (int i = 0; i < size; ++i) {
    // compute for basic type, then accumulate
    hash = 31 * hash + hash_dispatch(curr_element.child(i));
  }
}
```

ustcfy: > @ustcfy inspired by the approach `row_operators` Nghia pointed, it's not optimal to put all the ternimal item into stack. E.g.: put 1, 2, 3, 4, 5 to stack, then compute the hash value. We can detect if current element is list of basic types, if it is, then use a `for` iterator to compute hash directly:
> 
> ```c++
> if (curr_list_element.child() is not nested type) {
>   val hash = SEED;
>   for (int i = 0; i < size; ++i) {
>     // compute for basic type, then accumulate
>     hash = 31 * hash + hash_dispatch(curr_list_element.slice(i));
>   }
> }
> ```
Plan to add these changes in the next release.

src/main/cpp/src/hive_hash.cu
ustcfy: This approach may require significant changes to the existing framework, as the calculations for these 5 columns are not independent.

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
       return this->hash_functor.template operator()<T>(col, row_index);
     }
 
+    /**
+     * @brief A structure to keep track of the computation for nested types.
+     */
+    struct col_stack_frame {
+     private:
+      cudf::column_device_view _element;  // the column has only one row
```
ttnghia, line 216: The comment is wrong, since when we push a (nested) child column of a lists column then it may have many rows. In addition, `element` should mean the row, so we actually should not rename this in the last time.
```suggestion
      cudf::column_device_view _column;
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
...
+      hive_hash_value_t _cur_hash;        // current hash value of the column
+      int _idx_to_process;                // the index of child or element to process next
```
ttnghia, line 218: :+1: 
```suggestion
      int _idx_to_process;                // the index of child or element to process next
```

ttnghia, line 218: :+1:

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
...
+        _cur_hash = _cur_hash * HIVE_HASH_FACTOR + hash;
+      }
+
+      __device__ hive_hash_value_t get_hash() { return _cur_hash; }
+
+      __device__ int get_and_inc_idx_to_process() { return _idx_to_process++; }
+
+      __device__ int get_idx_to_process() { return _idx_to_process; }
+
+      __device__ cudf::column_device_view get_element() { return _element; }
```
ttnghia, line 240:
```suggestion
      __device__ cudf::column_device_view get_column() { return _column; }
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
...
+     * This functor produces the same result as "HiveHash" in Spark for structs and lists.
+     * The pseudocode of Spark's HiveHash function for structs and lists is as follows:
+     *
+     * hive_hash_value_t hive_hash(NestedType element) {
+     *    hive_hash_value_t hash = HIVE_INIT_HASH;
+     *    for (int i = 0; i < element.num_child(); i++) {
+     *        hash = hash * HIVE_HASH_FACTOR + hive_hash(element.get_child(i));
+     *    }
+     *    return hash;
+     * }
```
ttnghia, line 257:
```suggestion
     * The pseudocode of Spark's HiveHash function for structs is as follows:
     *
     * hive_hash_value_t hive_hash(NestedType element) {
     *    hive_hash_value_t hash = HIVE_INIT_HASH;
     *    for (int i = 0; i < element.num_child(); i++) {
     *        hash = hash * HIVE_HASH_FACTOR + hive_hash(element.get_child(i));
     *    }
     *    return hash;
     * }
     *
     * In the cases of lists, the hash value is computed by a similar way but we iterate through the list 
     * elements instead of through the child elements.
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
...
+     *
+     * 1. Initialize the stack and push the root column into the stack.
+     * 2. While the stack is not empty:
+     *    a. Get the top element of the stack. Don't pop it until it is processed.
+     *    b. If the column is a nested column:
+     *        i.  If all child columns are processed, pop the element and update `cur_hash` of its
+     *            parent column.
+     *        ii. Otherwise, push the next child column into the stack.
+     *    c. If the column is a primitive column, compute the hash value, pop the element,
+     *       and update `cur_hash` of its parent column.
```
ttnghia, line 273:
```suggestion
     *    b. If the column is a structs column:
     *        i.  If all child columns are processed, pop the element and update `cur_hash` of its
     *            parent column.
     *        ii. Otherwise, push the next child column into the stack.
     *    c. If the column is a lists column, process it by a similar way as structs column but iterating through the list elements instead of child columns' elements.
     *    d. If the column is a primitive column, compute the hash value, pop the element,
     *       and update `cur_hash` of its parent column.
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
...
+     * null_mask: 101
+     *
+     * Int level i:
+     * |Index|Int |
+     * |-----|----|
+     * |0    |1   |
+     * |1    |0   |
+     * |2    |2   |
+     * |3    |null|
+     * length: 4
```
ttnghia, line 330:
```suggestion
     * Int level i:
     * |Index|Int |
     * |-----|----|
     * |0    |1   |
     * |1    |0   |
     * |2    |2   |
     * |3    |null|
     * length: 4
     * null_mask: 0111
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
...
+      __device__ hive_hash_value_t get_hash() { return _cur_hash; }
+
+      __device__ int get_and_inc_idx_to_process() { return _idx_to_process++; }
+
+      __device__ int get_idx_to_process() { return _idx_to_process; }
+
+      __device__ cudf::column_device_view get_element() { return _element; }
+    };
+
+    typedef col_stack_frame* col_stack_frame_ptr;
```
ttnghia, line 243: This is used only once, so technically we don't have to type alias it.
```suggestion
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
...
     __device__ hive_hash_value_t operator()(cudf::column_device_view const& col,
                                             cudf::size_type row_index) const noexcept
     {
-      CUDF_UNREACHABLE("Nested type is not supported");
+      cudf::column_device_view curr_col = col.slice(row_index, 1);
+      // The default constructor of `col_stack_frame` is deleted, so it can not allocate an array
+      // of `col_stack_frame` directly.
+      // Instead leverage the byte array to create the col_stack_frame array.
+      alignas(col_stack_frame) char stack_wrapper[sizeof(col_stack_frame) * MAX_NESTED_DEPTH];
+      col_stack_frame_ptr col_stack = reinterpret_cast<col_stack_frame_ptr>(stack_wrapper);
```
ttnghia, line 370:
```suggestion
      auto col_stack = reinterpret_cast<col_stack_frame*>(stack_wrapper);
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +208,222 @@ class hive_device_row_hasher {
...
+        // - For lists, it is when all elements in the list are processed.
+        // - For primitive types, it is when the hash value is computed.
+        if (curr_col.type().id() == cudf::type_id::STRUCT) {
+          if (top.get_idx_to_process() == curr_col.num_child_columns()) {
+            if (--stack_size > 0) { col_stack[stack_size - 1].update_cur_hash(top.get_hash()); }
+          } else {
+            // Push the next child into the stack
+            col_stack[stack_size++] =
+              col_stack_frame(cudf::detail::structs_column_device_view(curr_col).get_sliced_child(
+                top.get_and_inc_idx_to_process()));
```
ttnghia, line 389: This can be optimized further, in a similar way as lists column: 
```
while(child is primitive type) {
 // update hash directly without pushing child onto stack
 // move to next child
}
```

src/main/cpp/src/hive_hash.cu
ustcfy: For the `list column`, I only push one element of `child_col` each time.

ttnghia: No, for lists column with nested child, you push that unique child column only once.

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +210,227 @@ class hive_device_row_hasher {
       return this->hash_functor.template operator()<T>(col, row_index);
     }
 
+    /**
+     * @brief A structure to keep track of the computation for nested types.
+     */
+    struct col_stack_frame {
+     private:
+      cudf::column_device_view _column;  // the column has only one row
```
ttnghia, line 218: Again, this is still incorrect. The column has either:
 * One row, if this is the root column or nested under children of a structs column, or
 * Many rows which are the elements of a list if this is nested under child of a lists column.

ttnghia:
```suggestion
      // The column has either one row (if root column or nested under structs column),
      // or has rows that are elements of a list (if it is child of a lists column).
      cudf::column_device_view _column;
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -210,11 +210,227 @@ class hive_device_row_hasher {
...
+      hive_hash_value_t _cur_hash;       // current hash value of the column
+      int _idx_to_process;               // the index of child or element to process next
+
+     public:
+      __device__ col_stack_frame() =
+        delete;  // Because the default constructor of `cudf::column_device_view` is deleted
+
+      __device__ col_stack_frame(cudf::column_device_view col)
+        : _column(col), _idx_to_process(0), _cur_hash(HIVE_INIT_HASH)
```
ttnghia, line 227:
```suggestion
        : _column(std::move(col)), _idx_to_process(0), _cur_hash(HIVE_INIT_HASH)
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -323,14 +327,14 @@ class hive_device_row_hasher {
...
-     * |Index|Int |
+     * |Index|int |
      * |-----|----|
      * |0    |1   |
      * |1    |0   |
      * |2    |2   |
      * |3    |null|
      * length: 4
-     * null_mask: 0111
+     * null_mask: 1110
```
ttnghia, line 337: Bit mask is read from right to left. The lowest bits are valid.
```suggestion
* null_mask: 0111
```

src/main/cpp/src/hive_hash.cu
```diff
@@ -224,6 +440,40 @@ class hive_device_row_hasher {
...
   cudf::table_device_view const _table;
 };
+
+void check_nested_depth(cudf::table_view const& input)
+{
+  using column_checker_fn_t = std::function<int(cudf::column_view const&)>;
+
+  column_checker_fn_t get_nested_depth = [&](cudf::column_view const& col) {
+    if (col.type().id() == cudf::type_id::LIST) {
+      auto const& child_col = cudf::lists_column_view(col).child();
```
ttnghia, line 450: `child()` return `rvalue` (temporary) thus do not bind a reference to it.
```suggestion
      auto const child_col = cudf::lists_column_view(col).child();
```

src/main/cpp/src/hive_hash.cu
ustcfy: This column really only has one row. (

ttnghia: Ah I see, sorry about that: I missed one place in the code.

ttnghia (approved):
Great work!

ustcfy:
build

res-life:
build

PR #2541, opened 2024-10-28
fix max bytes dealloc bug

zpuller:
This is a follow up to https://github.com/NVIDIA/spark-rapids-jni/pull/2392 to fix how we handle the deallocation path.

I verified that the deallocate behavior looks correct now, that is, if a job spills and reallocs, the metric never reports a watermark above the memory limit. The unit test change also verifies this case.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -1780,6 +1780,7 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     auto const thread = threads.find(tid);
     if (thread != threads.end()) {
       log_status("DEALLOC", tid, thread->second.task_id, thread->second.state);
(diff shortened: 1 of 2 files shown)


src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1802,7 +1805,6 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
             if (is_for_cpu == t_state.is_cpu_alloc) {
               transition(t_state, thread_state::THREAD_ALLOC_FREE);
             }
-            if (!is_for_cpu) { t_state.gpu_memory_allocated_bytes -= num_bytes; }
```
zpuller, line 1805: I didn't notice in the original change that this code path is only running on *other* threads which is incorrect for the tracking behavior (see line 1787)

revans2 (approved):

zpuller:
build

PR #2559, opened 2024-11-01
Introduce kudo writer.

liurenjie1024:
This pr is part of https://github.com/NVIDIA/spark-rapids-jni/pull/2532 , which introduces writer part of kudo format.


jlowe (commented):
There should be some unit tests added (e.g.: serializing row-count only, serializing a simple table of a single int column, serializing a list, serializing a struct, etc.) that examine the bytes serialized to ensure it matches the expected format.

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoSerializer.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+    return paddedBytes;
+  }
+
+  static long padFor64byteAlignment(long orig) {
+    return ((orig + 63) / 64) * 64;
+  }
+
+  static int safeLongToInt(long value) {
+    assert value >= Integer.MIN_VALUE  : "Value is too small to fit in an int";
```
revans2, line 260: Do we ever want to write negative numbers for any of these values?

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableHeader.java
```diff
@@ -0,0 +1,199 @@
...
+
+  private boolean initialized = false;
+
+
+  public KudoTableHeader(DataInputStream din) throws IOException {
+    readFrom(din);
+  }
+
+  KudoTableHeader(long offset, long numRows, long validityBufferLen, long offsetBufferLen,
+                  long totalDataLen, int numColumns, byte[] hasValidityBuffer) {
```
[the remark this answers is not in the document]
revans2: With that if any of these values should have valid ranges we should check for those ranges here too. (like negative values should not be allowed)

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableHeader.java
```diff
@@ -33,37 +34,61 @@ public final class KudoTableHeader {
...
+    byte[] hasValidityBuffer = new byte[validityBufferLength];
+    din.readFully(hasValidityBuffer);
 
-  public KudoTableHeader(DataInputStream din) throws IOException {
-    readFrom(din);
+    return new KudoTableHeader(offset, numRows, validityBufferLen, offsetBufferLen, totalDataLen, numColumns,
+        hasValidityBuffer);
   }
 
   KudoTableHeader(long offset, long numRows, long validityBufferLen, long offsetBufferLen,
```
revans2, line 78: Still confused why most of these are longs if we are going to make sure that they fit in the range of ints. Especially if this is a package private API that we are the consumers/producers of. That means that `safeLongToNonNegativeInt`, really would just become something to ensure that the value is not-negative.

revans2 (approved):
It looks good enough to ship for me. Especially because the API I have issues with is package private so we can change it later. But I want to hear from Jason on this too.

jlowe (approved):

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoSerializer.java
```diff
@@ -0,0 +1,293 @@
...
+   * @param table table to write
+   * @param out output stream
+   * @param rowOffset row offset in original table
+   * @param numRows number of rows to write
+   * @return number of bytes written
+   */
+  public long writeToStream(Table table, OutputStream out, int rowOffset, int numRows) {
+    HostColumnVector[] columns = null;
+    try {
+      columns = IntStream.range(0, table.getNumberOfColumns())
```
abellina, line 177: This looks like we are writing the whole `Table` (not sliced? but this is probably what I am missing) for every single partition of such table. Is that true? In the past we had 1 copy per column, not 200 copies per column. I am not seeing where this is getting called so that could explain it.

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoSerializer.java
```diff
@@ -0,0 +1,293 @@
...
+
+  static long padForHostAlignment(DataWriter out, long bytes) throws IOException {
+    final long paddedBytes = padForHostAlignment(bytes);
+    if (paddedBytes > bytes) {
+      out.write(PADDING, 0, (int) (paddedBytes - bytes));
+    }
+    return paddedBytes;
+  }
+
+  static long padFor64byteAlignment(long orig) {
```
abellina, line 284: this looks to be unused.

src/test/java/com/nvidia/spark/rapids/jni/KudoSerializerTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class KudoSerializerTest {
+
+  @Test
+  public void testRowCountOnly() throws Exception {
```
abellina, line 36: this doesn't test a no-rows case, but it would be nice to add.

src/test/java/com/nvidia/spark/rapids/jni/KudoSerializerTest.java
```diff
@@ -0,0 +1,109 @@
...
+    listBuilder.addColumn(DType.INT32, "c1");
+
+    Schema.Builder structBuilder =  builder.addColumn(DType.STRUCT, "d");
+    structBuilder.addColumn(DType.INT8, "d1");
+    structBuilder.addColumn(DType.INT64, "d2");
+
+    return builder.build();
+  }
+
+  private static Table buildSimpleTable() {
```
abellina, line 94: it would be nice to add a deeply nested case as well. The integration tests in spark-rapids are likely to cover some of this, but as a serialization format, I could see a bit more permutations of columns added to the tests.

src/test/java/com/nvidia/spark/rapids/jni/KudoSerializerTest.java
```diff
@@ -0,0 +1,109 @@
...
+    HostColumnVector.StructType st = new HostColumnVector.StructType(
+        true,
+        new HostColumnVector.BasicType(true, DType.INT8),
+        new HostColumnVector.BasicType(true, DType.INT64)
+    );
+    return new Table.TestBuilder()
+        .column(1, 2, 3, 4)
```
abellina, line 101: could we have a case where the number of rows is `Int.MaxValue`? We have tests for a small number of things, and 0 rows, but not the limit.

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoSerializer.java
abellina: ah yes, that would be great to mark as "test only". It wasn't obvious to me.

PR #2567, opened 2024-11-04
Make the zone DB shutdown/load cycle re-doable

revans2:
I swear I ran all of the unit tests in spark rapids before merging the last [GpuTimeZoneDB changes](https://github.com/NVIDIA/spark-rapids-jni/pull/2561). But when I ran them again after merging it I hit one case where a unit test shut down the database and another tried to initialize it again. So I ma just going to make it so that the code does not care how many times it is shut down are started up again. The goal is to just make it work.


revans2:
build

jlowe (approved):

gerashegalov (approved):

Issue #2574, opened 2024-11-06
[FEA] XXhash64 Supports nested data type required by approx_count_distinct


res-life:
**Is your feature request related to a problem? Please describe.**
Spark approx_count_distinct needs to get a xxhash64 hash for all types of data.
And one customer requires approx_count_distinct on nested type.

Link to this [Rapids issue](https://github.com/NVIDIA/spark-rapids/issues/5199)

PR #2575, opened 2024-11-06
Add `xxhash64` support for nested types

ustcfy:
Contribute to https://github.com/NVIDIA/spark-rapids-jni/issues/2574

This PR adds support for `xxhash64` with `struct` , `list` and `map` types.

The hardcoded expected values in the tests are generated using Spark. 
For example, you can produce the same results as in `testXXHash64ListOfStruct` using the following code.
``` scala
import org.apache.spark.sql.functions._
import org.apache.spark.sql.Row

val schema = StructType(Seq(
    StructField("listofstruct",
        ArrayType(
            StructType(Seq(
                StructField("string", StringType, true),
                StructField("int", IntegerType, true),
                StructField("double", DoubleType, true),
                StructField("float", FloatType, true),
                StructField("bool", BooleanType, true)
            )),
            true
        )
    )
))

val data = Seq(
    Row(List()),
    Row(List(Row("a", 0, 0.0, 0f, true))),
    Row(List(Row("B\n", 100, 100.0, 100f, false), Row("dE\"\u0100\t\u0101 \ud720\ud721", -100, -100.0, -100f, null))),
    Row(List(Row("A very long (greater than 128 bytes/char string) to test a multi hash-step data point " +
        "in the MD5 hash function. This string needed to be longer.", Int.MinValue, java.lang.Double.longBitsToDouble(0x7ff0000000000001L), java.lang.Float.intBitsToFloat(0xff800001), false))),
    Row(List(Row(null, Int.MaxValue, java.lang.Double.longBitsToDouble(0x7fffffffffffffffL), java.lang.Float.intBitsToFloat(0xffffffff), true), Row(null, null, null, null, null))),
    Row(null)
)

val df = spark.createDataFrame(spark.sparkContext.parallelize(data), schema)
df.columns.foreach(c => println(s"$c => ${df.select(xxhash64(col(c))).collect.mkString(",")}"))
```
You will get the following output:
``` scala
listofstruct => [42],[7451748878409563026],[948372773124634350],[8444697026100086329],[-5888679192448042852],[42]
```

### DIFF
--- a/src/main/cpp/src/xxhash64.cu
+++ b/src/main/cpp/src/xxhash64.cu
@@ -404,7 +404,8 @@ class device_row_hasher {
      * - If the current column is a primitive column, it computes the hash value.
      *
      * For example, consider the following nested column: `List<Struct<int, float>>`.
-     * list_of_struct_column = [ [(1, 2.0), (3, 4.0)], ], which has only one item [(1, 2.0), (3, 4.0)]
+     * list_of_struct_column = [ [(1, 2.0), (3, 4.0)] ], which has only one item
+     * [(1, 2.0), (3, 4.0)].
      *
      *            L1            List<Struct<int, float>>
      *            |
--- a/src/main/cpp/src/xxhash64.cu
+++ b/src/main/cpp/src/xxhash64.cu
@@ -34,6 +34,8 @@ namespace {
 using hash_value_type = int64_t;
 using half_size_type  = int32_t;
 
+constexpr int MAX_NESTED_DEPTH = 8;
+
 constexpr __device__ inline int64_t rotate_bits_left_signed(hash_value_type h, int8_t r)
 {
   return (h << r) | (h >> (64 - r)) & ~(-1 << r);
@@ -295,9 +297,8 @@ hash_value_type __device__ inline XXHash_64<numeric::decimal128>::operator()(
  * handling, refer to the SparkXXHash64 functor.
  *
  * @tparam Nullate A cudf::nullate type describing whether to check for nulls.
- * @tparam MAX_NESTED_DEPTH The maximum nested depth allowed in the input table.
  */
-template <typename Nullate, int MAX_NESTED_DEPTH>
+template <typename Nullate>
 class device_row_hasher {
  public:
   device_row_hasher(Nullate nulls, cudf::table_device_view const& t, hash_value_type seed)
@@ -367,25 +368,26 @@ class device_row_hasher {
       return hasher.template operator()<T>(col, row_index);
     }
 
-    class col_stack_element {
+    struct col_stack_frame {
      private:
-      cudf::column_device_view column;  // current column
-      int child_idx;  // index of the child column to process next, initialized as 0
+      cudf::column_device_view _column;  // the column to process
+      int _idx_to_process;               // the index of child or element to process next
 
      public:
-      __device__ col_stack_element() =
+      __device__ col_stack_frame() =
         delete;  // Because the default constructor of `cudf::column_device_view` is deleted
 
-      __device__ col_stack_element(cudf::column_device_view col) : column(col), child_idx(0) {}
+      __device__ col_stack_frame(cudf::column_device_view col)
+        : _column(std::move(col)), _idx_to_process(0)
+      {
+      }
 
-      __device__ int get_and_inc_child_idx() { return child_idx++; }
+      __device__ int get_and_inc_idx_to_process() { return _idx_to_process++; }
 
-      __device__ int cur_child_idx() { return child_idx; }
+      __device__ int get_idx_to_process() { return _idx_to_process; }
 
-      __device__ cudf::column_device_view get_column() { return column; }
+      __device__ cudf::column_device_view get_column() { return _column; }
     };
-
-    typedef col_stack_element* col_stack_element_ptr;
 
     /**
      * @brief Functor to compute hash value for nested columns.
@@ -403,9 +405,9 @@ class device_row_hasher {
      *      next struct element would be pushed into the stack.
      * - If the current column is a primitive column, it computes the hash value.
      *
-     * For example, consider the following nested column: `List<Struct<int, float>>`.
-     * list_of_struct_column = [ [(1, 2.0), (3, 4.0)] ], which has only one item
-     * [(1, 2.0), (3, 4.0)].
+     * For example, consider that the input column is of type `List<Struct<int, float>>`.
+     * Assume that the element at `row_index` is: [(1, 2.0), (3, 4.0)].
+     * The sliced column is noted as L1 here.
      *
      *            L1            List<Struct<int, float>>
      *            |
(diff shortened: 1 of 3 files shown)


src/main/cpp/src/xxhash64.cu
```diff
@@ -296,27 +320,187 @@ class device_row_hasher {
...
+      cudf::column_device_view column;  // current column
+      int child_idx;  // index of the child column to process next, initialized as 0
+
+     public:
+      __device__ col_stack_element() =
+        delete;  // Because the default constructor of `cudf::column_device_view` is deleted
+
+      __device__ col_stack_element(cudf::column_device_view col) : column(col), child_idx(0) {}
+
+      __device__ int child_idx_inc_one() { return this->child_idx++; }
```
res-life, line 384: Nit:
```suggestion
      __device__ int inc_child_idx() { return this->child_idx++; }
```

src/main/cpp/src/xxhash64.cu
```diff
@@ -271,9 +272,32 @@ hash_value_type __device__ inline XXHash_64<numeric::decimal128>::operator()(
...
+ * - Serially using the output hash as an input seed for the next item
+ * - Ignorance of null values
+ *
+ * The serial use of hashes as seeds means that data of different nested types
+ * can exhibit hash collisions. For example, a row of an integer column
+ * containing a 1 will have the same hash as a lists column of integers
+ * containing a list of [1] and a struct column of a single integer column
+ * containing a struct of {1}.
+ *
+ * As a consequence of ignoring null values, inputs like [1], [1, null], and
```
res-life, line 287: [1] and [1, null] have the same hash value?
If item is null, then return the seed(e.g.: 42).

src/main/cpp/src/xxhash64.cu
```diff
@@ -296,27 +320,187 @@ class device_row_hasher {
...
+   * element is the input seed (the previous element's hash).
    */
   class element_hasher_adapter {
    public:
-    template <typename T, CUDF_ENABLE_IF(cudf::column_device_view::has_element_accessor<T>())>
+    class element_hasher {
+     private:
+      Nullate const _check_nulls;
+      hash_value_type const _seed;
+      hash_value_type const _null_hash;
```
res-life, line 337: _seed and _null_hash are always the same, so we can remove one.

src/main/cpp/src/xxhash64.cu
```diff
@@ -296,27 +320,187 @@ class device_row_hasher {
...
+      cudf::column_device_view column;  // current column
+      int child_idx;  // index of the child column to process next, initialized as 0
+
+     public:
+      __device__ col_stack_element() =
+        delete;  // Because the default constructor of `cudf::column_device_view` is deleted
+
+      __device__ col_stack_element(cudf::column_device_view col) : column(col), child_idx(0) {}
+
+      __device__ int get_and_inc_child_idx() { return this->child_idx++; }
```
res-life, line 384:
```suggestion
      __device__ int get_and_inc_child_idx() { return child_idx++; }
```

src/main/cpp/src/xxhash64.cu
ustcfy: > _seed and _null_hash are always the same, so we can remove one.

This is to maintain consistency with the definition of `cudf::experimental::row::hash::element_hasher`.

src/main/cpp/src/xxhash64.cu
res-life: Can we omit this: `this->`?

src/main/cpp/src/xxhash64.cu
res-life: No need to keep consistent. Because it's a new class and we can remove one variable.

src/main/cpp/src/xxhash64.cu
```diff
@@ -296,27 +320,187 @@ class device_row_hasher {
...
+     *   non-list child since null values can be ignored in the xxhash64 computation.
+     * - If the current column is a struct column, there are two cases:
+     *    a. If the struct column has only one row, it would be treated as a struct element. The
+     *      children of the struct element would be pushed into the stack.
+     *    b. If the struct column has multiple rows, it would be treated as a struct column. The
+     *      next struct element would be pushed into the stack.
+     * - If the current column is a primitive column, it computes the hash value.
+     *
+     * For example, consider the following nested column: `List<Struct<int, float>>`.
+     * list_of_struct_column = {[1, 2.0], [3, 4.0]}
```
res-life, line 410: Nit: Using the following?
[    [(1, 2.0), (3, 4.0)],    ]

[] notes array
() notes structure

src/main/cpp/src/xxhash64.cu
ustcfy: > Can we omit this: `this->`?

Yes, thanks for the guidance.

src/test/java/com/nvidia/spark/rapids/jni/HashTest.java
```diff
@@ -385,6 +385,152 @@ void testXXHash64Mixed() {
     }
   }
 
+  @Test
```
res-life, line 388: Please add more cases:
- decimal, date, time ...
- Please test float/double NaN, -0.0

src/main/cpp/src/xxhash64.cu
ustcfy: > [1] and [1, null] have the same hash value? If item is null, then return the seed(e.g.: 42).

- For the first element, the seed is `42L`。
- For the other elements, the seed is the previous element's hash value.

res-life:
Please add more cases, LGTM for the function.

res-life:
LGTM

res-life:
build

res-life (dismissed):

src/main/cpp/src/xxhash64.cu
```diff
@@ -325,6 +507,35 @@ class device_row_hasher {
...
 };
 
+void check_nested_depth(cudf::table_view const& input, int max_nested_depth)
+{
+  using column_checker_fn_t = std::function<int(cudf::column_view const&)>;
+
+  column_checker_fn_t get_nested_depth = [&](cudf::column_view const& col) {
+    if (col.type().id() == cudf::type_id::LIST) {
+      // list column will be replaced by its most inner non-list child
+      return get_nested_depth(cudf::lists_column_view(col).child());
```
firestarman, line 517: Does List not take a depth slot ?

src/main/cpp/src/xxhash64.cu
```diff
@@ -296,27 +320,185 @@ class device_row_hasher {
...
+        : _check_nulls(check_nulls), _seed(seed)
+      {
+      }
+
+      template <typename T, CUDF_ENABLE_IF(cudf::column_device_view::has_element_accessor<T>())>
+      __device__ hash_value_type operator()(cudf::column_device_view const& col,
+                                            cudf::size_type row_index) const noexcept
+      {
+        if (_check_nulls && col.is_null(row_index)) { return _seed; }
+        return XXHash_64<T>{_seed}(col.element<T>(row_index));
```
revans2, line 349: Can we have a follow on issue to make the hash function a template? We have the same issues with all hash functions we want to support, so having generic code would be nice.

revans2 (commented):
looks good, but I am no expert on the CUDA side of things.

src/main/cpp/src/xxhash64.cu
ustcfy: > Does List not take a depth slot ?

Yes.

src/main/cpp/src/xxhash64.cu
```diff
@@ -296,27 +320,185 @@ class device_row_hasher {
...
+      Nullate const _check_nulls;
+      hash_value_type const _seed;
+
+     public:
+      __device__ element_hasher(Nullate check_nulls, hash_value_type seed)
+        : _check_nulls(check_nulls), _seed(seed)
+      {
+      }
+
+      template <typename T, CUDF_ENABLE_IF(cudf::column_device_view::has_element_accessor<T>())>
```
firestarman, line 344: NIT:
These two methods seems being copied from `element_hasher`, and I prefer to leveraging `element_hasher` here instead of implementing them here again.

firestarman (commented):
Look good to me, but a nit.

src/main/cpp/src/xxhash64.cu
ustcfy: In `cudf::experimental::row::hash::element_hasher`, the seed is of type `uint32_t`, whereas in `xxhash64`, the seed is of type `int64_t`.

ttnghia (changes requested):
Sorry, this PR has similar implementation with the HiveHash one, so both need to address the issue with shallow stack size unless we want to target this for 24.12.

ttnghia (approved):

ttnghia:
build

ustcfy:
build

src/main/cpp/src/xxhash64.cu
```diff
@@ -325,6 +509,40 @@ class device_row_hasher {
   hash_value_type const _seed;
 };
 
+void check_nested_depth(cudf::table_view const& input)
+{
+  using column_checker_fn_t = std::function<int(cudf::column_view const&)>;
+
+  column_checker_fn_t get_nested_depth = [&](cudf::column_view const& col) {
+    if (col.type().id() == cudf::type_id::LIST) {
+      auto const& child_col = cudf::lists_column_view(col).child();
```
ttnghia, line 518:
```suggestion
      auto const child_col = cudf::lists_column_view(col).child();
```

ttnghia (changes requested):

ttnghia (approved):

res-life:
[Spark-Rapids PR](https://github.com/NVIDIA/spark-rapids/pull/11859) verified the correctness,  we can merge it.

res-life (approved):

res-life:
build

PR #2578, opened 2024-11-07
Introduce kudo reader.

liurenjie1024:
This pr is part of https://github.com/NVIDIA/spark-rapids-jni/pull/2532 , which introduces reader part of kudo format.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
@@ -139,7 +139,7 @@ class KudoTableMerger extends MultiKudoTableVisitor<Void, Void, KudoHostMergeRes
 
           startRow += sliceInfo.getRowCount();
         }
-        return toIntExact(nullCountTotal);
+        return nullCountTotal;
       }
     } else {
       return 0;
--- a/src/main/java/com/nvidia/spark/rapids/jni/kudo/MultiKudoTableVisitor.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/kudo/MultiKudoTableVisitor.java
@@ -56,6 +56,7 @@ abstract class MultiKudoTableVisitor<T, P, R> implements SchemaVisitor<T, P, R> 
     this.currentOffsetOffsets = new long[tables.size()];
     this.currentDataOffset = new long[tables.size()];
     this.sliceInfoStack = new Deque[tables.size()];
+    long totalRowCount = 0L;
     for (int i = 0; i < tables.size(); i++) {
       this.currentValidityOffsets[i] = 0;
       KudoTableHeader header = tables.get(i).getHeader();
@@ -63,8 +64,8 @@ abstract class MultiKudoTableVisitor<T, P, R> implements SchemaVisitor<T, P, R> 
       this.currentDataOffset[i] = header.getValidityBufferLen() + header.getOffsetBufferLen();
       this.sliceInfoStack[i] = new ArrayDeque<>(16);
       this.sliceInfoStack[i].add(new SliceInfo(header.getOffset(), header.getNumRows()));
+      totalRowCount += header.getNumRows();
     }
-    long totalRowCount = tables.stream().mapToLong(t -> t.getHeader().getNumRows()).sum();
     this.totalRowCountStack = new ArrayDeque<>(16);
     totalRowCountStack.addLast(toIntExact(totalRowCount));
     this.hasNull = true;
@@ -88,7 +89,12 @@ abstract class MultiKudoTableVisitor<T, P, R> implements SchemaVisitor<T, P, R> 
   public T visitStruct(Schema structType, List<T> children) {
     updateHasNull();
     T t = doVisitStruct(structType, children);
-    updateOffsets(false, false, false, -1);
+    updateOffsets(
+        false, // Update offset buffer offset
+        false, // Update data buffer offset
+        false, // Update slice info
+        -1 // element size in bytes, not used for struct
+    );
     currentIdx += 1;
     return t;
   }
@@ -99,7 +105,12 @@ abstract class MultiKudoTableVisitor<T, P, R> implements SchemaVisitor<T, P, R> 
   public P preVisitList(Schema listType) {
     updateHasNull();
     P t = doPreVisitList(listType);
-    updateOffsets(true, false, true, Integer.BYTES);
+    updateOffsets(
+        true, // update offset buffer offset
+        false, // update data buffer offset
+        true, // update slice info
+        Integer.BYTES // element size in bytes
+    );
     currentIdx += 1;
     return t;
   }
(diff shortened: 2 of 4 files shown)


src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
```diff
@@ -0,0 +1,348 @@
...
+      }
+
+      curDestBitIdx += appendCount;
+      if (curDestBitIdx == 8) {
+        curDestBitIdx = 0;
+        curDestByteIdx += 1;
+      }
+
+      curIdx += appendCount;
+    }
```
jlowe, line 264: Same byte-at-a-time and conditionals-in-while-loop performance problems as noted above.
```suggestion
    if (curDestBitIdx > 0) {
      int numBits = 8 - curDestBitIdx;
      int mask = ((1 << numBits) - 1) << curDestBitIdx;
      dest.setByte(curDestByteIdx, (byte) (dest.getByte(curDestByteIdx) | mask));
      curDestByteIdx += 1;
      numRows -= numBits;
    }
    if (numRows > 0) {
      int numBytes = (numRows + 7) / 8;
      dest.setMemory(curDestByteIdx, numBytes, (byte) 0xFF);
    }
```

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
```diff
@@ -0,0 +1,348 @@
...
+          .order(ByteOrder.LITTLE_ENDIAN)
+          .asIntBuffer();
+
+      int accumulatedDataLen = 0;
+
+      for (int tableIdx = 0; tableIdx < getTableSize(); tableIdx += 1) {
+        SliceInfo sliceInfo = sliceInfoOf(tableIdx);
+
+        if (sliceInfo.getRowCount() > 0) {
+          int rowCnt = toIntExact(sliceInfo.getRowCount());
```
jlowe, line 284: sliceInfo shouldn't have row counts that are longs

abellina:
I will review this today as well

src/main/java/com/nvidia/spark/rapids/jni/kudo/ColumnViewInfo.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.Optional;
+
+import static java.lang.Math.toIntExact;
+
+class ColumnViewInfo {
+    private final DType dtype;
+    private final ColumnOffsetInfo offsetInfo;
+    private final long nullCount;
+    private final long rowCount;
```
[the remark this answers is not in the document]
jlowe: That's exactly my point.  These _cannot_ be larger than than an int in practice, yet this doesn't check for the overflow.  This class doesn't itself perform the accumulation, so IMO this should only deal with ints, and the code performing the accumulation should deal with any overflow if it is trying to convert a long to an int in order to construct this.  There are other cases where callers are not accumulating, which is why I don't think it's appropriate to take longs or check in this class.

src/main/java/com/nvidia/spark/rapids/jni/kudo/TableBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+
+import static com.nvidia.spark.rapids.jni.Preconditions.ensure;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This class is used to build a cudf table from a list of column view info, and a device buffer.
+ */
+class TableBuilder implements SchemaVisitor<Object, Table>, AutoCloseable {
+    // Current column index
+    private int curIdx;
```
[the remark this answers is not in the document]
jlowe: The reason I suggested the code above is that the comment is unnecessary when using a self-explanatory name.  If the new name is still cryptic enough that it warrants a comment, I'd suggest naming it currentColumnIndex so it's obvious what it is not only where it's declared but also where it's used.

src/main/java/com/nvidia/spark/rapids/jni/kudo/MultiKudoTableVisitor.java
```diff
@@ -0,0 +1,264 @@
...
+ * <br/>
+ * The schema used when visiting these kudo tables must be same as the schema used when creating these kudo tables.
+ */
+abstract class MultiKudoTableVisitor<T, R> implements SchemaVisitor<T, R> {
+    private final List<KudoTable> tables;
+    private final long[] currentValidityOffsets;
+    private final long[] currentOffsetOffsets;
+    private final long[] currentDataOffset;
+    private final Deque<SliceInfo>[] sliceInfoStack;
+    private final Deque<Long> totalRowCountStack;
```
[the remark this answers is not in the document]
jlowe: We cannot support a row count larger than an int.  I'm fine while we're accumulating to use a long to capture overflow conditions (e.g.: within the constructor when it accumulates and checks for overflow).  I see no reason to track computed  total rows that are larger than an int because the overflow check should have already happened in the constructor and in  updateOffsets which is missing an overflow check.

pxLi:
build

pxLi:
re-kicked, previous trigger failed due to github server-side reset the connection
```
Received RST_STREAM: Stream not processed
```

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
```diff
@@ -0,0 +1,332 @@
...
+            nullCountTotal += copyValidityBuffer(validityBuffer, startRow,
+                memoryBufferOf(tableIdx), toIntExact(validityOffset),
+                sliceInfo);
+          } else {
+            appendAllValid(validityBuffer, startRow, sliceInfo.getRowCount());
+          }
+
+          startRow += sliceInfo.getRowCount();
+        }
+        return toIntExact(nullCountTotal);
```
jlowe, line 142: nullCountTotal is already an int, so this is wasteful.  There should already be a row count overflow check elsewhere (and we cannot overflow nullcounts without also overflowing row counts), so seems like this should simply return nullCountTotal directly.

jlowe (approved):
Minor nit but otherwise lgtm.  Would be good to hear from @abellina before merging since he mentioned he intended to review.

src/main/java/com/nvidia/spark/rapids/jni/kudo/MultiKudoTableVisitor.java
```diff
@@ -0,0 +1,265 @@
...
+  public T visit(Schema primitiveType) {
+    updateHasNull();
+    if (primitiveType.getType().hasOffsets()) {
+      // string type
+      updateDataLen();
+    }
+
+    T t = doVisit(primitiveType);
+    if (primitiveType.getType().hasOffsets()) {
+      updateOffsets(true, true, false, -1);
```
abellina, line 131: could we add comments for each argument here?

```
updateOffsets(/*updateOffset*/ true, /*updateData*/ true, /*updateSliceInfo*/ false, /*sizeInBytes* -1)
```

abellina (commented):
I'd still like to do more passes, but don't block for me.

abellina (approved):

PR #2610, opened 2024-11-20
Add checkstyle in maven build

### LINKED ISSUE #2591: [FEA] Add checkstyle for building.
Currently we are missing format/style check during build. 

[checkstyle](https://checkstyle.org/) is among the most popular formatters/linters in java ecosystem, and we should add support for it during build process.

No.

**Additional context**
No.

liurenjie1024:
Close #2591.

The style check only checks modified files in each pr.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -770,7 +770,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
-        <version>3.5.0</version>
+        <version>3.3.0</version>
         <configuration>
           <configLocation>dev/checkstyle.xml</configLocation>
           <encoding>UTF-8</encoding>


jlowe (commented):
This PR needs to run checkstyle on the entire eligible codebase and update accordingly match the desired style.  Otherwise we're going to have a parade of PRs where the change they _want_ to make is relatively small, but the change is dwarfed by the style fixes required to pass CI.  This would also help sanity check that the configured style checks are doing what we desire in this PR.

dev/checkstyle.sh
```diff
@@ -0,0 +1,44 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#!/bin/bash
+
+set -ex
+
+# Assuming you are in the root of your git repository
+MODIFIED_FILES=$(git diff --name-only "origin/${GITHUB_BASE_REF}")
```
pxLi, line 23: agree with Jason, We would like to not only check style for incremental changes.

and this script is current for github action only with the ENV provided by github. 
can we just check against all eligible code in the change?

and we do not need below in action def then
```
      with:
        submodules: true
        fetch-depth: 0

    - name: Fetch branches and tags
      run: git fetch --all
```

jlowe:
> I ran the checkstyle which detected 744 errors, and most of them not easy to fix since it's not simple formatting issue, for example, missing docs, naming abbreviations, etc. The effort to fix them all is quite huge.

Then this PR should not go in like this.  It adds a significant burden to all future PRs, because if some PR needs to touch many files, now the burden you didn't want to perform is forced on an unrelated PR.  Like I said, we do not want any required fixes to be performed "on the fly" via unrelated PRs.  It's mixing purposes and makes those PRs harder to write and harder to read.

Can we turn on only some of the checkstyle rules for this first version and update any errant files accordingly?  We can then turn on more and more checks via foilowup PRs to make the changes manageable.

pxLi:
> > This PR needs to run checkstyle on the entire eligible codebase and update files accordingly to match the desired style. Otherwise we're going to have a parade of PRs where the change they _want_ to make is relatively small, but the change is dwarfed by the style fixes required to pass CI. This would also help sanity check that the configured style checks are doing what we desire in this PR.
> 
> I ran the checkstyle which detected 744 errors, and most of them not easy to fix since it's not simple formatting issue, for example, missing docs, naming abbreviations, etc. The effort to fix them all is quite huge.

Can we try disabling some rules (e.g., docs) initially and fixing those critical ones only within this change?

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up JDK
+      uses: actions/setup-java@v4.5.0
+      with:
+        java-version: '8'
+        distribution: 'adopt'
+
+    - uses: actions/checkout@v4
+      with:
```
pxLi, line 33: nit: no need to fetch submodule

src/main/java/com/nvidia/spark/rapids/jni/CaseWhen.java
```diff
@@ -16,59 +16,57 @@
...
- *
+ * SQL is:
+ * select
+ * case
+ * when bool_1_expr then "value_1"
+ * when bool_2_expr then "value_2"
+ * when bool_3_expr then "value_3"
+ * else "value_else"
+ * end
+ * from tab
```
jlowe, line 37: The original way the comment was written was easier to read.  IMO style checks should not be messing with leading whitespace within a comment, as it's likely intentional.  This comment applies to other places where leading whitespace of indented blocks was removed.

src/main/java/com/nvidia/spark/rapids/jni/DecimalUtils.java
```diff
@@ -180,29 +182,13 @@ public static Table add128(ColumnView a, ColumnView b, int targetScale) {
     return new Table(add128(a.getNativeView(), b.getNativeView(), targetScale));
   }
 
-  /**
-   * A class to store the result of a cast operation from floating point values to decimals.
-   * <p>
-   * Since the result column may or may not be used regardless of the value of hasFailure, we
-   * need to keep it and let the caller to decide.
-   */
-  public static class CastFloatToDecimalResult {
```
jlowe, line 189: This is a huge PR, and a large reason for that is the enforcement of a strict declaration order of inner classes and methods.  I think that should be deferred to a followup since I'm not sure everyone on the team would agree the new order is better.

Once that is removed, this PR should be relatively small once whitespace diffs are ignored.

PR #2612, opened 2024-11-20
[submodule-sync] bot-submodule-sync-branch-24.12 to branch-24.12 [skip ci] [bot]

nvauto:
submodule-sync to create a PR keeping thirdparty/cudf up-to-date.
HEAD commit SHA: 85354cc55e5a10b35e1a52f4e9d79276c9b38421, cudf commit SHA: https://github.com/rapidsai/cudf/commit/3111aa4723150cb09b88c6968a51afb681b1ab6a

This PR will be auto-merged if test passed. If failed, it will remain open until test pass or manually fix.

PR #2613, opened 2024-11-20
Fix auto merge conflict 2612

hyperbolic2346:
PR https://github.com/NVIDIA/spark-rapids-jni/pull/2612 failed due to `Server error` after all tests passed. This PR merges those changes.


jlowe (commented):
Looks like #2612 has now merged after a subsequent sync job run, so it looks like we don't need this.

PR #2626, opened 2024-11-26
Optimize validity buffer concat.

### LINKED ISSUE #2579: [FEA] Optimize kudo when merging validity buffer.

Following the discussion here: https://github.com/NVIDIA/spark-rapids-jni/pull/2532#discussion_r1815662126 , there is another approach to improve performance of kudo when merging validity buffers.

liurenjie1024:
Close #2579

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
@@ -16,17 +16,6 @@
 
 package com.nvidia.spark.rapids.jni.kudo;
 
-import ai.rapids.cudf.HostMemoryBuffer;
-import ai.rapids.cudf.Schema;
-import com.nvidia.spark.rapids.jni.Arms;
-import com.nvidia.spark.rapids.jni.schema.Visitors;
-
-import java.nio.ByteOrder;
-import java.nio.IntBuffer;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.OptionalInt;
-
 import static com.nvidia.spark.rapids.jni.Preconditions.ensure;
 import static com.nvidia.spark.rapids.jni.kudo.ColumnOffsetInfo.INVALID_OFFSET;
 import static com.nvidia.spark.rapids.jni.kudo.KudoSerializer.getValidityLengthInBytes;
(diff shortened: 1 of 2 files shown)


src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
```diff
@@ -155,80 +141,64 @@ private static int copyValidityBuffer(HostMemoryBuffer dest, int startBit,
...
-        nullCount += (appendCount - NUMBER_OF_ONES[srcByte & 0xFF]);
+      // This is safe since we always have validity buffer 4 bytes padded
+      int srcInt = src.getInt(curSrcOffset);
+      srcInt = srcInt >>> curSrcBitIdx;
 
-        // Sets the bits in destination buffer starting from curDestBitIdx to 0
-        byte destByte = dest.getByte(curDestByteIdx);
-        destByte = (byte) (destByte & ((1 << curDestBitIdx) - 1));
+      if (dest.getLength() >= (curDestOffset + Integer.BYTES)) {
+        // We have enough room to get an int
```
jlowe, line 163: Don't we _always_ have enough space to get an integer from the destination buffer because it's at least 4-byte padded?

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
```diff
@@ -155,80 +141,64 @@ private static int copyValidityBuffer(HostMemoryBuffer dest, int startBit,
...
+      } else {
+        int destBufRemBytes = toIntExact(dest.getLength() - curDestOffset);
+        byte[] destBytes = new byte[4];
+        dest.getBytes(destBytes, 0, curDestOffset, destBufRemBytes);
+        int destInt = ByteBuffer.wrap(destBytes).order(ByteOrder.LITTLE_ENDIAN).getInt();
+        destInt &= (1 << curDestBitIdx) - 1;
+        destInt |= srcInt << curDestBitIdx;
+
+        ByteBuffer.wrap(destBytes).order(ByteOrder.LITTLE_ENDIAN).putInt(destInt);
+        dest.setBytes(curDestOffset, destBytes, 0, destBufRemBytes);
```
jlowe, line 188: Similar to above, can call setInt here and leverage ByteOrder to determine when we need to swap bytes or not.

ttnghia:
Please update the PR description. It will be displayed in the commit log. Simply saying "closes XXX" will just show that words, which is difficult to track the changes through commit log.

gerashegalov:
> It will be displayed in the commit log

To be fair we do not have the automation (yet) of checking in PRs using the PR description. I follow the convention of copying the PR description as the commit message but it's not mandated and not followed widely in NVIDIA/spark* repos.

res-life:
Could we copy by long instead of int?  We can avoid to use `toIntExact` and use `Long.bitCount`. This may be more fast.

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
```diff
@@ -155,80 +141,64 @@ private static int copyValidityBuffer(HostMemoryBuffer dest, int startBit,
...
+    int totalRowCount = toIntExact(sliceInfo.getRowCount() + sliceInfo.getValidityBufferInfo().getBeginBit());
+    int curSrcIdx = sliceInfo.getValidityBufferInfo().getBeginBit();
+    int curDestIdx = startBit;
 
-        nullCount += (appendCount - NUMBER_OF_ONES[srcByte & 0xFF]);
 
-        // Sets the bits in destination buffer starting from curDestBitIdx to 0
-        byte destByte = dest.getByte(curDestByteIdx);
-        destByte = (byte) (destByte & ((1 << curDestBitIdx) - 1) & 0xFF);
+    while (curSrcIdx < totalRowCount) {
```
ttnghia, line 149: So does this concatenate the validity buffers, one word (32 bits) at a time, in a serial manner, in Java?

Can't we use C++/CUDA for accelerating this?

jlowe: There's a separate effort by @nvdbaranec to provide GPU kernels for Kudo buffer concatenation.  We need both CPU and GPU implementations.  In some cases, the GPU will be the bottleneck for a stage, and CPUs will be idle while waiting for the GPU semaphore.  It will be more efficient to concatenate on the CPU, even if it's slower than the GPU, when the CPU can complete the operation within the time otherwise spent waiting for the GPU.

ttnghia: I've never seen any parallel Java code in our plugin/JNI. Assuming that it is doable, would we consider doing so for the operations like this?

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
```diff
@@ -155,80 +141,64 @@ private static int copyValidityBuffer(HostMemoryBuffer dest, int startBit,
...
+      int leftRowCount = totalRowCount - curSrcIdx;
 
-        // Update destination byte with the bits from source byte
-        destByte = (byte) ((destByte | (srcByte << curDestBitIdx)) & 0xFF);
-        dest.setByte(curDestByteIdx, destByte);
+      int curDestOffset = (curDestIdx / 32) * Integer.BYTES;
```
gerashegalov, line 152: consider using mnemonic constants 
```suggestion
      int curDestOffset = (curDestIdx / Integer.SIZE) * Integer.BYTES;
```
or even 
```suggestion
      int curDestOffset = curDestIdx / Byte.SIZE;
```

PR #2629, opened 2024-11-26
Fix cudf submodule sync

jlowe:
rapidsai/cudf#17427 where nvtx/ranges.hpp was being implicitly included from another header where it was removed, so updated the source to include this directly.
This also updates to breaking changes in rapidsai/rmm#1722 where we need to explicitly link against rmm's logger impl to restore logging functionality.


jlowe:
build

pxLi (approved):
thanks!

PR #2634, opened 2024-11-27
Remove explicit rmm logger link as it is implicit from libcudf

jlowe:
After libcudf fixed their rmm logger link, we need to remove our explicit link to avoid duplicate definition errors.  The rmm logger will be picked up implicitly from libcudf.


jlowe:
build

pxLi (approved):

PR #2660, opened 2024-12-06
Implement date/time truncation functions

ttnghia:
This implements the functions to truncate date/timestamp to some specific component, matching the Spark SQL function `trunc` and `date_trunc`.

Due to changes to related code, the existing module `DateTimeRebase` is changed to `DateTimeUtils` (without breaking), which will also contain the newly implemented function.


ttnghia:
build

src/main/java/com/nvidia/spark/rapids/jni/DateTimeUtils.java
```diff
@@ -0,0 +1,99 @@
...
+   *  - {@code "MINUTE"}: zero out the second with fraction part.<br>
+   *  - {@code "SECOND"}: zero out the second fraction part.<br>
+   *  - {@code "MILLISECOND"}: zero out the microseconds.<br>
+   *  - {@code "MICROSECOND"}: keep everything.<br>
+   *
+   * @param datetime The input date/time
+   * @param format The time component to truncate to
+   * @return The truncated date/time
+   */
+  public static ColumnVector truncate(ColumnView datetime, ColumnView format) {
```
revans2, line 89: Could we have a version that supports a scalar format? I just think that will be the most common use case and if we can save memory and time by not exploding the scalar into a column I think that would be great.

I think that can be done as a follow on issue if you want.

revans2 (approved):
Just a possible improvement that is minor.

src/main/java/com/nvidia/spark/rapids/jni/DateTimeUtils.java
ttnghia: I thought about that, however, either parameter can be Scalar thus if doing so we will have to implement 4 JNI overloads for this function. Instead, the code supports columns in which any of them can have one row and can achieve the same performance. 

In plugin code, I just "convert"  Scalar into a column of one row. Such conversion should not be very expensive. But I can add more overload to support Scalar if that can help anything better.

revans2: Yes either can be a scalar, but fmt is the one that is really common to be a scalar. Most of the time you want to truncate to a very specific value so you know how to deal with the resulting timestamp/date.

PR #2679, opened 2024-12-11
Fetch `spdlog` without depending on `cudf`

ttnghia:
After https://github.com/rapidsai/cudf/pull/17307, cudf no longer provide `spdlog` in its dependency chain. As such, we need to adapt to it. 

This PR remove the dependency on `cudf` for fetching `spdlog`. Instead, `spark-rapids-jni` will fetch `spdlog` on its own.

### DIFF
--- a/src/main/cpp/faultinj/faultinj.cu
+++ b/src/main/cpp/faultinj/faultinj.cu
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-#include <cudf/logger.hpp>
-
 #include <cuda.h>
 
 #include <assert.h>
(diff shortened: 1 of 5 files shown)


src/main/cpp/faultinj/faultinj.cu
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include <cudf/logger.hpp>
```
jlowe, line 17: The faultinj project does not link against any cudf libs, including the rapids_logger.  I think we need to add an explicit dependency on spdlog for this project fetching it if necessary.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -14,18 +14,16 @@
  * limitations under the License.
  */
 
+#include <cudf/logger.hpp>
```
jlowe, line 17: This is a semantic change to how logging is performed in this class.  Currently we disable libcudf logging, but that does not disable the logging in this class.  If we leverage the cudf logger here, then we will only support logging in this class if we support logging in libcudf.

Like the faultinj project, I think we need to fetch our own spdlog here and use it as our own logger.

ttnghia:
build

jlowe (approved):

Issue #1307, opened 2023-07-31
[FEA] Extend the Scalar API with functionality to allow access to native handles.


nvdbaranec:
This PR https://github.com/NVIDIA/spark-rapids-jni/pull/1303 had to do use a package trampoline to get around lack of access to `Scalar.getScalarHandle()` (access to the native `cudf::list_scalar*`) and the Scalar constructor that takes a native handle.  These would be useful to have in general `Scalar(DType type, long scalarHandle)`

mattahrens:
Need to file corresponding cudf issue.

PR #2682, opened 2024-12-11
Remove `CudfAccessor.java` and adopt new public methods from `Scalar.java`

ttnghia:
Since the `Scalar` class now publishes its constructor and `getScalarHandle()` method, we can now access these directly without needing the workaround of `CudfAccessor` utility class.

This removes the class `CudfAccessor` completely and replaces its usage by the methods from the `Scalar` class.

Closes https://github.com/NVIDIA/spark-rapids-jni/issues/1307.

Depends on:
 * https://github.com/rapidsai/cudf/pull/17580


res-life (approved):
LGTM after building is passed.

ttnghia:
build

PR #2687, opened 2024-12-12
Optimize date/time truncation for scalar format

ttnghia:
In the current implementation, the input format string is always parsed in each thread. In the cases when it is given as just one literal string, we can optimize the kernel time (this PR) by parsing such string once on the CPU before launching the kernel.


ttnghia:
build

revans2 (approved):

ttnghia:
build

revans2 (approved):

Issue #2699, opened 2024-12-15
[BUG] rapids_integration-scala213-dev-github tests failing


sameerz:
**Describe the bug**
13 test failures related to com.nvidia.spark.rapids.jni.Hash.getMaxStackDepth

<details><summary>Example failure here</summary>
<p>

[2024-12-14T21:08:57.541Z] answer = 'xro516042'
[2024-12-14T21:08:57.541Z] gateway_client = <py4j.clientserver.JavaClient object at 0x7f6ed7cc3280>
[2024-12-14T21:08:57.541Z] target_id = 'o82', name = 'sql'
[2024-12-14T21:08:57.541Z] 
[2024-12-14T21:08:57.541Z]     [94mdef[39;49;00m [92mget_return_value[39;49;00m(answer, gateway_client, target_id=[94mNone[39;49;00m, name=[94mNone[39;49;00m):[90m[39;49;00m
[2024-12-14T21:08:57.541Z]     [90m    [39;49;00m[33m"""Converts an answer received from the Java gateway into a Python object.[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m    For example, string representation of integers are converted to Python[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m    integer, string representation of objects are converted to JavaObject[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m    instances, etc.[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m    :param answer: the string returned by the Java gateway[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m    :param gateway_client: the gateway client used to communicate with the Java[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m        Gateway. Only necessary if the answer is a reference (e.g., object,[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m        list, map)[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m    :param target_id: the name of the object from which the answer comes from[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m        (e.g., *object1* in `object1.hello()`). Optional.[39;49;00m
[2024-12-14T21:08:57.541Z]     [33m    :param name: the name of the member from which the answer comes from[39;49;00m
[...]

PR #2700, opened 2024-12-16
Fix UnsatisfiedLinkError in some rare cases

res-life:
closes https://github.com/NVIDIA/spark-rapids-jni/issues/2699

## why this issue occurs
The original code is:
``` java
public class Hash {
  public static final int MAX_STACK_DEPTH = getMaxStackDepth();

  static {
    NativeDepsLoader.loadNativeDeps();
  }

  private static native int getMaxStackDepth();
}
```
The JVM will first execute:
```java
public static final int MAX_STACK_DEPTH = getMaxStackDepth();
```
then:
```java
  static {
    NativeDepsLoader.loadNativeDeps();
  }
```
`getMaxStackDepth` is a native function which is in the JNI so file.
If do not load native so first, the `getMaxStackDepth` can not be used and it reports: `UnsatisfiedLinkError`

## why this issue do not occur in most cases
If other classes already loaded the JNI so file, then it does not report this error.
e.g.:
```java
JSONUtils.MAX_PATH_DEPTH
Hash.MAX_STACK_DEPTH
```
The above call seq will not report error, becasue `JSONUtils.MAX_PATH_DEPTH` will trigger loading JNI so file.

## fix
Load the JNI so file first in `Hash`

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/Hash.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/Hash.java
@@ -22,14 +22,14 @@ import ai.rapids.cudf.CudfException;
 import ai.rapids.cudf.NativeDepsLoader;
 
 public class Hash {
+  static {
+    NativeDepsLoader.loadNativeDeps();
+  }
+
   // there doesn't appear to be a useful constant in spark to reference. this could break.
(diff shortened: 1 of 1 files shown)


res-life:
build

ustcfy (approved):

PR #2733, opened 2025-01-07
Improve struct pruning logic for Parquet Schema Filtering

nartal1:
This PR refines the logic used to prune Parquet schemas when filtering out columns that do not match the query. Previously, an entire struct could be added prematurely to the schema_map and remain there even if none of its children matched. This could result in an incorrect schema when reading Parquet files.

This pull request includes updates to the `NativeParquetJni.cpp` file to improve the handling of schema pruning in the `column_pruner` class. The changes focus on ensuring accurate child count updates and proper struct inclusion or exclusion based on the presence of matching children.

Improvements to schema pruning logic:

* Updated the logic to save the current struct index and handle child count placeholders accurately. This ensures that parent-before-children ordering is maintained in `schema_map`.
* Introduced a mechanism to track the size of `schema_map` before and after processing each child. This helps determine if a child was retained and accurately updates the child count.
* Added logic to remove structs from `schema_map` and `schema_num_children` if none of their children are included, ensuring that only relevant structs are retained.

I tested this PR with these issues - https://github.com/NVIDIA/spark-rapids/issues/11619,  https://github.com/NVIDIA/spark-rapids/issues/11620, https://github.com/NVIDIA/spark-rapids/issues/11621, https://github.com/NVIDIA/spark-rapids/issues/11628 and https://github.com/NVIDIA/spark-rapids/issues/11629. I don't see the test failures anymore. I will close those issues after enabling the tests which would require the PR in spark-rapids.


revans2 (dismissed):
The changes look good to me. Are there more changes needed on the java side? There were issues there too, if I remember correctly.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -204,13 +204,23 @@ class column_pruner {
...
-    int our_num_children_index = schema_num_children.size();
-    schema_num_children.push_back(0);
+
+    // Save the current index (the position of this struct in the full schema).
+    // We'll need this to add it to the pruned schema map if it qualifies.
+    int struct_index = current_input_schema_index;
+
+    // We add the Struct to schema_map (with a placeholder for child count).
+    // But we might remove it later if no children match. This ensures parent-before-children
+    // ordering in schema_map.
```
mythrocks, line 214: 👏

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -221,21 +231,36 @@ class column_pruner {
...
+        // If the size of schema_map has increased, it means that child was successfully
+        // included
+        if (schema_map.size() > before_child_size) {
+          ++matched_children_count;
+          // Update the child count for the struct
+          schema_num_children[struct_map_position]++;
+        }
       } else {
-        // No match was found so skip the child.
+        // No match was found so skip the child
```
mythrocks, line 253: Absolute nit:  We are inconsistent with punctuation, in this change.  Note that line#235 has a period, but we opt to remove the period here.

My vote would be to punctuate deliberately, and add the period back.

src/main/cpp/src/NativeParquetJni.cpp
```diff
@@ -221,21 +231,36 @@ class column_pruner {
       auto found       = children.find(name);
 
       if (found != children.end()) {
-        // found a match so update the number of children that passed the filter and ask it to
-        // filter itself.
-        ++schema_num_children[our_num_children_index];
+        // Record the current size of schema_map before processing the child
+        // This helps determine if the child was retained after processing.
+        std::size_t before_child_size = schema_map.size();
```
mythrocks, line 236: Nit:  We can get away with making this `const`.
```suggestion
        auto const before_child_size = schema_map.size();
```

mythrocks (dismissed):
Good work, @nartal1.  LGTM!  A couple of minor nits picked.

nartal1:
>  Are there more changes needed on the java side? There were issues there too, if I remember correctly.

Thanks @revans2 for the review. For struct schema evolution issues, we don't need any changes on the java side. I tested with multiple nested structs and it works fine. We can enable the tests mentioned in the description once this PR is merged and we have a new jar including this.
I suppose there could still be some issues if there are maps or lists involved i.e empty structs within the maps/lists but have to do some testing to be certain.

src/main/cpp/src/NativeParquetJni.cpp
nartal1: Thanks @mythrocks for the review. Updated it. PTAL.

nartal1:
build

mythrocks (approved):
LGTM. Thanks for making the code changes suggested in the review.

PR #2759, opened 2025-01-13
Add an option to disable disable measuring kudo buffer copy.

liurenjie1024:
Part 2 of https://github.com/NVIDIA/spark-rapids/issues/11995 .

When used in shuffle cases with large partition numbers, each call to `System.nanoTime` is still quite expensive compared with a small memory copy. This pr adds an option to disable this measurement and disable it by default.


revans2 (commented):
Can we please tie this into visibility of the copy buffer time metric? If this is off, but the metric still shows up on the UI it will be confusing for customers.

binmahone (approved):

PR #2760, opened 2025-01-13
Add support for `org.apache.spark.sql.catalyst.expressions.Bin`

ustcfy:
Contribute to https://github.com/NVIDIA/spark-rapids/issues/11648

This PR adds support for `org.apache.spark.sql.catalyst.expressions.Bin`.
What it does is taking a `long` and outputting its binary representation.

### DIFF
--- a/src/main/cpp/src/cast_long_to_binary_string.cu
+++ b/src/main/cpp/src/cast_long_to_binary_string.cu
@@ -41,7 +41,7 @@ struct long_to_binary_string_fn {
     auto const value = d_longs.element<int64_t>(idx);
     char* d_buffer   = d_chars + d_offsets[idx];
     for (auto i = d_sizes[idx] - 1; i >= 0; --i) {
-      *d_buffer++ = '0' + ((value & (1LL << i)) >> i);
+      *d_buffer++ = '0' + ((value & (1UL << i)) >> i);
     }
   }
 
--- a/src/main/cpp/src/cast_long_to_binary_string.cu
+++ b/src/main/cpp/src/cast_long_to_binary_string.cu
@@ -41,9 +41,7 @@ struct long_to_binary_string_fn {
 
   __device__ cudf::size_type compute_output_size(LongType value)
   {
-    auto const size = 64 - __clzll(value);
-    // If the value is 0, the size should be 1
-    return size > 0 ? size : 1;
+    return max(64 - __clzll(value), 1);  // If the value is 0, the output size should be 1
   }
 
   __device__ void long_to_binary_string(cudf::size_type idx)
--- a/src/main/cpp/src/cast_long_to_binary_string.cu
+++ b/src/main/cpp/src/cast_long_to_binary_string.cu
@@ -41,7 +41,7 @@ struct long_to_binary_string_fn {
     auto const value = d_longs.element<int64_t>(idx);
     char* d_buffer   = d_chars + d_offsets[idx];
     for (auto i = d_sizes[idx] - 1; i >= 0; --i) {
-      *d_buffer++ = value & (1LL << i) ? '1' : '0';
+      *d_buffer++ = '0' + ((value & (1LL << i)) >> i);
     }
   }
 
--- a/src/main/cpp/src/cast_long_to_binary_string.cu
+++ b/src/main/cpp/src/cast_long_to_binary_string.cu
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cast_string.hpp"
+
+#include <cudf/column/column_device_view.cuh>
+#include <cudf/column/column_factories.hpp>
+#include <cudf/detail/null_mask.hpp>
+#include <cudf/detail/nvtx/ranges.hpp>
+#include <cudf/strings/detail/strings_children.cuh>
+#include <cudf/utilities/default_stream.hpp>
+#include <cudf/utilities/type_dispatcher.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/exec_policy.hpp>
+
+namespace spark_rapids_jni {
+
+namespace detail {
+namespace {
+
+template <typename LongType>
+struct long_to_binary_string_fn {
+  cudf::column_device_view d_longs;
+  cudf::size_type* d_sizes;
+  char* d_chars;
+  cudf::detail::input_offsetalator d_offsets;
+
+  __device__ cudf::size_type compute_output_size(LongType value)
+  {
+    auto const size = 64 - __clzll(value);
+    // If the value is 0, the size should be 1
+    return size > 0 ? size : 1;
+  }
+
+  __device__ void long_to_binary_string(cudf::size_type idx)
+  {
+    auto const value = d_longs.element<LongType>(idx);
+    char* d_buffer   = d_chars + d_offsets[idx];
+    for (auto i = d_sizes[idx] - 1; i >= 0; --i) {
+      *d_buffer++ = value & (1LL << i) ? '1' : '0';
+    }
+  }
+
+  __device__ void operator()(cudf::size_type idx)
+  {
+    if (d_longs.is_null(idx)) {
+      if (d_chars == nullptr) { d_sizes[idx] = 0; }
+      return;
+    }
+    if (d_chars != nullptr) {
+      long_to_binary_string(idx);
+    } else {
+      d_sizes[idx] = compute_output_size(d_longs.element<LongType>(idx));
+    }
+  }
+};
+
+struct dispatch_long_to_binary_string_fn {
+  template <typename LongType, CUDF_ENABLE_IF(std::is_same_v<LongType, std::int64_t>)>
+  std::unique_ptr<cudf::column> operator()(cudf::column_view const& input,
+                                           rmm::cuda_stream_view stream,
+                                           rmm::device_async_resource_ref mr) const
+  {
+    auto const d_column = cudf::column_device_view::create(input, stream);
+
+    auto [offsets, chars] = cudf::strings::detail::make_strings_children(
+      long_to_binary_string_fn<LongType>{*d_column}, input.size(), stream, mr);
+
+    return cudf::make_strings_column(input.size(),
+                                     std::move(offsets),
+                                     chars.release(),
+                                     input.null_count(),
+                                     cudf::detail::copy_bitmask(input, stream, mr));
+  }
+
+  template <typename LongType, CUDF_ENABLE_IF(not std::is_same_v<LongType, std::int64_t>)>
+  std::unique_ptr<cudf::column> operator()(cudf::column_view const&,
+                                           rmm::cuda_stream_view,
+                                           rmm::device_async_resource_ref) const
+  {
+    CUDF_FAIL("Values for long_to_binary_string function must be a long type.");
+  }
+};
+
+}  // namespace
+
+std::unique_ptr<cudf::column> long_to_binary_string(cudf::column_view const& input,
+                                                    rmm::cuda_stream_view stream,
+                                                    rmm::device_async_resource_ref mr)
+{
+  if (input.is_empty()) return cudf::make_empty_column(cudf::type_id::STRING);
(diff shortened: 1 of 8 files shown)


src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -0,0 +1,120 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cast_string.hpp"
+
+#include <cudf/column/column_device_view.cuh>
+#include <cudf/column/column_factories.hpp>
+#include <cudf/detail/null_mask.hpp>
+#include <cudf/detail/nvtx/ranges.hpp>
```
thirtiseven, line 22: Useless include (not sure)?

src/main/cpp/tests/cast_long_to_binary_string.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+using namespace cudf;
+
+constexpr cudf::test::debug_output_level verbosity{cudf::test::debug_output_level::FIRST_ERROR};
+
+struct LongToBinaryStringTests : public cudf::test::BaseFixture {};
+
+TEST_F(LongToBinaryStringTests, FromLongToBinary)
+{
+  auto const longs = cudf::test::fixed_width_column_wrapper<int64_t>{0L, 1L, 10L, -0L, -1L};
```
thirtiseven, line 34: Could we add some edge case like null, LONG_MAX, LONG_MIN?

thirtiseven (commented):
LGTM, love to see some Java tests

src/main/cpp/tests/cast_long_to_binary_string.cpp
binmahone: do we have a test case at plugin level to make sure Bin(13.3) returns 1101 ?

src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -0,0 +1,120 @@
...
+  cudf::column_device_view d_longs;
+  cudf::size_type* d_sizes;
+  char* d_chars;
+  cudf::detail::input_offsetalator d_offsets;
+
+  __device__ cudf::size_type compute_output_size(LongType value)
+  {
+    auto const size = 64 - __clzll(value);
+    // If the value is 0, the size should be 1
+    return size > 0 ? size : 1;
```
thirtiseven, line 46: nit: we can first check if the value == 0 to save a `__clzll` call in this case.

src/main/cpp/tests/cast_long_to_binary_string.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+  auto results = spark_rapids_jni::long_to_binary_string(longs, cudf::get_default_stream());
+
+  auto const expected = cudf::test::strings_column_wrapper{
+    "0", "1", "1010", "0", "1111111111111111111111111111111111111111111111111111111111111111"};
```
wjxiz1992, line 39: there you go: https://github.com/apache/spark/blob/master/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/MathExpressionsSuite.scala#L470-L490.

src/main/cpp/tests/cast_long_to_binary_string.cpp
revans2: This is really a Spark test not a good test for binary ops. Spark only accepts a Long as the input to `bin`

https://github.com/apache/spark/blob/3569e768e657d4e28ee7520808ec910cdff2b099/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/mathExpressions.scala#L1010

So any floating point input gets a cast to long inserted in before bin is called. So that test is really a test that Spark is doing the right thing. Even then it would probably be something that we would want to put in the integration tests if we did test it at all. 

Note that you can also pass in a string as an input and it will still try to cast it to a long before calling `bin`.

revans2 (dismissed):
The main thing for me is going to be performance. Do you have any numbers comparing the CPU to the GPU? If you want to wait for a plugin patch that is fine. But we need to do it somewhere.

src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -0,0 +1,120 @@
...
+#include <rmm/exec_policy.hpp>
+
+namespace spark_rapids_jni {
+
+namespace detail {
+namespace {
+
+template <typename LongType>
+struct long_to_binary_string_fn {
+  cudf::column_device_view d_longs;
```
sperlingxx, line 37: I don't think we have to copy the view. It seems all right if we just pass the pointer into the functor.

src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -0,0 +1,120 @@
...
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/exec_policy.hpp>
+
+namespace spark_rapids_jni {
+
+namespace detail {
+namespace {
+
+template <typename LongType>
```
sperlingxx, line 35: Is it really necessary to keep this template variable? Since the functor only supports `int64_t`

src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -0,0 +1,120 @@
...
+    // If the value is 0, the size should be 1
+    return size > 0 ? size : 1;
+  }
+
+  __device__ void long_to_binary_string(cudf::size_type idx)
+  {
+    auto const value = d_longs.element<LongType>(idx);
+    char* d_buffer   = d_chars + d_offsets[idx];
+    for (auto i = d_sizes[idx] - 1; i >= 0; --i) {
+      *d_buffer++ = value & (1LL << i) ? '1' : '0';
```
sperlingxx, line 54: `*d_buffer++ = '0' + ((value & (1LL << i)) >> i);` perhaps this approach is more efficient since it avoids branching, which might degrade performance on GPUs with warp divergence.

sperlingxx: But I am not sure if it is a good practice which is really effective. I would like to hear your opinions on this issue @res-life @ttnghia .

src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  __device__ void operator()(cudf::size_type idx)
+  {
+    if (d_longs.is_null(idx)) {
```
sperlingxx, line 60: NIT: Just some nice-to-have improvement, use `constexpr if` instead of `if` and add an extra template variable `nullable` for this functor. Because we already knew whether the column_view is nullable or NOT.

src/main/cpp/src/cast_long_to_binary_string.cu
thirtiseven: maybe `*d_buffer++ = '0' + ((value & (1LL << i)) != 0);`? It will be (very slightly) cheaper and easier to read.

sperlingxx: @thirtiseven 
Yes, this one is also a Branch-Free expression since the compiler shall use `setne` instruction avoids branching by directly setting a register based on the zero flag (ZF):
```assembly
cmp rax, 0
setne al
and al, 1
add eax, 48
```
The corresponding codes of my alternative would be translated into:
```assembly
sar     rax, cl
add     rax, 48
```

src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -0,0 +1,118 @@
...
+}
+
+}  // namespace detail
+
+// external API
+std::unique_ptr<cudf::column> long_to_binary_string(cudf::column_view const& input,
+                                                    rmm::cuda_stream_view stream,
+                                                    rmm::device_async_resource_ref mr)
+{
+  CUDF_FUNC_RANGE();
```
res-life, line 114: Assert input is long type here?

src/main/cpp/src/cast_long_to_binary_string.cu
revans2: But wouldn't that introduce thread divergence? I understand how that can be an advantage on a CPU, but I don't really see it on a GPU.

ustcfy: The PTX code translated from the code is as follows:
``` C++
auto const size = 64 - __clzll(value);
return size > 0 ? size : 1;

// PTX
clz.b64         %r1, %rd1;           // %r1: number of leading zeros in %rd1 (value)
mov.u32         %r2, 64;             // %r2: constant value 64
sub.s32         %r3, %r2, %r1;       // %r3: size = 64 - %r1
max.s32         %r4, %r3, 1;         // %r4: max(size, 1) , the return value
```
``` C++
if (!value) return 1;
return 64 - __clzll(value);

// PTX
setp.eq.s64     %p1, %rd1, 0;        // %p1: set to true if %rd1 (value) is zero
clz.b64         %r1, %rd1;           // %r1: number of leading zeros in %rd1 (value)
mov.u32         %r2, 64;             // %r2: constant value 64
sub.s32         %r3, %r2, %r1;       // %r3: size = 64 - %r1
selp.b32        %r4, 1, %r3, %p1;    // %r4: if %p1 is true, return 1; otherwise, return %r3 (final size)
```

thirtiseven: > But wouldn't that introduce thread divergence? I understand how that can be an advantage on a CPU, but I don't really see it on a GPU.

Yes, I was thinking there must be an if else to check size == 0 case so we can put it earlier than clz to save some calls without introducing new branch. But it looks like the compiler will optimise the `size > 0 ? size : 1;` to a `max.s32` so it's branch less then the original approach looks better in anyway.

ustcfy: I have now changed it to this:` return max(64 - __clzll(value), 1);`

res-life:
About the performance, we may try the following approach.
1.  Use warp level parallel, this means one warp(32 threads) handles a long value. Currently one thread handles a long value. 
2. Use unrolling approach:
```cpp
for (auto i = d_sizes[idx] - 1; i >= 0; --i) {
      *d_buffer++ = value & (1LL << i) ? '1' : '0';
```
==>>
```cpp
create a buffer in the stack:
char buf[64];

// totally have 64 lines:
buf[0] = value & (1LL << 0) ? '1' : '0';
buf[1] = value & (1LL << 1) ? '1' : '0';
buf[2] = value & (1LL << 2) ? '1' : '0';
...
buf[62] = value & (1LL << 62) ? '1' : '0';
buf[63] = value & (1LL << 63) ? '1' : '0';
// some of the above lines is useless, but keep them to avoid GPU divergence. 

// write out by 4 chars
num_chars = x
while(num_chars > 4) {
  int* d_int_buf = (int*)(d_buffer)
  int* int_buf = (int*)buf
  d_int_buf[i] = int_buf[i]
  num_chars -= 4
}
write out left chars

```

res-life:
I found an example, maybe it's useful:
```cpp
void intToBin(uint64_t i, char* out) {
	assert((out % sizeof(uint64_t)) == 0); // check alignment
	uint64_t zeros   = 0x3030303030303030ULL; // "00000000"
	uint64_t mask    = 0x0101010101010101ULL;
	uint64_t *out_64 = reinterprent_cast<uint64_t*>(out);

	uint64_t bit_index = 0;
	_BitScanForward64(&bit_index, i)
	i = i << bit_index;

	out_64[0] = zeros | _pdep_u64(i >> 56, mask);
	out_64[1] = zeros | _pdep_u64(i >> 48, mask);
	out_64[2] = zeros | _pdep_u64(i >> 40, mask);
	out_64[3] = zeros | _pdep_u64(i >> 32, mask);
	out_64[4] = zeros | _pdep_u64(i >> 24, mask);
	out_64[5] = zeros | _pdep_u64(i >> 16, mask);
	out_64[6] = zeros | _pdep_u64(i >>  8, mask);
	out_64[7] = zeros | _pdep_u64(i >>  0, mask);
	
	_BitScanReverse64(&bit_index, i)
	out[64 - bit_index] = '\0';
}
```

CUDA supports _pdep_u64

src/main/cpp/src/cast_long_to_binary_string.cu
ustcfy: > `*d_buffer++ = '0' + ((value & (1LL << i)) >> i);` perhaps this approach is more efficient since it avoids branching, which might degrade performance on GPUs with warp divergence.

Yes, this approach is more efficient.
image

ustcfy: > maybe `*d_buffer++ = '0' + ((value & (1LL << i)) != 0);`? It will be (very slightly) cheaper and easier to read.

This will generate the same PTX as my code.

revans2:
build

src/main/cpp/src/cast_long_to_binary_string.cu
ustcfy: > `*d_buffer++ = '0' + ((value & (1LL << i)) >> i);` perhaps this approach is more efficient since it avoids branching, which might degrade performance on GPUs with warp divergence.

Oh, there's a bug here when the highest bit is 1.

ustcfy:
build

thirtiseven (approved):

res-life (approved):
LGTM

PR #2763, opened 2025-01-14
Add zero copy data writer for kudo.

liurenjie1024:
1. Add `reserve` method in `DataWriter` so that kudo data writer could reserve memory before actual writing happens. This helps avoiding unnecessay allocation and copy.
2. Add `OpenByteArrayOutputStream` and its corresponding data writer. `OpenByteArrayOutputStream` could be used in customized shuffle manager to save memory.
3. Add `ByteArrayOutputStreamWriter` which helps saving memory copy without introducing any changes to shuffle manager.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/kudo/OpenByteArrayOutputStream.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/kudo/OpenByteArrayOutputStream.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


src/main/java/com/nvidia/spark/rapids/jni/kudo/OpenByteArrayOutputStream.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Increases the capacity if necessary to ensure that it can hold
+   * at least the number of elements specified by the minimum
+   * capacity argument.
+   *
+   * <br/>
+   *
+   * This code is copied from jdk's implementation.
```
binmahone, line 73: should we state at the begining of the class that most of the class code is copied from jdk, and highlight what is added on ? for both copywrite and mainenance considerations

binmahone (commented):
LGTM

sameerz:
Closing this PR as will not merge.  @liurenjie1024 please redo without GPL code.

PR #2781, opened 2025-01-21
Optimize `long_to_binary_string` function using warp parallel

ustcfy:
This PR optimizes the performance of the `long_to_binary_string` function. 

The main optimization is to split `make_strings_children` into two kernels: one for calculating the length of each string and another for filling in the string content.

In the original implementation, filling the strings used a loop, with loop iterations potentially reaching up to 64 times (the number of bits used by `long`), which is inefficient on the GPU. In the current implementation, each string is processed using multiple threads (currently 32).

The benchmark results are as follows:
```
|  num_rows  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |         Diff |   %Diff |  Status  |
|------------|------------|-------------|------------|-------------|--------------|---------|----------|
|   100000   | 374.211 us |      17.21% | 250.415 us |      17.28% |  -123.796 us | -33.08% |   FAST   |
|   500000   |   1.155 ms |       6.16% | 739.668 us |       6.57% |  -415.445 us | -35.97% |   FAST   |
|  1000000   |   2.031 ms |      23.46% |   1.373 ms |       4.14% |  -657.439 us | -32.38% |   FAST   |
|  5000000   |   9.525 ms |       2.42% |   6.599 ms |       5.68% | -2926.032 us | -30.72% |   FAST   |
|  10000000  |  18.042 ms |       2.46% |  12.400 ms |      22.33% | -5641.755 us | -31.27% |   FAST   |
```

### DIFF
--- a/src/main/cpp/src/cast_long_to_binary_string.cu
+++ b/src/main/cpp/src/cast_long_to_binary_string.cu
@@ -43,10 +43,10 @@ CUDF_KERNEL void compute_output_size_kernel(cudf::column_device_view d_longs,
   d_sizes[tid] = max(1, 64 - __clzll(d_longs.element<int64_t>(tid)));
 }
 
+template <int num_threads_per_row>
 CUDF_KERNEL void long_to_binary_string_kernel(cudf::column_device_view d_longs,
                                               char* d_chars,
-                                              cudf::detail::input_offsetalator d_offsets,
-                                              cudf::size_type num_threads_per_row)
+                                              cudf::detail::input_offsetalator d_offsets)
 {
   auto const tid     = cudf::detail::grid_1d::global_thread_id();
   auto const row_idx = tid / num_threads_per_row;
@@ -55,10 +55,10 @@ CUDF_KERNEL void long_to_binary_string_kernel(cudf::column_device_view d_longs,
   auto const str_len = d_offsets[row_idx + 1] - d_offsets[row_idx];
   if (str_len == 0) { return; }
 
-  auto const value               = d_longs.element<int64_t>(row_idx);
-  auto const lane_idx            = tid % num_threads_per_row;
-  auto const num_bits_per_thread = 64 / num_threads_per_row;
-  auto const first_byte_index    = lane_idx * num_bits_per_thread;
+  constexpr int num_bits_per_thread = 64 / num_threads_per_row;
+  auto const value                  = d_longs.element<int64_t>(row_idx);
+  auto const lane_idx               = tid % num_threads_per_row;
+  auto const first_byte_index       = lane_idx * num_bits_per_thread;
 
   char* d_buffer = d_chars + d_offsets[row_idx];
   for (auto byte_index = first_byte_index;
@@ -103,11 +103,11 @@ std::unique_ptr<cudf::column> long_to_binary_string(cudf::column_view const& inp
   char* d_chars = chars.data();
 
   // Fill in the chars data
-  auto constexpr num_threads_per_row = 32;
+  constexpr int num_threads_per_row = 32;
   auto new_grid = cudf::detail::grid_1d{strings_count * num_threads_per_row, block_size};
   if (bytes > 0) {
-    long_to_binary_string_kernel<<<new_grid.num_blocks, block_size, 0, stream.value()>>>(
-      *d_column, d_chars, d_offsets, num_threads_per_row);
+    long_to_binary_string_kernel<num_threads_per_row>
+      <<<new_grid.num_blocks, block_size, 0, stream.value()>>>(*d_column, d_chars, d_offsets);
   }
 
   return cudf::make_strings_column(input.size(),
--- a/src/main/cpp/src/cast_long_to_binary_string.cu
+++ b/src/main/cpp/src/cast_long_to_binary_string.cu
@@ -61,9 +61,9 @@ CUDF_KERNEL void long_to_binary_string_kernel(cudf::column_device_view d_longs,
   auto const first_byte_index    = lane_idx * num_bits_per_thread;
 
   char* d_buffer = d_chars + d_offsets[row_idx];
-  for (auto i = 0; i < num_bits_per_thread; ++i) {
-    auto const byte_index = first_byte_index + i;
-    if (byte_index >= str_len) { return; }
+  for (auto byte_index = first_byte_index;
+       byte_index < first_byte_index + num_bits_per_thread && byte_index < str_len;
+       ++byte_index) {
     auto const num_shifts = str_len - 1 - byte_index;
     d_buffer[byte_index]  = '0' + ((value & (1UL << num_shifts)) >> num_shifts);
   }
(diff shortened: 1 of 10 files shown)


src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -30,35 +30,44 @@ namespace spark_rapids_jni {
...
-      if (d_chars == nullptr) { d_sizes[idx] = 0; }
-      return;
-    }
-    if (d_chars != nullptr) {
-      long_to_binary_string(idx);
-    } else {
-      // If the value is 0, the size should be 1
-      d_sizes[idx] = max(1, 64 - __clzll(d_longs.element<int64_t>(idx)));
-    }
+CUDF_KERNEL void long_to_binary_string_kernel(cudf::column_device_view d_longs,
```
res-life, line 46: num_threads_per_row is a const value, we can make num_threads_per_row as a template parameter.

src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -30,35 +30,44 @@ namespace spark_rapids_jni {
...
+  auto const tid     = cudf::detail::grid_1d::global_thread_id();
+  auto const row_idx = tid / num_threads_per_row;
+  if (row_idx >= d_longs.size()) { return; }
+
+  auto const str_len = d_offsets[row_idx + 1] - d_offsets[row_idx];
+  if (str_len == 0) { return; }
+
+  auto const value               = d_longs.element<int64_t>(row_idx);
+  auto const lane_idx            = tid % num_threads_per_row;
+  auto const num_bits_per_thread = 64 / num_threads_per_row;
```
res-life, line 60:
```suggestion
  constexpr int num_bits_per_thread = 64 / num_threads_per_row;
```

src/main/cpp/src/cast_long_to_binary_string.cu
res-life, line 46: Add comment to describe this kernel, like:
Cast long to string using warp parallell algorithm, each warp hanles one long value, each thread handles two bits of long.

res-life:
LGTM, Only some nits.

src/main/cpp/src/cast_long_to_binary_string.cu
```diff
@@ -30,35 +30,44 @@ namespace spark_rapids_jni {
...
+  auto const value                  = d_longs.element<int64_t>(row_idx);
+  auto const lane_idx               = tid % num_threads_per_row;
+  auto const first_byte_index       = lane_idx * num_bits_per_thread;
+
+  char* d_buffer = d_chars + d_offsets[row_idx];
+  for (auto byte_index = first_byte_index;
+       byte_index < first_byte_index + num_bits_per_thread && byte_index < str_len;
+       ++byte_index) {
+    auto const num_shifts = str_len - 1 - byte_index;
+    d_buffer[byte_index]  = '0' + ((value & (1UL << num_shifts)) >> num_shifts);
```
res-life, line 68: 1UL or 1ULL?
From chatGPT:
```
Size of Types
unsigned long (UL):
Typically, unsigned long is 4 bytes (32 bits) on many platforms, especially on 32-bit systems.
On 64-bit systems, it is commonly 8 bytes (64 bits), but this can vary based on the compiler and platform.
unsigned long long (ULL):
unsigned long long is guaranteed to be at least 8 bytes (64 bits) on all platforms, as per the C++ standard.
```

res-life (dismissed):
LGTM

ustcfy:
build

res-life:
The main process is similar to [make_strings_children](https://github.com/rapidsai/cudf/blob/v24.12.00/cpp/include/cudf/strings/detail/strings_children.cuh#L220)

ustcfy:
build

ustcfy:
@revans2

ustcfy:
build

res-life (approved):

res-life:
For the current warp parallel approach, the 32 threads first handles 32 bits: [0,2,4,....62] concurrently, then [1,3,5,7...63].
Not sure if the following approach get improvement: 32 threads first handling bits [0,32), then bits [32, 64), in this way, GPU may reduce the divergence.

ustcfy:
> For the current warp parallel approach, the 32 threads first handle 32 bits: [0,2,4,....62] concurrently, then [1,3,5,7...63]. Not sure if the following approach gets improvement: 32 threads first handle bits [0,32), then bits [32, 64), in this way, GPU may reduce the divergence.

It seems there is not much difference. 🤔
image

sperlingxx (approved):
LGTM! Great attempt!

sperlingxx:
AS a beginner of CUDA programing, I tried another approach which assign one entire byte for each CUDA thread, which achieved nearly same performance as the approach introduced by this MR in the newly-added benchmark:

```CUDA
CUDF_KERNEL void long_to_binary_string_kernel2(cudf::size_type row_size,
                                               char const* __restrict__ long_data,
                                               char* __restrict__ d_chars,
                                               cudf::detail::input_offsetalator d_offsets)
{
  auto const tid       = cudf::detail::grid_1d::global_thread_id();
  auto const row_idx   = tid >> 3;
  auto const byte_idx  = tid % 8;
  auto const bit_size  = d_offsets[row_idx + 1] - d_offsets[row_idx];
  auto const byte_size = (bit_size + 7) >> 3;

  if (row_idx >= row_size || byte_idx >= byte_size) {
    return;
  }

  char const working_byte = *(long_data + tid);
  char *d_buffer = d_chars + d_offsets[row_idx];
  if (bit_size % 8 != 0 && byte_idx == byte_size - 1) {
    auto const lp_size = bit_size % 8;
    for (auto i = 0; i < lp_size; i++) {
      auto const mv_bits = lp_size - 1 - i;
      d_buffer[i] = '0' + ((working_byte & (1 << mv_bits)) >> mv_bits);
    }
  } else {
    d_buffer    += bit_size - 8 - (byte_idx << 3);
    d_buffer[0]  = '0' + ((working_byte & 128) >> 7);
    d_buffer[1]  = '0' + ((working_byte & 64) >> 6);
    d_buffer[2]  = '0' + ((working_byte & 32) >> 5);
    d_buffer[3]  = '0' + ((working_byte & 16) >> 4);
    d_buffer[4]  = '0' + ((working_byte & 8) >> 3);
    d_buffer[5]  = '0' + ((working_byte & 4) >> 2);
    d_buffer[6]  = '0' + ((working_byte & 2) >> 1);
    d_buffer[7]  = '0' + (working_byte & 1);
  }
}



// Caller
  auto new_grid = cudf::detail::grid_1d{strings_count << 3, block_size};
  if (bytes > 0) {
    long_to_binary_string_kernel2<<<new_grid.num_blocks, block_size, 0, stream.value()>>>(
      strings_count,
      d_column->data<char>(),
      d_chars,
      d_offsets);
  }
```

revans2 (approved):

revans2:
@sperlingxx if you want to put up another PR with your change in 25.04 that is fine with me. I merged this one in so that it would make it into 25.02

PR #2799, opened 2025-01-24
Port of cudf::contiguous_split into spark_rapids_jni

nvdbaranec:
This is a setup PR for the implementation of `shuffle_split` and `shuffle_assemble`. The split half starts with this version of `cudf::contiguous_split` as a basis, so for ease of diffs with the main PR I'm bringing this over separately.  This is not called or in any way used in `spark_rapids_jni` at the moment.

### DIFF
--- a/src/main/cpp/src/shuffle_split.cu
+++ b/src/main/cpp/src/shuffle_split.cu
@@ -1,0 +1,1292 @@
+/*
+ * Copyright (c) 2024, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cudf/column/column_device_view.cuh>
+#include <cudf/column/column_view.hpp>
+#include <cudf/contiguous_split.hpp>
+#include <cudf/detail/contiguous_split.hpp>
+#include <cudf/detail/copy.hpp>
+#include <cudf/detail/iterator.cuh>
+#include <cudf/detail/null_mask.hpp>
+#include <cudf/detail/nvtx/ranges.hpp>
+#include <cudf/detail/utilities/cuda.cuh>
+#include <cudf/lists/lists_column_view.hpp>
+#include <cudf/structs/structs_column_view.hpp>
+#include <cudf/table/table_view.hpp>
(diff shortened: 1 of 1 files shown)


src/main/cpp/src/shuffle_split.cu
```diff
@@ -0,0 +1,1292 @@
+/*
+ * Copyright (c) 2024, NVIDIA CORPORATION.
```
sameerz, line 2: Update copyright

revans2: You also need to sign your PR.

src/main/cpp/src/shuffle_split.cu
```diff
@@ -0,0 +1,1292 @@
...
+ */
+struct dst_buf_info {
+  // constant across all copy commands for this buffer
+  std::size_t buf_size;  // total size of buffer, including padding
+  int num_elements;      // # of elements to be copied
+  int element_size;      // size of each element in bytes
+  int num_rows;  // # of rows to be copied(which may be different from num_elements in the case of
+                 // validity or offset buffers)
+
+  int src_element_index;   // element index to start reading from from my associated source buffer
```
revans2, line 109: nit: A extra "from" was added to this comment.

src/main/cpp/src/shuffle_split.cu
```diff
@@ -0,0 +1,1292 @@
...
+};
+
+/**
+ * @brief Copy a single buffer of column data, shifting values (for offset columns),
+ * and validity (for validity buffers) as necessary.
+ *
+ * Copies a single partition of a source column buffer to a destination buffer. Shifts
+ * element values by value_shift in the case of a buffer of offsets (value_shift will
+ * only ever be > 0 in that case).  Shifts elements bitwise by bit_shift in the case of
+ * a validity buffer (bif_shift will only ever be > 0 in that case).  This function assumes
```
revans2, line 126: nit: bif_shift should be bit_shift

revans2 (approved):
I am no expert on this code, so all I did was compare it to the version in cudf. 

I assume the version we should compare against is https://github.com/rapidsai/cudf/blob/7575e8da54499990b51535a0f975acd02a493144/cpp/src/copying/contiguous_split.cu

Is that correct?

I did notice a few fixes that are in the latest version that are not in this one, but they are all nits so you can ignore them.

abellina:
The port of contiguous split here doesn't make use of the MR we introduced for chunked pack. I do wonder if it would make sense to pipe it here, or if there is a reason not to use it? In the future, I could imagine us using the kudo format for spill under "no memory" circumstances, where we do not want to go to RMM for an allocation in the middle of spill.

nvdbaranec:
> The port of contiguous split here doesn't make use of the MR we introduced for chunked pack. I do wonder if it would make sense to pipe it here, or if there is a reason not to use it? In the future, I could imagine us using the kudo format for spill under "no memory" circumstances, where we do not want to go to RMM for an allocation in the middle of spill.

Jason and I thought it would easier to go with this version as the code is considerably simpler.  We could add it later if need be, but the cranked contiguous-split is quite a bit more complex.

revans2 (approved):

revans2:
build

PR #2800, opened 2025-01-24
GPU implementation of the Kudo shuffle format.

nvdbaranec:
This PR provides the GPU implementation of the split and assemble halves of the Kud0 shuffle format.  It is dependent on https://github.com/NVIDIA/spark-rapids-jni/pull/2799. That PR exists solely to help reduce the amount of code that needs to get reviewed for this PR.  `shuffle_split.cu` should be a little easier to review once it gets merged.

`shuffle_split` implements the sender side half of the Kud0 format.  Splitting an input table up into the final output binary and associated metadata.  `shuffle_split.cu` is where it is implemented, and is based on `cudf::contiguous_split`

`shuffle_assemble` implements the receiver side of the process - reassembling one or more blobs produced by one or more `shuffle_split` calls.

This is feature complete and in a reviewable state, but there is a known issue with `struct<list>` and `struct<string>` columns that I have to fix.  I don't expect it to alter the code much. I also need to add more tests.  But in the interest of getting reviews going, I'm putting this up in draft form.


src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,96 @@
...
+  }
+};
+
+/**
+ * @brief Header for each partition.
+ */
+struct partition_header {
+  uint32_t magic_number;
+  uint32_t version;
+  uint64_t offset;
```
revans2, line 60: In the java code this is a uint32_t not a unit64_t. Again I would prefer it if we could have the code match for both to start out with, and then we can make adjustments afterwards. I am fine if we have to throw an exception if the input does not fit.

https://github.com/NVIDIA/spark-rapids-jni/blob/74e33d948442d72cd45dee2f3e062603d4953284/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableHeader.java#L73

https://github.com/NVIDIA/spark-rapids-jni/blob/74e33d948442d72cd45dee2f3e062603d4953284/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoSerializer.java#L82

src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,96 @@
...
+  uint64_t num_rows;
+  uint64_t validity_size;
+  uint64_t offset_size;
+  uint64_t data_size;
```
revans2, line 64: All of these should be uint32_t not uint64_t if this is going to be compatible with the java code.

src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,96 @@
...
+};
```
revans2, line 65: We appear to be missing a `hasValidityBuffer` entry at the end here. I know it is variable length (based on the number of columns), so is it just not returned to the user, or is it not a part of this code base yet?

https://github.com/NVIDIA/spark-rapids-jni/blob/74e33d948442d72cd45dee2f3e062603d4953284/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoSerializer.java#L112-L124

src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+// padding values for each validity type, as applied at the end of that data type
+// in each partition. so for example all of the grouped-together validity buffers for
+// a given partition will have a final 4 byte pad applied before the offset buffers begin
+constexpr size_t validity_pad = 4;
+constexpr size_t offset_pad   = 8;
```
revans2, line 71: The kudo serializer indicates that this and the data padding should be 4 bytes, not 8. Is there a reason for 8, like the decimal128?

https://github.com/NVIDIA/spark-rapids-jni/blob/74e33d948442d72cd45dee2f3e062603d4953284/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoSerializer.java#L129-L138

revans2 (commented):
I am no where near done with the review. I am planning on writing some JNI APIs for this and put together some kudo interoperability tests. Would you prefer to have them be a separate PR based on this one, or should I make a PR to your branch so they can be in one place?

src/main/cpp/src/shuffle_split.hpp
```diff
@@ -0,0 +1,156 @@
...
+{
+  return cudf::type_dispatcher(type, is_fixed_point_impl{});
+}
+
+// per-column information, stored in the metadata header
+struct shuffle_split_col_data {
+  cudf::type_id type;
+
+  shuffle_split_col_data(cudf::type_id _type, cudf::size_type _param)
+    : type(_type), _num_children(_param)
```
revans2, line 55: nit: This is a bit confusing and I have to do a double take to know what was happening with the union and all. A comment would be nice.

src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,96 @@
...
+  {
+    return sizeof(cudf::device_storage_type_t<T>);
+  }
+};
+
+/**
+ * @brief Header for each partition.
+ */
+struct partition_header {
+  uint32_t magic_number;
```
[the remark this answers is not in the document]
nvdbaranec: Exposing it in what sense? This struct is here because it's used by both the reader and writer side.

src/main/cpp/src/shuffle_split_detail.hpp
nvdbaranec: It's in there, it's just not part of the fixed-sized part of the header (the `partition_header` struct).

https://github.com/NVIDIA/spark-rapids-jni/blob/8a8e920f771529f9d353f03fdcd3c307adc4aac9/src/main/cpp/src/shuffle_split.cu#L642

src/main/cpp/src/shuffle_split_detail.hpp
nvdbaranec: Yeah it was just an optimal number I chose.  The original design doc didn't specify. I'll tweak them down - the code doesn't care: the offset copy only requires 4 bytes and the data copy uses the cub batched-memcpy which doesn't care at all. 

The above header sizes all came from the original design doc, so I guess that's a little behind. Easy tweaks.

src/main/cpp/src/shuffle_split_detail.hpp
nvdbaranec: Looks like I need to swizzle them to big-endian too.

src/main/cpp/src/shuffle_split.hpp
nvdbaranec: Done.

revans2:
@nvdbaranec your other PR was merged in. Could you upmerge so I can better see the changes.

nvdbaranec:
Upmerged. It...kind of helps I guess.  There's a few big obvious blocks of code that got removed. Lots of little changes spread throughout though.

src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,95 @@
...
+  uint32_t row_index;  // row index in the source table that this partition started at
+  uint32_t num_rows;
+  uint32_t validity_size;
+  uint32_t offset_size;
+  uint32_t data_size;
+};
+
+// padding values for each validity type, as applied at the end of that data type
+// in each partition. so for example all of the grouped-together validity buffers for
+// a given partition will have a final 4 byte pad applied before the offset buffers begin
```
hyperbolic2346, line 68: It isn't 4 bytes of padding as much as it is padded to align on a 4 byte boundary, right? I wonder if there is better wording we could use here.

src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,95 @@
...
+constexpr size_t validity_pad = 4;
+constexpr size_t offset_pad   = 4;
+constexpr size_t data_pad     = 4;
+
+/**
+ * @brief Compute per-partition metadata size.
+ */
+constexpr size_t compute_per_partition_metadata_size(size_t total_columns)
+{
+  auto const has_validity_length = (total_columns + 7) / 8;  // has-validity bit per column
```
hyperbolic2346, line 78: I believe you should prefer `bitmask_allocation_size_bytes()` here, but it isn't `constexpr`. Maybe use `div_rounding_up_safe` directly?

src/main/cpp/tests/shuffle_split.cu
```diff
@@ -0,0 +1,467 @@
...
+                                             "yellow",
+                                             "red",
+                                             "black",
+                                             "white",
+                                             "gray",
+                                             "aquamarine",
+                                             "mauve",
+                                             "ultraviolet"}};
+    cudf::test::strings_column_wrapper col2{
+      {"left", "up", "right", "down", "", "space", "", "delete", "end", "insert"}};
```
hyperbolic2346, line 180: I'm disturbed by your lack of Contra cheat code.

hyperbolic2346 (changes requested):
Nits and some comments/questions for now. Still working through this.

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1771 @@
...
+      auto& cinstance_info = column_instance_info[col_instance_index];
+
+      uint8_t const* const buf_start =
+        reinterpret_cast<uint8_t const*>(partitions + partition_offsets[partition_index]);
+      partition_header const* const pheader = reinterpret_cast<partition_header const*>(buf_start);
+
+      bitmask_type const* const has_validity_buf =
+        reinterpret_cast<bitmask_type const*>(buf_start + sizeof(partition_header));
+      cinstance_info.has_validity =
+        has_validity_buf[col_index / 32] & (1 << (col_index % 32)) ? 1 : 0;
```
hyperbolic2346, line 411: Here is that calculation again. Should we bump this out into a function?

src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * @brief Header for each partition.
+ */
+struct partition_header {
+  uint32_t magic_number;
+  uint32_t row_index;  // row index in the source table that this partition started at
+  uint32_t num_rows;
+  uint32_t validity_size;
+  uint32_t offset_size;
+  uint32_t data_size;
```
revans2, line 63: The last value in here is supposed to be the number of columns.
`uint32_t num_columns;` This feels redundant, seeing how we need the schema for this to be read back in. So should I remove it from the CPU code or should we add it here?

revans2 (commented):
Still looking at this. I have a very basic java API and I am just getting started to test conf compatibility with the two.

src/main/cpp/src/shuffle_split_detail.hpp
revans2: @liurenjie1024 the numColummns field in the header looks like it should be something that we do not need. Is there a reason we are including it instead of relying on the schema to get it?

revans2:
I am seeing another difference between this and the CPU code. It looks like this is padding the data after the header, whereas the CPU is not doing that. I think is because the header is parsed totally on the host, and then only the data part is sent to the GPU. I was able to remove the columnCount from the header, so not I am going to try and add in padding too. But I would like some feedback from @liurenjie1024 and @nvdbaranec on what we want a final solution to look like.

revans2:
@liurenjie1024 can you please explain to me why we have different padding for concat vs over the wire? It is very confusing to see that some are padded to 4 bytes and others are padded to 64 bytes.

nvdbaranec:
> @nvdbaranec can you please explain to me how padding works for this PR? It looks like there is padding after the header, but there is no padding in between columns? Is that right?

- Fixed-size component of the header
- (no padding)
- variable-length component of the header (has-validity bit for each column), written as bytes
- (no padding)
- all validity buffers for a partition, no padding between them
- (pad to 4 bytes)
- all offsets buffers for a partition. 4 byte padding implicitly between them
- (pad to 4 bytes)
- add data buffers for a partition. no padding between them
- (pad to 4 bytes). end of the partition

revans2:
> Doing 64 byte padding when concating is learned from jcudf, I thought it was a required for gpu? 

64-bytes was me being overly cautious.  Cuda malloc does 64-byte alignment (I think), because like malloc it does not know that data types so it cannot make any assumption about how it needs to be aligned. In reality we need each column to be aligned based on how it is accessed.

 * byte/bool = 1 byte aligned (no padding before it needed)
 * short = 2 byte aligned
 * int/float/timestamp_days/decimal_32 = 4 byte aligned
 * long/double/timestamp_micros/decimal_64/validity = 8 byte aligned
 * decimal_128 = 16 byte aligned

Offsets we could have as 4 byte aligned, but because CUDF supports longer offsets we might want to think about 8 byte aligned (unless we have a way to store what type of offset is being stored in the header)

In JCUDF serialization I thought about reordering the columns to reduce the amount of padding that would be needed, but I never did it because it didn't feel like it saved that much, but I wasn't looking at it from thousands of partitions.

@nvdbaranec how is the padding needed for the kernels in this patch to work? Would it be alright if there was no padding at all? Or is the data copied in a non-byte wise way?

src/main/cpp/tests/shuffle_split.cu
```diff
@@ -0,0 +1,583 @@
...
+      {static_cast<uint8_t*>(split_data.partitions->data()), split_data.partitions->size()},
+      split_data.offsets,
+      cudf::get_default_stream(),
+      rmm::mr::get_current_device_resource());
+
+    CUDF_TEST_EXPECT_TABLES_EQUAL(tbl, *result);
+  }
+}
+
+TEST_F(ShuffleSplitTests, Simple)
```
pmattione-nvidia, line 151: We need tests for bools and decimal32/64/128. Especially 128 if the data is really only going to be 8-byte-aligned.

revans2:
@nvdbaranec I was looking for the hasValidity part of the header and I am not finding it in this code, but it is in the CPU code, just like the numColumns part.

https://github.com/NVIDIA/spark-rapids-jni/blob/ef7bb9265d6ffec2e3a6bc6a52d415aa586b85e6/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableHeader.java#L79-L82

Is this intentional or is it an oversight?

src/main/cpp/src/shuffle_split.cu
```diff
@@ -770,212 +539,244 @@ struct dst_offset_output_iterator {
...
+  if (partition_index >= num_partitions) { return; }
+  auto const col_index = tid % threads_per_partition;
+
+  // start of the metadata buffer for this partition
+  uint8_t* buf_start        = out_buffer + out_buffer_offsets[partition_index];
+  partition_header* pheader = reinterpret_cast<partition_header*>(buf_start);
+
+  // first thread in each partition stores constant stuff
+  if (col_index == 0) {
+    pheader->magic_number = cudf::hashing::detail::swap_endian(magic);
```
pmattione-nvidia, line 622: Why are we doing endian swaps here?

revans2: We do it because that is what the CPU code does when it writes the values out. That is a relic of how java's DataOutputStream writes out values. I would prefer to see it dropped, but it should change in both implementations at the same time.

revans2:
> > @nvdbaranec can you please explain to me how padding works for this PR? It looks like there is padding after the header, but there is no padding in between columns? Is that right?
> 
> * Fixed-size component of the header
> * (no padding)
> * variable-length component of the header (has-validity bit for each column), written as bitpacked bytes
> * (no padding)
> * all validity buffers for a partition, no padding between them
> * (pad to 4 bytes)
> * all offsets buffers for a partition. 4 byte padding implicitly between them
> * (pad to 4 bytes)
> * all data buffers for a partition. no padding between them
> * (pad to 4 bytes). end of the partition

That sounds reasonable to me. The CPU version has the following padding

 * Fixed-size component of the header
 * (no padding)
 * variable-length component of the header (has-validity bit for each column), written as bitpacked bytes
 * (no padding)
 * all validity buffers for a partition, padded to 4 byte boundary at the end of each buffer
 * (no extra padding)
 * all offsets buffers for a partition. padded to 4 byte boundary at the end of each buffer 
 * (no extra padding)
 * all data buffers for a partition. padded to 4 byte boundary at the end of each buffer
 * (no extra padding). end of the partition

@nvdbaranec I modified your patch to have a num_columns field.

```diff
diff --git a/src/main/cpp/src/shuffle_split.cu b/src/main/cpp/src/shuffle_split.cu
index 92f7dfc92..4ecbb3eef 100644
--- a/src/main/cpp/src/shuffle_split.cu
+++ b/src/main/cpp/src/shuffle_split.cu
@@ -637,6 +637,8 @@ __global__ void pack_per_partition_metadata_kernel(uint8_t* out_buffer,
     pheader->offset_size =
       cudf::hashing::detail::swap_endian(static_cast<uint32_t>(psize.offset_size));
     pheader->data_size = cudf::hashing::detail::swap_endian(static_cast<uint32_t>(psize.data_size));
+    pheader->num_flattened_columns =
+      cudf::hashing::detail::swap_endian(static_cast<uint32_t>(columns_per_partition));
   }
 
   bitmask_type* has_validity =
diff --git a/src/main/cpp/src/shuffle_split_detail.hpp b/src/main/cpp/src/shuffle_split_detail.hpp
index da1841e80..d53cfeeb4 100644
--- a/src/main/cpp/src/shuffle_split_detail.hpp
+++ b/src/main/cpp/src/shuffle_split_detail.hpp
@@ -61,6 +61,7 @@ struct partition_header {
   uint32_t validity_size;
   uint32_t offset_size;
   uint32_t data_size;
+  uint32_t num_flattened_columns; // used by the CPU version, but not the GPU version
 };
 
 // padding values for each validity type, as applied at the end of that data type
@@ -92,4 +93,4 @@ enum class buffer_type { VALIDITY = 0, OFFSETS = 1, DATA = 2 };
 
 }  // namespace detail
 
-}  // namespace spark_rapids_jni
\ No newline at end of file
+}  // namespace spark_rapids_jni
```

I have been trying to modify the CPU code to do padding like you are doing, but it is proving to be a little difficult. I'll keep trying to see if I can make this work.

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1779 @@
...
+ */
+std::unique_ptr<cudf::table> shuffle_assemble(shuffle_split_metadata const& metadata,
+                                              cudf::device_span<uint8_t const> partitions,
+                                              cudf::device_span<size_t const> partition_offsets,
+                                              rmm::cuda_stream_view stream,
+                                              rmm::device_async_resource_ref mr)
+{
+  // if the input is empty, just generate an empty table
+  if (partition_offsets.size() == 1) { return build_empty_table(metadata.col_info, stream, mr); }
+
```
pmattione-nvidia, line 1745: Should we have a shortcut here for there being only one partition

pmattione-nvidia:
Has this been tested yet on complex customer data?

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1779 @@
...
+          break;
+        // all other types jump back up to the nearest struct/branch point, as they represent the
+        // end of a chain of children
+        case cudf::type_id::STRING:
+          col_inst.num_chars = offsets[num_rows] - offsets[0];
+          offsets += (num_rows + 1);
+          // fallthrough
+        default: {
+          if (rc_stack_pos >= 0) {
+            src_row_index           = rc_stack[(rc_stack_pos * 2) + 2];
```
pmattione-nvidia, line 273: Doesn't this need to be rc_stack_pos * 3 instead of * 2?

revans2:
@nvdbaranec I got an Illegal Memory Access issue when I added the following test (note it was in java, but I ported it to C++)

```diff
diff --git a/src/main/cpp/tests/shuffle_split.cpp b/src/main/cpp/tests/shuffle_split.cpp
index 97f68f0e3..0d9a59c66 100644
--- a/src/main/cpp/tests/shuffle_split.cpp
+++ b/src/main/cpp/tests/shuffle_split.cpp
@@ -104,6 +104,23 @@ TEST_F(ShuffleSplitTests, Strings)
   }
 }
 
+TEST_F(ShuffleSplitTests, SimpleWithStrings)
+{
+  cudf::test::fixed_width_column_wrapper<int8_t> col0(
+      {0, 0xF0, 0x0F, 0xAA, 0},
+      {1,    0,    0,    0, 1});
+  cudf::test::strings_column_wrapper col1(
+    {"0xFF", "", "0x0F", "0xAA", "0x55"},
+    {0,       1,      0,      0,      0});
+
+  // 2 columns split once
+  {
+    cudf::table_view tbl{{static_cast<cudf::column_view>(col0),
+                          static_cast<cudf::column_view>(col1)}};
+    run_split(tbl, {5});
+  }
+}
+
 TEST_F(ShuffleSplitTests, Lists)
 {
   // list<uint64_t>
@@ -360,4 +377,4 @@ TEST_F(ShuffleSplitTests, NestedTypes)
     run_split(tbl, {2});
     run_split(tbl, {2, 4});
   }
-}
\ No newline at end of file
+}
```

```
Running main() from /.../spark-rapids-jni/target/jni/cmake-build/_deps/gtest-src/googletest/src/gtest_main.cc
[==========] Running 10 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 10 tests from ShuffleSplitTests
[ RUN      ] ShuffleSplitTests.Simple
[       OK ] ShuffleSplitTests.Simple (347 ms)
[ RUN      ] ShuffleSplitTests.Strings
[       OK ] ShuffleSplitTests.Strings (51 ms)
[ RUN      ] ShuffleSplitTests.SimpleWithStrings
========= Invalid __global__ read of size 4 bytes
=========     at void cub::CUB_200700_860_NS::detail::MultiBlockBatchMemcpyKernel<cub::CUB_200700_860_NS::detail::DeviceBatchMemcpyPolicy<unsigned int, unsigned int>::Policy700, unsigned int, void **, void **, unsigned long *, unsigned int *, cub::CUB_200700_860_NS::ScanTileState<unsigned int, (bool)1>, unsigned int, (bool)1>(T3, T4, T5, T6, T7, T8)+0x1800
=========     by thread (64,0,0) in block (0,0,0)
=========     Address 0x404 is out of bounds
=========     and is 134,934,914,137,084 bytes before the nearest allocation at 0x7ab8fba00000 of size 64 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x2b765f]
=========                in /lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame:__cudart1071 [0x360835a]
=========                in /.../spark-rapids-jni/target/jni/cmake-build/libcudf.so
=========     Host Frame:cudaLaunchKernel_ptsz [0x3647fa7]
=========                in /.../spark-rapids-jni/target/jni/cmake-build/libcudf.so
```

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1779 @@
...
+      leading_bytes != 0
+        ? (batch.src[0] | (batch.src[1] << 8) | (batch.src[2] << 16) | (batch.src[3] << 24))
+        : (reinterpret_cast<bitmask_type const*>(batch.src))[0];
+    bitmask_type const relevant_row_mask = ((1 << rows_in_batch) - 1);
+
+    // shift and mask the incoming word so that bit 0 is the first row we're going to store.
+    word = (word >> batch.src_bit_shift) & relevant_row_mask;
+
+    // any bits that are not being stored in the current dest word get overflowed to the next copy
+    prev_word[0] = word >> (32 - batch.dst_bit_shift);
```
pmattione-nvidia, line 1313: Do any of the existing test trigger these misaligned cases?  I just want to make sure our tests are actually hitting them.  Do printfs in here show all kinds of alignments? Ditto for the main loop below.

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1779 @@
...
+ * This returns the size of the buffer -without- padding. Just the size of
+ * the raw bytes containing the actual data.
+ *
+ * Note that there is an odd edge case here. We may have a partition that contains 8 rows, which
+ * would translate to 1 byte of validity data. However, we copy from byte boundaries. So if this
+ * partition started at row index 2, we would need to account for those extra 2 bits that will be
+ * included in the buffer. So the source buffer will be 2 bytes (10 bits) even though the we are
+ * only using 8 bits of that data in the output. Therefore, ths size returned from this functor is
+ * only valid for -source- data.
+ *
```
pmattione-nvidia, line 556: Is there a guarantee that the src buffer validity bits are aligned to the beginning of a byte?  Suppose we have 10 rows and an offset of 7, wouldn't we need 3 bytes?  E.g.: `xxxxxxxB BBBBBBBB Bxxxxxxx`

src/main/cpp/src/shuffle_split_detail.hpp
```diff
@@ -0,0 +1,95 @@
...
+  }
+};
+
+/**
+ * @brief Header for each partition.
+ */
+struct partition_header {
+  uint32_t magic_number;
+  uint32_t row_index;  // row index in the source table that this partition started at
+  uint32_t num_rows;
```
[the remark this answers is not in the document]
nvdbaranec: This is what the kud0 spec is at the moment.  I think it originally was 8 byte sizes, but it's been shrunk for now.

src/main/cpp/src/shuffle_split_detail.hpp
nvdbaranec: Changed it a bit.

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1771 @@
...
+ */
+__global__ void compute_offset_child_row_counts(
+  cudf::device_span<int const> offset_columns,
+  cudf::device_span<shuffle_split_col_data const> column_metadata,
+  cudf::device_span<assemble_column_info> column_instances,
+  cudf::device_span<uint8_t const> partitions,
+  cudf::device_span<size_t const> partition_offsets,
+  size_t per_partition_metadata_size)
+{
+  if (threadIdx.x != 0) { return; }
```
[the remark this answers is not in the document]
nvdbaranec: It does parallelize as one-block-per-partition (one warp per block), so we're at least parallel on the number of partitions.

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1771 @@
...
+          break;
+        // all other types jump back up to the nearest struct/branch point, as they represent the
+        // end of a chain of children
+        case cudf::type_id::STRING:
+          col_inst.num_chars = offsets[num_rows] - offsets[0];
+          offsets += (num_rows + 1);
+          // fallthrough
+        default: {
+          if (rc_stack_pos >= 0) {
+            auto const num_children = --rc_stack[(rc_stack_pos * 2) + 1];
```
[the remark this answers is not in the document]
nvdbaranec: Done.

src/main/cpp/src/shuffle_assemble.cu
nvdbaranec: Done.

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1771 @@
...
+
+/**
+ * @brief Compute size of a bitmask in bytes, including padding
+ *
+ * @param number_of_bits Number of bits in the bitmask
+ * @param pad Padding for the final buffer size
+ *
+ * @return Size of the required allocation in bytes
+ */
+constexpr size_t bitmask_allocation_size_bytes(size_type number_of_bits, int pad)
```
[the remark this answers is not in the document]
nvdbaranec: Yep.

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1771 @@
...
+
+    // each block needs to maintain a small stack of information of size max_depth when traversing
+    // the column hierarchy. we'll use dynamic shared memory to do it.
+    int const shmem_per_block = static_cast<int>(sizeof(size_type) * max_branch_depth * 2);
+    int device_id;
+    CUDF_CUDA_TRY(cudaGetDevice(&device_id));
+    int shmem_limit_per_block;
+    CUDF_CUDA_TRY(cudaDeviceGetAttribute(
+      &shmem_limit_per_block, cudaDevAttrMaxSharedMemoryPerBlock, device_id));
+    CUDF_EXPECTS(shmem_per_block <= shmem_limit_per_block,
```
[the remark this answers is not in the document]
nvdbaranec: So, the size needed here is proportional to the max nesting depth of any one column, where "nesting" here just means "structs".  12 bytes per level, so even for a 64k shared memory GPU, that would require a nesting of 5000+ deep. `row_bit_count` does the same thing and we haven't hit this yet.

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -0,0 +1,1771 @@
...
+  uint8_t* dst;
+  size_t size;                   // bytes
+  buffer_type btype;
+  int value_shift;               // amount to shift values down by (for offset buffers)
+  int src_bit_shift;             // source bit (right) shift. easy way to think about this is
+                                 // 'the number of rows at the beginning of the buffer to ignore'.
+                                 // we need to ignore them because the split-copy happens at
+                                 // byte boundaries, not bit/row boundaries. so we may have
+                                 // irrelevant rows at the very beginning.
+  int dst_bit_shift;             // dest bit (left) shift
```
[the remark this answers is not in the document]
nvdbaranec: No because the source and destination buffers can all be different boundaries.  For example, 

- in the source, we could have started at row 5, but because we copy on byte boundaries, we've got 5 leading bits we need to ignore. So shift those right.
- at the destination, we could be placing these same bits starting at row 18.  Also, we work in 32 bit words here, so we're modding by a different value.

`copy_validity` is maximally bananas because of all of this.

src/main/cpp/src/shuffle_split.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/exec_policy.hpp>
+
+#include <vector>
+
+namespace spark_rapids_jni {
+
+// TODO: this is duplicated from cudf because the cudf function is not marked as constexpr, so it
+// cannot be called on the gpu there is an issue filed against cudf to make
```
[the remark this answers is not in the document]
nvdbaranec: Mike pointed this out too.  I think this was a fat-finger on the comment (related to bitmask-allocation_size_bytes).  There is no open issue.  The cudf guys didn't want to expose this.

nvdbaranec:
> > Doing 64 byte padding when concating is learned from jcudf, I thought it was a required for gpu?
> 
> 64-bytes was me being overly cautious. Cuda malloc does 64-byte alignment (I think), because like malloc it does not know the data types so it cannot make any assumption about how it needs to be aligned. In reality we need each column to be aligned based on how it is accessed.
> 
>     * byte/bool = 1 byte aligned (no padding needed)
> 
>     * short = 2 byte aligned
> 
>     * int/float/timestamp_days/decimal_32 = 4 byte aligned
> 
>     * long/double/timestamp_micros/decimal_64/validity = 8 byte aligned
> 
>     * decimal_128 = 16 byte aligned
> 
> 
> Offsets we could have as 4 byte aligned, but because CUDF supports longer offsets we might want to think about 8 byte aligned (unless we have a way to store what type of offset is being stored in the header)
> 
> In JCUDF serialization I thought about reordering the columns to reduce the amount of padding that would be needed, but I never did it because it didn't feel like it saved that much, but I wasn't looking at it from thousands of partitions.
> 
> @nvdbaranec how is the padding needed for the kernels in this patch to work? Would it be alright if there was no padding at all? Or is the data copied in a non-byte wise way?

For the data buffer, it wouldn't matter at all.  We use the cub batched memcpy which doesn't care.  The only thing we explicitly need alignment on is the offset buffers. See:

https://github.com/NVIDIA/spark-rapids-jni/pull/2800#issuecomment-2644167454

So we could either eliminate the 4 byte pad before the data buffers or change it to be per-type.

src/main/cpp/tests/shuffle_split.cu
nvdbaranec: Done

src/main/cpp/src/shuffle_assemble.cu
nvdbaranec: Yes.  Adding tests has gradually shaken out all the permutations.

src/main/cpp/src/shuffle_assemble.cu
nvdbaranec: Yep. That was a bug.

src/main/cpp/src/shuffle_assemble.cu
nvdbaranec: No, there's no guarantee.  That's what this comment is saying.  Even though we've got 8 bits of actual data, because we're starting at row 2, there will be 2 bytes of data in the source buffer.  That's what this extra bit of code is doing

https://github.com/NVIDIA/spark-rapids-jni/blob/9ea606ef4aef51fbe6ee7c1e49a6fbc3fdb48125/src/main/cpp/src/shuffle_assemble.cu#L569

But in the destination, we're only interested in the 8 bits / 1 byte of data.  So we can't just say that, for example, the destination buffer size is just the sum of all the sizes of all the sources.

src/main/cpp/src/shuffle_split.cu
```diff
@@ -984,64 +785,37 @@ std::vector<packed_table> contiguous_split(cudf::table_view const& input,
...
-                       packed_columns{std::make_unique<std::vector<uint8_t>>(pack_metadata(
-                                        empty_inputs, static_cast<uint8_t const*>(nullptr), 0)),
-                                      std::make_unique<rmm::device_buffer>()}};
-                   });
-
-    return result;
-  }
+  auto temp_mr = cudf::get_current_device_resource_ref();
+
+  size_t const num_partitions   = splits.size() + 1;
```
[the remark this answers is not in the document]
nvdbaranec: Yes.  You'll get two partitions. One containing the whole input table and one that's empty.  `contiguous_split` behaves the same way.

src/main/cpp/src/shuffle_assemble.cu
nvdbaranec: What shortcut?  It still has to unpack everything.

nvdbaranec:
> ```diff
> +TEST_F(ShuffleSplitTests, SimpleWithStrings)
> +{
> +  cudf::test::fixed_width_column_wrapper<int8_t> col0(
> +      {0, 0xF0, 0x0F, 0xAA, 0},
> +      {1,    0,    0,    0, 1});
> +  cudf::test::strings_column_wrapper col1(
> +    {"0xFF", "", "0x0F", "0xAA", "0x55"},
> +    {0,       1,      0,      0,      0});
> +
> +  // 2 columns split once
> +  {
> +    cudf::table_view tbl{{static_cast<cudf::column_view>(col0),
> +                          static_cast<cudf::column_view>(col1)}};
> +    run_split(tbl, {5});
> +  }
> +}
> ```

I believe this has since been fixed.  I'll push this test up.

src/main/cpp/src/shuffle_split.cu
```diff
@@ -770,212 +539,247 @@ struct dst_offset_output_iterator {
...
+        ? split_indices[partition_index + 1] - split_indices[partition_index]
+        : 0;
+    pheader->num_rows =
+      cudf::hashing::detail::swap_endian(static_cast<uint32_t>(partition_num_rows));
+
+    auto const& psize = partition_sizes[partition_index];
+    pheader->validity_size =
+      cudf::hashing::detail::swap_endian(static_cast<uint32_t>(psize.validity_size));
+    pheader->offset_size =
+      cudf::hashing::detail::swap_endian(static_cast<uint32_t>(psize.offset_size));
```
revans2, line 639: The validity and offset sizes in the CPU version include the padding at the end.

src/main/cpp/src/shuffle_split.cu
```diff
@@ -770,212 +539,247 @@ struct dst_offset_output_iterator {
...
+    pheader->data_size = cudf::hashing::detail::swap_endian(static_cast<uint32_t>(psize.data_size));
```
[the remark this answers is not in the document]
nvdbaranec: @liurenjie1024

src/main/cpp/src/shuffle_split.cu
revans2: @nvdbaranec Just as a side note I was able to update the CPU code to match the padding that you are doing on the GPU code. So the only thing that is left to really make these cross compatible is the changes for the lengths.

revans2:
@nvdbaranec I have been doing more testing and I ran into a new issue. I don't have it in c++ yet, but here is the java code for the table. The schema is a table with two columns in it, both are LIST<STRING>.  When the data is serialized the data section is gone. The length is 0.

```
  static Table buildStringListTable() {
    return new Table.TestBuilder()
        .column(strings("*"), strings("*"), strings("****"),
            strings("", "*", null))
        .column(strings(null, null, null, null), strings(),
            strings(null, null, null), strings())
        .build();
  }
```

src/main/cpp/src/shuffle_split.cu
nvdbaranec: Added the padding to the sizes and changed `data_size` to `total_size`, representing (validity+offsets+data).

src/main/cpp/src/shuffle_split.cu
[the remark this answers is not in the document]
nvdbaranec: Updated.

revans2:
Thanks for the padding. Most of my tests can now be read by the CPU after being written by the GPU. I am still seeing issues with the list of strings, and I am also seeing some issues with validity for structs.

If I have a struct with two child columns in it that are both int8, and I only set one value to null, it looks like the order of the has validity and the order of the validity are messed up relative to the CPU.

```
  static Table buildByteStructTable() {
    HostColumnVector.StructType bsType = new HostColumnVector.StructType(false,
            new HostColumnVector.BasicType(false, DType.INT8),
            new HostColumnVector.BasicType(true, DType.INT8));

    return new Table.TestBuilder()

        .column(dateType, struct((byte)1, (byte)2), struct((byte)3, null), struct((byte)4, (byte)5))
        .build();
  }
```

I get the following output from by debug scripts for the GPU output vs the CPU output.

```
testComplexSinglePartWriteCPURead COMP GPU(40) VS CPU(40)
ALIGNED 0000:     0x4B 075 VS 0x4B 075 <-- MAGIC START
        0001:     0x55 085 VS 0x55 085
        0002:     0x44 068 VS 0x44 068
        0003:     0x30 048 VS 0x30 048 <-- MAGIC END
ALIGNED 0004:     0x00 000 VS 0x00 000 <-- OFFSET START
        0005:     0x00 000 VS 0x00 000
        0006:     0x00 000 VS 0x00 000
        0007:     0x00 000 VS 0x00 000 <-- OFFSET END (0/0)
ALIGNED 0008:     0x00 000 VS 0x00 000 <-- NUM_ROWS START
        0009:     0x00 000 VS 0x00 000
        0010:     0x00 000 VS 0x00 000
        0011:     0x03 003 VS 0x03 003 <-- NUM_ROWS END (3/3)
ALIGNED 0012:     0x00 000 VS 0x00 000 <-- VALIDITY_BUF_LEN START
        0013:     0x00 000 VS 0x00 000
        0014:     0x00 000 VS 0x00 000
        0015:     0x03 003 VS 0x03 003 <-- VALIDITY_BUF_LEN END (3/3)
ALIGNED 0016:     0x00 000 VS 0x00 000 <-- OFFSET_BUF_LEN START
        0017:     0x00 000 VS 0x00 000
        0018:     0x00 000 VS 0x00 000
        0019:     0x00 000 VS 0x00 000 <-- OFFSET_BUF_LEN END (0/0)
ALIGNED 0020:     0x00 000 VS 0x00 000 <-- TOTAL_DATA_LEN START
        0021:     0x00 000 VS 0x00 000
        0022:     0x00 000 VS 0x00 000
        0023:     0x0B 011 VS 0x0B 011 <-- TOTAL_DATA_LEN END (11/11)
ALIGNED 0024:     0x00 000 VS 0x00 000 <-- NUM_COL START
        0025:     0x00 000 VS 0x00 000
        0026:     0x00 000 VS 0x00 000
        0027:     0x03 003 VS 0x03 003 <-- NUM_COL END (3/3)
ALIGNED 0028: *** 0x04 004 VS 0x02 002 <-- HAS_VALID_BUF
        0029:     0xFD -03 VS 0xFD -03 <-- VALIDITY_BUFFERS START
        0030:     0x00 000 VS 0x00 000
        0031:     0x00 000 VS 0x00 000 <-- VALIDITY_BUFFERS END
ALIGNED 0032:     0x01 001 VS 0x01 001 <-- DATA_BUFFERS START
        0033:     0x03 003 VS 0x03 003
        0034:     0x04 004 VS 0x04 004
        0035:     0x02 002 VS 0x02 002
ALIGNED 0036:     0xDB -37 VS 0xDB -37
        0037:     0x05 005 VS 0x05 005
        0038:     0x00 000 VS 0x00 000
        0039:     0x00 000 VS 0x00 000 <-- DATA_BUFFERS END
```

The CPU code puts the validity for the struct first compared to the validity for its children, both in the has validity buffer in the header, but also for the order of the validity buffers when read. If I read the GPU kudo buffer with the CPU reader I get an output like

```
{1, 2}
null
{4, 5}
```

It looks like the second columns validity was interpreted as the validity for the parent.

src/main/cpp/src/shuffle_split.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/exec_policy.hpp>
+
+#include <vector>
+
+namespace spark_rapids_jni {
+
+// TODO: this is duplicated from cudf because the cudf function is not marked as constexpr, so it
+// cannot be called on the gpu. there is an issue filed against cudf to make
```
hyperbolic2346, line 27: link issue

hyperbolic2346 (approved):

winningsix:
Great job. Do we have any performance number comparing to CPU version of Kudo?

revans2:
@winningsix and @nvdbaranec there is some more work that I need to do to get them both producing identical results. I also have not completed testing the merge code. But the numbers for the GPU are not looking that great. @Dave when I try to write out a table with 10 long columns in it and 10,000,000 rows (split 200 ways) the GPU utilization does not go above 14%. I get that I am doing an end to end like test where I copy the data to the host afterwards (all 763 MiB of it), but the speedup over a single CPU core in that case is only about 13%.

Here are the full numbers that I collected for serialization. Number of slices is vertically in the table and the number of rows in the table is horizontally. The times reported are the median of 5 runs.

CPU time (ns) |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | --
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 791,236 | 359,489 | 332,298 | 479,777 | 1,736,173 | 42,538,222 | 524,669,564
200 | N/A | 1,369,029 | 1,247,158 | 1,403,504 | 2,957,372 | 43,618,445 | 530,119,592
1,000 | N/A | N/A | 5,247,451 | 5,679,069 | 5,679,069 | 48,772,070 | 528,532,823
10,000 | N/A | N/A | N/A | 16,722,159 | 18,189,142 | 60,551,873 | 555,724,861
100,000 | N/A | N/A | N/A | N/A | 265,128,153 | 241,922,576 | 688,284,764
  |   |   |   |   |   |   |  
GPU time (ns) |   |   |   |   |   |   |  
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 155,214 | 156,466 | 200,679 | 323,111 | 1,065,234 | 46,292,922 | 461,698,868
200 | N/A | 155,313 | 188,667 | 286,782 | 1,235,606 | 46,390,496 | 468,152,352
1,000 | N/A | N/A | 165,633 | 300,398 | 1,091,033 | 46,190,278 | 465,485,100
10,000 | N/A | N/A | N/A | 847,653 | 1,757,773 | 46,213,682 | 468,127,258
100,000 | N/A | N/A | N/A | N/A | 18,790,960 | 65,125,481 | 469,566,332
  |   |   |   |   |   |   |  
size (bytes) |   |   |   |   |   |   |  
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 1,120 | 16,320 | 80,320 | 800,320 | 8,000,320 | 80,000,320 | 800,000,320
200 | N/A | 22,400 | 86,400 | 806,400 | 8,006,400 | 80,006,400 | 800,006,400
1,000 | N/A | N/A | 112,000 | 832,000 | 8,032,000 | 80,032,000 | 800,032,000
10,000 | N/A | N/A | N/A | 1,120,000 | 8,320,000 | 80,320,000 | 800,320,000
100,000 | N/A | N/A | N/A | N/A | 11,200,000 | 83,200,000 | 803,200,000
  |   |   |   |   |   |   |  
Speedup |   |   |   |   |   |   |  
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 5.1 | 2.3 | 1.66 | 1.48 | 1.63 | 0.92 | 1.14
200 | N/A | 8.81 | 6.61 | 4.89 | 2.39 | 0.94 | 1.13
1,000 | N/A | N/A | 31.68 | 18.91 | 5.21 | 1.06 | 1.14
10,000 | N/A | N/A | N/A | 19.73 | 10.35 | 1.31 | 1.19
100,000 | N/A | N/A | N/A | N/A | 14.11 | 3.71 | 1.47

nvdbaranec:
> @winningsix and @nvdbaranec there is some more work that I need to do to get them both producing identical results. I also have not completed testing the merge code. But the numbers for the GPU are not looking that great. @dave when I try to write out a table with 10 long columns in it and 10,000,000 rows (split 200 ways) the GPU utilization does not go above 14%. I get that I am doing an end to end like test where I copy the data to the host afterwards (all 763 MiB of it), but the speedup over a single CPU core in that case is only about 13%.

I'll have a look. This definitely seems slow.

revans2:
@nvdbaranec and @winningsix I have to apologize. I was too fast in writing the test and I included a device to host copy for both set of tests, but it was not pinned. So I am going to update the test to remove the copy for both and just look at the computation side of things.

revans2:
I have some updated numbers for splitting the data. This has removed the data transfer to the CPU.  This is just processing the data, so for the CPU all of the data is already on the CPU. For the GPU it is still on the GPU and the results are not transferred back yet.

In general both the CPU and the GPU don't have great numbers when the number of slices is close to the number of rows, or when the input data is relatively small. The CPU, on my setup which is single threaded, tops out at about 11 GiB/sec where as the GPU (a6000 without any RMM pool) tops out at about 100 GiB/sec.

There were two test cases where the CPU beat the GPU. 10 slices on 10,000 rows and 200 slices on 100,000 rows. I am going to get some nsys traces of these to see if I can figure out what is happening. Especially the 200 slices for 100,000 rows because it does not appear to follow the curve for the values around it. Also this is the median of 9 runs.

CPU time (ns) |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | --
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 472,995 | 135,346 | 172,777 | 175,812 | 724,650 | 6,749,277 | 67,354,438
200 | N/A | 1,175,483 | 1,126,190 | 1,173,720 | 1,684,997 | 8,087,709 | 68,422,749
1,000 | N/A | N/A | 5,106,830 | 5,208,693 | 5,809,790 | 13,285,050 | 73,633,236
10,000 | N/A | N/A | N/A | 15,429,857 | 16,670,053 | 24,018,133 | 132,202,987
100,000 | N/A | N/A | N/A | N/A | 154,347,396 | 168,751,994 | 245,993,977
  |   |   |   |   |   |   |  
GPU time (ns) |   |   |   |   |   |   |  
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 116,260 | 124,144 | 133,172 | 210,007 | 325,005 | 2,476,635 | 8,441,568
200 | N/A | 125,557 | 156,536 | 181,514 | 2,110,262 | 2,374,983 | 7,389,518
1,000 | N/A | N/A | 129,575 | 188,196 | 396,780 | 2,703,995 | 7,388,556
10,000 | N/A | N/A | N/A | 685,406 | 980,815 | 2,893,743 | 12,665,589
100,000 | N/A | N/A | N/A | N/A | 18,090,009 | 20,366,516 | 24,278,636
  |   |   |   |   |   |   |  
size (bytes) |   |   |   |   |   |   |  
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 1,120 | 16,320 | 80,320 | 800,320 | 8,000,320 | 80,000,320 | 800,000,320
200 | N/A | 22,400 | 86,400 | 806,400 | 8,006,400 | 80,006,400 | 800,006,400
1,000 | N/A | N/A | 112,000 | 832,000 | 8,032,000 | 80,032,000 | 800,032,000
10,000 | N/A | N/A | N/A | 1,120,000 | 8,320,000 | 80,320,000 | 800,320,000
100,000 | N/A | N/A | N/A | N/A | 11,200,000 | 83,200,000 | 803,200,000
  |   |   |   |   |   |   |  
Speedup |   |   |   |   |   |   |  
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 4.07 | 1.09 | 1.3 | 0.84 | 2.23 | 2.73 | 7.98
200 | N/A | 9.36 | 7.19 | 6.47 | 0.8 | 3.41 | 9.26
1,000 | N/A | N/A | 39.41 | 27.68 | 14.64 | 4.91 | 9.97
10,000 | N/A | N/A | N/A | 22.51 | 17. | 8.3 | 10.44
100,000 | N/A | N/A | N/A | N/A | 8.53 | 8.29 | 10.13
  |   |   |   |   |   |   |  
GiB/Sec CPU |   |   |   |   |   |   |  
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 0.002 | 0.112 | 0.433 | 4.240 | 10.282 | 11.039 | 11.062
200 | N/A | 0.018 | 0.071 | 0.640 | 4.425 | 9.213 | 10.889
1,000 | N/A | N/A | 0.020 | 0.149 | 1.288 | 5.610 | 10.119
10,000 | N/A | N/A | N/A | 0.068 | 0.465 | 3.114 | 5.638
100,000 | N/A | N/A | N/A | N/A | 0.068 | 0.459 | 3.041
  |   |   |   |   |   |   |  
GiB/Sec GPU |   |   |   |   |   |   |  
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 0.009 | 0.122 | 0.562 | 3.549 | 22.925 | 30.084 | 88.261
200 | N/A | 0.166 | 0.514 | 4.138 | 3.533 | 31.374 | 100.827
1,000 | N/A | N/A | 0.805 | 4.117 | 18.853 | 27.565 | 100.844
10,000 | N/A | N/A | N/A | 1.522 | 7.90 | 25.850 | 58.849
100,000 | N/A | N/A | N/A | N/A | 0.577 | 3.805 | 30.811

src/main/cpp/src/shuffle_assemble.cu
pmattione-nvidia: Ah I see, the `src_row_index % 8` handles the first byte and `bitmask_allocation_size_bytes()` handles the last one. ok.

revans2:
Looks like cuda malloc and free were slowing me down a lot in the tests. The new numbers for the GPU are below.

The GPU now tops out at about 400 GiB/sec and for 200 splits the perf improvement over the CPU is between 4.5x faster to 32.8x faster.  I am noticing from the traces that when we start to have lots of splits (see the 100,000 row) most of the time on the GPU split is happening in the CPU code before it runs any GPU kernels.  I think there is something there that is scaling with the number of buffers. But it is still relatively fast so I think we can leave that as follow on work.

GPU time (ns) |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | --
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 94,479 | 105,570 | 102,303 | 187,535 | 105,289 | 262,516 | 2,361,426
200 | N/A | 102,013 | 140,145 | 162,067 | 354,150 | 282,314 | 2,004,661
1,000 | N/A | N/A | 114,757 | 169,631 | 198,676 | 633,889 | 1,856,621
10,000 | N/A | N/A | N/A | 283,035 | 352,978 | 450,141 | 2,546,085
100,000 | N/A | N/A | N/A | N/A | 17,306,810 | 17,473,646 | 19,533,060

Speedup |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | --
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 3.33 | 1.47 | 1.51 | 1.07 | 6.69 | 24.53 | 27.12
200 | N/A | 11.47 | 7.98 | 7.19 | 4.58 | 27.27 | 32.8
1,000 | N/A | N/A | 44.44 | 30.63 | 28.19 | 20.63 | 38.5
10,000 | N/A | N/A | N/A | 54.01 | 46.35 | 118.97 | 37.31
100,000 | N/A | N/A | N/A | N/A | 8.99 | 9.68 | 11.39

Output GiB/Sec GPU |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | --
slices\rows | 10 | 200 | 1,000 | 10,000 | 100,000 | 1,000,000 | 10,000,000
10 | 0.011 | 0.144 | 0.731 | 3.974 | 70.766 | 283.815 | 315.512
200 | N/A | 0.204 | 0.574 | 4.634 | 21.055 | 263.932 | 371.666
1,000 | N/A | N/A | 0.909 | 4.568 | 37.651 | 117.585 | 401.314
10,000 | N/A | N/A | N/A | 3.685 | 21.952 | 166.179 | 292.746
100,000 | N/A | N/A | N/A | N/A | 0.603 | 4.434 | 38.296

nvdbaranec:
>32.8x faster. I am noticing from the traces that when we start to have lots of splits (see the 100,000 row) most of the time on the GPU split is happening in the CPU code before it runs any GPU kernels. I think there is something there that is scaling with the number of buffers. But it is still relatively fast so I think we can leave that as follow on work.

Yes, there is a nontrivial amount of work that happens, especially on the assemble side.  Although a lot of that is also small thrust calls.  One optimization approach here is to see how much of that setup work being done via thrust/GPU would actually be faster on the CPU.  It's a similar problem to the parquet reader where doing it the thrust way causes a lot of back-and-forth memcpying. Also, it's pageable memory right now, so even though the sizes are small it's not quick.  Switching to pinned may help.  I'll do some additional tests here, but yeah I think it's followup optimization.

revans2:
I am getting another test failure. This time it is when I try to turn a single slice back into a table. The schema is two columns. The first is a `LIST<STRING>` and the second is a `LIST<STRUCT<INT32,INT32>>`. Each of them have two rows. Both of the string lists are empty.

```
    HostColumnVector.ListType listDateType = new HostColumnVector.ListType(true,
        new HostColumnVector.StructType(false,
            new HostColumnVector.BasicType(false, DType.INT32),
            new HostColumnVector.BasicType(true, DType.INT32)));

    return new Table.TestBuilder()
        .column(strings(), strings())
        .column(listDateType, singletonList(struct(-210, 293)), singletonList(struct(311, 992)))
        .build();
```

Both the CPU and the GPU produce the same 72 byte serialized output. But the GPU crashes turning it back into a table by trying to allocate 18446744073709550784 bytes of memory (looks like an overflow/underflow).

```
java.lang.OutOfMemoryError: Could not allocate native memory: std::bad_alloc: out_of_memory: CUDA error (failed to allocate 18446744073709550784 bytes) at: .../rmm/mr/device/cuda_memory_resource.hpp:62: cudaErrorMemoryAllocation out of memory
```

nvdbaranec:
> I am getting another test failure. 

On it.

revans2:
@nvdbaranec I have found another inconsistency between the GPU and CPU implementations. The CPU code has an optimization in it, where it will not output a validity buffer if the row count for that buffer is 0.

https://github.com/NVIDIA/spark-rapids-jni/blob/2ebbc69990efeaa1242135d77477f8ddff479b16/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableHeaderCalc.java#L89

There are some inconsistencies for this that I plan on trying to fix, but technically the CPU output could be smaller so I am inclined to use their version if we can.  Is that something you think you could add to the GPU version?

nvdbaranec:
@revans2  Your string bug and the validity inconsistency should be fixed.

revans2:
@liurenjie1024 I have some much more interesting numbers now. They are not perfect because it is a contrived test for a single column shuffle. I plan to expand this out to more columns soon.  But I have compared JCUDF, Kudo CPU and Kudo GPU.  Here are the results I have.

Essentially the speedup for Kudo is all on the write. The read is actually slower. But overall the results in decent speedup over JCUDF for data sizes and split numbers I think are reasonable.  Note that the sizes of the batches are 2k, 64k, 1m, 10m, 100m, and 1g.

E2E Speedup CPU vs JCUDF |   |   |   |   |   |   |   | Encode Write CPU vs JCUDF |   |   |   |   |   |   |   | Decode Read CPU vs JCUDF |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728 |   | slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728 |   | slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728
1 | 1.17 | 1.20 | 0.73 | 0.60 | 1.11 | 1.13 |   | 1 | 1.90 | 3.04 | 1.67 | 1.47 | 1.37 | 1.35 |   | 1 | 0.81 | 0.83 | 0.63 | 0.46 | 1.07 | 1.09
10 | 1.58 | 3.23 | 1.14 | 0.63 | 1.11 | 1.13 |   | 10 | 3.71 | 18.21 | 5.28 | 1.56 | 1.39 | 1.35 |   | 10 | 0.63 | 0.70 | 0.73 | 0.48 | 1.06 | 1.09
200 | 10.58 | 15.62 | 5.53 | 0.98 | 0.95 | 1.09 |   | 200 | 24.71 | 50.13 | 39.28 | 3.12 | 1.52 | 1.38 |   | 200 | 0.82 | 0.77 | 0.73 | 0.62 | 0.86 | 1.00
1,000 | N/A | 21.04 | 3.58 | 1.44 | 0.94 | 1.08 |   | 1,000 | N/A | 59.76 | 11.66 | 7.91 | 2.14 | 1.45 |   | 1,000 | N/A | 0.75 | 0.55 | 0.49 | 0.69 | 0.97
10,000 | N/A | N/A | 13.88 | 11.90 | 3.74 | 1.21 |   | 10,000 | N/A | N/A | 38.23 | 30.80 | 12.65 | 2.12 |   | 10,000 | N/A | N/A | 0.94 | 1.50 | 1.04 | 0.94
100,000 | N/A | N/A | 18.58 | 18.20 | 7.02 | 2.57 |   | 100,000 | N/A | N/A | 39.96 | 41.15 | 33.86 | 8.27 |   | 100,000 | N/A | N/A | 1.26 | 1.42 | 0.61 | 0.97
  |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
E2E Speedup GPU vs CPU |   |   |   |   |   |   |   | Encode Write GPU vs CPU |   |   |   |   |   |   |   | Decode Read GPU vs CPU |   |   |   |   |   |  
slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728 |   | slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728 |   | slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728
1 | 0.39 | 0.40 | 1.17 | 2.04 | 2.07 | 2.07 |   | 1 | 0.40 | 0.23 | 0.60 | 0.89 | 0.98 | 0.97 |   | 1 | 0.39 | 0.46 | 1.28 | 2.58 | 2.56 | 2.57
10 | 1.41 | 0.51 | 1.13 | 1.95 | 2.06 | 2.03 |   | 10 | 1.54 | 0.26 | 0.63 | 0.85 | 0.96 | 0.96 |   | 10 | 0.84 | 0.60 | 1.23 | 2.48 | 2.54 | 2.52
200 | 2.12 | 1.25 | 1.25 | 1.67 | 2.00 | 1.25 |   | 200 | 3.01 | 1.36 | 0.80 | 0.93 | 0.97 | 0.97 |   | 200 | 1.26 | 1.17 | 1.33 | 1.95 | 2.43 | 1.37
1,000 | N/A | 2.56 | 2.45 | 2.22 | 1.61 | 1.29 |   | 1,000 | N/A | 3.59 | 2.23 | 1.08 | 0.99 | 0.98 |   | 1,000 | N/A | 2.01 | 2.47 | 2.65 | 1.84 | 1.43
10,000 | N/A | N/A | 2.63 | 1.53 | 1.31 | 1.10 |   | 10,000 | N/A | N/A | 5.13 | 1.97 | 1.15 | 1.01 |   | 10,000 | N/A | N/A | 1.99 | 1.35 | 1.35 | 1.14
100,000 | N/A | N/A | 3.21 | 2.72 | 3.09 | 1.19 |   | 100,000 | N/A | N/A | 9.07 | 8.71 | 2.22 | 1.16 |   | 100,000 | N/A | N/A | 2.05 | 1.77 | 3.38 | 1.20
  |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
E2E Sppedup GPU vs JCUDF |   |   |   |   |   |   |   | Encode Write GPU vs JCUDF |   |   |   |   |   |   |   | Decode Read GPU vs JCUDF |   |   |   |   |   |  
slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728 |   | slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728 |   | slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728
1 | 0.46 | 0.48 | 0.85 | 1.24 | 2.30 | 2.33 |   | 1 | 0.76 | 0.70 | 1.00 | 1.31 | 1.34 | 1.32 |   | 1 | 0.31 | 0.39 | 0.81 | 1.20 | 2.73 | 2.80
10 | 2.23 | 1.66 | 1.30 | 1.23 | 2.29 | 2.29 |   | 10 | 5.70 | 4.66 | 3.30 | 1.33 | 1.34 | 1.29 |   | 10 | 0.53 | 0.43 | 0.89 | 1.19 | 2.70 | 2.75
200 | 22.46 | 19.53 | 6.93 | 1.64 | 1.91 | 1.36 |   | 200 | 74.30 | 68.40 | 31.47 | 2.89 | 1.47 | 1.34 |   | 200 | 1.04 | 0.90 | 0.96 | 1.21 | 2.09 | 1.37
1,000 | N/A | 53.77 | 8.76 | 3.21 | 1.51 | 1.39 |   | 1,000 | N/A | 214.28 | 25.97 | 8.53 | 2.13 | 1.42 |   | 1,000 | N/A | 1.51 | 1.37 | 1.30 | 1.27 | 1.38
10,000 | N/A | N/A | 36.51 | 18.20 | 4.91 | 1.34 |   | 10,000 | N/A | N/A | 196.19 | 60.55 | 14.53 | 2.14 |   | 10,000 | N/A | N/A | 1.88 | 2.02 | 1.41 | 1.07
100,000 | N/A | N/A | 59.70 | 49.56 | 21.65 | 3.06 |   | 100,000 | N/A | N/A | 362.34 | 358.30 | 75.14 | 9.55 |   | 100,000 | N/A | N/A | 2.59 | 2.53 | 2.05 | 1.16
  |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
  |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
Size kudo vs jcudf |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
slices\rows | 256 | 8,192 | 131,072 | 1,310,720 | 13,107,200 | 134,217,728 |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
1 | 99.90% | 100.00% | 100.00% | 100.00% | 100.00% | 100.00% |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
10 | 81.66% | 99.39% | 99.96% | 100.00% | 100.00% | 100.00% |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
200 | 42.14% | 99.45% | 99.81% | 99.89% | 100.00% | 100.00% |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
1,000 | N/A | 97.99% | 96.26% | 99.83% | 99.96% | 99.99% |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
10,000 | N/A | N/A | 84.04% | 96.26% | 99.83% | 99.98% |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
100,000 | N/A | N/A | 42.28% | 84.04% | 96.26% | 99.83% |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |

nvdbaranec:
Adding an issue for a potential optimization after this is merged
https://github.com/NVIDIA/spark-rapids-jni/issues/3020

revans2:
@nvdbaranec I have another test failure for you.

This is for a `List<List<Struct<STRING, STRING>>>`

The data is something like
```
list(list(struct("k1", "v1"))),
list(list(struct("k4", "v4"))),
null,
null,
null,
list(list(struct("k8", "v8"))),
list(list(struct("k14", "v14"))),
null,
null,
null,
null,
list(list(struct("k16", "v16"))),
list(list(struct("k19", "v19"))),
list(list(struct("k22", "v22"))),
list(null),
list(list(struct("k22", null))), 
null,
null
```

So 18 rows for a single column.

I split it with `[2, 4, 6, 8, 10, 12, 14, 16]` and then just combine them back together again (encoding and decoding are both on the GPU) and I get a different result. The data is corrupted.

The CPU and the GPU produce slightly different serialized versions. The CPU is not including validity for slices with no rows in them, but the GPU is. But I don't think this is the issue because the CPU can successfully decode what the GPU wrote and the GPU gets the same error trying to decode what the CPU wrote.

revans2:
@nvdbaranec just to give you a little more info it looks like it gets confused trying to decode the very last offset of the inner list. The GPU version outputs a final offset of 0 instead of 8. After that all of the other columns end up being empty. All of the string columns have no data or validity buffers.  There was also no exception thrown to say that it decoded things incorrectly. i.e. there are lists pointing to offsets in string columns that do not exist.

pmattione-nvidia (dismissed):

nvdbaranec:
> @nvdbaranec just to give you a little more info it looks like it gets confused trying to decode the very last offset of the inner list. The GPU version outputs a final offset of 0 instead of 8. After that all of the other columns end up being empty. All of the string columns have no data or validity buffers. There was also no exception thrown to say that it decoded things incorrectly. i.e. there are lists pointing to offsets in string columns that do not exist.

Fixes posted.  Issue was the code that was determining "which copy batch writes the terminating partition" was just assuming "the last partition", which isn't correct if we have empty trailing partitions (since we don't send the offsets for those any more).

I also fixed the validity inconsistency.  Issue there was the code that was dealing with misaligned addresses when determining how many bytes to copy for validity.  Even when we had 0 rows, it would add in the expected number of leading bits due to alignment, which could lead to a > 0 number of bytes to copy. 

https://github.com/NVIDIA/spark-rapids-jni/blob/a821f8db5f50aab2e322a4f691a1c471b193c4ab/src/main/cpp/src/shuffle_split.cu#L948

hyperbolic2346 (approved):

pmattione-nvidia (approved):

nvdbaranec:
build

hyperbolic2346:
🚀

PR #2841, opened 2025-01-30
Fix a race condition that happens on shutdown of the SparkResourceAdaptor

revans2:
This fixes https://github.com/NVIDIA/spark-rapids/issues/12038

This does not fix all potential race conditions, but it does fix this one and helps to mitigate the others a little more. But the others should never occur in practice unless the spark worker is shutting down in an error case already.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
@@ -22,11 +22,28 @@ import ai.rapids.cudf.RmmDeviceMemoryResource;
 import ai.rapids.cudf.RmmEventHandlerResourceAdaptor;
 import ai.rapids.cudf.RmmWrappingDeviceMemoryResource;
 
+/**
+ * This is an internal class that provides an interface to a C++ spark_resource_adaptor class that
+ * provides the ability to roll back threads when OOMs happen and spill does not take care of it.
+ */
 public class SparkResourceAdaptor
     extends RmmWrappingDeviceMemoryResource<RmmEventHandlerResourceAdaptor<RmmDeviceMemoryResource>> {
   static {
     NativeDepsLoader.loadNativeDeps();
   }
+  /*
+   * Please note that this class itself is not 100% thread safe. Most of the thread safety is handled
+   * by RmmSpark, as no one else should interact with this class directly. There are a few functions
+   * inside RmmSpark that cannot be called with a lock held. They try to protect against this being
+   * called after it is shut down, but they are not 100% perfect. This is by design as these methods
+   * might block internally until other methods are called to wake them up. A lock would put us in
+   * a deadlock situation. This is okay because RmmSpark will not set the event handler except at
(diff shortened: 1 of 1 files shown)


revans2:
build

revans2:
build

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
```diff
@@ -56,9 +72,10 @@ public SparkResourceAdaptor(RmmEventHandlerResourceAdaptor<RmmDeviceMemoryResour
   public SparkResourceAdaptor(RmmEventHandlerResourceAdaptor<RmmDeviceMemoryResource> wrapped,
       String logLoc) {
     super(wrapped);
-    watchDog = new Thread(() -> {
+    // We are going to exit, so ignore the exception
```
zpuller, line 75: nit: this comment seems misplaced - it makes more sense in line 83 where it is I guess copied from

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
```diff
@@ -79,12 +96,12 @@ public SparkResourceAdaptor(RmmEventHandlerResourceAdaptor<RmmDeviceMemoryResour
   }
 
   @Override
-  public long getHandle() {
+  public synchronized long getHandle() {
```
zpuller, line 99: Is it possible someone can get the handle and then we can close it underneath them? I understand the PR says it's not trying to fix all possible races, but just wondering.

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
```diff
@@ -116,8 +133,13 @@ public void endRetryBlock(long threadId) {
     endRetryBlock(getHandle(), threadId);
   }
 
-  public void checkAndBreakDeadlocks() {
-    checkAndBreakDeadlocks(getHandle());
+  public synchronized void checkAndBreakDeadlocks() {
```
zpuller, line 136: Is there a possible performance penalty for making this synchronized?

zpuller (dismissed):
Couple questions but LGTM

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
revans2: yes. Technically the API is public, but it is used as if it is a private API.

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
revans2: Yes there is, but it should be very small. The C++ code has locking internally. The java code wrapping this call also has locking. So the only path where there was not locking was the one that we are trying to fix. The warchdog thread.

revans2:
build

revans2:
@zpuller please take another look. I have removed the extra comment. Thanks for finding it.

gerashegalov (approved):
LGTM

src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java
gerashegalov: IMO the other instance of this comment https://github.com/NVIDIA/spark-rapids-jni/blob/e943b86be513593675bbd56f3c3c461c17030daa/src/main/java/com/nvidia/spark/rapids/jni/SparkResourceAdaptor.java#L82 is not correct either. Ensuring a normal exit on InterruptedException is handling it rather than ignoring it.

revans2: @gerashegalov I agree it is not a great comment, but I am going to leave it because I didn't touch that line in this PR, and because your approved the PR.

gerashegalov: I did not expect a fix, just could not help but make a pedantic comment :) @revans2

mythrocks (approved):

zpuller (approved):

PR #2891, opened 2025-02-07
Enhance DataWriter to save memory during kudo serialization.

### LINKED ISSUE #2890: [FEA] Enhance DataWriter in kudo to save memory copy.

When working on actual custome workload, we found that `DataWriter` interface and its implementation could be enhanced to save memory copy.

1. Add `reserve` method to avoid repeated memory allocation.
2. Copy memory directly from host memory to `ByteArrayOutputStream` to save memory copy.

No.

**Additional context**

liurenjie1024:
Close #2890 

1. Add `reserve` method in DataWriter so that kudo data writer could reserve memory before actual writing happens. This helps avoiding unnecessay allocation and copy.
2. Add `OpenByteArrayOutputStream` and its corresponding data writer. `OpenByteArrayOutputStream` could be used in customized shuffle manager to save memory.
3. Add `ByteArrayOutputStreamWriter` which helps saving memory copy without introducing any changes to shuffle manager.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/kudo/ByteArrayOutputStreamWriter.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/kudo/ByteArrayOutputStreamWriter.java
@@ -26,8 +26,8 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 
 /**
- * Adapter class which helps to save memory copy when shuffle manager uses {@link ByteArrayOutputStream} during
- * serialization.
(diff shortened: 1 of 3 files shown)


binmahone (commented):
LGTM

revans2 (approved):
The changes look fine. I think that there are probably simpler ways to do this, but with the current code where we need to maintain compatibility with the JCUDF Serialization for the time being I think this is okay. It would be nice to have a follow on issue to come back and clean things up when we drop support for jcudf serialization.

PR #2892, opened 2025-02-07
Disable shared memmory aggregations in CUDF

revans2:
This is a work around to https://github.com/NVIDIA/spark-rapids/issues/11835 it does not really fix the underlying issue, but it should let us release 25.02 with the feature disabled while CUDF decides what to do themsleves.

PR #2909, opened 2025-02-11
Revert "Disable shared memmory aggregations in CUDF"

pmattione-nvidia:
Reverts NVIDIA/spark-rapids-jni#2892

This is no longer needed since [a fix](https://github.com/rapidsai/cudf/pull/17985) got merged into cudf.


ttnghia:
Indeed this has to be included in https://github.com/NVIDIA/spark-rapids-jni/pull/2910 for submodule sync.

pmattione-nvidia:
Oh you integrated this directly into the other PR?  Why not keep them separate?

ttnghia:
Because we cannot sync without this revert. Libcudf has changed its code thus we cannot patch anymore. As such, in order to sync, we have to remove the patch.

PR #2918, opened 2025-02-12
Put in updated patch to disable shared memory aggregations

revans2:
This will disable the shared memory feature in CUDF again. The patch was slightly updated to match the new code in CUDF.


revans2:
build

revans2:
Please do not merge this yet. We want to discuss a bit more about the bug fix that went into CUDF before we make a final decision on if it should be disabled or not.

revans2:
After more discussion I am going to close this instead.

PR #2920, opened 2025-02-13
Apply groupby patch based on rapidsai/cudf#17985 to branch-24.12

pmattione-nvidia:
This PR applies a patch to 24.12 that contains the core fix to the groupby algorithm that [was merged](https://github.com/rapidsai/cudf/pull/17985) into cuDF 25.02.

### DIFF
--- a/patches/fix_groupby.patch
+++ b/patches/fix_groupby.patch
@@ -1,0 +1,12 @@
+diff --git a/cpp/src/groupby/hash/compute_shared_memory_aggs.cu b/cpp/src/groupby/hash/compute_shared_memory_aggs.cu
+index f0361ccced..8c446d4256 100644
+--- a/cpp/src/groupby/hash/compute_shared_memory_aggs.cu
++++ b/cpp/src/groupby/hash/compute_shared_memory_aggs.cu
+@@ -222,6 +222,7 @@ CUDF_KERNEL void single_pass_shmem_aggs_kernel(cudf::size_type num_rows,
+   block.sync();
+ 
+   while (col_end < num_cols) {
++    block.sync();
+     if (block.thread_rank() == 0) {
+       calculate_columns_to_aggregate(col_start,
+                                      col_end,


patches/fix_groupby.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/cpp/src/groupby/hash/compute_shared_memory_aggs.cu b/cpp/src/groupby/hash/compute_shared_memory_aggs.cu
+index f0361ccced..8c446d4256 100644
+--- a/cpp/src/groupby/hash/compute_shared_memory_aggs.cu
++++ b/cpp/src/groupby/hash/compute_shared_memory_aggs.cu
+@@ -222,6 +222,7 @@ CUDF_KERNEL void single_pass_shmem_aggs_kernel(cudf::size_type num_rows,
+   block.sync();
+ 
+   while (col_end < num_cols) {
++    block.sync();
```
ttnghia, line 9: Can you remove the first `block.sync()` please?

ttnghia: Or I think that such sync should be added at the end of the loop instead.

Nevertheless, I'm fine to keep this to make it consistent with the merged patch to avoid any surprising indicent.

pmattione-nvidia: No, that would break it. That sync is needed to sync the change to `col_end` directly above it.  This is how the code was fixed in the cuDF PR and I'd really rather not try to optimize it and accidentally introduce more problems in this patch.

ttnghia: I believe that it is not needed since:
```
sync() <= redundant, since sync() is called again immediately below
while() {
sync()
...
}
```
But yeah I agree that we can keep this for now to avoid accidentally introducing more problems.

pmattione-nvidia: No, the while statement itself is the key.  That check there was the bug.  col_end is shared and we must sync any changes to it around the while statement.

gerashegalov (approved):

ttnghia:
Since we are applying to 24.12, I'm not sure how to merge this. Maybe it needs ops team to work on it.

pmattione-nvidia:
> Since we are applying to 24.12, I'm not sure how to merge this. Maybe it needs ops team to work on it.

The plan at the moment is to not merge it.  Perhaps that will change, but we won't do that for now.

ttnghia (approved):

sameerz:
We are going to try for a 24.12.2 patch release with this PR.  Please comment as needed, or approve if it looks ok. .  
@ttnghia 
@gerashegalov

patches/fix_groupby.patch
gerashegalov, line 1: I missed in the first review that this file should have the name starting with `1_` in order to make sure that this patch is applied after the `0_` patch. 

However, the patches have no conflict. The order is still established given 'f' > '0'.  But we need to be careful if more patches are required.

gerashegalov (approved):

pxLi:
I will go revert the patch in branch-25.02 for this patch after auto-merge

pxLi:
> > I will go revert the patch in branch-25.02 for this patch after auto-merge if needed
> 
> @pxLi This patch is not wanted in branch-25.02

yes, we have automerge from old branch to newer branches automatically, so we need to revert the merged unrequired patch in 25.02 from 24.12

PR #2928, opened 2025-02-14
Sort the entries for each map in map column according to the keys in each map

### LINKED ISSUE #2927: [FEA] Sort the entries for each map in map column according to the keys in each map

Customer specific requirement: Group by map column which requires sort the maps first.
Although vanilla Spark does not support group by map, adding implementinto as a utility to this repo is harmless.

Implement the support in this repo, but do not use it in Spark-Rapids open source repo.
This implement will be upmerged to Customer specific spark-rapids-jni repo.

**Additional context**
## Vanilla Spark does not support group by map

```scala
import org.apache.spark.sql._

val schema = StructType(Array(
  StructField("id", IntegerType, true),
  StructField("properties", MapType(StringType, StringType), true)
))
val data = Seq(
  Row(2, Map("color" -> "blue", "size" -> "L")),
  Row(3, Map("color" -> "green", "size" -> "S"))
)
spark.createDataFrame(sc.parallelize(data), schema).createOrReplaceTempView("tab")
spark.sql("select properties, sum(id) from tab group by properties").show(truncate=false)
```

Error:  
org.apache.spark.sql.AnalysisException: expression tab.properties cannot be used as a grouping expression because its data type map<string,string> is not an orderable data type.;

res-life:
closes #2927

Map in CUDF is List<struct<KEY, VALUE>>.
First sort struct<KEY, VALUE> using `cudf::segmented_sort_by_key`, then construct list column.

### DIFF
--- a/src/main/cpp/src/map.cu
+++ b/src/main/cpp/src/map.cu
@@ -34,7 +34,6 @@ namespace spark_rapids_jni {
 
 std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& map_column,
                                               cudf::order sort_order,
-                                              cudf::null_order null_order,
                                               rmm::cuda_stream_view stream,
                                               rmm::device_async_resource_ref mr)
 {
@@ -52,18 +51,18 @@ std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& map_colum
                                             cudf::table_view{{keys}},
                                             segments,
                                             {sort_order},
-                                            {null_order},
+                                            {},  // Map keys MUST be not null
                                             stream,
                                             mr);
   stream.synchronize();
   std::vector<std::unique_ptr<cudf::column>> k_v = sorted->release();
 
+  // Note: The keys in a map MUST not be null, so the struct<Key, Value> MUST not be null too.
+  // Of course, the null count is zero for struct<Key Value> column
   auto mask = cudf::create_null_mask(keys.size(), cudf::mask_state::UNALLOCATED, stream, mr);
   stream.synchronize();
-
   auto sorted_struct =
     cudf::make_structs_column(keys.size(), std::move(k_v), 0, std::move(mask), stream, mr);
-  stream.synchronize();
 
   // clone segments
   auto copied_segements = cudf::make_numeric_column(cudf::data_type(segments.type().id()),
@@ -72,15 +71,12 @@ std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& map_colum
                                                     stream,
                                                     mr);
   stream.synchronize();
-
   CUDF_CUDA_TRY(cudaMemcpyAsync(copied_segements->mutable_view().data<int32_t>(),
                                 segments.data<int32_t>(),
                                 segments.size() * sizeof(int32_t),
                                 cudaMemcpyDeviceToDevice,
                                 stream.value()));
   stream.synchronize();
-
-  printf("mydebug: success \n");
 
   return cudf::make_lists_column(lists_of_structs.size(),
                                  std::move(copied_segements),  // offsets
--- a/src/main/cpp/src/map.cu
+++ b/src/main/cpp/src/map.cu
@@ -14,31 +14,29 @@
  * limitations under the License.
  */
 
-#include "cudf_jni_apis.hpp"
-#include "jni_utils.hpp"
 #include "map.hpp"
 
 #include <cudf/column/column_factories.hpp>
-#include <cudf/column/column_view.hpp>
 #include <cudf/lists/lists_column_view.hpp>
 #include <cudf/null_mask.hpp>
 #include <cudf/sorting.hpp>
+#include <cudf/table/table.hpp>
 #include <cudf/table/table_view.hpp>
-#include <cudf/utilities/default_stream.hpp>
 #include <cudf/utilities/error.hpp>
-
-#include <rmm/cuda_stream_view.hpp>
-#include <rmm/resource_ref.hpp>
 
 namespace spark_rapids_jni {
 
-std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& map_column,
+std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& input,
                                               cudf::order sort_order,
                                               rmm::cuda_stream_view stream,
                                               rmm::device_async_resource_ref mr)
 {
-  auto const& lists_of_structs = cudf::lists_column_view(map_column);
-  auto const structs           = lists_of_structs.child();
+  CUDF_EXPECTS(input.type().id() == cudf::type_id::LIST,
+               "maps_column_view input must be LIST type");
+  if (input.size() == 0) { return cudf::make_empty_column(input.type()); }
+
+  auto const lists_of_structs = cudf::lists_column_view(input);
+  auto const structs          = lists_of_structs.child();
   CUDF_EXPECTS(structs.type().id() == cudf::type_id::STRUCT,
                "maps_column_view input must have exactly 1 child (STRUCT) column.");
   CUDF_EXPECTS(structs.num_children() == 2,
@@ -49,7 +47,7 @@ std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& map_colum
   CUDF_EXPECTS(keys.null_count() == 0, "maps_column_view keys must have no null.");
   auto segments = lists_of_structs.offsets();
 
-  auto sorted = cudf::segmented_sort_by_key(cudf::table_view{{keys, values}},
+  auto sorted = cudf::segmented_sort_by_key(cudf::table_view{{structs}},
                                             cudf::table_view{{keys}},
                                             segments,
                                             {sort_order},
@@ -57,14 +55,7 @@ std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& map_colum
                                             stream,
                                             mr);
   stream.synchronize();
-  std::vector<std::unique_ptr<cudf::column>> k_v = sorted->release();
-
-  // Note: The keys in a map MUST not be null, so the struct<Key, Value> MUST not be null too.
-  // Of course, the null count is zero for struct<Key Value> column
-  auto mask = cudf::create_null_mask(keys.size(), cudf::mask_state::UNALLOCATED, stream, mr);
-  stream.synchronize();
-  auto sorted_struct =
-    cudf::make_structs_column(keys.size(), std::move(k_v), 0, std::move(mask), stream, mr);
+  std::vector<std::unique_ptr<cudf::column>> one_item_vec = sorted->release();
 
   // clone segments
   auto copied_segements = cudf::make_numeric_column(cudf::data_type(segments.type().id()),
@@ -82,7 +73,7 @@ std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& map_colum
 
   return cudf::make_lists_column(lists_of_structs.size(),
                                  std::move(copied_segements),  // offsets
-                                 std::move(sorted_struct),     // child column
+                                 std::move(one_item_vec[0]),   // child column
                                  lists_of_structs.null_count(),
                                  cudf::copy_bitmask(lists_of_structs.parent(), stream, mr),
                                  stream,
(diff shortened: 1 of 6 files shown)


res-life:
build

res-life:
build

src/main/cpp/src/map.hpp
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cudf/column/column_view.hpp>
+#include <cudf/table/table_view.hpp>
```
ttnghia, line 18:
```suggestion
```

src/main/cpp/src/map.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <cudf/utilities/default_stream.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/resource_ref.hpp>
```
ttnghia, line 22: Replace this by `cudf/utilities/memory_resource.hpp`.

src/main/cpp/src/map.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace spark_rapids_jni {
+
+std::unique_ptr<cudf::column> sort_map_column(
+  cudf::column_view const& map_column,
+  cudf::order sort_order,
+  rmm::cuda_stream_view stream      = cudf::get_default_stream(),
+  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource());
```
ttnghia, line 30:
```suggestion
  rmm::device_async_resource_ref mr = cudf::get_current_device_resource());
```

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,92 @@
...
+#include "jni_utils.hpp"
+#include "map.hpp"
+
+#include <cudf/column/column_factories.hpp>
+#include <cudf/column/column_view.hpp>
+#include <cudf/lists/lists_column_view.hpp>
+#include <cudf/null_mask.hpp>
+#include <cudf/sorting.hpp>
+#include <cudf/table/table_view.hpp>
+#include <cudf/utilities/default_stream.hpp>
```
ttnghia, line 27:
```suggestion
```

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,92 @@
...
+#include <cudf/utilities/error.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/resource_ref.hpp>
```
ttnghia, line 31:
```suggestion
```

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,92 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cudf_jni_apis.hpp"
+#include "jni_utils.hpp"
+#include "map.hpp"
+
+#include <cudf/column/column_factories.hpp>
+#include <cudf/column/column_view.hpp>
```
ttnghia, line 22:
```suggestion
```

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cudf_jni_apis.hpp"
+#include "jni_utils.hpp"
```
ttnghia, line 18:
```suggestion
```

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,92 @@
...
+#include <rmm/resource_ref.hpp>
+
+namespace spark_rapids_jni {
+
+std::unique_ptr<cudf::column> sort_map_column(cudf::column_view const& map_column,
+                                              cudf::order sort_order,
+                                              rmm::cuda_stream_view stream,
+                                              rmm::device_async_resource_ref mr)
+{
+  auto const& lists_of_structs = cudf::lists_column_view(map_column);
```
ttnghia, line 40: Reference to temporary.
```suggestion
  auto const lists_of_structs = cudf::lists_column_view(map_column);
```

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,92 @@
...
+  auto const structs           = lists_of_structs.child();
+  CUDF_EXPECTS(structs.type().id() == cudf::type_id::STRUCT,
+               "maps_column_view input must have exactly 1 child (STRUCT) column.");
+  CUDF_EXPECTS(structs.num_children() == 2,
+               "maps_column_view key-value struct must have exactly 2 children.");
+  auto keys   = structs.child(0);
+  auto values = structs.child(1);
+  CUDF_EXPECTS(structs.null_count() == 0, "maps_column_view key-value struct must have no null.");
+  CUDF_EXPECTS(keys.null_count() == 0, "maps_column_view keys must have no null.");
+  auto segments = lists_of_structs.offsets();
```
ttnghia, line 50: We should consider the case when the input is empty, and return empty column accordingly.

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,92 @@
...
+                                            mr);
+  stream.synchronize();
+  std::vector<std::unique_ptr<cudf::column>> k_v = sorted->release();
+
+  // Note: The keys in a map MUST not be null, so the struct<Key, Value> MUST not be null too.
+  // Of course, the null count is zero for struct<Key Value> column
+  auto mask = cudf::create_null_mask(keys.size(), cudf::mask_state::UNALLOCATED, stream, mr);
+  stream.synchronize();
+  auto sorted_struct =
+    cudf::make_structs_column(keys.size(), std::move(k_v), 0, std::move(mask), stream, mr);
```
ttnghia, line 67: Instead of sorting the table `{keys, values}`, we can just sort the table `{structs}`. Am I missing something?

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,92 @@
...
+                                                    cudf::mask_state::UNALLOCATED,
+                                                    stream,
+                                                    mr);
+  stream.synchronize();
+  CUDF_CUDA_TRY(cudaMemcpyAsync(copied_segements->mutable_view().data<int32_t>(),
+                                segments.data<int32_t>(),
+                                segments.size() * sizeof(int32_t),
+                                cudaMemcpyDeviceToDevice,
+                                stream.value()));
+  stream.synchronize();
```
ttnghia, line 81: This is just:
```
auto output_segments = std::make_unique<cudf::column>(segments);
```

res-life: IIUC: The passed in map ColumnView is managed by Java side ColumnVector via  refCount in it. If the refCount is zero, the Java side will release the whole map column including the `offsets` sub column.  If we use `offsets` directly, we may reference released memory.

ttnghia: No we are not using `offsets` in the final output. We copy it by the constructor of `cudf::column` as my code above.

res-life:
build

res-life:
build

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,73 @@
...
+               "maps_column_view input must have exactly 1 child (STRUCT) column.");
+  CUDF_EXPECTS(structs.num_children() == 2,
+               "maps_column_view key-value struct must have exactly 2 children.");
+  auto keys   = structs.child(0);
+  auto values = structs.child(1);
+  CUDF_EXPECTS(structs.null_count() == 0, "maps_column_view key-value struct must have no null.");
+  CUDF_EXPECTS(keys.null_count() == 0, "maps_column_view keys must have no null.");
+  auto segments = lists_of_structs.offsets();
+
+  auto sorted = cudf::segmented_sort_by_key(cudf::table_view{{structs}},
```
revans2, line 50: nit: Could you also add in a comment that we can assume that maps do not have duplicate keys in them so we don't have to worry about it?

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,73 @@
...
+                                            cudf::table_view{{keys}},
+                                            segments,
+                                            {sort_order},
+                                            {},  // Map keys MUST be not null
+                                            stream,
+                                            mr);
+  stream.synchronize();
```
revans2, line 57: Why do we need to synchronize the stream here and below? I get it if the deep copy blow might use the wrong stream, but creating one_item_vec is not doing a copy so I am a little confused.

revans2 (commented):
I assume that if we had a Map<String,Map<String,String>> or something similar that the spark could would handle applying the this method first to the child map and then to the parent map so that the normalization is correct?

src/main/cpp/src/map.cu
```diff
@@ -0,0 +1,73 @@
...
+
+  // Deep copying segments
+  auto output_segments = std::make_unique<cudf::column>(segments);
+  stream.synchronize();
+
+  return cudf::make_lists_column(lists_of_structs.size(),
+                                 std::move(output_segments),  // offsets
+                                 std::move(one_item_vec[0]),  // child column
+                                 lists_of_structs.null_count(),
+                                 cudf::copy_bitmask(lists_of_structs.parent(), stream, mr),
```
ttnghia, line 68:
```suggestion
  return cudf::make_lists_column(input.size(),
                                 std::move(std::make_unique<cudf::column>(segments)),  // offsets
                                 std::move(sorted->release().front()),  // child column
                                 input.null_count(),
                                 cudf::copy_bitmask(input, stream, mr),
```

res-life:
> I assume that if we had a Map<String,Map<String,String>> or something similar that the spark could would handle applying the this method first to the child map and then to the parent map so that the normalization is correct?

Map type in key as a child is not allowed:
Customer code is using `TypeUtils.getInterpretedOrdering` to sort key.
```scala
private lazy val keyOrdering: Ordering[Any] = TypeUtils.getInterpretedOrdering(keyType)
```

TypeUtils.getInterpretedOrdering:
```scala
  def getInterpretedOrdering(t: DataType): Ordering[Any] = {
    t match {
      case i: AtomicType => i.ordering.asInstanceOf[Ordering[Any]]
      case a: ArrayType => a.interpretedOrdering.asInstanceOf[Ordering[Any]]
      case s: StructType => s.interpretedOrdering.asInstanceOf[Ordering[Any]]
      case udt: UserDefinedType[_] => getInterpretedOrdering(udt.sqlType)
    }
  }
```
The map type in child will causes MatchError error:
```
scala.MatchError: MapType(IntegerType,IntegerType,true) (of class org.apache.spark.sql.types.MapType)
  at org.apache.spark.sql.catalyst.util.TypeUtils$.getInterpretedOrdering(TypeUtils.scala:83)
  ... 49 elided

ArrayType.interpretedOrdering and StructType.interpretedOrdering both do not supports map as a child.

**Comment by res-life** [comment]
build

File: src/main/cpp/src/map.cu (hunk 14 above)
Comment (res-life) [reply src/main/cpp/src/map.cu]: Done

File: src/main/cpp/src/map.cu (hunk 13 above)
Comment (res-life) [reply src/main/cpp/src/map.cu]: Done.

File: src/main/cpp/src/map.cu (hunk 12 above)
Comment (res-life) [reply src/main/cpp/src/map.cu:50]: Done.

File: src/main/cpp/src/map.cu [hunk 15]
```diff
@@ -0,0 +1,67 @@
...
+                                            {sort_order},
+                                            {},  // Map keys MUST not be null
+                                            stream,
+                                            mr);
+
+  return cudf::make_lists_column(lists_of_structs.size(),
+                                 std::make_unique<cudf::column>(segments),  // copy segment offsets
+                                 std::move(sorted->release().front()),      // child column
+                                 lists_of_structs.null_count(),
+                                 cudf::copy_bitmask(lists_of_structs.parent(), stream, mr),
```
ttnghia, line 62: `lists_of_structs.parent()` is `input`.
```suggestion
                                 input.null_count(),
                                 cudf::copy_bitmask(input, stream, mr),
```

res-life:
build

src/main/cpp/src/map.cu
res-life: Done.

ttnghia (approved):

PR #2944, opened 2025-02-17
Remove unnecessary kudo metrics

liurenjie1024:
Part 2 of https://github.com/NVIDIA/spark-rapids/issues/12145

Remove some unnecessary metrics of kudo.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/kudo/WriteMetrics.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/kudo/WriteMetrics.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, NVIDIA CORPORATION.
+ * Copyright (c) 2024-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoSerializer.java
```diff
@@ -302,20 +312,32 @@ public Pair<KudoHostMergeResult, MergeMetrics> mergeOnHost(List<KudoTable> kudoT
...
+    }
+  }
+
+
+  /**
+   * See {@link #mergeToTable(KudoTable[])}.
+   *
+   * @deprecated Use {@link #mergeToTable(KudoTable[])} instead.
+   */
+  @Deprecated
```
binmahone, line 340: are we planning to delete this in another PR ?

binmahone: Is it mandatory to approach it in such a tedious way?  Can we just make changes at both spark-rapids-jni side and spark-rapids side, then merge the PR simultaneously? The reason why I ask this is that I also have some incompatible (so I cannot use this Deprecate annotation way) PR changes in spark-rapids-jni at hand, and I'm wondering what's the best way to check in such kind of changes, also cc @revans2

binmahone (approved):

PR #2976, opened 2025-02-21
Fix a corner case in the SparkResourceAdaptor state machine

revans2:
This fixes https://github.com/NVIDIA/spark-rapids/issues/12158

But it is not there are no real test changes as a part of this because it we are not set up for testing spill in spark-rapids-jni.


binmahone (dismissed):

binmahone:
The code change looks good, I'm okay with it. And after it is checked in I'll try to add my test case as mentioned in https://github.com/NVIDIA/spark-rapids/issues/12158

BTW, I tend to avoid defining this issue as a corner case, actually it's quite common when I'm working on CPU OOM issues.

revans2:
@binmahone please take another look. I have not added any tests because spilling is hard to simulate in this repo, but I have manually tested it and it appears to be working.

PR #3025, opened 2025-03-11
(JNI part) fix premature concensus on "all blocked" in OOM state machine

binmahone:
This PR contributes to https://github.com/NVIDIA/spark-rapids/issues/12311, it should work together with https://github.com/NVIDIA/spark-rapids/pull/12312


revans2 (changes requested):
This will not work unless we have a way to reliably detect if a thread is blocked. The java thread state is not good enough in all cases. A lot of java IO currently will block in a native method and the thread state appears as runnable.

This makes things especially difficult for python UDFs where up until version 4.0 there are separate read threads and write threads. If we block the read thread because of low memory the write thread will eventually block when all of the memory buffers between the processes fill up. But it will not appear to be blocked because it is blocked in a native method trying to write.

This change will cause that case to deadlock.

binmahone:
@revans2 , the PR is updated to take care of the python UDF case as well

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -599,6 +599,34 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     }
   }
 
+  /**
+   * Bind two threads together so that if one is blocked the other should be blocked too.
+   */
+  void bind_propagate_threads(long const first, long const second)
```
revans2, line 605: See https://github.com/NVIDIA/spark-rapids/pull/12312#discussion_r1995602239

I would rather flip this around and have the default be any blocked, and inform the state machine of thread pool threads instead.

binmahone:
build

binmahone:
hi @revans2 , the PR is updated to meet the requirements in https://github.com/NVIDIA/spark-rapids/pull/12312#discussion_r1995602239. Actually it reuses the pre existing `RmmSpark.poolThreadWorkingOnTask` for the threadpool threads of gpu orc/parquet/avro scan. 

For other places using thread pool, they may have  similar issue with https://github.com/NVIDIA/spark-rapids/issues/12311, but this PR have not covered all of them. That's why this PR will not close https://github.com/NVIDIA/spark-rapids/issues/12311

binmahone:
build

revans2 (approved):

PR #3045, opened 2025-03-19
Fix bugs in HyperLogLogPlusPlus(approximate count distinct)

res-life:
- Bug 1:
Fix reduce bug: does not handle all sketch longs
Result is diff when precision is 12, 4096 register values, e.g.:  block size is 256, it needs 410 longs to save a sketch.
Each thread MUST handle multiple long columns instead of 1 long column.

Refer to the 1st commit
- Bug 2:
Fix a bug in test case
Typo error
Refer to the 2nd commit

- Bug 3:
Fix a bug which access out bound memory when precision = 4
Did not exit out bound threads in advance.
Refer to the 3rd commit

- Bug 4:
Fix shared mem is not enough bug when precision >= 14.
Refer to the 4th commit.

### DIFF
--- a/src/main/cpp/src/hyper_log_log_plus_plus.cu
+++ b/src/main/cpp/src/hyper_log_log_plus_plus.cu
@@ -46,13 +46,13 @@ namespace {
  * is bigger than this value, then leverage global memory to do the reduction to avoid the
  * limitation of shared memory.
  */
-constexpr int MAX_SHARED_MEM_PRECISION = 12
+constexpr int MAX_SHARED_MEM_PRECISION = 12;
 
-  /**
-   * @brief The seed used for the XXHash64 hash function.
-   * It's consistent with Spark
-   */
-  constexpr int64_t SEED = 42L;
+/**
+ * @brief The seed used for the XXHash64 hash function.
+ * It's consistent with Spark
+ */
+constexpr int64_t SEED = 42L;
 
 /**
  * @brief 6 binary MASK bits: 111-111


res-life:
build

res-life:
These bugs were discovered by test cases in [link](https://github.com/NVIDIA/spark-rapids/pull/11638/files#diff-75e727c69138ceb6a9fbf2f10c9c91e33542199799b20ea42277a9d0e5526e85)
```python
# precision = Math.ceil(2.0d * Math.log(1.106d / relativeSD) / Math.log(2.0d)).toInt
_relativeSD = [
    0.3,   #  precision 4
    0.25,  #  precision 5
    0.15,  #  precision 6
    0.1,   #  precision 7
    0.08,  #  precision 8
    0.05,  #  precision 9
    0.04,  #  precision 10
    0.03,  #  precision 11
    0.02,  #  precision 12
    0.015, #  precision 13
    0.01,  #  precision 14
    0.008, #  precision 15
    0.005, #  precision 16
    0.004, #  precision 17
    0.003, #  precision 18
]

@ignore_order(local=True)
@pytest.mark.parametrize('relativeSD', _relativeSD, ids=idfn)
def test_hllpp_precisions_reduce(relativeSD):
    assert_gpu_and_cpu_are_equal_sql(
        lambda spark: unary_op_df(spark, int_gen),
        "tab",
        f"select APPROX_COUNT_DISTINCT(a, {relativeSD}) from tab")

@ignore_order(local=True)
@pytest.mark.parametrize('relativeSD', _relativeSD, ids=idfn)
def test_hllpp_precisions_groupby(relativeSD):
    assert_gpu_and_cpu_are_equal_sql(
        lambda spark: gen_df(spark, [("c1", int_gen), ("c2", int_gen)]),
        "tab",
        f"select c1, APPROX_COUNT_DISTINCT(c2, {relativeSD}) from tab group by c1")

```

res-life:
@ttnghia  Could you please help review this PR, thanks.

src/main/cpp/src/hyper_log_log_plus_plus.cu
```diff
@@ -675,7 +680,7 @@ __launch_bounds__(block_size) CUDF_KERNEL
 
   // init tmp data
   for (int i = tid; i < num_registers_per_sketch; i += block_size) {
-    shared_data[i] = 0;
+    cache[i] = 0;
```
ttnghia, line 683:
```suggestion
    memset(cache, 0, num_registers_per_sketch * sizeof(int));
```

src/main/cpp/src/hyper_log_log_plus_plus.cu
```diff
@@ -663,9 +664,13 @@ template <int block_size>
...
                            cudf::device_span<int64_t*> output,
-                           int precision)
+                           int precision,
+                           int32_t* mem_cache)
 {
-  __shared__ int32_t shared_data[block_size];
+  extern __shared__ int32_t shared_mem_cache[];
+
+  int32_t* cache = shared_mem_cache;
+  if (precision > 12) { cache = mem_cache; }
```
ttnghia, line 673: Instead of hardcoding `12` in several places, please use a global variable, for example: `constexpr int MAX_SHARED_MEM_PRECISION = 12`.

ttnghia, line 673: This should be better and less error-prone:
```suggestion
  int32_t* cache = mem_cache ? mem_cache : shared_mem_cache; 
```

src/main/cpp/src/hyper_log_log_plus_plus.cu
res-life: memset does not work here, each thread does ont set a continuous block of data:
```cpp
i += block_size
```

res-life:
build

ttnghia (approved):

res-life:
build

firestarman (approved):
Approve this PR because of only a quite simple change (adding a `;`) after the previous apporval to unblock the following work.

PR #3049, opened 2025-03-20
Expose getter functions for KudoHostMergeResult

binmahone:
contributes to https://github.com/NVIDIA/spark-rapids/issues/12345
This is the JNI part.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoHostMergeResult.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoHostMergeResult.java
@@ -30,6 +30,18 @@ public class KudoHostMergeResult implements AutoCloseable {
   private final ColumnViewInfo[] columnInfoList;
   private HostMemoryBuffer hostBuf;
 
+  public Schema getSchema() {
+    return schema;
+  }
+
+  public ColumnViewInfo[] getColumnInfoList() {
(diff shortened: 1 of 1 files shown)


src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoHostMergeResult.java
```diff
@@ -61,6 +73,11 @@ public long getDataLength() {
    * @return the cudf table
    */
   public Table toTable() {
+    return toTableStatic(hostBuf, schema, columnInfoList);
+  }
+
+  public static Table toTableStatic(HostMemoryBuffer hostBuf,
```
firestarman, line 79: NIT
Not it is an API with 3 separate arguments, so better have some comment for the arguments and what will happen if these arguments mismatch.

firestarman (approved):
LGTM, only a nit

firestarman:
build

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoHostMergeResult.java
binmahone: resolved

firestarman: Good to me

PR #3072, opened 2025-03-26
Add Timezone Rule

warrickhe:
Adds timezone transitions to Spark, as part of https://github.com/NVIDIA/spark-rapids/issues/6840.

Not to be merged until all the following is completed.

- [x] Complete implementation on spark-rapids side
- [x] Set the end year to be user configurable


warrickhe:
build

warrickhe:
Things to consider: how to inform user that CPU fallback occurred?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,63 +122,114 @@ private static synchronized void closeResources()  {
...
+  // enforce that all timestamps, regardless of timezone, be less than the desired date
+  private static boolean isValidInput(ColumnVector input, ZoneId zoneId){
+    if (zoneId.getRules().isFixedOffset()){
+      return true;
+    }
+    boolean isValid = false;
+    long scaleFactor = getScalefactor(input);
+    try (Scalar targetTimestamp = Scalar.timestampFromLong(input.getType(), maxTimestamp*scaleFactor);
+         ColumnVector isGreater = input.binaryOp(BinaryOp.GREATER, targetTimestamp, DType.BOOL8)) {
+      isValid = !isGreater.any().getBoolean();
```
revans2, line 161: This leaks the scalar value that is returned. Also if all of the input values are null, then any returns a null and the getBoolean value is invalid.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,63 +122,114 @@ private static synchronized void closeResources()  {
...
+          timestamp = instant.atZone(targetTimeZone).toLocalDateTime().atZone(currentTimeZone)
+            .toInstant().getEpochSecond();
+          timestamp = timestamp * scaleFactor + unitOffset;
+          builder.append(timestamp);
+        }
+        
+        resultCV = builder.buildAndPutOnDevice();
+      }
+    } catch (Exception e) {
+      throw e;
```
revans2, line 198: nit: If we are just going to rethrow it, don't bother to catch it.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -55,18 +60,23 @@ public class GpuTimeZoneDB {
   private static Map<String, Integer> zoneIdToTable;
 
   // use this reference to indicate if time zone cache is initialized.
-  private static HostColumnVector fixedTransitions;
+  private static HostColumnVector transitions;
+  // initial year set to 1900 because some transition rules start early
+  private static final int initialTransitionYear = 1900;
+  private static int finalTransitionYear = 2200;
```
revans2, line 66: I would prefer to have this set in the cache API and never have a default.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,63 +122,114 @@ private static synchronized void closeResources()  {
...
+        
+        resultCV = builder.buildAndPutOnDevice();
+      }
+    } catch (Exception e) {
+      throw e;
+    }
+    return resultCV;
+  }
+
+  public static ColumnVector fromTimestampToUtcTimestamp(ColumnVector input, ZoneId currentTimeZone, int maxYear) {
```
revans2, line 203: I get what you are trying to do by passing in `maxYear`, but I think it is not great API design.

The issue is that first wins when it comes to setting this value. By having it in the API it implies that that `maxYear` can be changed for each API call, which is not true. It only changes when `cacheDatabaseAsync` is called, assuming that there are no really bad race conditions on startup.

Could we do one of two things?

1. Have cacheDatabaseImpl check if `transitions != null &&  maxYear != finalTransitionYear`, output a warning that this is not supported. Then add in javadocs to each of these APIs explaining that first one wins.
2. Remove `maxYear` from these APIs and instead of `cacheDatabase` we call `verifyDatabaseCached`, where it would not try and load the database, it would just synchronize and then verify that `transitions != null`

revans2 (commented):
In general it looks good.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
warrickhe: In the second option, how would we pass in a maxYear?

warrickhe: The first option is easy, but I much prefer to remove the parameter entirely. Or if we have maxYear, we could make it so that a greater value for maxYear will increase the range of transitions.

warrickhe: So for the second approach, just add cacheDatabaseAsync(maxYear) and have that be the only place where we cache all times? From what I can see, it isn't added anywhere yet.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +122,94 @@ private static synchronized void closeResources()  {
...
+    try {
+      getZoneId(zoneId); // check that zoneId is a valid zone
+      return true;
+    } catch (ZoneRulesException e) {
+      return false;
+    }
+  }
+
+  // enforce that all timestamps, regardless of timezone, be less than the desired date
+  private static boolean isValidInput(ColumnVector input, ZoneId zoneId){
```
jihoonson, line 153: The function name seems misleading a bit. It makes me think that the `input` will be invalid if this function returns false. But it seems to actually indicate that the computation should be done in CPU instead. Perhaps `fallBackToCpu` is a better name?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -193,12 +261,25 @@ private static synchronized void loadData() {
...
-        if (!zoneRules.isFixedOffset() && !zoneRules.getTransitionRules().isEmpty()) {
-          continue;
-        }
         if (!zoneIdToTable.containsKey(zoneId.getId())) {
-          List<ZoneOffsetTransition> transitions = zoneRules.getTransitions();
+          List<ZoneOffsetTransition> zoneOffsetTransitions = new ArrayList<>(zoneRules.getTransitions());
+          // It is desired to get lastTransitionEpochSecond because some rules don't start until late (e.g. 2007)
+          long lastTransitionEpochSecond = Long.MIN_VALUE;
+          if (!zoneOffsetTransitions.isEmpty()) {
+            long transitionInstant = zoneOffsetTransitions.get(zoneOffsetTransitions.size()-1).getInstant().getEpochSecond();
```
jihoonson, line 269: It seems like that the Java contract does not guarantee that `zoneRules.getTransitions()` would always return a sorted result. We should not make such an assumption unless it is clearly stated in the contract. I would recommend computing `lastTransitionEpochSecond` after `zoneOffsetTransitions` is sorted as we are going to sort it anyway.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -261,7 +342,7 @@ private static synchronized Table getTransitions() {
   }
 
   private static synchronized ColumnVector getFixedTransitions() {
-    return fixedTransitions.copyToDevice();
+    return transitions.copyToDevice();
```
jihoonson, line 345: I understand this was missing previously, but it is a good practice to check if `transitions` is null before it accesses `transitions`. Could you add `verifyDatabaseCached()` here?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -280,7 +361,7 @@ static synchronized List getHostFixedTransitions(String zoneId) {
     if (idx == null) {
       return null;
     }
-    return fixedTransitions.getList(idx);
+    return transitions.getList(idx);
```
jihoonson, line 364: Same comment. Would be nice to add `verifyDatabaseCached()` before it reads `transitions.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +122,94 @@ private static synchronized void closeResources()  {
...
+          if (hostCV.isNull(i)) {
+            builder.appendNull();
+            continue;
+          }
+          
+          long timestamp = hostCV.getLong(i);
+          long unitOffset = timestamp % scaleFactor;
+          timestamp /= scaleFactor;
+          Instant instant = Instant.ofEpochSecond(timestamp);
+          timestamp = instant.atZone(targetTimeZone).toLocalDateTime().atZone(currentTimeZone)
```
jihoonson, line 191: Could you please add a comment explaining this conversion? It would help people to understand this code.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +122,94 @@ private static synchronized void closeResources()  {
...
+  }
+
+  // enforce that all timestamps, regardless of timezone, be less than the desired date
+  private static boolean isValidInput(ColumnVector input, ZoneId zoneId){
+    if (zoneId.getRules().isFixedOffset()){
+      return true;
+    }
+    boolean isValid = false;
+    long scaleFactor = getScalefactor(input);
+    try (Scalar targetTimestamp = Scalar.timestampFromLong(input.getType(), maxTimestamp*scaleFactor);
```
jihoonson, line 159: `maxTimestamp` is static private and seems to be default initialized, which will be set to 0. Did you intend to have it passed as a parameter of this function instead?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +122,94 @@ private static synchronized void closeResources()  {
...
+      // don't need to throw error, can try CPU processing
+      return false;
+    }
+    return isValid;
+  }
+
+  private static ColumnVector cpuChangeTimestampTz(ColumnVector input, ZoneId currentTimeZone, ZoneId targetTimeZone) {
+    ColumnVector resultCV = null;
+    try (HostColumnVector hostCV = input.copyToHost()) {
+      // assuming we don't have more than 2^31-1 rows
```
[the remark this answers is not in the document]
warrickhe: Dated comment from a prior implementation where I stored columns in an array first. This meant that the size could only be an integer amount. Will remove

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
warrickhe: I will remove this and move this code to the try statement in getTransitions. getTransition calls are only used in fromTimestampToUtcTimestamp. I can just move it so that the verify calls are within the getTransition instead

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
warrickhe: Ahhhhh, good catch, I didn't update when addressing for Bobby's comments. I will update it

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +124,103 @@ private static synchronized void closeResources()  {
...
+    try {
+      getZoneId(zoneId); // check that zoneId is a valid zone
+      return true;
+    } catch (ZoneRulesException e) {
+      return false;
+    }
+  }
+
+  // enforce that all timestamps, regardless of timezone, be less than the desired date
+  private static boolean shouldFallbackToCpu(ColumnVector input, ZoneId zoneId){
```
NVnavkumar, line 155: This method is not named according to the logic. It actually returns **true** for inputs where the computation should run on GPU, and **false** for inputs that will be copied to host and the computation will run on the CPU.

I would suggest either renaming it to `canRunOnGpu` or reversing the logic to return true for CPU and false for GPU.

src/test/java/com/nvidia/spark/rapids/jni/TimeZoneTest.java
```diff
@@ -44,10 +44,10 @@ static void cleanup() {
...
     // Check for a few timezones
-    List transitions = GpuTimeZoneDB.getHostFixedTransitions("UTC+8");
+    List transitions = GpuTimeZoneDB.getHostTransitions("UTC+8");
     assertNotNull(transitions);
     assertEquals(1, transitions.size());
-    transitions = GpuTimeZoneDB.getHostFixedTransitions("Asia/Shanghai");
+    transitions = GpuTimeZoneDB.getHostTransitions("Asia/Shanghai");
     assertNotNull(transitions);
     ZoneId shanghai = ZoneId.of("Asia/Shanghai").normalized();
     assertEquals(shanghai.getRules().getTransitions().size() + 1, transitions.size());
```
NVnavkumar, line 53: Please add a check for a non-fixed timezone here. Should verify that the cached rule was created.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +124,103 @@ private static synchronized void closeResources()  {
...
+          if (hostCV.isNull(i)) {
+            builder.appendNull();
+            continue;
+          }
+          
+          long timestamp = hostCV.getLong(i);
+          long unitOffset = timestamp % scaleFactor;
+          timestamp /= scaleFactor;
+          Instant instant = Instant.ofEpochSecond(timestamp);
+          /***
```
NVnavkumar, line 192: Minor nit, any reason for the javadoc style comment here? Shouldn't this be moved above the method definition in that case?

NVnavkumar (commented):
Mostly looks good, just a few things

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
warrickhe: Jihoon noted that this piece of logic was a little confusing, so I added this to help clarify the logic for anyone who might touch the code later. I think it's mainly confusing because you start by setting the zone to the targetTimeZone instead of the current one.

NVnavkumar: Yeah it's the CPU analogue of what we actually do on the GPU when we support the timezone.  I assume you based it off what Apache Spark itself does. Can we link to the Apache Spark code on GitHub as reference point for this algorithm? That might help resolve the confusion aspect.

The original comment here is more of stylistic question. In general javadoc style comments with `/**` should typically be prefixing the definitions of methods and classes. So in this case since the comment is more inline with code, we should use either a standard block comment `/*...*/` or double-slash `//`.

warrickhe: I see, I'll make the suggested changes. I'll keep the comment here since the spark source code also doesn't really explain it's logic

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +132,118 @@ private static synchronized void closeResources()  {
...
+    } catch (ZoneRulesException e) {
+      return false;
+    }
+  }
+
+  /**
+   * TO BE DELETED: added to not break spark-rapids build
+   * Will be removed after spark-rapids changes are merged,
+   * since all timezones will be supported.
+   * This function will not be called after spark-rapids changes
```
jihoonson, line 166: nit: this is OK, but you can use the `@Deprecated` annotation next time.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -77,13 +88,25 @@ public static void cacheDatabaseAsync() {
...
+    cacheDatabaseImpl(maxYear);
+  }
+
+  /**
+   * TO BE DELETED: added to not break spark-rapids build
+   * Will be removed after spark-rapids changes are committed
+   */
   public static void cacheDatabase() {
-    cacheDatabaseImpl();
+    cacheDatabaseImpl(9999);
```
jihoonson, line 109: Wonder if this will impact any CI jobs which may have small memory space. I would suggest to use the default max year in https://github.com/NVIDIA/spark-rapids/pull/12502 to avoid any unnecessary hassles.

warrickhe:
Changes pass all testcases on both the 25.06 branch and the https://github.com/NVIDIA/spark-rapids/pull/12502 PR

warrickhe:
build

warrickhe:
Can someone trigger blossom for this repo, I don't have the permissions here

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +132,118 @@ private static synchronized void closeResources()  {
...
+    if (zoneId.getRules().isFixedOffset()){
+      return false;
     }
+    boolean isValid = false;
+    long scaleFactor = getScaleFactor(input);
+    try (Scalar targetTimestamp = Scalar.timestampFromLong(input.getType(), maxTimestamp*scaleFactor);
+         ColumnVector compareCv = input.binaryOp(BinaryOp.GREATER, targetTimestamp, DType.BOOL8);
+         Scalar isGreater = compareCv.any() ) {
+      if (!isGreater.isValid()) isValid = false;
+      else isValid = isGreater.getBoolean();
```
res-life, line 185:
```suggestion
      if (!isGreater.isValid()) {
        isValid = false;
      }
      else {
        isValid = isGreater.getBoolean();
      }  
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +132,118 @@ private static synchronized void closeResources()  {
...
+  // enforce that all timestamps, regardless of timezone, be less than the desired date
+  private static boolean shouldFallbackToCpu(ColumnVector input, ZoneId zoneId){
+    if (zoneId.getRules().isFixedOffset()){
+      return false;
     }
+    boolean isValid = false;
+    long scaleFactor = getScaleFactor(input);
+    try (Scalar targetTimestamp = Scalar.timestampFromLong(input.getType(), maxTimestamp*scaleFactor);
+         ColumnVector compareCv = input.binaryOp(BinaryOp.GREATER, targetTimestamp, DType.BOOL8);
+         Scalar isGreater = compareCv.any() ) {
```
res-life, line 183:
```suggestion
         Scalar isGreater = compareCv.any()) {
```

Please format using spark-rapids/docs/dev/idea-code-style-settings.xml

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -247,23 +361,20 @@ private static synchronized void loadData() {
...
           masterTransitions.toArray(new List[0]));
     } catch (Exception e) {
       throw new IllegalStateException("load time zone DB cache failed!", e);
     }
   }
 
   private static synchronized Table getTransitions() {
-    try (ColumnVector fixedTransitions = getFixedTransitions()) {
+    verifyDatabaseCached();
+    try (ColumnVector fixedTransitions = transitions.copyToDevice();) {
```
res-life, line 373:
```suggestion
    try (ColumnVector fixedTransitions = transitions.copyToDevice()) {
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: For every if/else, please add brackets

res-life:
Please refer to this [link](https://github.com/NVIDIA/spark-rapids/pull/12502#issuecomment-2811986136)

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -109,23 +132,118 @@ private static synchronized void closeResources()  {
...
+   * This function will not be called after spark-rapids changes
+   */
+  public static boolean isSupportedTimeZone(ZoneId desiredTimeZone) {
+    return desiredTimeZone != null &&
+      (desiredTimeZone.getRules().isFixedOffset() ||
+      desiredTimeZone.getRules().getTransitionRules().isEmpty());
+  }
+
+  // enforce that all timestamps, regardless of timezone, be less than the desired date
+  private static boolean shouldFallbackToCpu(ColumnVector input, ZoneId zoneId){
```
res-life, line 175: For the non Daylight Saving timezone(DST) : tz.getRules().getTransitionRules().isEmpty(), if there are timestamp > fall_back_limitation, it will use CPU. But original behavior is to compute on GPU.
Although it's rare case, but it does changed the original behavior.
And the extra out-of-range checking will slow down the speed for non-DST tzs.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -55,18 +60,24 @@ public class GpuTimeZoneDB {
...
    * If `shutdown` was called ever, then will not load the cache
    */
-  public static void cacheDatabaseAsync() {
+  public static void cacheDatabaseAsync(int maxYear) {
     // start a new thread to load
     Runnable runnable = () -> {
       try {
-        cacheDatabaseImpl();
+        cacheDatabaseImpl(maxYear);
+        maxTimestamp = LocalDateTime.of(maxYear+1, 1, 1, 0, 0, 0)
```
res-life, line 79:
```suggestion
        maxTimestamp = LocalDateTime.of(maxYear-1, 1, 1, 0, 0, 0)
```

Here we use maxYear + 1, but for the transition items, we check:
```scala
for (int year = initialTransitionYear; year <= finalTransitionYear; year++)
```
The initialTransitionYear is maxYear.
In `shouldFallbackToCpu` we actually use  maxYear + 1.
We should use a greater value for transition items, right?

res-life:
For the casting string with timezone to timestamp function, e.g.:

```
>>> spark.sql("select cast('2000-01-05 01:02:03 America/Los_Angeles' as timestamp)").show()
+----------------------------------------------------------+
|CAST(2000-01-05 01:02:03 America/Los_Angeles AS TIMESTAMP)|
+----------------------------------------------------------+
|                                       2000-01-05 17:02:03|
+----------------------------------------------------------+
```

For a string column, the time zones in each line could be different, so we should considier add functions like:

```scala
// the zones maybe the indices of cached GPU time zones.
ColumnVector fromTimestampToUtcTimestamp(ColumnVector input, ColumnVector zones)
```

jihoonson:
> For a string column, the time zones in each line could be different, so we should considier add functions like:
> 
> ```scala
> // the zones maybe the indices of cached GPU time zones, or just strings
> ColumnVector fromTimestampToUtcTimestamp(ColumnVector input, ColumnVector zones)
> ```

It is an excellent point @res-life. It seems like a good feature we could support after this change gets in. I'd suggest to file the issue on github and do it as a follow-up.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
warrickhe: Unfortunately I use VS code, is there an equivalent?

res-life:
Do we have performance test number compared to CPU? 
The functionality is Okay, it's better to conduct a performance test.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: There are two converts: `fromTimestampToUtcTimestamp` and `fromUtcTimestampToTimestamp`, and there are two kinds of timezones: plus and nagative, e.g.: +10:00, -10:00. If the `input`  represents a local time and time zone is -10:00, then its UTC timestamp value is original value + 10 hours which will be out of range. Using a greater value for transition items can make sure there is no problem. I did not test, but I guess there might be a problem. Would you please test the corner cases.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: When it's non-DST and there is value > 2200 year, now it runs on CPU.
But before it runs on GPU, do we need to chang this behavior?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
warrickhe: You mean in the case that the date time is 9999? I suppose I did not account for that edge case properly, I will set it to be maxYear+1,1,2 then, this should cover it.

warrickhe:
microbenchmark takes average 0.3823045721054077 seconds after 1000 retries, (max year 9999), gpu enabled
microbenchmark takes average 0.38564225149154663 seconds after 1000 retries , (max year 9999), gpu disabled
microbenchmark takes average 0.35852471661567686 seconds after 1000 retries , (max year 2199), gpu enabled
microbenchmark takes average 0.37146811437606814 seconds after 1000 retries, (max year 2199), gpu disabled
This is for America/Los_Angeles. Results don't look too promising (4% improvement only?). Perhaps my testing methodology was poor, or the kernel needs work

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Yes, this is done.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Cosidering the following:
```scala
maxYear = 2200
maxTimestamp = LocalDateTime.of(maxYear+1, 1, 1, 0, 0, 0)
for (int year = initialTransitionYear; year <= finalTransitionYear; year++)
shouldFallbackToCpu(cv)
```

When calling `shouldFallbackToCpu`, the data in `cv` may contain a equivalent timestamp: `2201-01-01 08:00:00`  cosidering of time zone offset, and its literal value is < `2201-01-01`,  but the max transition offset rule is only up to `2201-01-01 00:00:00`

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -55,18 +60,24 @@ public class GpuTimeZoneDB {
...
    * If `shutdown` was called ever, then will not load the cache
    */
-  public static void cacheDatabaseAsync() {
+  public static void cacheDatabaseAsync(int maxYear) {
     // start a new thread to load
     Runnable runnable = () -> {
       try {
-        cacheDatabaseImpl();
+        cacheDatabaseImpl(maxYear);
+        maxTimestamp = LocalDateTime.of(maxYear+1, 1, 2, 0, 0, 0)
```
res-life, line 79:
```suggestion
        maxTimestamp = LocalDateTime.of(maxYear, 12, 30, 0, 0, 0)
```
Considering the timezone offset is max 18 hours, use a less value 12-30 instead of 12-31 to make sure GPU has the transitions for the max timestamp.

res-life:
LGTM, except the last comment.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
warrickhe: I'd prefer we support the whole maxYear, so it's easier for the user to understand that any timestamp that has the year maxYear or less is supported. An analysis of all timezones an timezone rules show that there is transition rule happening in the first week of January, except for some fixed ones happening before 2100.

jihoonson (approved):

jihoonson:
build

res-life (approved):

res-life:
build

res-life:
Checked with @warrickhe , I'll merge it because of Warrick does not have privilege.

PR #3074, opened 2025-03-30
Improve docker image to enable `gcc-toolset` by default

ttnghia:
This modify the docker image to enable `gcc-toolset` by default for non-interactive containers. By doing so, we can invoke the container's `cmake ` with the correct version of `gcc`:
```
❯ docker run spark-rapids-jni-build:11.8.0-devel-rockylinux8 bash -c "cmake --version && g++ --version"
==========
== CUDA ==
==========

CUDA Version 11.8.0
...
cmake version 3.28.6
...
g++ (GCC) 11.2.1 20220127 (Red Hat 11.2.1-9)
```

Previously, this call executes `cmake` with the default system's `gcc` version `8.5`.

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -23,17 +23,23 @@
 ARG CUDA_VERSION=11.8.0
 ARG OS_RELEASE=8
 ARG TARGETPLATFORM=linux/amd64
+
+# Software versions for development
+ARG TOOLSET_VERSION=11
+ARG CMAKE_VERSION=3.30.4
+ARG CCACHE_VERSION=4.6
+
 # multi-platform build with: docker buildx build --platform linux/arm64,linux/amd64 <ARGS> on either amd64 or arm64 host
 # check available official arm-based docker images at https://hub.docker.com/r/nvidia/cuda/tags (OS/ARCH)
 FROM --platform=$TARGETPLATFORM nvidia/cuda:$CUDA_VERSION-devel-rockylinux$OS_RELEASE
-ARG TOOLSET_VERSION=11
+
 ### Install basic requirements
 # pin urllib3<2.0 for https://github.com/psf/requests/issues/6432
 RUN dnf --enablerepo=powertools install -y scl-utils gcc-toolset-${TOOLSET_VERSION} python39 zlib-devel maven tar wget patch ninja-build git && \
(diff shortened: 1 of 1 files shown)


ttnghia:
build

ttnghia:
build

hyperbolic2346 (approved):

abellina (dismissed):
Looks fine. Just to confirm, then I assume this is going to be simplified in a follow up? https://github.com/NVIDIA/spark-rapids-jni/blob/branch-25.04/build/run-in-docker#L84

ttnghia:
> Looks fine. Just to confirm, then I assume this is going to be simplified in a follow up? https://github.com/NVIDIA/spark-rapids-jni/blob/branch-25.04/build/run-in-docker#L84

Indeed there are more changes need to be done to cleanup that in the follow up work: https://github.com/search?q=repo%3ANVIDIA%2Fspark-rapids-jni%20gcc-toolset&type=code

ttnghia:
build

hyperbolic2346 (approved):

abellina (approved):

revans2 (approved):
looks fine to me. Not sure if this is going to impact other CI jobs we have, but it looks to have minimal impact.

ci/Dockerfile
```diff
@@ -23,29 +23,42 @@
...
 ARG OS_RELEASE=8
 ARG TARGETPLATFORM=linux/amd64
+
 # multi-platform build with: docker buildx build --platform linux/arm64,linux/amd64 <ARGS> on either amd64 or arm64 host
 # check available official arm-based docker images at https://hub.docker.com/r/nvidia/cuda/tags (OS/ARCH)
 FROM --platform=$TARGETPLATFORM nvidia/cuda:$CUDA_VERSION-devel-rockylinux$OS_RELEASE
+
+### Software versions for development
 ARG TOOLSET_VERSION=11
+ARG CMAKE_VERSION=3.30.4
```
gerashegalov, line 33: nit: some comment about the version choice sine I recalls there is always some caveat why a particular version is used

gerashegalov (approved):

ci/Dockerfile
ttnghia: Will do that in the follow up work.

PR #3120, opened 2025-04-10
Fix path to `rapids_config.cmake`

ttnghia:
As cudf has move its `rapids_config.cmake`, we need to update the reference path to it.


ttnghia:
build

jihoonson (approved):

PR #3126, opened 2025-04-14
Fixed a discrepancy between CPU and GPU kud0 validity data.

nvdbaranec:
The GPU kud0 path was incorrectly setting the `has_validity` flag for individual partitions that contain no rows, when the overall column does contain rows and validity.  This can happen when splitting a list column and ending up with an empty partition on the inner column.  For example:

```
rows      {2, 2, 2},  {3, 3, 3}, {4, 4, 4}, {}
validity  {1, 1, 1},  {1, 1, 1}, {1, 1, 1}, {}
```

The inner int column contains rows and validity. The resulting column when reassumbled will contain validity.  However, if we split it at row 3, the last partition will contain no rows.  The code was using the `has_validity` bit for the entire column instead of computing it per-partition for each column, resulting in us setting the flag for partition 3 and then just not adding any validity data.


hyperbolic2346 (approved):

nvdbaranec:
build

PR #3133, opened 2025-04-15
Fix bugs in CMake build

ttnghia:
This fixes 3 bugs in CMake build system:
 * Fix double fetching `cccl` while configuring CMake. `cccl` is already installed by `cudf` so we should not fetch it again then include both into the `spark-rapids-jni` target.
 * Fix CMake issues with wrong path when finding package `flatbuffers`.
 * Fix missing CMake function definition due to not including the corresponding source.


gerashegalov:
This looks disruptive. GPU_ARCHS is ingrained in my memory and maybe scripts. I would prefer to keep it. I don't mind if we can also set via CMAKE_CUDA_ARCHITECTURES with higher precedence as well.

This PR should focus on cleanup and  parameter rename deserves a dedicated issue and PR

ttnghia:
> This PR should focus on cleanup and parameter rename deserves a dedicated issue and PR

That's fine. Now this PR has only cleanup. I'll push another PR separately.

ttnghia:
build

ttnghia:
build

src/main/cpp/cmake/get_flatbuffers.cmake
```diff
@@ -24,6 +24,7 @@ function(find_and_configure_flatbuffers VERSION)
     GIT_SHALLOW TRUE
```
hyperbolic2346, line 24: Need a copyright update in this file.

gerashegalov: cc @YanxuanLiu @pxLi to cover cmake files in the license header PR check.

gerashegalov (approved):

src/main/cpp/cmake/get_flatbuffers.cmake
ttnghia: How could I forget this....

ttnghia:
build

gerashegalov (approved):

PR #3139, opened 2025-04-16
Incorporate C++ stacktrace into Java stacktrace

ttnghia:
Till now, whenever there is an exception being thrown from `cudf`, we only see the error message of that exception in Java/Spark event log. There is nothing else. It is very difficult to debug such exception without having its stack trace.

This PR, with the support of its dependency (https://github.com/rapidsai/cudf/pull/18512), allows to generate a C++ stack frame of the being thrown exception and concatenates that into the error message of the corresponding Java exception. By doing so, it is much easier to debug Spark jobs.

Depends on:
 * https://github.com/rapidsai/cudf/pull/18512
 * https://github.com/NVIDIA/spark-rapids-jni/pull/3133


ttnghia:
Close this as the new implementation is totally different from this prototype.

PR #3154, opened 2025-04-21
Implement `CentralMomentHostUDF` aggregations

ttnghia:
This reimplements the cudf's groupby aggregations `M2` and `MERGE_M2`, modifying the original code such that the type of `count` quantity is changed from `size_type` to `double`. Such change is to fix the overflow issue reported in https://github.com/NVIDIA/spark-rapids/issues/12468. In addition to that, the newly introduced aggregations `CentralMomentHostUDF` in this PR also paves the way for further optimizing the implementation of `stddev` and `variance` SQL functions, which will be implemented in the follow up work.


ttnghia:
This PR also modifies/renames some existing files, just moving function declarations around.

ttnghia:
Close as reconsider to implement this in cudf. Will go back to this approach if it is not possible.

Issue #1721, opened 2024-01-24
[BUG] `GpuTimeZoneDB` missed some time zones which are not normalized.


res-life:
**Describe the bug**
GpuTimeZoneDB only index normalized time zones, but the original time zone names are ignored, user can use original time zone.

**Steps/Code to reproduce bug**
```java
import java.util._
import java.time._
val ids = TimeZone.getAvailableIDs.map(e => ZoneId.of(e, ZoneId.SHORT_IDS).normalized.getId)
val ids2 = TimeZone.getAvailableIDs

for (id <- ids2) {
    if (!ids.contains(id)) {
        println("In normalized, but missed in general: " + id)
    }
}

Spark:
>>> spark.sql("select cast('1970-01-01 00:00:00 Etc/GMT' as timestamp)").show()
+----------------------------------------------+
|CAST(1970-01-01 00:00:00 Etc/GMT AS TIMESTAMP)|
+----------------------------------------------+
|                           1970-01-01 08:00:00|
+----------------------------------------------+

```

```
In normalized, but missed in general: Etc/GMT
In normalized, but missed in general: Etc/GMT+0
In normalized, but missed in general: Etc/GMT+1

```

res-life:
I'm not planing work on this for release 24.04
@sameerz can we move it to next release?

PR #3160, opened 2025-04-23
Add utils for cast string to timestamp

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/10035
closes https://github.com/NVIDIA/spark-rapids-jni/issues/1721

### Description
Casting string to timestamp requires more info in `GpuTimeZoneDB`:

1. If a timezone is DST

  Requires checking timezones in timestamp strings are all non-DST
  if has any timestamp is greater than the threshold e.g.: 2200,  it means do not support for all timezones.  We then should check if there is any non-DST timezone, if has: 1. fallback to cpu. 2. throw exception not support. 

2. Require non-normalized timezone names
Currently GpuTimeZoneDB only records normalized timezone names.
But timestamp strings can contains non-normalized timezone names.

E.g.:
  ZoneId.of("Etc/GMT").normalized.getId = Z;
  ZoneId.of("Etc/GMT+0").normalized.getId = Z
Both '2025-01-01 00:00:00 Etc/GMT' and '2025-01-01 00:00:00 Z' are valid timestamp strings.
`zoneIdToTable` should have entries:
Z -> index_of_Z
Etc/GMT -> index_of_Z
Etc/GMT+0 -> index_of_Z
Currently only has the first entry.

3. `zoneIdToTable` should also record short IDs.

ZoneId.SHORT_IDS

4. Make sure all the timezones are loaded.

### Changes
1.  Add timeZoneInfo which can query DST and transition index
```scala
  // host column STRUCT<tz_name: string, index_to_transition_table: int, is_DST: int8>,
  // sorted by time zone, is used to query index to transition table and if tz is DST
  private static HostColumnVector timeZoneInfo;
```

2. Let `zoneIdToTable` record non-normalized timezones

3. Record shoort IDs

```scala
zoneId = ZoneId.of(tz).normalized()
```
=>
```scala
zoneId = ZoneId.of(nonNormalizedTz, ZoneId.SHORT_IDS).normalized()
```

4. Make sure all the timezones are loaded.
Casting string with timezone to timestamp needs loading all timezone is successful. All valid time zones may be included in the timestamp strings.
If it has a timezone unloaded, then kernel can not decide if the missed timezone is valid.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
@@ -32,12 +32,7 @@ import java.time.zone.ZoneOffsetTransition;
 import java.time.zone.ZoneOffsetTransitionRule;
 import java.time.zone.ZoneRules;
 import java.time.zone.ZoneRulesException;
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TimeZone;
+import java.util.*;
 import java.util.concurrent.Executors;
 
 /**
@@ -59,12 +54,24 @@ public class GpuTimeZoneDB {
   //   LIST<STRUCT<utcInstant: int64, localInstant: int64, offset: int32>>
   private static Map<String, Integer> zoneIdToTable;
 
+  // host column STRUCT<tz_name: string, index_to_transition_table: int, is_DST: int8>,
+  // sorted by time zone, is used to query index to transition table and if tz is DST
+  private static HostColumnVector timeZoneInfo;
+
   // use this reference to indicate if time zone cache is initialized.
   private static HostColumnVector transitions;
   // initial year set to 1900 because some transition rules start early
   private static final int initialTransitionYear = 1900;
   private static long maxTimestamp;
   private static final ZoneId utcZoneId = ZoneId.of("UTC");
+
+  // Casting string with timezone to timestamp needs loading all timezone is successful.
+  // All valid time zones may be included in the timestamp strings.
+  // Kernel behavior: if timezone is invalid and ansi is off, then return null.
+  // If it has a timezone unloaded, then kernel can not decide if the missed timezone is valid.
+  // So `zoneIdToTable` and `timeZoneInfo` MUST contain all the valid timezones, except fixed
+  // timezone like: +08:00
+  private static boolean loadError = false;
 
   /**
    * This should be called on startup of an executor.
@@ -89,7 +96,9 @@ public class GpuTimeZoneDB {
   }
 
   public static synchronized void verifyDatabaseCached() {
-    if (transitions == null) throw new IllegalStateException("Time Zone DB not loaded!");
+    if (transitions == null || loadError) {
+      throw new IllegalStateException("Time Zone DB not loaded!");
+    }
   }
 
   /**
@@ -120,7 +129,9 @@ public class GpuTimeZoneDB {
     if (transitions == null) {
       try {
         loadData(maxYear);
+        loadError = false;
       } catch (Exception e) {
+        loadError = true;
         closeResources();
         throw e;
       }
@@ -135,6 +146,10 @@ public class GpuTimeZoneDB {
     if (transitions != null) {
       transitions.close();
       transitions = null;
+    }
+    if (timeZoneInfo != null) {
+      timeZoneInfo.close();
+      timeZoneInfo = null;
     }
   }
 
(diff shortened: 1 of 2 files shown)


res-life:
build

res-life:
build

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -89,7 +96,9 @@ public static void cacheDatabaseAsync(int maxYear) {
   }
 
   public static synchronized void verifyDatabaseCached() {
-    if (transitions == null) throw new IllegalStateException("Time Zone DB not loaded!");
+    if (transitions == null || loadError) {
+      throw new IllegalStateException("Time Zone DB not loaded!");
+    }
```
jihoonson, line 101: nit: I'm not sure why we need another variable to track the load state. In any case, do we want to allow a state where `transitions` is not null but `loadError` is true? If not, we can just set `transitions` to null when an error occurs.

jihoonson (approved):
Looks good. Thanks @res-life. I left only one minor question.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
warrickhe: I'm guessing it's to cover a case where there is an error during loading and transitions has some stuff added but not all the data. Not sure when exactly that would happen, but it would cover that.

warrickhe (approved):
LGTM. Same comment as Jihoon.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -284,20 +299,42 @@ public static ZoneId getZoneId(String timeZoneId) {
...
+      //
+
+      // get and sort time zones
+      String[] timeZones = TimeZone.getAvailableIDs();
+      List<String> sortedTimeZones = new ArrayList<>(Arrays.asList(timeZones));
+      // Note: Z is a special normalized time zone from UTC: ZoneId.of("UTC").normalized = Z
+      // TimeZone.getAvailableIDs does not contain Z
+      // Should add Z to `zoneIdToTable`
+      sortedTimeZones.add("Z");
+      Collections.sort(sortedTimeZones);
```
mythrocks, line 327: I haven't completely grasped why `sortedTimeZones` needs to be sorted.  

Would there be value in deduping `sortedTimeZones` _a priori_?  Something like:

```java
timeZones = Stream.concat(
              Arrays.stream(TimeZone.getAvailableIDs()),
              Stream.of("Z")) // UTC
            .map( x -> ZoneId.of(x, ...).normalized().getId())
            .distinct();
```

mythrocks (dismissed):
LGTM. 👍  Just a couple of nitpicks and clarifications.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: `timeZoneInfo` records all available timezones and map info:
```
   tz => index_to_transition_table
   tz => is DST
```
The coming kernel for cast string to timestamp will use binary search to look up the above info.
So we sort all the `TimeZone.getAvailableIDs` timezones.
 
`timeZoneInfo`:
 ```scala
  // host column STRUCT<tz_name: string, index_to_transition_table: int, is_DST: int8>,
  // sorted by time zone, is used to query index to transition table and if tz is DST
  private static HostColumnVector timeZoneInfo;
 ```

We should not dedup sortedTimeZones in prior as you posted above.
Both normalized and non-normalized timezones can be in the timestamp strings.
E.g.: 
Normalization:
```
 ACT => Australia/Darwin
 AET => Australia/Sydney
```
timestamp strings:
```
2025-01-01 00:00:00 Australia/Darwin
2025-01-01 00:00:00 Australia/Sydney
2025-01-01 00:00:00 ACT
2025-01-01 00:00:00 AET
```
The dedup will remove `ACT` and `AET`.
The ideal solution is that only do dedup in `transitions` to save memory because of `ACT` and `Australia/Darwin` have the same `transition`
```scala
private static HostColumnVector transitions;
```
In the `Map<String, Integer> zoneIdToTable`,   should have `ACT` and `AET`, or we do not know ACT is valid timezone.
Also in `timeZoneInfo` , should also have  `ACT` and `AET`.

Before this PR,  `ACT` and `AET` are not recorded.
This PR is to record all the non-normalized timezones like `ACT` and `AET`.

BTW:
TimeZone.getAvailableIDs contains all the normalized and non-normalized timezones.

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Fixed, please see the commit.
Removed the loadError, now check the `timeZoneInfo == null` instead of `transitions == null`, because `timeZoneInfo` is the last variable to construct in `loadData`.

mythrocks (approved):

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
mythrocks: Thank you for the explanation.

PR #3171, opened 2025-04-25
Cleanup `gcc-toolset` in docker build

ttnghia:
This removes the calls to `gcc-toolset-11` from several places in build scripts, as it is no longer needed. The toolset should already be enabled by default.

This cleanup will make future upgrading for the toolset straightforward.


ttnghia:
build

ttnghia:
build

ttnghia:
build

build/run-in-docker
```diff
@@ -81,4 +81,4 @@ $DOCKER_CMD run $DOCKER_GPU_OPTS $DOCKER_RUN_EXTRA_ARGS -u $(id -u):$(id -g) --r
   -e VERBOSE \
   $DOCKER_OPTS \
   $SPARK_IMAGE_NAME \
-  scl enable gcc-toolset-11 "$RUN_CMD"
+  bash -c "$RUN_CMD"
```
abellina, line 84: could we add a comment that `bash` is really `scl enable gcc-toolset-11 bash`? just for the future, and so I don't have to go to the Dockerfile to realize that `bash` isn't just `bash`.

ttnghia:
build

abellina (approved):

Issue #1654, opened 2023-12-15
[FEA] Support parsing dates and timestamps with timezone IDs in the string


res-life:
**Is your feature request related to a problem? Please describe.**
This is from https://github.com/NVIDIA/spark-rapids/issues/6846
Needs a kernel to parse dates and timestamps with timezone IDs in the string

- [x] https://github.com/NVIDIA/spark-rapids-jni/issues/1721
- [x] Parse to timestamp
- [x] Parse Just time https://github.com/NVIDIA/spark-rapids/issues/10221
- [x] Parse to date
- [x] Check boundary/overflow of long value.   Spark saves ts as long.
Spark code uses `Math.addExact` to check overflow:
```
  def instantToMicros(instant: Instant): Long = {
    val secs = instant.getEpochSecond
    if (secs == MIN_SECONDS) {
      val us = Math.multiplyExact(secs + 1, MICROS_PER_SECOND)
      Math.addExact(us, NANOSECONDS.toMicros(instant.getNano) - MICROS_PER_SECOND)
    } else {
      val us = Math.multiplyExact(secs, MICROS_PER_SECOND)
      Math.addExact(us, NANOSECONDS.toMicros(instant.getNano))
    }
  }
```

res-life:
Alfred is co-working on this.

res-life:
I'm not planing work on this for release 24.04
@sameerz can we move it to next release?

res-life:
PRs:

https://github.com/NVIDIA/spark-rapids-jni/pull/3174
https://github.com/res-life/spark-rapids-jni/pull/3

res-life:
- [x] casting string to timestamp part: https://github.com/NVIDIA/spark-rapids-jni/pull/3174
- [ ]  casting string to date part

res-life:
For casting string to date, it is not related to timezone.
Will update the title and close this issue.

res-life:
- [ ] https://github.com/NVIDIA/spark-rapids-jni/issues/3292

PR #3174, opened 2025-04-26
Support casting string with timezone to timestamp

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/10035
contributes to https://github.com/NVIDIA/spark-rapids-jni/issues/1654
contributes to https://github.com/NVIDIA/spark-rapids/issues/6846
closes https://github.com/NVIDIA/spark-rapids/issues/10221

This is the part one  of the whole feature.

### changes
Parse casting string with timezone to intermediate result.
- [x] https://github.com/NVIDIA/spark-rapids-jni/pull/3160,  prepare all timezone info.
- [x] parse string
- [x] query timezone info, get timezone index, get if timezone is DST and generate intermediate result.
Intermediate result is a struct column with 7 sub-columns:
   * Parse Result type: 0 Success, 1 invalid e.g. year is 7 digits 1234567
   * seconds part of parsed UTC timestamp
   * microseconds part of parsed UTC timestamp
   * Timezone type: 0 unspecified, 1 fixed type, 2 other type, 3 invalid
   * Timezone offset for fixed type, only applies to fixed type
   * Timezone is DST, only applies to other type
   * Timezone index to `GpuTimeZoneDB.transitions` table

#### Part 2 
- [x] If has timestamp > 2200 and it has any DST timezone in strings, it means kernel can not handle, fallback to CPU. Refer to https://github.com/NVIDIA/spark-rapids-jni/pull/3072, this PR convert DST timezone to non-DST by computing the transitions < 2200 year in advance, and thus has limitation.
- [x] Kernel:  convert intermediate result to final timestamp column. For each row each timezone, find the transition according to index, convert intermediate timestamp column to final.
- [x] If ansi and has invalid, return null.

### Performance
Generate file:
```scala
import org.apache.spark.sql.functions
val path = "/home/chongg/c.parquet"
spark.range(10000000).repartition(17).withColumn("m", functions.expr("CASE WHEN id % 4 == 0 THEN '2025-01-05' WHEN id % 4 == 1 THEN '2025-01' WHEN id % 4 = 2 THEN '2025' ELSE '2025-01-05 00:01:02' END")).drop("id").write.mode("overwrite").parquet(path)
```

Perf code:
```scala
$SPARK_HOME/bin/spark-shell --master local[*] --jars ${PLUGIN_JAR} --conf spark.plugins=com.nvidia.spark.SQLPlugin --conf spark.rapids.sql.explain=ALL --driver-java-options '-ea -Duser.timezone=UTC '

spark.conf.set("spark.rapids.sql.hasExtendedYearValues", false)
spark.conf.set("spark.rapids.sql.castStringToTimestamp.enabled", true)
val path = "/home/chongg/c.parquet"
spark.read.parquet(path).createOrReplaceTempView("tab")
spark.time(spark.sql("select sum(cast(cast(m as timestamp) as long)) from tab").collect())
```

|lecacy GPU (hot runs ms)|new GPU kernel (hot runs ms) | speedup |
|:-|:-|:-|
|318 276 272|231 207 185|0.39x|


res-life:
build

res-life:
build

src/main/cpp/src/utils.hpp
```diff
@@ -0,0 +1,109 @@
...
+   * std::chrono::year range is [-32,767 , 32,767], here can not use std::chrono::year_month_day
+   */
+  __device__ int32_t to_epoch_day() const
+  {
+    int32_t y          = year - (month <= 2);
+    const int32_t era  = (y >= 0 ? y : y - 399) / 400;
+    const uint32_t yoe = static_cast<uint32_t>(y - era * 400);                           // [0, 399]
+    const uint32_t doy = (153 * (month > 2 ? month - 3 : month + 9) + 2) / 5 + day - 1;  // [0, 365]
+    const uint32_t doe = yoe * 365 + yoe / 4 - yoe / 100 + doy;  // [0, 146096]
+    return era * 146097 + doe - 719468;
```
ttnghia, line 51: Did you check out the `cuda::std::chrono` class? It has a lot of implementation for date/time operations.

res-life: I checked the Java implementation and the above impl,  they are equivalent, I also tested the all the days in  [-300000, 300000]  years, they have the identical results. And this is more concise compared to Java.

Note: [-300000, 300000]  years are the approximately range Spark can handle, and it's large than Spark year range.

Java impl, LocalDate.toEpochDay
```java
  public long toEpochDay() {
        long y = year;
        long m = month;
        long total = 0;
        total += 365 * y;
        if (y >= 0) {
            total += (y + 3) / 4 - (y + 99) / 100 + (y + 399) / 400;
        } else {
            total -= y / -4 - y / -100 + y / -400;
        }
        total += ((367 * m - 362) / 12);
        total += day - 1;
        if (m > 2) {
            total--;
            if (isLeapYear() == false) {
                total--;
            }
        }
        return total - DAYS_0000_TO_1970;
    }
```

ttnghia: Yeah I digged into it today and found this:
```
_LIBCUDACXX_HIDE_FROM_ABI constexpr days year_month_day::__to_days() const noexcept
{
  static_assert(std::numeric_limits<unsigned>::digits >= 18, "");
  static_assert(std::numeric_limits<int>::digits >= 20, "");

  // nvcc doesn't allow ODR using constexpr globals. Therefore,
  // make a temporary initialized from the global
  auto constexpr __Feb = February;
  const int __yr       = static_cast<int>(__y) - (__m <= __Feb);
  const unsigned __mth = static_cast<unsigned>(__m);
  const unsigned __dy  = static_cast<unsigned>(__d);

  const int __era      = (__yr >= 0 ? __yr : __yr - 399) / 400;
  const unsigned __yoe = static_cast<unsigned>(__yr - __era * 400); // [0, 399]
  const unsigned __doy =
    static_cast<unsigned>((153 * (__mth + static_cast<unsigned>(__mth > 2 ? -3 : 9)) + 2) / 5 + __dy - 1); // [0, 365]
  const unsigned __doe = __yoe * 365 + __yoe / 4 - __yoe / 100 + __doy; // [0, 146096]
  return days{__era * 146097 + static_cast<int>(__doe) - 719468};
}
```

So please try to reuse the implementation from `cuda::std::chrono` as much as possible.

src/main/cpp/src/utils.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include <cudf/types.hpp>
+
+namespace spark_rapids_jni {
+
+/**
+ * Represents timestamp with microsecond accuracy.
+ * Max year is six-digits, approximately [-300000, 300000]
+ */
+struct ts_segments {
```
ttnghia, line 27: We would probably be able to remove this completely and just use the classes from `std::cuda::chrono`.

src/main/cpp/src/utils.hpp
res-life: Thanks, @ttnghia ,  the `__to_days` is equivalent to the current impl and it has no overflow for six digit years.
But it's a private mothod `__to_days`, I do not know how to call intto this function.

I tried to use `cuda::std::chrono`, but it gets wrong days for years > 32767.
```cpp
# include <cuda/std/chrono>
# include <cstdio>

__global__ void t() {
    auto const ymd =
        cuda::std::chrono::year_month_day(cuda::std::chrono::year{32768},
                                        cuda::std::chrono::month{static_cast<uint32_t>(1)},
                                        cuda::std::chrono::day{static_cast<uint32_t>(1)});
    auto const days = cuda::std::chrono::sys_days(ymd).time_since_epoch().count();
    printf("days %u\n", days);
}

int main() {
  t<<<1,1>>>(); 
  cudaDeviceSynchronize();
  return 0;
}
```
days 4282279502

Scala:
```scala
scala> import java.time.LocalDate
import java.time.LocalDate

scala> LocalDate.of(32768,1,1).toEpochDay
val res0: Long = 11248738
```
The result 4282279502 from cuda::std::chrono is wrong, while Scala gets 11248738.

Is there a way to replace `cuda::std::chrono::sys_days(ymd).time_since_epoch().count()` and get days without overflow?

ttnghia: `__to_days()` should be public:
```
public:
....
  _LIBCUDACXX_HIDE_FROM_ABI static constexpr year_month_day __from_days(days __d) noexcept;
  _LIBCUDACXX_HIDE_FROM_ABI constexpr days __to_days() const noexcept;
};
```

ttnghia: In addition, I don't remember exactly but probably year/month in chrono are int16_t which should not operate on values bigger than `32768`. We would never need such year.

res-life: Can not get correct result using `year_month_day.__to_days()`. The root cause is that the `year` in `year_month_day` is short type. Passing 32768 to short will cause overflow.

```cpp
class year
{
private:
  short __y;
// ......
_LIBCUDACXX_HIDE_FROM_ABI explicit constexpr year(int __val) noexcept
      : __y(static_cast<short>(__val))  ///////// This cast to short causes overflow ////////// 
  {}

}
......
```

Spark does support years > 32768,  up to year `294247`, e.g.:

```
>>> spark.sql("select cast('294247-01-10 00:00:00' as timestamp)").show(truncate = False)
+----------------------------------------+
|CAST(294247-01-10 00:00:00 AS TIMESTAMP)|
+----------------------------------------+
|+294247-01-10 00:00:00                  |
+----------------------------------------+
```

src/main/cpp/src/utils.hpp
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Is the segments of the timestamp valid?
+   */
+  __device__ bool is_valid_ts() const { return is_valid_date() && is_valid_time(); }
+
+  /**
+   * Get epoch day. Can handle years in the range [-1,000,000, 1,000,000].
+   * Refer to https://howardhinnant.github.io/date_algorithms.html#days_from_civil
+   * Spark year is approximately in range [-300,000, 300,000]
+   * std::chrono::year range is [-32,767 , 32,767], here can not use std::chrono::year_month_day
```
revans2, line 42: If this is true, does that mean any time we use a year outside of this range it is highly probably that std::chrono and hence CUDF will return the wrong result? We often limit our tests to years within a valid python range, which is smaller than both of these.

ttnghia: Yes that is true. If you look into cudf code such as `convert_datetime.cu`, `cuda::std::chrono::year` is used in many places and should have the same restriction.

ttnghia: Due to such issue, I don't think we should pursue supporting large year numbers as Spark does - large years may not be used in the real world. Otherwise, it would be not a simple task to fix all the related code.

res-life: For this casting string to timestamp PR, it supports large years. We can file a follow up issue to track other related code.

res-life:
build

res-life:
Note:  `to_epoch_day` is from https://howardhinnant.github.io/date_algorithms.html#days_from_civil.
I locally tested all the possible input against Java, and get correct result.

src/main/java/com/nvidia/spark/rapids/jni/BooleanUtils.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  /**
+   * Get the number of true values in a boolean/int8/uint8 column.
+   */
+  public static long trueCount(ColumnView input) {
+    return input.getRowCount() - falseCount(input);
+  }
+
+  private static native long falseCount(long handle) throws CudfException;
```
ttnghia, line 41: I don't think we need any of them in JNI. These can be computed on the fly in the plugin:
 * For `trueCount`, we can compute it as `sum(DType.INT32)`.
 * `falseCount = rowCount - trueCount()`.

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -270,4 +270,26 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromInteger
   }
   CATCH_CAST_EXCEPTION(env, 0);
 }
+
+JNIEXPORT jlong JNICALL
+Java_com_nvidia_spark_rapids_jni_CastStrings_parseTimestampStrings(JNIEnv* env,
```
ttnghia, line 275: How about `toTimestamp`? `CastStrings.toTimestamp()` sound better than `CastStrings.parseTimestampStrings`.

res-life: This JNI is parsing strings to intermediate result, a struct:
```java
  /**
   * Trims and parses strings to intermediate result.
   * This is the first phase of casting string with timezone to timestamp.
   * Intermediate result is a struct column with 7 sub-columns:
   * - Parse Result type: 0 Success, 1 invalid e.g. year is 7 digits 1234567
   * - seconds part of parsed UTC timestamp
   * - microseconds part of parsed UTC timestamp
   * - Timezone type: 0 unspecified, 1 fixed type, 2 other type, 3 invalid
   * - Timezone offset for fixed type, only applies to fixed type
   * - Timezone is DST, only applies to other type
   * - Timezone index to `GpuTimeZoneDB.transitions` table
   *
   *
   * @param input The input String column contains timestamp strings
   * @param defaultTimeZoneIndex The default timezone index to `GpuTimeZoneDB`
   *   transition table.
   * @param defaultEpochDay Default epoch day to use if just time, e.g.: epoch day of
   *   "2025-05-05", then "T00:00:00" is "2025-05-05T00:00:00Z"
   * @param timeZoneInfo Timezone info column:
   *   STRUCT<tz_name: string, index_to_transition_table: int, is_DST: int8>,
   *   Refer to `GpuTimeZoneDB` for more details.
   * @return a struct column constains 7 columns described above.
   */
  static ColumnVector parseTimestampStrings(
```

There are two phases:
- Parse to intermediate result
- Convert intermediate to final timestamp column

```java
private static ColumnVector convertToTimestampOnGpu
-> GpuTimeZoneDB.fromTimestampToUtcTimestampWithTzCv
```

res-life:
build

res-life:
build

res-life:
build

src/main/cpp/src/cast_string.hpp
```diff
@@ -138,4 +138,33 @@ std::unique_ptr<cudf::column> long_to_binary_string(
   rmm::cuda_stream_view stream,
   rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource());
 
+/**
+ * @brief Parse a timestamp string column into a intermediate struct column.
```
ttnghia, line 142: Typo.
```suggestion
 * @brief Parse a timestamp string column into an intermediate struct column.
```

src/main/cpp/src/cast_string.hpp
```diff
@@ -138,4 +138,33 @@ std::unique_ptr<cudf::column> long_to_binary_string(
...
+ * The output column is a struct column with 7 children:
+ * - Parse Result type: 0 Success, 1 invalid e.g. year is 7 digits 1234567
```
ttnghia, line 144: "result type" sounds misleading. "result code" seems better.
```suggestion
 * - Parse result code: 0 - success, 1 - invalid (e.g. year is 7 digits 1234567)
```

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+enum class TS_TYPE : uint8_t { NOT_JUST_TIME = 0, JUST_TIME = 1 };
+
+/**
+ * Represents local date time in a timezone.
+ * Spark stores timestamp into Long in microseconds.
+ * A Long is able to represent a timestamp within [+-]200 thousand years.
+ * So year is 4-6 digits, formula is: Long.MaxValue/microseconds_per_year.
+ */
+
+enum class TZ_TYPE : uint8_t {
```
ttnghia, line 65:
```suggestion
 */
enum class TZ_TYPE : uint8_t {
```

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+/**
+ * If the timestamp string only contains the time part, TS_TYPE is JUST_TIME.
+ * E.g.:
+ *   For timestamp string T18:01:01 : TS_TYPE = JUST_TIME
+ *   For timestamp string 2020-01-01T12:00:00 : TS_TYPE = NOT_JUST_TIME
+ */
+enum class TS_TYPE : uint8_t { NOT_JUST_TIME = 0, JUST_TIME = 1 };
+
+/**
+ * Represents local date time in a timezone.
```
ttnghia, line 59: Seems wrong comment? As this is an enum instead of a local datetime.

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+/**
+ * Is white space, consistent with Spark UTF8String.trimAll for char input
+ */
+__device__ bool is_whitespace(char const c)
+{
+  // Keep consistent with Java: Character.isWhitespace(c) ||
+  // Character.isISOControl(c),
+  // 0-31 is control characters, 32 is space, 127 is delete
+  return (c >= 0 && c <= 32) || c == 127;
+}
```
ttnghia, line 139: This will lead to the same error as in https://github.com/NVIDIA/spark-rapids-jni/issues/3227.

Maybe copy this? https://github.com/NVIDIA/spark-rapids-jni/pull/3250/files#diff-3f32312ca5b95c20e7c0aeee7fd0d6271e17474fc013e97ffcd057e6f410918bR48-R56

However, why do we care about code 127 here?

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+ * @param max_digits the maximum digits to parse, if exceeds, return false
+ * @return true if the input string is valid, false otherwise
+ */
+__device__ bool parse_int(
+  char const* const ptr, int& pos, int const end_pos, int& v, int min_digits, int max_digits)
+{
+  v          = 0;
+  int digits = 0;
+  while (pos < end_pos) {
+    int const parsed_value = static_cast<int32_t>(ptr[pos] - '0');
```
ttnghia, line 159: `char` type is not guaranteed to be signed. It can be unsigned thus we must cast to `int` before doing subtraction.
```suggestion
    int const parsed_value = static_cast<int32_t>(ptr[pos]) - '0';
```

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+ * @param[out] pos the pointer to the string when parsing
+ * @param end_pos the end pointer of the string
+ * @param[out] v the parsed value
+ */
+__device__ void parse_microseconds(char const* const ptr, int& pos, int const end_pos, int& v)
+{
+  v          = 0;
+  int digits = 0;
+  while (pos < end_pos) {
+    int const parsed_value = static_cast<int32_t>(ptr[pos] - '0');
```
ttnghia, line 187: Similar for this.
```suggestion
    int const parsed_value = static_cast<int32_t>(ptr[pos]) - '0';
```

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+    if (parsed_value >= 0 && parsed_value <= 9) {
+      if (++digits <= 6) {
+        v = v * 10 + parsed_value;
+      } else {
+        // ignore/truncate the rest digits
```
ttnghia, line 192: `break;`?

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+ * @param max_digits the maximum digits to parse
+ * @returns the number of digits parsed.
+ */
+__device__ int parse_digits(
+  char const* const ptr, int& pos, int const end_pos, int& v, int const max_digits)
+{
+  v          = 0;
+  int digits = 0;
+  while (pos < end_pos) {
+    int const parsed_value = static_cast<int32_t>(ptr[pos] - '0');
```
ttnghia, line 230:
```suggestion
    int const parsed_value = static_cast<int32_t>(ptr[pos]) - '0';
```

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+    if (parsed_value >= 0 && parsed_value <= 9) {
+      v = v * 10 + parsed_value;
+      ++pos;
+      if (++digits == max_digits) { break; }
+    } else {
+      // meets non-digit
```
ttnghia, line 236: So there is no error report needed for parsing invalid digits?

src/main/cpp/src/cast_string_to_timestamp.cu
```diff
@@ -0,0 +1,876 @@
...
+
+  // parsed timezone info
+  uint8_t* tz_types;
+  int32_t* tz_fixed_offsets;
+  uint8_t* is_DSTs;
+  int32_t* tz_indices;
+
+  __device__ void operator()(cudf::size_type const idx) const
+  {
+    auto const str         = d_strings.element<cudf::string_view>(idx);
```
ttnghia, line 723: Is there any null check needed for this function?

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +121,144 @@ auto convert_timestamp_tz(column_view const& input,
...
+      output[row_idx]      = cudf::timestamp_us{cudf::duration_us{0L}};
+      output_mask[row_idx] = 0;
+      return;
+    }
+
+    int64_t epoch_seconds      = input_seconds[row_idx];
+    int64_t epoch_microseconds = static_cast<int64_t>(input_microseconds[row_idx]);
+
+    // 2. fixed offset conversion
+    if (tz_type[row_idx] == /*Fixed TZ*/ 1) {
```
ttnghia, line 159: Better to compare the exact values:
```suggestion
    if (static_cast<...>(tz_type[row_idx]) == FIXED_TZ) {
```

src/main/cpp/src/timezones.cu
```diff
@@ -118,6 +121,144 @@ auto convert_timestamp_tz(column_view const& input,
...
+                                               tz_indices.begin<int32_t>(),
+                                               result->mutable_view().begin<cudf::timestamp_us>(),
+                                               null_mask->mutable_view().begin<uint8_t>()});
+
+  auto [output_bitmask, null_count] = cudf::detail::valid_if(null_mask->view().begin<uint8_t>(),
+                                                             null_mask->view().end<uint8_t>(),
+                                                             cuda::std::identity{},
+                                                             stream,
+                                                             mr);
+  result->set_null_mask(std::move(output_bitmask), null_count);
```
ttnghia, line 258: Typically we should avoid keeping the null mask if there is no nulls. Some downstream functions may take extra time to process that null mask without doing any meaningful work.
```suggestion
  if (null_count) { result->set_null_mask(std::move(output_bitmask), null_count); }
```

src/main/cpp/src/utils.hpp
```diff
@@ -0,0 +1,155 @@
+/*
```
ttnghia, line 1: Oh what `utils` in the file name? The code seems date/time related. We already have `datetime_utils.hpp` for the function declaration, but here we have device code thus it should be something like `datetime_utils.cuh`.

src/main/cpp/src/utils.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+struct size_type_iterator {
+  cudf::size_type const* values;
+  __device__ cudf::size_type operator()(cudf::size_type const& idx) const { return values[idx]; }
+};
+
+struct const_size_type_iterator {
+  cudf::size_type const value;
+  __device__ cudf::size_type operator()(cudf::size_type const&) const { return value; }
+};
```
ttnghia, line 33: Unused?

src/main/cpp/src/cast_string_to_timestamp.cu
res-life: UTF8String.trimAll treat ascii 127 as space.
`Character.isISOControl(127) = true`

Details:
https://github.com/apache/spark/blob/v3.5.0/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala#L412
```scala
def parseTimestampString
...
  val bytes = s.trimAll().getBytes
```

https://github.com/apache/spark/blob/v3.5.0/common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java
```java
  public UTF8String trimAll() {
    ... 
    while (s < this.numBytes && isWhitespaceOrISOControl(getByte(s))) s++;
  }

  private boolean isWhitespaceOrISOControl(int codePoint) {
    return Character.isWhitespace(codePoint) || Character.isISOControl(codePoint);
  }
```

src/main/cpp/src/cast_string_to_timestamp.cu
res-life: No, allow tailing pattern like: ".123456000000000000 PST"

src/main/cpp/src/cast_string_to_timestamp.cu
res-life: Yes, no error report.
The `parse_digits` is similar to `parse_int`, the diff is that `parse_digits` returns the parsed number of parsed digits.

In `parse_tz_from_sign`, it needs to check the parsed digits number:
```cpp
        if (!(m_digits == 2 && (s_digits == 0 || s_digits == 2) && eof(pos, end_pos))) {
          // not: [+-]h[h][mm] or [+-]h[h][mm][ss]
          return make_invalid_tz();
        }
```

src/main/cpp/src/utils.hpp
res-life: Removed

ttnghia (approved):
C++ approval. If somebody else can take a look more at Java code then it's great.

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -156,6 +159,163 @@ public static ColumnVector fromIntegersWithBase(ColumnView cv, int base) {
...
+      ColumnView tzIndex = parseResult.getChildColumnView(6);
+
+      // 3. fallback to cup if has any DST timezone and has any timestamp exceeds max
+      // year
+      ColumnView hasDSTCv = parseResult.getChildColumnView(5); // DST col
+      boolean exceedsMaxYearThresholdOfDST = GpuTimeZoneDB.exceedsMaxYearThresholdOfDST(tsSeconds);
+      boolean hasDST;
+      try (Scalar s = hasDSTCv.sum(DType.INT32)) {
+        hasDST = s.isValid() && s.getInt() > 0;
+      }
```
firestarman, line 307:
```suggestion
      boolean hasDST;
      try (Scalar s = hasDSTCv.any()) {
        hasDST = s.isValid() && s.getBoolean();
      }
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -91,12 +94,21 @@ public static void cacheDatabaseAsync(int maxYear) {
...
   }
 
-  public static synchronized void verifyDatabaseCached() {
+  private static synchronized void verifyDatabaseCachedSync() {
     if (timeZoneInfo == null) {
       throw new IllegalStateException("Timezone DB is not loaded, or the loading was failed.");
     }
   }
 
+  public static void verifyDatabaseCached() {
```
firestarman, line 103: I don't know how this API is used.
But since it is public, personally it would be better return a boolean instead or comment that it would throw out an exception when database is missing.

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
res-life: I remember Nghia said that cuDF logical operator is slow than math operator. So here use sum instead of any.

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -156,6 +159,163 @@ public static ColumnVector fromIntegersWithBase(ColumnView cv, int base) {
...
+    }
+
+    long defaultEpochDay = LocalDate.now(ZoneId.of(defaultTimeZone)).toEpochDay();
+
+    // 2. parse to intermediate result
+    try (ColumnVector tzInfo = GpuTimeZoneDB.getTimeZoneInfo();
+        ColumnVector parseResult = parseTimestampStrings(
+            input, defaultTimeZoneIndex, defaultEpochDay, tzInfo)) {
+
+      ColumnView invalid = parseResult.getChildColumnView(0);
```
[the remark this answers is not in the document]
res-life: These column views are sub views of struct column vector `parseResult`, parseResult is in try block and will be closed. So the sub views will not leak.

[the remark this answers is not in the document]
firestarman: yeah, the buffer is closed, but the native object of the column view is not. You can refer to the code in our Plugin.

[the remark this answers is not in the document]
firestarman: e.g.
```
  override def doColumnar(input: GpuColumnVector): ColumnVector = {
    val normalizedStringCV = withResource(HashUtils.normalizeInput(input.getBase))
    { normalized =>
      // at the moment sha1 on CuDF doesn't support lists, so have to translate
      // BinaryType into StringType first before we operate on it
      withResource(normalized.getChildColumnView(0)) { dataCol =>
        withResource(new ColumnView(DType.STRING, normalized.getRowCount,
          Optional.of[java.lang.Long](normalized.getNullCount),
          dataCol.getData, normalized.getValid, normalized.getOffsets)) { cv =>
          cv.copyToColumnVector()
        }
      }
    }
    withResource(normalizedStringCV) { normalizedStringCV => 
      withResource(ColumnVector.sha1Hash(normalizedStringCV)) { fullResult =>
        // necessary because cudf treats nulls as "" for hashing
        fullResult.mergeAndSetValidity(BinaryOp.BITWISE_AND, normalizedStringCV)
      }
    }
  }
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: `verifyDatabaseCachedSync` is a `static synchronized` method.  Normally it does not throw an exception. Normally it waits loading is done. And it's expected that exception occurs when database is missing.

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -156,6 +159,159 @@ public static ColumnVector fromIntegersWithBase(ColumnView cv, int base) {
...
+      }
+      if (exceedsMaxYearThresholdOfDST && hasDST) {
+        // run on CPU
+        return convertToTimestamp(input.getNullCount(), invalid, tsSeconds, tsMicroseconds,
+            tzType, tzOffset, tzIndex, ansi_enabled, /* runOnGpu */ false);
+      }
+
+      // 4. convert to timestamp
+      return convertToTimestamp(input.getNullCount(), invalid, tsSeconds, tsMicroseconds,
+          tzType, tzOffset, tzIndex, ansi_enabled, /* runOnGpu */ true);
```
firestarman, line 311:
```suggestion
      // 4. convert to timestamp
      return convertToTimestamp(input.getNullCount(), invalid, tsSeconds, tsMicroseconds,
          tzType, tzOffset, tzIndex, ansi_enabled, !(exceedsMaxYearThresholdOfDST && hasDST));
```

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -156,6 +159,159 @@ public static ColumnVector fromIntegersWithBase(ColumnView cv, int base) {
...
+        if (ansi_enabled && result.getNullCount() > originInputNullcount) {
+          // has new nulls, means has any invalid data,
+          // e.g.: format is invalid, year is not supported 7 digits
+          // protocol: if ansi mode and has any invalid data, return null
+          return null;
+        } else {
+          return result.incRefCount();
+        }
+      }
+    }
```
firestarman, line 229:
```suggestion
    ColumnVector result;
    if (runOnGpu) {
      result = GpuTimeZoneDB.fromTimestampToUtcTimestampWithTzCv(
          invalid, ts_seconds, ts_microseconds, tzType, tzOffset, tzIndex);
    } else { // run on CPU
      result = GpuTimeZoneDB.cpuChangeTimestampTzWithTimezones(
          invalid, ts_seconds, ts_microseconds, tzType, tzOffset, tzIndex);
    }
   
    try (ColumnVector tmp = result) {
      if (ansi_enabled && tmp.getNullCount() > originInputNullcount) {
        // has new nulls, means has any invalid data,
        // e.g.: format is invalid, year is not supported 7 digits
        // protocol: if ansi mode and has any invalid data, return null
        return null;
      } else {
        return tmp.incRefCount();
      }
    }
```

res-life:
build

sperlingxx:
LGTM. Great job!

firestarman (approved):
LGTM for the java part

res-life:
Thanks all for your review.

PR #3180, opened 2025-04-29
Override CTK with HOST_CUDA_PATH

### LINKED ISSUE #3123: [FEA] Provide an opt-in override to use a host CUDA Toolkit inside the CI container
Sometimes we need to test build against an upcoming CTK version 

On an experimental host we want to mount CTK path from the host instead of using the one in the Docker image. 
The interface could be something like setting an environment variable HOST_CUDA_TOOLKIT_DIR whose value is mounted into Docker and passed as `CUDA_BIN_PATH` or `CUDA_TOOLKIT_ROOT_DIR` to the container https://cmake.org/cmake/help/latest/module/FindCUDA.html

Internal repo

**Additional context**
Testing

gerashegalov:
Closes #3123

If requested by defining the HOST_CUDA_PATH environment this PR mounts this path replacing the container content under /usr/local/cuda. This is the minimum change because /usr/local/cuda/bin is already part of the PATH inside the container

This PR switches from volumes to bind mounts, the --mount syntax is recommended. 

Example 
```
HOST_CUDA_PATH=/usr/local/cuda-12.8 ./build/build-in-docker clean package -DCPP_PARALLEL_LEVEL=$(nproc) -DskipTests -DGPU_ARCHS=native
```

### DIFF
--- a/build/run-in-docker
+++ b/build/run-in-docker
@@ -60,46 +60,37 @@ else
   RUN_CMD="${@@Q}"
 fi
 
-HOST_MOUNT_TYPE=bind
-RO_MOUNTS=(
+MNT_ARGS=()
+
+RO_SRC=(
   "/etc/group"
   "/etc/passwd"
   "/etc/shadow"
   "/etc/sudoers.d"
 )
+RO_DST=("${RO_SRC[@]}")
+if [[ "$HOST_CUDA_PATH" != "" ]]; then
+  RO_SRC+=("$HOST_CUDA_PATH")
+  RO_DST+=("/usr/local/cuda")
+fi
+for (( i=0; i<${#RO_SRC[@]}; i++)); do
+  MNT_ARGS+=(--mount type=bind,src=${RO_SRC[$i]},dst=${RO_DST[$i]},ro)
+done
 
-RW_MOUNTS=(
+RW_SRC=(
   "$GIT_COMMON_DIR"
   "$WORKDIR"
   "$LOCAL_CCACHE_DIR"
   "$LOCAL_MAVEN_REPO"
 )
+for (( i=0; i<${#RW_SRC[@]}; i++)); do
+  MNT_ARGS+=(--mount type=bind,src=${RW_SRC[$i]},dst=${RW_SRC[$i]})
+done
 
-function mount_args() {
-  local mount_mode=$1
-  shift
-  local mounts=("$@")
-  for m in "${mounts[@]}"
-  do
-    if [[ "$mount_mode" == "ro" ]]
-    then
-      printf -- ' --mount type='$HOST_MOUNT_TYPE',src=%s,dst=%s,%s\n' "$m" "$m" "$mount_mode"
-    else
-      # rw is default
-      printf -- ' --mount type='$HOST_MOUNT_TYPE',src=%s,dst=%s\n' "$m" "$m"
-    fi
-  done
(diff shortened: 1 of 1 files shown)


revans2 (dismissed):
The change looks fine, just a little confused.

1) Why do we need to add in mount_args, when the previous code was smaller? I am fine that it makes it more consistent, but that leads to my second question
2) Why are we not using mount_args for HOST_CUDA_BIN_PATH processing? My guess at why we need mount_args is to make the processing consistent and less prone to typos, but then why is this extra one hard coded?

gerashegalov:
> The change looks fine, just a little confused.
> 
> 1. Why do we need to add in mount_args, when the previous code was smaller? I am fine that it makes it more consistent, but that leads to my second question
> 2. Why are we not using mount_args for HOST_CUDA_BIN_PATH processing? My guess at why we need mount_args is to make the processing consistent and less prone to typos, but then why is this extra one hard coded?

The difference between 1 and  2 is that the latter has a different destination option whereas the former has different special cases. The original idea was to make it more concise with xargs but I did not get it done before last night meeting with the CUDA team. So I opted out for a brute-force working solution. I'll try to clean it up

build/run-in-docker
```diff
@@ -60,16 +60,45 @@ else
...
+    then
+      printf -- ' --mount type=bind,src=%s,dst=%s,%s\n' "$m" "$m" "$mount_mode"
+    else
+      # rw is default
+      printf -- ' --mount type=bind,src=%s,dst=%s\n' "$m" "$m"
+    fi
+  done
+}
+
+if [[ "$HOST_CUDA_BIN_PATH" != "" ]]
```
ttnghia, line 93: I think the `BIN_PATH` is not accurate, as we need to specify the entire CTK path including `bin` path and `lib` and `lib64` paths etc. So saying `BIN` is misleading.
```suggestion
if [[ "$HOST_CUDA_ROOT_PATH" != "" ]]
```

ttnghia: Then how about `HOST_CUDA_TOOLKIT_ROOT_DIR`? The suffix `_ROOT_DIR` is more accurate than `_BIN_PATH`.

ttnghia: I did a quick search and found this:
```
Make sure that after installation that you have CUDA_BIN_PATH, CUDA_INC_PATH and CUDA_LIB_PATH set to the location of where CUDA was installed (e.g., C:\CUDA\bin, C:\CUDA\include, C:\CUDA\lib) 
```
So indeed, `CUDA_BIN_PATH` should point to the `CUDA_TOOLKIT_ROOT_PATH/bin` directory. The CMake doc seems incorrect.

ttnghia: Oh wow, also found another env variable: `CUDA_PATH`, which is shorter and seems to be better than `CUDA_TOOLKIT_ROOT_DIR`: https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html.

So how about `HOST_CUDA_PATH`?

mythrocks: 👍 FWIW, I liked the idea of `HOST_CUDA_TOOLKIT_ROOT_DIR`, personally.  It is explicit about being the "root".

But I'm happy to go with what you two work out.

ttnghia (approved):

ttnghia:
build

PR #3186, opened 2025-04-30
Leverage cudf's compilation flags and definitions

ttnghia:
We are using libcudf for building our code base, thus we should try to use the same set of compiler flags and definitions as them. After libcudf exports their compiler flags and definitions, we just need to retrieve and use these, instead of maintaining our own versions of flags/definitions that get out of sync over time.

This completely replaces our compiler flags and definitions with those exported from libcudf.

Depends on:
 * https://github.com/rapidsai/cudf/pull/18604


hyperbolic2346:
How will this impact the debug build scenario? Would this mean to debug spark-rapids-jni that cudf would need to be built in debug as well?

ttnghia:
> How will this impact the debug build scenario? Would this mean to debug spark-rapids-jni that cudf would need to be built in debug as well?

Yes, this would happen as we won't have separate build types for libcudf and spark-rapids-jni. But I don't see this is a big problem since we do not need to build in `Debug` mode to debug CUDA code.

abellina:
> > How will this impact the debug build scenario? Would this mean to debug spark-rapids-jni that cudf would need to be built in debug as well?
> 
> Yes, this would happen as we won't have separate build types for libcudf and spark-rapids-jni. But I don't see this is a big problem since we do not need to build in `Debug` mode to debug CUDA code.

That said, we could still do stuff like define custom flags for a single file we want to compile in the future (debug only my .cpp file please). That way we don't have to build all of cuDF as Debug.

ttnghia:
> That said, we could still do stuff like define custom flags for a single file we want to compile in the future (debug only my .cpp file please). That way we don't have to build all of cuDF as Debug.

Yes, we can do that entirely. Here we only retrieve the flags/definitions from libcudf, copying them into local CMake variables (https://github.com/NVIDIA/spark-rapids-jni/pull/3186/files#diff-384ae7f84921adb12774a1067bae7995145fa959216de970c34ad47a5b41d414R147-R150). We can do whatever we want with these variables here.

ttnghia:
Note that we set the build mode for libcudf by a dedicated flag `CUDF_BUILD_TYPE`. So we can entirely build libcudf in Release mode while spark-rapids-jni in Debug mode. I'll add more CMake code to separate the libcudf's flags/definitions if we have a different build mode.

abellina (approved):

ttnghia:
build

ttnghia:
build

gerashegalov:
it might have been originally done to quickly override the flags in CI

What is our new mechanism of changing the flags for our binaries independently from cudf

ttnghia:
> What is our new mechanism of changing the flags for our binaries independently from cudf

If you look into `buildcpps.sh`, the section building `spark-rapids-jni` is:
```
echo "Configuring spark-rapids-jni native libs"
...
  cmake \
```
Here, in order to specify the build type independently from libcudf, we can add another line such as:
```
 cmake \
   -DCMAKE_BUILD_TYPE="$SPARK_RAPIDS_JNI_BUILD_TYPE" \
```

And the variable `SPARK_RAPIDS_JNI_BUILD_TYPE` needs to be added across the build pipeline.

gerashegalov:
I would wait for DevOps to chime in

ttnghia:
Yeah that is fine, more ideas are always welcome. Nevertheless, this is more on the software side (CMake/C++) rather than DevOps side.

ttnghia:
@pxLi @yinqingh do you have any comments on this?

pxLi (approved):
+1

PR #3190, opened 2025-05-01
Kudo short validity fix

nvdbaranec:
Fixes an edge case in the validity copy kernel:   When we have an incoming bit shift (row split position essentially) large enough that the number of rows in the first batch of bits to be copied does not fit entirely in one output word, but we _also_ don't have enough bits to require looping, there is an early out in the code that wasn't copying trailing bits.

More concretely:   if we had 12 rows/bits to store for a partition, but they start at row 22 in the destination, we have two extra bits (12 + 22 == 34) that overflow into the second bitmask word.  We were computing this correctly, but not storing it because of an incomplete early out case.

### DIFF
--- a/src/main/cpp/tests/shuffle_split.cu
+++ b/src/main/cpp/tests/shuffle_split.cu
@@ -396,6 +396,22 @@ TEST_F(ShuffleSplitTests, ShortNulls)
     run_split(tbl, {2, 5});
     run_split(tbl, {0, 2, 4});
   }
+
+  // 34 rows (two input words) split such that the last two rows of validity are written as overflow
+  // bits in the copy kernel.
+  // - split at row 22, causing the second copy to have 12 bits to write.
+  // - the first 10 of those 12 bits goes in the first word, the last 2 bits goes into the second
+  // word. this is an
+  //   edge case in the copy kernel involving an early-out.
+  {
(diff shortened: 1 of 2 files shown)


ttnghia (approved):

src/main/cpp/tests/shuffle_split.cu
```diff
@@ -396,6 +396,22 @@ TEST_F(ShuffleSplitTests, ShortNulls)
     run_split(tbl, {2, 5});
     run_split(tbl, {0, 2, 4});
   }
+
+  // 34 rows (two input words) split such that the last two rows of validity are written as overflow
+  // bits in the copy kernel.
+  // - split at row 22, causing the second copy to have 12 bits to write.
+  // - the first 10 of those 12 bits goes in the first word, the last 2 bits goes into the second
+  // word. this is an
```
mythrocks, line 404: Is there an inadvertent new-line in the comments here?

nvdbaranec: Yeah, this is what pre-commit did.  🤷

revans2:
build

mythrocks (approved):
Sorry for the delay.  Forgot about this one.  👍

revans2 (approved):
This fixed the issues I was seeing

revans2:
build

PR #3175, opened 2025-04-26
Upgrade `ccache` in Dockerfile

ttnghia:
This upgrades `ccache-4.6` to `ccache-4.11.2` in Dockerfile.

PR #3198, opened 2025-05-05
Drop cuda11 and make cuda12 as default

pxLi:
Update build default to cuda12


pxLi:
build

pxLi:
build

pxLi:
build

ttnghia (commented):
I have a similar (but not enough changes) proposal in https://github.com/NVIDIA/spark-rapids-jni/pull/3175. How about upgrading CUDA to `12.8.1` instead of `12.8.0`?

pxLi:
> I have a similar (but not enough changes) proposal in #3175. How about upgrading CUDA to `12.8.1` instead of `12.8.0`?

It's fine to update the cuda version, your change will be auto-merged into 25.08 if no conflict. For this change, the purpose is switch the default only

yinqingh (approved):

PR #3211, opened 2025-05-06
more loggings in OOM state machine

binmahone:
add some loggings in OOM state machine so that it will become more friendly for troubleshooting GPU/CPU OOM because of state machine issues.

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -1674,7 +1674,7 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
 
       logger->info(
         "deadlock state is reached with all_task_ids: {} ({}), blocked_task_ids: {} ({}), "
-        "bufn_task_ids: {} ({}), threads: ({})",
+        "bufn_task_ids: {} ({}), threads: {} ({})",
         setToString(all_task_ids), all_task_ids.size(),
         setToString(blocked_task_ids), blocked_task_ids.size(),
         setToString(bufn_task_ids), bufn_task_ids.size(),
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -1551,6 +1551,24 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     return ret;
   }
 
+  // Function to convert a set (ordered or unordered) of long long to a concatenated string
+  template <typename SetType>
+  std::string setToString(const SetType& longSet, const std::string& separator = ",") {
+    // Use std::ostringstream for efficient string building.
+    std::ostringstream oss;
+
+    oss << "{";
+    // Iterate through the set.
+    for (auto it = longSet.begin(); it != longSet.end(); ++it) {
+      oss << *it;
+      if (std::next(it) != longSet.end()) {
+        oss << separator;
+      }
+    }
+    oss << "}";
+    return oss.str();
+  }
+
   bool is_in_deadlock(std::map<long, long>& pool_bufn_task_thread_count,
                       std::map<long, long>& pool_task_thread_count,
                       std::unordered_set<long>& bufn_task_ids,
@@ -1637,13 +1655,19 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     // Now if all of the tasks are blocked, then we need to break a deadlock
     bool ret = all_task_ids.size() == blocked_task_ids.size() && !all_task_ids.empty();
     if (ret) {
+
+      std::set<int> threadsKeySet;
+      std::transform(threads.begin(), threads.end(),
+                     std::inserter(threadsKeySet, threadsKeySet.begin()),
+                     [](const auto& pair) { return pair.first; });
+
       logger->info(
-        "deadlock state is reached with all_task_ids size: {}, blocked_task_ids: {}, "
-        "bufn_task_ids: {}, threads size: {}",
-        all_task_ids.size(),
-        blocked_task_ids.size(),
-        bufn_task_ids.size(),
-        threads.size());
+        "deadlock state is reached with all_task_ids: {} {}, blocked_task_ids: {} {}, "
+        "bufn_task_ids: {} {}, threads size: {} {}",
+        setToString(all_task_ids), all_task_ids.size(),
+        setToString(blocked_task_ids), blocked_task_ids.size(),
+        setToString(bufn_task_ids), bufn_task_ids.size(),
+        setToString(threadsKeySet),threads.size());
     }
     return ret;
   }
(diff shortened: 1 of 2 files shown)


binmahone:
build

revans2 (commented):
I am concerned about the scala logging adding more at info level. I think it is okay, because it is small and really about threads starting/tasks starting/stopping. But I would prefer it if it was a debug level.

The other logging I am also not super happy about. The original code has a strict pattern for what was logged at INFO level

https://github.com/NVIDIA/spark-rapids-jni/blob/460de414f0b34a49781e02dbc502ee0936eca326/src/main/cpp/src/SparkResourceAdaptorJni.cpp#L426

This has changed all of that so now automatically parsing the data is not simple. Could we keep them at debug level and expose a way to set the logger to produce DEBUG logs?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1637,17 +1665,38 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+                     threads.end(),
+                     std::inserter(threadsKeySet, threadsKeySet.begin()),
+                     [](const auto& pair) { return pair.first; });
+
       logger->info(
-        "deadlock state is reached with all_task_ids size: {}, blocked_task_ids: {}, "
-        "bufn_task_ids: {}, threads size: {}",
+        "deadlock state is reached with all_task_ids: {} ({}), blocked_task_ids: {} ({}), "
+        "bufn_task_ids: {} ({}), threads: {} ({})",
+        setToString(all_task_ids),
```
abellina, line 1677: seems like this is only used when there is a deadlock: hope it's not often. The reason I mention this is that even if the logger level is set to warn/error we will still call `setToString` and build a string that won't be printed anywhere.

binmahone:
> I am concerned about the scala logging adding more at info level. I think it is okay, because it is small and really about threads starting/tasks starting/stopping. But I would prefer it if it was a debug level.
> 
> The other logging I am also not super happy about. The original code has a strict pattern for what was logged at INFO level
> 
> https://github.com/NVIDIA/spark-rapids-jni/blob/460de414f0b34a49781e02dbc502ee0936eca326/src/main/cpp/src/SparkResourceAdaptorJni.cpp#L426
> 
> This has changed all of that so now automatically parsing the data is not simple. Could we keep them at debug level and expose a way to set the logger to produce DEBUG logs?

Hi Bobby, for the scala logging part, like you said since it wouldn't generate too much verbose logs, I suggest to keep it INFO level, so that we don't have to manually switch log level to DEBUG, run the query again, to get the log we want.

For the spdlog part, in my understanding all of the spdlog logs are already "DEBUG" level per se : even if in cpp code we're using INFO level, a query running in default configs will not print any spdlog logs unless spark.rapids.memory.gpu.state.**debug** is set. So why bother to introduce another debug log level when all of the log is already debug level? Considering your automatically parsing requirement, how about I add a prefix ([DETAIL]) for all of the non-uniform logs like this:


23:17:55.338425,**[DETAIL]** deadlock state is reached with all_task_ids: {39058} 1, blocked_task_ids: {39058} 1, bufn_task_ids: {39058} 1, threads: {-1460631808,-1412360448} 2
23:17:55.338444,TRANSITION,139713873778432,139713873778432,39058,THREAD_RUNNING,THREAD_ALLOC,

then you might be able to do some invert match on [DETAIL] (e.g. grep -v "[DETAIL]") and get all the transition logs you want? what you think ?

src/main/cpp/src/SparkResourceAdaptorJni.cpp
binmahone: hi @abellina ,  I will protect them with `if(is_log_enabled)`

revans2:
@binmahone If you want to provide a way to separate out the two, that is fine. Or you could have your new/updated logs correspond to the format of the existing log, lines. It might be nice to know that thread 5 discovered that there was a deadlock, or if it was the watchdog thread. The very last field is "notes" and you can put more or less anything you want in there.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
binmahone: fixed

binmahone:
@revans2 I have updated the PR so that everywhere is using existings log_status, pls take another look

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1551,6 +1566,23 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
     return ret;
   }
 
+  // Function to convert a set (ordered or unordered) to a concatenated string
+  template <typename SetType>
+  std::string setToString(const SetType& set, const std::string& separator = ",")
```
ttnghia, line 1571: Just call it `to_string` for simplicity.
```suggestion
  std::string to_string(const SetType& set, const std::string& separator = ",")
```

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1637,17 +1669,41 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
     bool ret = all_task_ids.size() == blocked_task_ids.size() && !all_task_ids.empty();
     if (ret) {
-      logger->info(
-        "deadlock state is reached with all_task_ids size: {}, blocked_task_ids: {}, "
-        "bufn_task_ids: {}, threads size: {}",
-        all_task_ids.size(),
-        blocked_task_ids.size(),
-        bufn_task_ids.size(),
-        threads.size());
+      std::set<long> threadsKeySet;
```
ttnghia, line 1672: C++ names are `snake_case`.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1637,17 +1669,41 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+      std::transform(threads.begin(),
+                     threads.end(),
+                     std::inserter(threadsKeySet, threadsKeySet.begin()),
+                     [](const auto& pair) { return pair.first; });
```
ttnghia, line 1676:
```suggestion
      for(auto const& [k, v] : threads) { thread_key_set.insert(k); }
```

ttnghia: In addition, this `std::set` is unused if `is_log_enabled` is `false`, so it should better be declared inside that `if` branch.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
ttnghia: In addition, I don't see this function is used with both type of set (order + unorder) so for simplicity, it is better to just declare the exact type instead of template:
```
std::string to_string(std::set<....> const& set, ...);
```

binmahone:
@ttnghia can you please take a look again?

ttnghia (approved):

PR #3221, opened 2025-05-06
Reuse `libcudf` dependencies when they are available

ttnghia:
This modifies CMake build to search for dependency packages that have been installed alongside with libcudf. Having this, we can avoid pulling dependencies multiple times when they are already available. There are also a few minor cleanup are performed in this work.

Depends on:
 * https://github.com/rapidsai/cudf/pull/18682


src/main/cpp/CMakeLists.txt
```diff
@@ -26,13 +22,12 @@ set(CUDFJNI_BUILD_DIR
...
 )
-set(CUDF_CPP_BUILD_DIR
-  "$ENV{CUDF_CPP_BUILD_DIR}"
-  CACHE STRING "path to libcudf build root"
-)
+
 # libcudf's kvikio dependency requires this to be set when statically linking CUDA runtime
 set(CUDA_STATIC_RUNTIME ON)
 
+set(CUDF_DIR "${CMAKE_SOURCE_DIR}/../../../thirdparty/cudf")
```
gerashegalov, line 29: Relative paths, and duplicate definitions of the paths by different means make it more error-prone. Why not propagate and rely on `$ENV{CUDF_PATH}` from Maven?

ttnghia: Humnn, that is also possible if we need to build against external cudf repo. I'll revert that.

ttnghia: Done. I've added a comment to clarify the purpose, so future refactor will not do this again.

build/buildcpp.sh
```diff
@@ -138,8 +139,7 @@ cmake --build "$LIBCUDFJNI_BUILD_PATH" "-j$CPP_PARALLEL_LEVEL"
 mkdir -p "$SPARK_JNI_BUILD_PATH"
 cd "$SPARK_JNI_BUILD_PATH"
 echo "Configuring spark-rapids-jni native libs"
-CUDF_CPP_BUILD_DIR="$LIBCUDF_BUILD_PATH" \
```
ttnghia, line 141: This is unused.

ttnghia:
build

gerashegalov (approved):

Issue #3227, opened 2025-05-07
[BUG] nightly compilation fails error #186-D: pointless comparison of unsigned integer with zero in arm64


pxLi:
**Describe the bug**
First seen in spark-rapids-jni_nightly-dev, run: 1095 (edbbc787dca46617036f8252dee38ae8326a04ee branch-25.06)

Note: this is failing in arm64 (cuda11.8 and 12.8, driver 535) compilation only, the X86 build (cuda11.8&12.8, driver 570) still works well
```
[2025-05-07T03:11:06.463Z] [INFO]      [exec] FAILED: CMakeFiles/spark_rapids_jni.dir/src/cast_string_to_float.cu.o 

[...]

PR #3242, opened 2025-05-08
Fix compile warnings when comparing unsigned char to zero

ttnghia:
This fixes the (potential) warnings when comparing unsigned char to zero. Most of the times, the warnings do not show up because `char` is a signed type on `x86` platform but it can be unsigned on `arm64` platform (`char` can be signed or unsigned, it is not specified by C++ standard).

Fixes https://github.com/NVIDIA/spark-rapids-jni/issues/3227.


ttnghia:
build

revans2 (approved):

abellina (approved):

PR #3250, opened 2025-05-09
Fix compiler warning on `arm64` platform

ttnghia:
This fixes the compiler warning when comparing an `unsigned char <= 0`. The warnings only show up when building on `arm64` platform. There was an attempt to fix it in https://github.com/NVIDIA/spark-rapids-jni/pull/3242 but that seems didn't work.

### DIFF
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -46,19 +46,15 @@ constexpr auto NUM_THREADS{256};
  */
 constexpr bool is_whitespace(char const chr)
 {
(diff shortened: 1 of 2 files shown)


ttnghia:
build

pxLi (approved):

ttnghia:
build

pxLi (approved):
Confirmed there is no more compilation error w/ this change in both x86 and arm64, thanks

I will merge once the pre-merge is finished

pxLi:
build

pxLi:
rekicked due to internal 500

PR #3286, opened 2025-05-14
Add centralized priority for SparkResourceAdaptor

revans2:
This keeps the priority calculations unchanged to avoid issues with tests in the rapids plugin


revans2:
build

PR #3290, opened 2025-05-14
Change the priority to be first seen is highest

revans2:
This does not really have a big impact yet. I ran NDS and saw no improvement or loss, even if I reduce the amount of memory available to the processes by a lot.  The improvements will come when I get spill to also have the same priorities, and then we start to run into situations where there is a lot of spilling.


revans2:
build

revans2:
build

src/main/cpp/src/TaskPriorityJni.cpp
```diff
@@ -20,14 +20,39 @@
 
 namespace spark_rapids_jni {
 
+// Track the next priority to assign and maintain a map of attempt_id to priority
+static long next_task_priority = std::numeric_limits<long>::max() - 1;
+static std::mutex priority_mutex;
+static std::unordered_map<long, long> attempt_priorities;
```
ttnghia, line 26: These static variables are directly under the `spark_rapids_jni` namespace, which are visible and can be mixed with all other JNI functions. This is not a good encapsulation practice. How about put them into some other namespace, like this?
```
namespace task_priority {
static long ....;
}
```

ttnghia: Another option is that, we put them into an anonymous namespace, which guarantees them to be visible **only** to the function in this source file. I think this is a better option because they should not be visible to any other functions in other places (shouldn't they?).

src/main/cpp/src/TaskPriorityJni.cpp
```diff
@@ -20,14 +20,39 @@
...
+
+  std::lock_guard<std::mutex> lock(priority_mutex);
+  auto it = attempt_priorities.find(attempt_id);
+  if (it != attempt_priorities.end()) {
+    // Return existing priority for this attempt_id
+    return it->second;
+  }
+
+  // Assign new priority for this attempt_id
+  long priority                  = next_task_priority--;
```
ttnghia, line 43: `next_task_priority` only decreases over time. Is this on purpose?

revans2:
build

revans2:
@ttnghia I have addressed your comments please take another look

src/main/cpp/src/TaskPriorityJni.cpp
```diff
@@ -18,16 +18,43 @@
...
+}  // namespace
+
 namespace spark_rapids_jni {
 
 long get_task_priority(long attempt_id)
 {
-  return std::numeric_limits<long>::max() - (attempt_id + 1);
+  if (attempt_id == -1) {
+    // Special case: -1 always gets highest priority
+    return std::numeric_limits<long>::max();
```
abellina, line 34: so -1 is meant to be higher than the first task? (which is max() - 1)? I assume this is the intent.

abellina (approved):

src/main/cpp/src/TaskPriorityJni.cpp
revans2: yes this matches the code that was there before. -1 is used to indicate a thread with no task associated with it.

abellina: Presumably if we have work going on in a thread not associated with the task we don't know how to prioritize it, yet. This is a default chosen to not block those tasks/and to not spill that data first. We should experiment with other defaults here.

It may be that we arrive at a place where not thread has -1 priority, but we need to experiment.

Issue #3292, opened 2025-05-15
[FEA] Support large years for casting date to string.


res-life:
**Is your feature request related to a problem? Please describe.**
PR adds large years (4-6 digits) support for string to ts casting
This issue is to add large years (4-6 digits) support for the reverse casting

PR #3298, opened 2025-05-16
Kernel for casting string to date

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/12737
closes https://github.com/NVIDIA/spark-rapids-jni/issues/3292

Kernel for casting string to date
Note: Spark date supports max 7 digits year, Spark timestamp supports max 6 digits year.

## changes
- Rename cast_string_to_timestamp.cu to cast_string_to_datetime.cu
- Kernel for casting string to date: support large years(7 digits).
- Refactor src/main/cpp/src/datetime_utils.cuh
- `to_epoch_day` return value changed to int64 
- Add 'to_date' which  converts epoch day to date segments, 'to_date' is not related to casting string to date, but related to the coming PR of casting date to string. I've verified all the possible value against Java API locally, this  'to_date' works well.


res-life:
build

res-life:
build

res-life:
build

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -23,91 +23,216 @@
...
+    const uint32_t m   = mp < 10 ? mp + 3 : mp - 9;                              // [1, 12]
+    year               = y + (m <= 2);
+    month              = static_cast<uint8_t>(m);
+    day                = static_cast<uint8_t>(d);
+  }
 
-  // 1-31; it is 29 for leap February, or 28 for regular February
-  int32_t day;
+  /**
+   * @brief Is the month/day is valid.
```
thirtiseven, line 92: How about "If the month/day is valid." of "Is the Spark date valid."
This also applies to to `is_valid_date_for_date`, `is_valid_date_for_timestamp`, and `is_valid_time `

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -23,91 +23,216 @@
...
+  __device__ static bool is_valid_month_day(int const year, int const month, int const day)
+  {
+    if (month < 1 || month > 12 || day < 1) {
+      return false;  // Invalid month or day
+    }
 
-  // 0-23
-  int32_t hour;
+    // Check for leap year, February has 29 days in leap year
+    if (month == 2 && is_leap_year(year)) { return (day <= 29); }
```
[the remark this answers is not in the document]
res-life: Good catch!

res-life:
build

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -23,91 +23,213 @@
...
-  __device__ int32_t to_epoch_day() const
+  __device__ static int64_t to_epoch_day(int const year, int const month, int const day)
   {
     int32_t y          = year - (month <= 2);
     const int32_t era  = (y >= 0 ? y : y - 399) / 400;
     const uint32_t yoe = static_cast<uint32_t>(y - era * 400);                           // [0, 399]
     const uint32_t doy = (153 * (month > 2 ? month - 3 : month + 9) + 2) / 5 + day - 1;  // [0, 365]
     const uint32_t doe = yoe * 365 + yoe / 4 - yoe / 100 + doy;  // [0, 146096]
-    return era * 146097 + doe - 719468;
+    return era * 146097L + doe - 719468L;
```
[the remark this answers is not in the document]
res-life: See the comment of this function: Refer to https://howardhinnant.github.io/date_algorithms.html

src/main/cpp/src/cast_string.hpp
```diff
@@ -167,4 +167,29 @@ std::unique_ptr<cudf::column> parse_timestamp_strings(
   rmm::cuda_stream_view stream      = cudf::get_default_stream(),
   rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
 
+/**
+ * @brief Parse date string column to date column, first trim the input strings.
+ *
+ *
```
ttnghia, line 173:
```suggestion
 * @brief Parse date string column to date column, first trim the input strings.
 *
```

src/main/cpp/src/cast_string.hpp
```diff
@@ -167,4 +167,29 @@ std::unique_ptr<cudf::column> parse_timestamp_strings(
...
+ * org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala#L298
+ *
+ * formats are allowed:
+ *
+ * `[+-]yyyy*`
+ * `[+-]yyyy*-[m]m`
+ * `[+-]yyyy*-[m]m-[d]d`
+ * `[+-]yyyy*-[m]m-[d]d `
+ * `[+-]yyyy*-[m]m-[d]d *`
+ * `[+-]yyyy*-[m]m-[d]dT*`
```
ttnghia, line 184:
```suggestion
 * Allowed formats:
 *   `[+-]yyyy*`
 *   `[+-]yyyy*-[m]m`
 *   `[+-]yyyy*-[m]m-[d]d`
 *   `[+-]yyyy*-[m]m-[d]d `
 *   `[+-]yyyy*-[m]m-[d]d *`
 *   `[+-]yyyy*-[m]m-[d]dT*`
```

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -857,6 +860,169 @@ std::unique_ptr<cudf::column> parse_ts_strings(cudf::strings_column_view const&
...
+  // input strings
+  cudf::column_device_view d_strings;
+
+  // output columns
+  uint8_t* null_mask;
+  cudf::timestamp_D* output;
+
+  __device__ void operator()(cudf::size_type const idx) const
+  {
+    auto const str                   = d_strings.element<cudf::string_view>(idx);
```
ttnghia, line 956: No need for null check?

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -857,6 +860,169 @@ std::unique_ptr<cudf::column> parse_ts_strings(cudf::strings_column_view const&
...
+      output[idx]    = cudf::timestamp_D{cudf::duration_D{0}};
+      return;
+    }
+
+    int64_t days = date_segments.to_epoch_day();
+    // Spark stores date int int, so need to check the range
+    if (days < cuda::std::numeric_limits<int32_t>::min() ||
+        days > cuda::std::numeric_limits<int32_t>::max()) {
+      // exceeds int range.
+      null_mask[idx] = 0;
```
ttnghia, line 977: Here we output validity instead of null mask. The null mask is set (bit `1`) when the row is null. So the variable needs to be renamed to `validity` instead.

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -857,6 +860,169 @@ std::unique_ptr<cudf::column> parse_ts_strings(cudf::strings_column_view const&
...
+      output[idx]    = cudf::timestamp_D{cudf::duration_D{0}};
+      return;
+    }
+
+    null_mask[idx] = 1;
+    output[idx]    = cudf::timestamp_D{cudf::duration_D{days}};
```
ttnghia, line 983: Indeed we don't care about the output value in case of null.
```suggestion
    output[idx] = cudf::timestamp_D{cudf::duration_D{days}};
    validity[idx] = days >= cuda::std::numeric_limits<int32_t>::min() &&
        days <= cuda::std::numeric_limits<int32_t>::max()    
```

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -857,6 +860,169 @@ std::unique_ptr<cudf::column> parse_ts_strings(cudf::strings_column_view const&
...
+
+  auto const d_input = cudf::column_device_view::create(input.parent(), stream);
+  auto result = cudf::make_timestamp_column(cudf::data_type{cudf::type_to_id<cudf::timestamp_D>()},
+                                            input.size(),
+                                            rmm::device_buffer{},
+                                            0,
+                                            stream,
+                                            mr);
+  auto null_mask = cudf::make_fixed_width_column(
+    cudf::data_type{cudf::type_id::UINT8}, num_rows, cudf::mask_state::UNALLOCATED, stream, mr);
```
ttnghia, line 1007:
```suggestion
  auto validity = rmm::device_uvector<bool>(...);
```

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -23,91 +23,213 @@
...
+  int32_t hour;
+
+  // 0-59
+  int32_t minute;
+
+  // 0-59
+  int32_t second;
+
+  // 0-999999, only parse 6 digits, ignore/truncate the rest digits
+  int32_t microseconds;
```
ttnghia, line 232: We may not need these all the time. How about separating them into a new/derived class for saving memory when they are not needed? In particular:
```
struct date_segment {
int year, month, day;
};

struct ts_segment: date_segment {
int hour, minute, second, microsecond;
};
```
So when we process only date, we won't have to allocate extra memory for the hour/minutes etc stuff that are never touched.

res-life:
build

src/main/cpp/src/datetime_utils.cuh
res-life: You are right, could it be a follow-up? And I would use composite types instead of inherited types.

```cpp
struct date_segments {
  int year, month, day;
};

struct time_segments {
  int hour, minute, second, microsecond;
}

struct timestamp_segments {
  date_segment date_segments;
  time_segment time_segments;
}

parse_int(ts.year) => parse_int(ts.date_segments.year)

```cpp

File: src/main/cpp/src/cast_string_to_datetime.cu (hunk 6 above)
Comment (res-life) [reply src/main/cpp/src/cast_string_to_datetime.cu:961]: Yes, not need. But I updated, checking null will be a little faster IMO.

```cpp
    // check null
    if (d_strings.is_null(idx)) {
      validity[idx] = false;
      return;
    }
```

res-life:
build

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -23,91 +23,213 @@
...
+    const uint32_t doe = static_cast<uint32_t>(z - era * 146097);                // [0, 146096]
+    const uint32_t yoe = (doe - doe / 1460 + doe / 36524 - doe / 146096) / 365;  // [0, 399]
+    const int32_t y    = static_cast<uint32_t>(yoe) + era * 400;
+    const uint32_t doy = doe - (365 * yoe + yoe / 4 - yoe / 100);                // [0, 365]
+    const uint32_t mp  = (5 * doy + 2) / 153;                                    // [0, 11]
+    const uint32_t d   = doy - (153 * mp + 2) / 5 + 1;                           // [1, 31]
+    const uint32_t m   = mp < 10 ? mp + 3 : mp - 9;                              // [1, 12]
+    year               = y + (m <= 2);
+    month              = static_cast<uint8_t>(m);
+    day                = static_cast<uint8_t>(d);
```
ttnghia, line 88: `month` and `day` are `int` type instead of `uint8_t`.

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -23,91 +23,213 @@
...
 
-  __device__ bool is_leap_year() const
+  /**
+   * @brief If the time is valid.
+   * Spark timestamp: hour 0-23, minute 0-59, second 0-59, microseconds 0-999999.
+   */
+  __device__ static bool is_valid_time(int const hour,
+                                       int const minute,
+                                       int const second,
+                                       int const microseconds)
```
ttnghia, line 136: Don't to put `const` qualifier for plain type to avoid compiler warning.

res-life:
build

res-life:
build

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -858,6 +861,163 @@ std::unique_ptr<cudf::column> parse_ts_strings(cudf::strings_column_view const&
...
+    // check parsed result
+    if (!result_success || !date_segments.is_valid_date()) {
+      validity[idx] = false;
+      return;
+    }
+
+    // calculate the epoch day and check 'days' fits in int32_t
+    int64_t days  = date_segments.to_epoch_day();
+    output[idx]   = cudf::timestamp_D{cudf::duration_D{static_cast<int32_t>(days)}};
+    validity[idx] = days >= cuda::std::numeric_limits<int32_t>::min() &&
```
sperlingxx, line 980: With the int32 limitation, the year > `Int.MaxValue / 365 = 5883516` will return null ?

res-life: Yes. Exactly year > (Int.MaxValue / 365 - 1970), 1970-01-01 is the epoch day.

ttnghia (approved):
Nit: Please use `east const` from lines 58-83.

src/main/cpp/src/cast_string_to_datetime.cu
sperlingxx: Looks good to me if Spark also returns null

PR #3302, opened 2025-05-16
Fix bug in casting string to timestamp

res-life:
Fix bug in casting string to timestamp

### bug
E.g.: "2020-01-01 00:00:00.101", the tailing ".101" represents 101000 microseconds,  because the digits in the microseconds part is fraction.
Before this PR the code gets 101, but the 101000  is the correct value.

### fix
Refer to the code.


res-life:
build

sperlingxx (approved):
LGTM

PR #3313, opened 2025-05-19
Fix bug in casting string to timestamp: the isDST is wrong when using default timezone

res-life:
### bug
The isDST is wrong when using a DST timezone as default timezone, e.g.:  America/Los_Angeles
America/Los_Angeles is DST.

GpuTimeZoneDB.getIndexToTransitionTable("America/Los_Angeles") returns 150,
But the corresponding row inex for America/Los_Angeles in `GpuTimeZoneDB.timeZoneInfo` is 153.
Before this PR, the code use 150 instead of 153, refer to the code:
```cpp
      auto const is_DST_col = tz_info.child(2);
      if (is_DST_col.element<uint8_t>(default_tz_index)) {  // default_tz_index is index to transition table, not tz_info
        // update is DST
        is_DSTs[idx] = 1;
      }
```
Thus get wrong info.

### analysis
GpuTimeZoneDB has the following 3 fields:
- GpuTimeZoneDB.transitions
- GpuTimeZoneDB.zoneIdToTable saves the mappings: timezone -> index in transition table.
- GpuTimeZoneDB.timeZoneInfo saves  the info:  STRUCT<tz_name: string, index_to_transition_table: int, is_DST: int8>

The `transitions` only saves transitions for normalized timezone to save memory.
The zoneIdToTable and timeZoneInfo have more rows than transitions, not only  saves normalized but also non-normalized timezone.

E.g.:
After normalization for PST, it's `America/Los_Angeles`.
The `PST` and `America/Los_Angeles` can both appear in timestamp strings.
For the two timezones, there is only one transition info in `GpuTimeZoneDB.transitions`.

Before this PR, it assumes the rows are the same.

### fix
Pass in the isDST boolean for the default timezone.

### DIFF
--- a/src/main/cpp/src/CastStringJni.cpp
+++ b/src/main/cpp/src/CastStringJni.cpp
@@ -276,6 +276,7 @@ Java_com_nvidia_spark_rapids_jni_CastStrings_parseTimestampStrings(JNIEnv* env,
                                                                    jclass,
                                                                    jlong input_column,
                                                                    int default_timezone_index,
+                                                                   bool is_default_timezone_dst,
                                                                    long default_epoch_day,
                                                                    jlong timezone_info_column)
 {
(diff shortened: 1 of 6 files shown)


ttnghia (approved):

res-life:
Thanks @ttnghia for your review.

Issue #3291, opened 2025-05-15
[BUG] Test fails KudoSerializer: cudaErrorLaunchFailure unspecified launch failure


pxLi:
**Describe the bug**
spark-rapids-jni_submodule-sync-dev run: 4741
The jobs was trying to update cudf ref to https://github.com/rapidsai/cudf/commit/9e3e5b4d6572e5f65487ae50443f3982b416a108 for branch-25.06

related to https://github.com/NVIDIA/spark-rapids-jni/pull/3100

```
[2025-05-15T02:20:21.202Z] [ERROR] testSerializeAndDeserializeTable  Time elapsed: 0.003 s  <<< ERROR!
[2025-05-15T02:20:21.202Z] java.lang.RuntimeException: ai.rapids.cudf.CudaFatalException: std::bad_alloc: CUDA error (failed to allocate 256 bytes) at: /home/jenkins/agent/workspace/jenkins-spark-rapids-jni_submodule-sync-dev-4741/target/libcudf-install/include/rmm/mr/device/cuda_memory_resource.hpp:62: cudaErrorLaunchFailure unspecified launch failure
[2025-05-15T02:20:21.202Z] 	at com.nvidia.spark.rapids.jni.kudo.KudoSerializerTest.testSerializeAndDeserializeTable(KudoSerializerTest.java:58)
[2025-05-15T02:20:21.202Z] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[2025-05-15T02:20:21.202Z] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[2025-05-15T02:20:21.202Z] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[2025-05-15T02:20:21.202Z] 	at java.lang.reflect.Method.invoke(Method.java:498)
[2025-05-15T02:20:21.202Z] 	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
[2025-05-15T02:20:21.202Z] 	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
[2025-05-15T02:20:21.202Z] 	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
[...]

PR #3325, opened 2025-05-20
Fix an issue with reading off the end of input data in the copy_validity kudo kernel.

nvdbaranec:
Fixes https://github.com/NVIDIA/spark-rapids-jni/issues/3291

Fixes an issue where we could read off the end of the incoming buffer when copying validity bits.  The new code only reads as many bytes as absolutely necessary (the number of actual rows of data), while ensuring that we still do aligned 4 byte reads for the bulk of the data.

Also re-enables `testEmptyStructRoundTrip()` in the kudo serialization test suite.

### DIFF
--- a/src/main/cpp/src/shuffle_assemble.cu
+++ b/src/main/cpp/src/shuffle_assemble.cu
@@ -1264,9 +1264,6 @@ __global__ void copy_validity(cudf::device_span<assemble_batch> batches)
 
   __shared__ bitmask_type prev_word[block_size];
 
-  // note that in several cases here, we are reading past the end of the actual input buffer. but
-  // this is safe because we are always guaranteed final padding of partitions out to 8 bytes.
-
   // how many leading bytes we have. that is, how many bytes will be read by the initial read, which
   // accounts for misaligned source buffers.
   int const leading_bytes = (4 - (reinterpret_cast<uint64_t>(batch.src) % 4));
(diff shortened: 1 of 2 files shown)


revans2 (approved):

src/main/cpp/src/shuffle_assemble.cu
```diff
@@ -1279,16 +1276,35 @@ __global__ void copy_validity(cudf::device_span<assemble_batch> batches)
...
+      uint8_t const* const src_b = reinterpret_cast<uint8_t const*>(_src);
+      if (num_bits > 24) {
+        // if we are aligned we can do a single read. this should be the most common case. only
+        // at the ends of the range of bits to copy will we hit the non-aligned cases.
+        return leading_bytes == 4
+                 ? (reinterpret_cast<bitmask_type const*>(_src))[0]
+                 : static_cast<bitmask_type>(src_b[0]) | static_cast<bitmask_type>(src_b[1] << 8) |
+                     static_cast<bitmask_type>(src_b[2] << 16) |
+                     static_cast<bitmask_type>(src_b[3] << 24);
+      } else if (num_bits > 16) {
```
ttnghia, line 1292: So `leading_bytes == 4` is only relevant for `num_bits > 24`?

ttnghia:
build

src/main/cpp/src/shuffle_assemble.cu
nvdbaranec: Correct. It means we've got a full 4, aligned 4 bytes of data at the start of the buffer.

mythrocks (approved):
This should be good to commit now, I think.

Issue #3348, opened 2025-05-22
[BUG] [FOLLOW-UP] cast string to timestamp:  should not use the default timezone if there is timezone in the string for just time string.


res-life:
**Describe the bug**
Should not use the spark session timezone if there is timezone in the string for just time string.

### example 1
Spark session timezone is  "UTC", Timestamp strings is "T00:00:00 America/Los_Angeles", 

The kernel should use the current date at the current time in the tz 'America/Los_Angeles', 
current kernel is using 'UTC' for all the input strings. So the diff can be 1 days.
E.g.:  
Current date at the current time in America/Los_Angeles is 2025-05-22
But the current date in 'UTC' is 2025-05-21

Spark use `LocalDate.now(zoneId)` to get current date.

### example 2
Spark session timezone is  "UTC", Timestamp strings is "T00:00:00", 
There is no tz in the ts string, should use UTC date.

**Expected behavior**
Update kernel

res-life:
cc @ttnghia, will post PR to fix

PR #3352, opened 2025-05-22
Fix bug: Using the default timezone is wrong if there is timezone in just time string

res-life:
closes https://github.com/NVIDIA/spark-rapids-jni/issues/3348

### bug
Refer to https://github.com/NVIDIA/spark-rapids-jni/issues/3348
Should not use the spark session timezone if there is timezone in the string for just time string.

### fix
- refactor timezones.cu to extract a rebase timezone method.
- If just time and has fixed tz,  rebase the current timestamp to the fixed tz, and get the current date to use.
- If just time and has specified tz,  rebase the current timestamp to the specified tz, and get the current date to use.

### DIFF
--- a/src/main/cpp/src/cast_string_to_datetime.cu
+++ b/src/main/cpp/src/cast_string_to_datetime.cu
@@ -826,7 +826,10 @@ struct parse_timestamp_string_fn {
             transitions,
             tz_indices[idx],
             /* to_utc */ false);
-          int64_t rebased                    = *reinterpret_cast<int64_t*>(&rebased_seconds);
+
+          auto const rebased = static_cast<int64_t>(
+            cuda::std::chrono::duration_cast<cudf::duration_s>(rebased_seconds.time_since_epoch())
+              .count());
           int64_t rebased_days_of_local_date = rebased / (24L * 3600L);
 
           // Step 2: add date part to the seconds
--- a/src/main/cpp/src/cast_string_to_datetime.cu
+++ b/src/main/cpp/src/cast_string_to_datetime.cu
@@ -32,6 +32,7 @@
 #include <thrust/iterator/counting_iterator.h>
 #include <thrust/transform.h>
 
+#include <chrono>
 #include <vector>
 
 /**
@@ -623,6 +624,8 @@ __device__ inline RESULT_TYPE to_long_check_max(ts_segments const& ts,
                                                 int64_t& seconds,
                                                 int32_t& microseconds)
 {
+  // if it's a just time timestamp string, then the `days` is 0, because
+  // default date in ts_segments is 1970-01-01
   int64_t const days = ts.to_epoch_day();
 
   // seconds part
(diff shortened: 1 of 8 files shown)


res-life:
build

res-life:
build

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -783,6 +811,27 @@ struct parse_timestamp_string_fn {
...
+          //   LocalDate.now("Asia/Shanghai")       = 2025-05-22
+
+          // Step 1: rebase `current_seconds_since_epoch` from utc timezone to local timezone
+          // to get the current date
+          auto rebased_seconds = spark_rapids_jni::convert_timestamp<cudf::timestamp_s>(
+            cudf::timestamp_s{cudf::duration_s{current_seconds_since_epoch}},
+            transitions,
+            tz_indices[idx],
+            /* to_utc */ false);
+          int64_t rebased                    = *reinterpret_cast<int64_t*>(&rebased_seconds);
```
ttnghia, line 829: Do not use this way (dereference a cast pointer), it is not guaranteed to work by the standard and may thrown compiler warning/error. There is a way to get the internal value of a timestamp.

res-life: Fixed:
```diff
-          int64_t rebased                    = *reinterpret_cast<int64_t*>(&rebased_seconds);
+
+          auto const rebased = static_cast<int64_t>(
+            cuda::std::chrono::duration_cast<cudf::duration_s>(rebased_seconds.time_since_epoch())
+              .count());

```

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -261,4 +265,41 @@ struct overflow_checker {
...
+  auto const list_size      = tz_transitions.size();
+
+  auto const transition_times = cudf::device_span<int64_t const>(
+    (to_utc ? tz_instants : utc_instants).data<int64_t>() + tz_transitions.element_offset(0),
+    static_cast<size_t>(list_size));
+
+  auto const it = thrust::upper_bound(
+    thrust::seq, transition_times.begin(), transition_times.end(), epoch_seconds);
+  auto const idx = static_cast<cudf::size_type>(thrust::distance(transition_times.begin(), it));
+  auto const list_offset = tz_transitions.element_offset(idx - 1);
```
ttnghia, line 299: Can `it == transition_times.end()`?

ttnghia (approved):

src/main/cpp/src/datetime_utils.cuh
res-life: Yes, it's intended. If it's end, the following code gets the transition index `end -1` which is not out of bound.

```scala
  auto const it = thrust::upper_bound(
    thrust::seq, transition_times.begin(), transition_times.end(), epoch_seconds);
  auto const idx = static_cast<cudf::size_type>(thrust::distance(transition_times.begin(), it));
  auto const list_offset = tz_transitions.element_offset(idx - 1);
```

E.g.:
```scala
print(java.time.ZoneId.of("Asia/Shanghai").getRules.getTransitions())
```
The output is:
```
Transition[Overlap at 1901-01-01T00:00+08:05:43 to +08:00]
...
Transition[Overlap at 1991-09-15T02:00+09:00 to +08:00]
```
For China time which is bigger than `1991-09-15T02:00+09:00`,  will use the last entry, and get time offset +08:00. So the `it` can be `transition_times.end()`.

PR #3367, opened 2025-05-27
Align kernel to each Spark version for cast string to timestamp

res-life:
Kernel before this PR matches Spark 321 and Spark 321+, but does not match 320.
This PR is to update kernel to match Spark 320.
After this PR, kernel will match Spark versions from 320 to 355.

### 1. diff getZoneId

Details:

For fixed offset timezone, e.g.: +08:00, Java requires hour, minute and second are 2 digits.
For Spark 320, Below `getZoneId` code replaces 1 digit hour to 2 digits hour, but does not replace minute:  h:xx => 0h:xx. Spark 321+ also replaces minute from 1 digits to 2 digits.

Spark 320 only supports: [+-]x:xx, [+-]xx:xx fixed timezones, e.g.:
+1:02
+01:02
UT-1:02
UT+01:02
GMT-1:02
GMT+01:02
...
Spark 320 does not supports: [+-]h:m, [+-]hh:m, but Spark 321+ does.

Spark 320 code:
https://github.com/apache/spark/blob/v3.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala#L53-L56
```scala
  def getZoneId(timeZoneId: String): ZoneId = {
    // To support the (+|-)h:mm format because it was supported before Spark 3.0.
    ZoneId.of(timeZoneId.replaceFirst("(\\+|\\-)(\\d):", "$10$2:"), ZoneId.SHORT_IDS)
  }
```
Spark 321 code :
Below `getZoneId` code replaces 1 digit hour to 2 digits hour, and 1 digit minute to 2 digits minute: x:x => xx:xx
So supports: [+-]x:x, [+-]x:xx, [+-]xx:x, [+-]xx:xx

https://github.com/apache/spark/blob/v3.2.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala#L53-L61
```scala
  def getZoneId(timeZoneId: String): ZoneId = {
    val formattedZoneId = timeZoneId
      // To support the (+|-)h:mm format because it was supported before Spark 3.0.
      .replaceFirst("(\\+|\\-)(\\d):", "$10$2:")
      // To support the (+|-)hh:m format because it was supported before Spark 3.0.
      .replaceFirst("(\\+|\\-)(\\d\\d):(\\d)$", "$1$2:0$3")

    ZoneId.of(formattedZoneId, ZoneId.SHORT_IDS)
  }
```
Note: the second `replaceFirst` can convert "01:2" to "01:02", but can not convert "01:2:03" to "01:02:03", so the "01:2:03" is special case, it's invalid.
Please notice the tailing '$' in the regexp.

#### diff parseTimestampString
https://github.com/apache/spark/blob/v3.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala#L251

https://github.com/apache/spark/blob/v3.2.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala#L256
diff2

### changes

- Add  `is_spark_320` boolean to kernel
- Port Spark 320 code and Spark 321 code
- Refactor `parse_tz_from_sign`
- Add test case

### DIFF
--- a/src/main/cpp/src/CastStringJni.cpp
+++ b/src/main/cpp/src/CastStringJni.cpp
@@ -279,7 +279,8 @@ Java_com_nvidia_spark_rapids_jni_CastStrings_parseTimestampStrings(JNIEnv* env,
                                                                    bool is_default_timezone_dst,
                                                                    long default_epoch_day,
                                                                    jlong timezone_info_column,
-                                                                   jlong transitions_table)
+                                                                   jlong transitions_table,
+                                                                   jboolean is_spark_320)
 {
   JNI_NULL_CHECK(env, input_column, "input column is null", 0);
   JNI_NULL_CHECK(env, timezone_info_column, "timezone info column is null", 0);
@@ -298,7 +299,8 @@ Java_com_nvidia_spark_rapids_jni_CastStrings_parseTimestampStrings(JNIEnv* env,
                                                 is_default_timezone_dst,
                                                 default_epoch_day,
                                                 *tz_info_view,
-                                                *transitions));
+                                                *transitions,
+                                                is_spark_320));
   }
   CATCH_STD(env, 0);
 }
--- a/src/main/cpp/src/cast_string_to_datetime.cu
+++ b/src/main/cpp/src/cast_string_to_datetime.cu
@@ -474,7 +474,7 @@ __device__ bool is_valid_digits(int segment, int digits)
 /**
  * Parse a string with timezone
  */
-template <bool is_spark_320 = false>
+template <bool is_spark_320>
 __device__ RESULT_TYPE parse_timestamp_string(unsigned char const* const ptr,
                                               unsigned char const* ptr_end,
                                               time_zone& tz,
@@ -658,6 +658,7 @@ __device__ RESULT_TYPE parse_timestamp_string(unsigned char const* const ptr,
 /**
  * Parse a string with timezone to a timestamp.
  */
+template <bool is_spark_320>
 struct parse_timestamp_string_fn {
   // inputs
   cudf::column_device_view d_strings;
@@ -703,8 +704,8 @@ struct parse_timestamp_string_fn {
     TS_TYPE just_time    = TS_TYPE::NOT_JUST_TIME;
 
     // parse the timestamp string
-    auto result_type =
-      parse_timestamp_string(str_ptr, str_end_ptr, tz, seconds, microseconds, just_time);
+    auto result_type = parse_timestamp_string<is_spark_320>(
+      str_ptr, str_end_ptr, tz, seconds, microseconds, just_time);
 
     // set result column
     result_types[idx]     = static_cast<uint8_t>(result_type);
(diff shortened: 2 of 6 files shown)


res-life:
build

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -279,7 +279,8 @@ Java_com_nvidia_spark_rapids_jni_CastStrings_parseTimestampStrings(JNIEnv* env,
                                                                    bool is_default_timezone_dst,
                                                                    long default_epoch_day,
                                                                    jlong timezone_info_column,
-                                                                   jlong transitions_table)
+                                                                   jlong transitions_table,
+                                                                   jboolean is_spark_320)
```
ttnghia, line 283: Since the changes target Spark 320 to 355, should this?
```suggestion
                                                                   jboolean is_from_spark_320)
```

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -651,52 +512,178 @@ __device__ RESULT_TYPE parse_timestamp_string(unsigned char const* const ptr,
...
+                          segments[6]) ||
+      !is_valid_tz(tz)) {
+    return RESULT_TYPE::INVALID;
+  }
 
-  if (negative_year_sign) { ts.year = -ts.year; }
-  return to_long_check_max(ts, seconds, microseconds);
+  seconds =
+    to_epoch_seconds(segments[0], segments[1], segments[2], segments[3], segments[4], segments[5]);
+  microseconds = segments[6];
```
ttnghia, line 679: Similarly, using enum to index should be more readable.

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -893,24 +881,45 @@ std::unique_ptr<cudf::column> parse_ts_strings(cudf::strings_column_view const&
...
+                                       *d_tz_info,
+                                       fixed_transitions,
+                                       parsed_result_type_col->mutable_view().begin<uint8_t>(),
+                                       parsed_utc_seconds_col->mutable_view().begin<int64_t>(),
+                                       parsed_utc_microseconds_col->mutable_view().begin<int32_t>(),
+                                       parsed_tz_type_col->mutable_view().begin<uint8_t>(),
+                                       parsed_tz_fixed_offset_col->mutable_view().begin<int32_t>(),
+                                       is_DST_tz_col->mutable_view().begin<uint8_t>(),
+                                       parsed_tz_index_col->mutable_view().begin<int32_t>()});
+  }
```
ttnghia, line 922: How about extracting into a template function like this?
```
template<typename F, typename... Args>
void parse_strings_column(num_rows, stream, F&& f, Args&&... arg) {
thrust::for_each_n(
      rmm::exec_policy_nosync(stream),
      thrust::make_counting_iterator(0),
      num_rows,
      F{td::forward<Ts>(args)...});
...

if (is_from_spark_320) { parse_strings_column(num_rows, stream, parse_timestamp_string_fn<true>{}); }
else { ...}
```

src/main/cpp/src/CastStringJni.cpp
ttnghia: If this applies to Spark 320 to 355 then should it be `is_from_spark_320`?

PR #3375, opened 2025-05-28
Adopt to changes in cudf JNI exception handling

ttnghia:
The exception handling in cudf jni is changed to prepare for support capturing native stacktrace when exception being thrown. That is breaking changes and this PR fixes it.

No new feature/implementation is added.

Depends on:
 * https://github.com/rapidsai/cudf/pull/18983

### DIFF
--- a/src/main/cpp/src/HostTableJni.cpp
+++ b/src/main/cpp/src/HostTableJni.cpp
@@ -26,8 +26,8 @@
 #include <cuda_runtime_api.h>
 
 #include <algorithm>
-#include <iterator>
 #include <memory>
+#include <numeric>
 #include <stdexcept>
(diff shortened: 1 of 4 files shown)


src/main/cpp/src/HostTableJni.cpp
```diff
@@ -26,8 +26,8 @@
 #include <cuda_runtime_api.h>
 
 #include <algorithm>
-#include <iterator>
 #include <memory>
+#include <numeric>
```
ttnghia, line 30: Change to fix compile error. The error is due to not including the necessary header here but such header was included indirectly somewhere else.

pxLi (approved):

PR #3384, opened 2025-05-29
Update `pow_ten` function to properly handle unsupported exponents

nartal1:
This is a small PR to fix pow_ten function to properly handle unsupported exponents

The `pow_ten` function in `decimal_utils.cu` currently returns an uninitialized `chunked256()` object when an exponent outside the supported range (0-76) is passed. 
Replace the return statement with `CUDF_UNREACHABLE("exp exceeds supported value")` to properly indicate when an unsupported exponent is used.


ttnghia:
build

PR #3393, opened 2025-05-29
add doc for Dedicated Threads

binmahone:
we're writing more background threads nowadays. this is a PR to add some docs reminding code contributors try to avoid making background threads "dedicated" to any task.


docs/memory_management.md
```diff
@@ -52,3 +52,10 @@ The selected thread transitions its state to `THREAD_BUFN_THROW`. Any threads th
...
+
+From the view of the OOM state machine, each task has one or more "dedicated threads", along with zero or more "pool threads". When checking whether a task is blocked, OOM state machine is lenient on dedicated threads (only require any one of the dedicated threads to be blocked), but stringent on pool threads (all pool threads must be blocked). Being treated leniently is not always a good thing, it increases the chance of being mistakenly identified as a block task, thus causing unnecessary deadlock resolution. So we don't want a thread to be treated as a dedicated thread unless it is really necessary. There are two ways of avoiding a thread being treated as a dedicated thread:
+
+1. Avoid calling TaskContext.setTaskContext() in the current thread, this will prevent OOM state machine connecting the current thread to the task as a dedicated thread. 
+2. If you have to set TaskContext, then it's also a good idea to proactively register thread itself as a pool thread instead of a dedicated thread. E.g. https://github.com/binmahone/spark-rapids/blob/581c1726b419ba93efb4302516370973ec465f52/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala#L2058
```
jihoonson, line 61: Can we use a link to the nvidia repo instead of yours?

jihoonson, line 61: Also this could be better to read 

```suggestion
2. If you have to set TaskContext, then it's also a good idea to proactively register thread itself as a pool thread instead of a dedicated thread. An example code can be found [here](https://github.com/binmahone/spark-rapids/blob/581c1726b419ba93efb4302516370973ec465f52/sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOrcScan.scala#L2058) for registering threads.
```

docs/memory_management.md
```diff
@@ -43,7 +43,7 @@ Shuffle threads have the highest priority to avoid priority inversion as the tas
 
 ### Deadlock Resolution
 
-Deadlocks occur when every active task has at least one thread blocked on memory allocation, either directly during its execution or indirectly by dependencies on shuffle blocked.
+Deadlocks occur when every active task has at least one dedicated thread blocked on memory allocation, and all of the pool threads working on that task are also blocked.
```
jihoonson, line 46: Question: does the deadlock occur when both conditions are met that there is one or more blocked dedicated threads _and_ all pooled threads are blocked?

docs/memory_management.md
binmahone: sure , good idea

binmahone:
@jihoonson can you pls check again?

binmahone:
build

docs/memory_management.md
```diff
@@ -52,3 +52,10 @@ The selected thread transitions its state to `THREAD_BUFN_THROW`. Any threads th
 After the rollback, all data of the thread will be spillable and the thread will be blocked before allocating more GPU memory until enough memory is freed up for other threads.
 2) If all threads are in the `THREAD_BUFN` state, the lowest priority thread is selected to split its data first and then retry.
 The selected thread transitions its state to `THREAD_SPLIT_THROW` and throws an exception to initiate the split-and-retry process.
+
+### Dedicated Threads vs. Pool Threads
+
+From the view of the OOM state machine, each task has one or more "dedicated threads", along with zero or more "pool threads". When checking whether a task is blocked, OOM state machine is lenient on dedicated threads (only require any one of the dedicated threads to be blocked), but stringent on pool threads (all pool threads must be blocked). Being treated leniently is not always a good thing, it increases the chance of being mistakenly identified as a block task, thus causing unnecessary deadlock resolution. So we don't want a thread to be treated as a dedicated thread unless it is really necessary. There are two ways of avoiding a thread being treated as a dedicated thread:
```
jihoonson, line 58: Perhaps this could be a bit easier to read.

```suggestion
### Dedicated Threads vs. Pool Threads

From the perspective of the OOM state machine, each task has one or more "dedicated threads" along with zero or more "pool threads". The OOM state machine treats these threads differently when checking the task OOM state. A task is considered as blocked if one or more dedicated threads are blocked _and_ all pooled threads are also blocked. As such, having more dedicated threads can increase the chance for the task to be considered as blocked. It is important to register threads as pooled threads unless it is absolutely necessary to avoid unnecessary deadlock resolution.

There are two ways of avoiding a thread being treated as a dedicated thread:
```

jihoonson (approved):
LGTM. Left a minor suggestion. Thanks @binmahone for making this improvement.

PR #3394, opened 2025-05-29
Check offset timezone in cast string to timestamp

res-life:
## bug
Does not check fixed timezone for Spark 320
fixed timezone checks:
- offset range is: [-18:00:00, 18:00:00], so hour > 18 is invalid
- minute/second range: : [0, 59]

## fix
Add the above fixed timezone checks

### DIFF
--- a/src/main/cpp/src/cast_string_to_datetime.cu
+++ b/src/main/cpp/src/cast_string_to_datetime.cu
@@ -267,9 +267,9 @@ __device__ time_zone parse_tz_from_sign(
     }
   }
 
-  if (minute > 59 || second > 59) { return make_invalid_tz(); }
+  // check hour, minute and second; max offset is 18:00:00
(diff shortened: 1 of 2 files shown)


thirtiseven (approved):

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -267,9 +267,9 @@ __device__ time_zone parse_tz_from_sign(
     }
   }
 
-  if (minute > 59 || second > 59) { return make_invalid_tz(); }
+  // check hour, minute and second; max offset is 18:00:00
+  if (hour > 18 || minute > 59 || second > 59) { return make_invalid_tz(); }
```
res-life, line 271: here hour, minute and second are >=0,   no need to check: < 0

Issue #2506, opened 2024-10-15
Use native test common lib to reduce compilation times


jlowe:
#2505 added the cudftestutil_impl dependency to each test which fixes the build, but it causes redundant compilation of the cudftestutil_impl sources.  It would save some build time on the native tests if we used a common test library to house these sources.  See rapidsai/cuspatial#1475 for one example on how to do this.

PR #3397, opened 2025-05-29
Use a common object library to avoid repeated compilation when building C++ tests

ttnghia:
Currently, building each test executable involves compiling all the source files from `cudf::cudftestutil_impl`. That is because `cudf::cudftestutil_impl` is just an interface library that does not compile but just provides a list of source files. This  PR adds an object library that compiles these source files, allowing them to be compiled just once instead of compiled for every build of a test executable.

Closes https://github.com/NVIDIA/spark-rapids-jni/issues/2506.

---
#### Before:
```
cat build.ninja | grep column_utilities.cu.o: | wc -l
17
```
That means the file is compiled 17 times.

#### After:
```
cat build.ninja | grep column_utilities.cu.o: | wc -l
1
```

### DIFF
--- a/src/main/cpp/tests/CMakeLists.txt
+++ b/src/main/cpp/tests/CMakeLists.txt
@@ -17,6 +17,20 @@
 ###################################################################################################
 # - compiler function -----------------------------------------------------------------------------
 
+# We need to include the source code defined in cudftestutil_impl, but it is an interface library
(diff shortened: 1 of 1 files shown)


ttnghia:
build

hyperbolic2346 (approved):

Issue #3401, opened 2025-05-30
[BUG] DB35 and Spark400 has a bug in casting string to timestamp for pattern: spaces + Thh:mm:ss


res-life:
**Describe the bug**
Caused by Spark issue: https://issues.apache.org/jira/browse/SPARK-52351
It's actually a Spark bug. But we plan to align to Spark.

There is a difference between Spark350 and Spark400 when handing pattern:  space followed by a `T`
### Spark350:

``` log
scala> spark.sql("select cast(' T23:17:50' as timestamp)").show()
+-----------------------------+
|CAST( T23:17:50 AS TIMESTAMP)|
+-----------------------------+
|          2025-05-30 23:17:50|
+-----------------------------+
```

### Spark400:

``` log
spark.conf.set("spark.sql.ansi.enabled", false)
scala> spark.sql("select cast(' T23:17:50' as timestamp)").show()
+-----------------------------+
|CAST( T23:17:50 AS TIMESTAMP)|
+-----------------------------+
|                         NULL|
+-----------------------------+
```

Seems DB35 ported Spark400 code.

#### Note
The behavior on Spark 400 is a bug

### Spark bug explain

[Code link](https://github.com/apache/spark/blob/v4.0.0/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/SparkDateTimeUtils.scala#L450-L468)
```scala
    var j = getTrimmedStart(bytes)          // ============== here j is not zero after trimed left
    val strEndTrimmed = getTrimmedEnd(j, bytes)

    if (j == strEndTrimmed) {
      return (Array.empty, None, false)
    }

    var digitsMilli = 0
    var justTime = false
    var yearSign: Option[Int] = None
    if (bytes(j) == '-' || bytes(j) == '+') {
      yearSign = if (bytes(j) == '-') Some(-1) else Some(1)
      j += 1
    }
    while (j < strEndTrimmed) {
      val b = bytes(j)
      val parsedValue = b - '0'.toByte
      if (parsedValue < 0 || parsedValue > 9) {
        if (j == 0 && b == 'T') {
          justTime = true  // ============== never reach to here if trimed left and pattern is Thh:mm:ss
          i += 3
        } else if (i < 2) {
```

res-life:
Spark issue link: https://issues.apache.org/jira/browse/SPARK-52351
Please keep this Spark-Rapids issue open util Spark fixed.

res-life:
[...]

PR #3402, opened 2025-05-30
Fix bug in casting string to timestamp: Spark400+ and DB35 do not support pattern: spaces + Thh:mm:ss

res-life:
contributes to https://github.com/NVIDIA/spark-rapids-jni/issues/3401

It's follow-up for https://github.com/NVIDIA/spark-rapids/pull/12598

## bug
Spark400+ and DB35 do not support pattern: spaces + Thh:mm:ss, e.g.:
cast('    T00:00:00') fails on Spark400 while works well on Spark35x
The changed PR https://github.com/apache/spark/pull/44463 introduced this change.
For more details, refer to https://github.com/NVIDIA/spark-rapids-jni/issues/3401

## fix
Although it's a bug in Spark, but we decided to align to Spark versions.
- Add version classes: Version, Platform.
- Pass in platform, major, minor, patch info into Kernel.
- If it's Spark400+ and DB35, fails to parse timestamp like: '   Thh:mm:ss'


src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -1295,4 +1293,51 @@ void castStringToTimestampOnGpuForSpark320() {
...
+      } else {
+        expectedTS.add(null);
+      }
+    }
+    // test spark 400+
+    Version v400 = new Version(Platform.SPARK, 4, 0, 0);
+    try (ColumnVector inputCv = ColumnVector.fromStrings(input.toArray(new String[0]));
+        // test spark 320
+        ColumnVector actual = CastStrings.toTimestamp(
+            inputCv, "Z", /* ansi */ false, v400);
```
res-life, line 1326: Here use new API to fail for:  cast('   T00:00:00' as timestamp)

res-life:
build

thirtiseven (approved):

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -697,8 +705,11 @@ __device__ result_type parse_timestamp_string(unsigned char const* const ptr,
 /**
  * Parse a string with timezone to a timestamp.
  */
-template <bool is_spark_320>
 struct parse_timestamp_string_fn {
+  // version info
+  bool is_spark_320;
```
sperlingxx, line 710: NIT: IMHO, it is better to use template variable to dispatch versions to avoid branch operation being executed in each CUDA thread.  Perhaps, we can handle the version dispatch more elegantly just like type dispatcher of CUDF (using template functions).

Of course, current implementation looks good enough for me. Maybe we can mark it as a TODO action?

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -272,15 +272,19 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromInteger
...
+  jlong input_column,
+  int default_timezone_index,
+  bool is_default_timezone_dst,
+  long default_epoch_day,
+  jlong timezone_info_column,
+  jlong transitions_table,
+  int platform,
+  int majorVersion,
+  int minorVersion,
+  int fixVersion)
```
ttnghia, line 287: Reformat?

src/main/cpp/src/version.hpp
```diff
@@ -0,0 +1,33 @@
...
+#pragma once
+
+namespace spark_rapids_jni {
+
+// NOTE: MUST keep sync with Platform.java
+constexpr int PLATFORM_SPARK      = 0;
+constexpr int PLATFORM_DATABRICKS = 1;
+constexpr int PLATFORM_CLOUDERA   = 2;
+
+enum class platform_type { SPARK, DATABRICKS, CLOUDERA };
```
ttnghia, line 26: Let's not using separate variables like that, it is redundant.
```suggestion
// NOTE: MUST keep sync with Platform.java
enum class platform_type { SPARK = 0, DATABRICKS, CLOUDERA };
```

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -1113,17 +1112,29 @@ std::unique_ptr<cudf::column> parse_timestamp_strings(cudf::strings_column_view
...
+                                                      int fix_version,
                                                       rmm::cuda_stream_view stream,
                                                       rmm::device_async_resource_ref mr)
 {
+  platform_type const platform = spark_rapids_jni::get_platform_type(platform_ordinal);
+  int v                        = major_version * 100 + minor_version * 10 + fix_version;
+  bool is_spark_320            = (platform == platform_type::SPARK && v == 320);
+  bool is_spark_400_or_later   = (platform == platform_type::SPARK && v >= 400);
+  int v_for_db                 = major_version * 100 + minor_version;
+  bool is_db_14_3_or_later     = (platform == platform_type::DATABRICKS && v_for_db >= 1403);
```
ttnghia, line 1127: For better portability and code reuse, let add a custom class in `version.hpp/cpp` for doing all those parsing:
```
struct spark_system {
  spark_system(platform, major, minor, fix) {...}
  platform_type platform() const {...}
  int spark_version() const { return _spark_version; }
private:
 platform_type platform;
 int _spark_version; // can be 320
}
```

src/main/cpp/src/version.hpp
ttnghia: In addition:
 * `platform_type` seems a bit confusing. Maybe `spark_platform`?\
 * `SPARK` seems not a good name for the opensource Spark. Maybe `APACHE` or `OPENSOURCE`?

src/main/java/com/nvidia/spark/rapids/jni/Version.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.nvidia.spark.rapids.jni;
+
+public class Version {
+  private final int platformOrdinal;
+  private final int major;
+  private final int minor;
+  private final int patch;
+
+  public Version(Platform platform, int major, int minor, int patch) {
```
ttnghia, line 25: In Java the last number is `patch` but in C++ it is `fix`. Let's make consistent. I think `patch` is better than `fix`.

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -1113,17 +1112,29 @@ std::unique_ptr<cudf::column> parse_timestamp_strings(cudf::strings_column_view
...
+                                                      int minor_version,
+                                                      int fix_version,
                                                       rmm::cuda_stream_view stream,
                                                       rmm::device_async_resource_ref mr)
 {
+  platform_type const platform = spark_rapids_jni::get_platform_type(platform_ordinal);
+  int v                        = major_version * 100 + minor_version * 10 + fix_version;
+  bool is_spark_320            = (platform == platform_type::SPARK && v == 320);
+  bool is_spark_400_or_later   = (platform == platform_type::SPARK && v >= 400);
+  int v_for_db                 = major_version * 100 + minor_version;
```
ttnghia, line 1126: Why for `v_for_db`, `minor_version` is not multiply by `10`?

res-life: For DB, its major version is 2 digits, so it has little probablity that minor and fix versions are also 2 digits.
So here use multiple by 100, it's similar to multiple by 10.

ttnghia: Alright, I think we should better store all the major and minor/patch version numbers. Then when we query, we can do like this:
```
struct spark_system {
....
private:
 int major, minor, patch;
};

auto const is_spark_320 = sys.platform() == platform_type::XXX && sys.major() == 3 && sys.minor() = 2 && sys.patch() == 0;

auto const is_db_14_3_or_later = sys.platform() == platform_type::DATABRICKS && sys.major() == 14 && sys.
minor() = 3;

etc.
```

src/main/cpp/src/version.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+  bool isDatabricks_14_3_or_later() const
+  {
+    return (isDatabricks() && (major > 14 || (major == 14 && minor >= 3)));
+  }
+
+ private:
+  spark_platform_type spark_platform() const { return static_cast<spark_platform_type>(platform); }
+
+  int platform, major, minor, patch;
```
ttnghia, line 65: Sorry I was wrong when suggested to store platform as number. It is more convenient to be stored in the enum type.
```suggestion
  spark_platform_type platform_type;
  int major, minor, patch;
```

src/main/cpp/src/cast_string_to_datetime.cu
```diff
@@ -549,11 +551,17 @@ __device__ result_type parse_timestamp_string(unsigned char const* const ptr,
     ++j;
   }
 
+  // Spark400+ and DB14.3+: do not support pattern: spaces + Thh:mm:ss
+  // Refer to https://github.com/NVIDIA/spark-rapids-jni/issues/3401
+  // Refer to https://issues.apache.org/jira/browse/SPARK-52351
+  // Check if Spark is Spark400+ or DB14.3+ and has left spaces
+  bool match_issue_52351 = is_spark_400_or_later_or_db_14_3_or_later && pos > 0;
```
ttnghia, line 558: I feel a bit awkward when hardcoding the issue into the code but don't have a better idea for it :(

res-life:
build

ttnghia (approved):

PR #3426, opened 2025-06-05
[DOC] explain more to convince explicitly using pool thread [skip ci]

binmahone:
add more doc to convince explicitly using pool thread


binmahone:
build

winningsix (approved):

Issue #3451, opened 2025-06-13
[BUG] Subtraction of BigDecimal is wrong.


res-life:
**Describe the bug**
`0.964808037102759 - 1` results in wrong result, should be negative value.
```
0.964808037102759 - 1 = 0.964809567597735
```

GPU is Quadro RTX 8000
Code branch: 25.08
CUDA Version: 12.2

This is from https://github.com/NVIDIA/spark-rapids/issues/12926

**Steps/Code to reproduce bug**
```java

  @Test
  void subTest1() {
    String lhString = "0.964808037102759";
    String rhString = "1";
    try (
        ColumnVector lhs = makeDec128Column(lhString);
        ColumnVector rhs = makeDec128Column(rhString);

        ColumnVector expectedValid = ColumnVector.fromBooleans(true);
        Table result = DecimalUtils.subtract128(lhs, rhs, -15)) {

      System.out.println("my debug: lhs precision " + lhs.getType().getDecimalMaxPrecision());
      System.out.println("my debug: lhs scale " + lhs.getType().getScale());
      System.out.println("my debug: rhs precision " + rhs.getType().getDecimalMaxPrecision());
      System.out.println("my debug: rhs scale " + rhs.getType().getScale());

      String ret = result.getColumn(1).copyToHost().getBigDecimal(0).toString();
      System.out.println(String.format("my debug: %s - %s = %s", lhString, rhString, ret));

      boolean isOverflow = result.getColumn(0).copyToHost().getBoolean(0);
      System.out.println("my debug: isOverflow = " + isOverflow);
    }
  }

```

Get output:
```

my debug: lhs precision 38
my debug: lhs scale -15
my debug: rhs precision 38
my debug: rhs scale 0

my debug: 0.964808037102759 - 1 = 0.964809567597735
my debug: isOverflow = false
```

PR #3452, opened 2025-06-13
Fix overflow in decimal scaling operation

ttnghia:
When doing decimal scaling, the scale factor (256bit) is incorrectly cast into `int` value, which cannot hold the 256 bit value. This fixes it, keeping the 256bit value without casting for further operation. A new Java test is added to verify this.

This also fixes a few memory leak issues in Java tests, and renames variables into a more accurate meaning.

Closes https://github.com/NVIDIA/spark-rapids-jni/issues/3451.

### DIFF
--- a/src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/DecimalUtilsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2024, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


ttnghia:
build

jihoonson (approved):
LGTM

src/main/cpp/src/decimal_utils.cu
```diff
@@ -535,13 +535,14 @@ __device__ chunked256 set_scale_and_round(chunked256 data, int old_scale, int ne
 {
   if (old_scale != new_scale) {
     if (new_scale < old_scale) {
-      int const raise      = old_scale - new_scale;
-      int const multiplier = pow_ten(raise).as_128_bits();
-      data                 = multiply(data, chunked256(multiplier));
+      int const raise       = old_scale - new_scale;
```
gerashegalov, line 538: why not `auto const raise` for consistency?

PR #3459, opened 2025-06-17
Use shared auto-merge workflow [skip ci]

pxLi:
follow-up of https://github.com/NVIDIA/spark-rapids-common/pull/40, target 25.06 to have some sanity check after merge as workflow update only

1. Migrated automerge strategy to the share action in the common repo
2. Removed deprecated script in JNI repo

This has been verified in forked repo, I will do post verification after merge and revert this change is any issue


pxLi:
build

PR #3463, opened 2025-06-17
Implement Cookie serialization format

ttnghia:
This implements `Cookie` serialization format, a fast and efficient data serialization that targets efficiency for both data serialization/deserialization and disk IO. From the input as an array of host buffers given as byte arrays, the serializer simply compresses these byte arrays (using CPU thread pool) and assembles the compressed data along with other metadata into one output byte array for efficient disk IO. Deserialization is performed in the reversed way.

Contribute to https://github.com/NVIDIA/spark-rapids/issues/12509.


nvauto:

PR #3466, opened 2025-06-18
Add JNI api for cudaDevAttrIntegrated

zpuller:
adds a java api for querying the `cudaDevAttrIntegrated` attribute, which specifies whether the host and device are sharing an integrated physical memory


revans2 (approved):
I agree this would be better in CUDF, but we can do that as follow on work in 25.12

Issue #3299, opened 2025-05-16
[BUG] ai.rapids.cudf.TableTest in cuda12 hang there forever in CI intermittently


pxLi:
**Describe the bug**
Previously ai.rapids.cudf.TableTest case run would cost only 2-3 mins,

Currently in nightly CI, it hangs for over 1 hour and timeout.

spark-rapids-jni_nightly-dev, run: 1103, 1104 (**cuda12 only**, repro on A100 and L4)
commit: db0c61fad99d07a5e55f2a938270a052cf63b20d, submodule cudf ref: https://github.com/rapidsai/cudf/commit/6fef829f0f04704d816a5f1b24d22235cb57523d

cuda11 and arm64 works fine

```
[INFO] Running ai.rapids.cudf.TableTest
```

**Steps/Code to reproduce bug**
with internal `<URM>/sw-spark-docker/plugin-jni:rockylinux8-cuda12.8.0-blossom`

**Expected behavior**

**Environment details (please complete the following information)**
 - Environment location: [Standalone, YARN, Kubernetes, Cloud(specify cloud provider)]
 - Spark configuration settings related to the issue

**Additional context**

revans2:
I think I was able to reproduce this hang. It happened only when I built the entire thing using the build script  

```
rm -rf target/surefire-reports/* ./sanitizer_for_pid_*.log; DOCKER_OPTS='-p 5005:5005' CUDA_VERSION=12.8.0 PARALLEL_LEVEL=18 USE_GDS=ON USE_SANITIZER=ON BUILD_FAULTINJ=ON ./build/run-in-docker ./ci/nightly-build.sh 2>&1 | tee log.txt
```

I did this on an a6000 GPU. For cuda 11 the test takes about 92 seconds to complete, and 60 seconds on my desktop, but with cuda12 it is now at 20 mins, and has still not finished. This matches with the time it took for the test to time out, which from the logs shows it took 19.6 mins before the timeout happened.

revans2:
So I was able to get some stack traces on where it is hung, and it is happening here.

https://github.com/rapidsai/cudf/blob/ada370f5a366c25e062ef1004645f5ced02f1e55/java/src/test/java/ai/rapids/cudf/TableTest.java#L9224

It appears to be stuck trying to run a shell command to check if `isSetsidSupported`

```
[...]

PR #3474, opened 2025-06-25
Revert "scl default" changes made previously given CI failures

abellina:
Closes https://github.com/NVIDIA/spark-rapids-jni/issues/3299

The issue here looks to be that we are enabling a devtoolset from within a shell that already has devtoolset enabled. TableTest, in particularly, will shell out (invoking bash, see https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Shell.java#L852), and this is the stack where we hang. 

It looks that enabling toolsets should be done once, and this nested case may be error prone. I have reverted the change locally and cannot reproduce the hang anymore. I believe this to be the root cause of our hangs in CI.

### DIFF
--- a/build/run-in-docker
+++ b/build/run-in-docker
@@ -20,6 +20,11 @@
 
 set -e
 
+# Set this environment variable to ON to build the docker image for local development,
+# with gcc-toolset enabled by default.
+# This is necessary especially if the docker container is used for local development with an IDE.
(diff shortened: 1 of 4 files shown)


ttnghia:
build

revans2 (approved):

abellina:
build

ttnghia:
Build shows "fail" as there were two "build" commands so the first one was aborted, but the second one is still running so let's wait...

PR #3490, opened 2025-07-01
Implement map_zip_with functionality

sdrp713:
**indices_of**
_Purpose_: Finds the indices of search keys within corresponding search value lists
_Input_: Two lists columns - one containing keys to search for, another containing values to search within
_Output_: A column containing the first matching index for each key in its corresponding value list (or 100 if no match found)
_Algorithm_: Performs cross-product comparisons between keys and values, then uses segmented reduction to find the minimum (first) match for each list

**map_zip**
_Purpose_: Merges two map columns by finding the union of keys and gathering corresponding values
_Input_: Two lists columns representing maps (each containing key-value pairs)
_Output_: A pair of columns containing the gathered values from both input maps for the union of all keys
_Algorithm_:
- Extracts keys and values from both input maps
- Finds the union of all unique keys
- Determines where each key appears in both maps using indices_of
- Gathers the corresponding values using the calculated indices

### DIFF
--- a/src/main/cpp/src/map_zip_with_utils.cu
+++ b/src/main/cpp/src/map_zip_with_utils.cu
@@ -73,6 +73,8 @@ std::unique_ptr<column> generate_labels(
  * 1. a `search_keys` column (keys being searched *for*)
  * 2. a `search_space` column (the space being searched)
  *
+ * Neither list should have repeated keys within each list row.
+ *
  * The function's output is a LISTS column, where each row[i] contains the list
  * indices of all matches between `search_keys[i]` and `search_space[i]`.
  * i.e.
@@ -94,7 +96,9 @@ std::unique_ptr<column> generate_labels(
  * results[i]        == { 1, 3, 4, ∅};
  *
  * The results column has as many (list) rows as search_keys or search_space.
- * Note that each result (list) row has as many indices as unique values in search_space.
+ * Note that each result (list) row has as many indices as values in the search_keys row.
+ * Also note that this function assumes that neither search_keys nor search_space
+ * has repetitions in the keys.
  *
  * @param search_keys Column containing lists of keys to search for
  * @param search_values Column containing lists of values to search through (i.e. search space)
@@ -113,7 +117,6 @@ std::unique_ptr<column> indices_of(
     cudf::get_current_device_resource_ref())  // Memory resource for allocations
 {
   /*
-  *
   * Example using two list columns and indexing
   *
   * search_keys     | search_values


src/main/cpp/tests/map_zip_with_utils.cpp
```diff
@@ -0,0 +1,483 @@
...
+      {48, 27, 3, 4, 25, 6, 31, 351, 9, 10, 351, 12}, {1, 1, 0, 0, 1, 0, 1, 1, 0, 0, 1, 0}};
+    auto const expected_vals2 = cudf::test::fixed_width_column_wrapper<size_type>{
+      {5, 6, 12, 35, 1, 25, 3, 3, 31, 351, 2, 351}, {0, 0, 1, 1, 0, 1, 0, 0, 1, 1, 0, 1}};
+    auto results = spark_rapids_jni::map_zip(cudf::lists_column_view(*list_col1),
+                                             cudf::lists_column_view(*list_col2));
+    CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT(
+      expected_vals1, cudf::lists_column_view(*results.first).child(), verbosity);
+    CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT(
+      expected_vals2, cudf::lists_column_view(*results.second).child(), verbosity);
+  }
```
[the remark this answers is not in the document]
sdrp713: Yeah that was my final plan. I was gonna take the pair of values thats currently being returned and pack them into a struct. Then I would pack the (v1, v2) pair along with the search keys to make the final map k : v1, v2. For right now I just had my function return a pair of v1, v2 columns for easier testing

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+{
+  auto labels = make_numeric_column(
+    data_type(type_to_id<size_type>()), n_elements, cudf::mask_state::UNALLOCATED, stream, mr);
+  auto const labels_begin = labels->mutable_view().template begin<size_type>();
+  cudf::detail::label_segments(
+    input.offsets_begin(), input.offsets_end(), labels_begin, labels_begin + n_elements, stream);
+  return labels;
+}
+
+[[maybe_unused]] std::unique_ptr<column> indices_of(
```
mythrocks, line 48: Maybe this belongs in the unnamed namespace, if it's not used anywhere beyond this .cu.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+  auto map1_indices = indices_of(search_keys_list, cudf::lists_column_view(map1_keys), stream, mr);
+  auto map1_indices_list = make_lists_column(search_keys_list.size(),
+                                             std::make_unique<column>(search_keys_list.offsets()),
+                                             std::move(map1_indices),
+                                             0,
+                                             rmm::device_buffer{0, stream, mr});
+
+  // Find the indices of each key in the second map
+  // This tells us where each key appears in the second map, or if it doesn't exist
+  auto map2_indices = indices_of(search_keys_list, cudf::lists_column_view(map2_keys), stream, mr);
```
mythrocks, line 257: `map2_indices` is a temp column.  We should be using the default mr here.
One option to keep things simple is to use pass the default mr explicitly as the argument to `indices_of()`.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+
+  // Use segmented reduction to find the minimum value (first match) for each list
+  // This gives us the first matching value index for each row, or Out of Bounds index if no match found
+  auto offsets_span = cudf::device_span<cudf::size_type const>(list_sizes_offsets->view());
+  auto results =
+    cudf::segmented_reduce(compare_results->view(),
+                           offsets_span,
+                           *cudf::make_min_aggregation<cudf::segmented_reduce_aggregation>(),
+                           cudf::data_type{cudf::type_to_id<size_type>()},
+                           cudf::null_policy::EXCLUDE);
```
mythrocks, line 193: Definitely use the user-specified mr for this call.  This is the actual output.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+  // Generate labels to map each key back to its corresponding list index
+  // This helps us know which list each key belongs to
+  auto keys_labels = generate_labels(search_keys, num_keys, stream, mr);
+
+  // For each key, get the size of the associated values list
+  // This tells us how many values we need to compare against for each key
+  auto associated_list_sizes = thrust::make_transform_iterator(
+    keys_labels->view().begin<size_type>(),
+    [list_sizes_ptr = list_sizes->view().begin<size_type>(), num_lists] __device__(
+      auto const offset_val) { return offset_val >= num_lists ? 0 : list_sizes_ptr[offset_val]; });
```
mythrocks, line 78:
```suggestion
      auto const key_label) { return offset_val >= num_lists ? 0 : list_sizes_ptr[offset_val]; });
```

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+
+  // Calculate cumulative offsets for the associated list sizes
+  // This gives us the starting position for each key's value comparisons
```
mythrocks, line 81: Consider including the ASCII art and/or examples here.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+  auto list_sizes = cudf::lists::count_elements(search_values, stream, mr);
+  auto keys_sizes = cudf::lists::count_elements(search_keys, stream, mr);
+
+  // Generate labels to map each key back to its corresponding list index
+  // This helps us know which list each key belongs to
+  auto keys_labels = generate_labels(search_keys, num_keys, stream, mr);
+
+  // For each key, get the size of the associated values list
+  // This tells us how many values we need to compare against for each key
+  auto associated_list_sizes = thrust::make_transform_iterator(
```
mythrocks, line 75: The name doesn't quite convey what this is for.

`search_key_to_num_search_values` or something might be clearer.

mythrocks (changes requested):

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+    keys_labels->view().begin<size_type>(),
+    [list_sizes_ptr = list_sizes->view().begin<size_type>(), num_lists] __device__(
+      auto const offset_val) { return offset_val >= num_lists ? 0 : list_sizes_ptr[offset_val]; });
+
+  // Calculate cumulative offsets for the associated list sizes
+  // This gives us the starting position for each key's value comparisons
+  auto list_sizes_offsets = make_numeric_column(
+    data_type(type_to_id<size_type>()), num_keys + 1, cudf::mask_state::UNALLOCATED, stream, mr);
```
mythrocks, line 83:
```suggestion
    data_type(type_to_id<size_type>()), num_keys + 1, cudf::mask_state::UNALLOCATED, stream);
```

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+                         associated_list_sizes,
+                         associated_list_sizes + num_keys + 1,
+                         d_list_sizes_offsets);
+
+  // Calculate the total number of comparisons needed for each row
+  // For each list, we need: number_of_keys * number_of_values comparisons
+  auto total_compares_per_row = thrust::make_transform_iterator(
+    thrust::make_counting_iterator(0),
+    [keys_sizes = keys_sizes->mutable_view().template data<size_type>(),
+     list_sizes = list_sizes->mutable_view().template data<size_type>(),
```
mythrocks, line 95: Am I reading this right?  These shouldn't be `mutable_view`, right?  These are only inputs.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+  // Get the number of lists (rows) in the input columns
+  auto const num_lists = search_values.size();
+
+  // Extract the child columns containing the actual key and value data
+  auto const all_keys   = search_keys.child();
+  auto const num_keys   = all_keys.size();
+  auto const all_values = search_values.child();
+
+  // Calculate the number of elements in each list for both keys and values
+  auto list_sizes = cudf::lists::count_elements(search_values, stream, mr);
```
mythrocks, line 66: I had to look further down to understand which "list" this is a reference to.

This should mention `search_space` in the variable name.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+      return offset_val >= num_lists ? 0 : keys_sizes[offset_val] * list_sizes[offset_val];
+    });
+  // Sum up all comparisons across all rows
+  auto total_compares = thrust::reduce(
+    rmm::exec_policy(stream), total_compares_per_row, total_compares_per_row + num_lists);
+
+  // Create an index array that maps each comparison to its corresponding key
+  // This tells us which key we're comparing in each comparison operation
+  auto key_index = make_numeric_column(
+    data_type(type_to_id<size_type>()), total_compares, cudf::mask_state::UNALLOCATED, stream, mr);
```
mythrocks, line 106:
```suggestion
    data_type(type_to_id<size_type>()), total_compares, cudf::mask_state::UNALLOCATED, stream);
```

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+  cudf::detail::label_segments(list_sizes_offsets->view().begin<size_type>(),
+                               list_sizes_offsets->view().end<size_type>(),
+                               d_key_index,
+                               d_key_index + total_compares,
+                               stream);
+
+  // Calculate offsets for the total comparisons per row
+  // This gives us the starting position for each row's comparisons
+  auto total_compares_offsets = make_numeric_column(
+    data_type(type_to_id<size_type>()), num_lists + 1, cudf::mask_state::UNALLOCATED, stream, mr);
```
mythrocks, line 119:
```suggestion
    data_type(type_to_id<size_type>()), num_lists + 1, cudf::mask_state::UNALLOCATED, stream);
```

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+  // Use label_segments to expand the key indices based on the list size offsets
+  cudf::detail::label_segments(list_sizes_offsets->view().begin<size_type>(),
+                               list_sizes_offsets->view().end<size_type>(),
+                               d_key_index,
+                               d_key_index + total_compares,
+                               stream);
+
+  // Calculate offsets for the total comparisons per row
+  // This gives us the starting position for each row's comparisons
+  auto total_compares_offsets = make_numeric_column(
```
mythrocks, line 118: Needs to mention that this is on the search space.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+       d_list_sizes_val_offsets = search_values.offsets_begin()] __device__(auto const idx) {
+        return d_list_sizes_offsets[d_key_index[idx]] > 0
+                 ? idx % d_list_sizes_offsets[d_key_index[idx]] +
+                     d_list_sizes_val_offsets[d_val_offsets[idx]]
+                 : idx;
+      }));
+
+  // Use row comparator to allow nested/NULL/NaN comparisons
+  auto const keys_tview  = cudf::table_view{{all_keys}};
+  auto const child_tview = cudf::table_view{{all_values}};
```
mythrocks, line 154:
```suggestion
  auto const values_tview = cudf::table_view{{all_values}};  // Search space.
```

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,308 @@
...
+  auto search_keys_list = cudf::lists_column_view(*search_keys);
+
+  // Find the indices of each key in the first map
+  // This tells us where each key appears in the first map, or if it doesn't exist
+  auto map1_indices = indices_of(search_keys_list, cudf::lists_column_view(map1_keys), stream, mr);
+  auto map1_indices_list = make_lists_column(search_keys_list.size(),
+                                             std::make_unique<column>(search_keys_list.offsets()),
+                                             std::move(map1_indices),
+                                             0,
+                                             rmm::device_buffer{0, stream, mr});
```
mythrocks, line 253:
```suggestion
                                             rmm::device_buffer{0, stream});
```

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,359 @@
...
+{
+  CUDF_EXPECTS(col1.child().type().id() == cudf::type_id::STRUCT,
+               "col1 must have exactly 1 child (STRUCT) column.");
+  CUDF_EXPECTS(col1.child().num_children() == 2,
+               "col1 key-value struct must have exactly 2 children.");
+  CUDF_EXPECTS(col2.child().type().id() == cudf::type_id::STRUCT,
+               "col2 must have exactly 1 child (STRUCT) column.");
+  CUDF_EXPECTS(col2.child().num_children() == 2,
+               "col2 key-value struct must have exactly 2 children.");
+
```
mythrocks, line 254: We're making an assumption that neither maps columns have null keys in them.  Spark wouldn't allow it.

It might be good to add an assert that neither keys column `.has_nulls()`.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,359 @@
...
+  // Extract the child columns containing the actual key and value data
+  auto const all_keys   = search_keys.child();
+  auto const num_keys   = all_keys.size();
+  auto const all_values = search_values.child();
+
+  // Calculate the number of elements in each list for both keys and values
+  auto values_sizes = cudf::lists::count_elements(search_values, stream);
+  auto values_nulls = cudf::is_null(*values_sizes, stream);
+  auto keys_sizes   = cudf::lists::count_elements(search_keys, stream);
+  auto keys_nulls   = cudf::is_null(*keys_sizes, stream);
```
mythrocks, line 98: These should be `const`-able.

mythrocks:
Build

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,368 @@
...
+#include <cudf/unary.hpp>
+#include <cudf/utilities/span.hpp>
+
+#include <thrust/scan.h>
+
+using namespace cudf;
+
+namespace spark_rapids_jni {
+
+// Taken from src/lists/utilities.cu
```
mythrocks, line 37: 1. Documentation needs to indicate what this function is for.  What does this function do?

mythrocks, line 37: 2. Also, this comment isn't accurate.  This was borrowed from CUDF.  Please change the document to reflect this.

mythrocks, line 37: 3. This is a utility function that does not need external visibility.  It is implementation specific.  Please put this in a nested, unnamed namespace to force internal linkage.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,368 @@
...
+ * [1, 2, 3]       | [10, 2, 3]
+ * [4, 5]          | [4, 14, 5, 16]
+ * [6, 7, 8, 9]    | [6, 8]
+
+ * search_keys child:      1, 2, 3, 4,  5,  6, 7, 8, 9
+ * search_values child:   10, 2, 3, 4, 14, 16, 6, 8
+ *
+ * This algorithm works by comparing each key and value parallelly,
+ * hence the need for separate indexes. The algorithm first calculates
+ * the total number of comparisons need, which is a row-wise dot product.
```
mythrocks, line 66:
```suggestion
 * the total number of comparisons needed, which is a row-wise dot product.
```

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,368 @@
...
+  auto labels = make_numeric_column(
+    data_type(type_to_id<size_type>()), n_elements, cudf::mask_state::UNALLOCATED, stream, mr);
+  auto const labels_begin = labels->mutable_view().template begin<size_type>();
+  cudf::detail::label_segments(
+    input.offsets_begin(), input.offsets_end(), labels_begin, labels_begin + n_elements, stream);
+  return labels;
+}
+
+/*
+ *
```
mythrocks, line 53:
```suggestion
 * This function takes two LISTS columns:
 * 1. a `search_keys` column (keys being searched *for*)
 * 2. a `search_space` column (the space being searched)
 *
 * The function's output is a LISTS column, where each row[i] contains the list
 * indices of all matches between `search_keys[i]` and `search_space[i]`.
 * i.e.
 * For each row (i.e. list) in the search-space, 
 * 1. If an element matches with an element in the search-keys, 
 *    its list index is included in the output row.
 * 2. If there are repeated values in the search-space row,
 *    the last element's index is returned.
 * 3. If there are values that simply aren't in the search-space row,
 *    -INTMAX (or whatever) is returned.
 * 
 * For example, consider the following rows at index `i`:
 * 
 * auto const ∅ = -INTMAX; // Indicates that the element isn't found.
 * search_keys[i]    == { B, C, D };
 * search_space[i]   == { A, B, C, C, D, E, F };
 * 
 *
 * results[i]        == { ∅, 1,    3, 4, ∅, ∅ };
 *
 * The results column has as many (list) rows as search_keys or search_space.
 * Note that each result (list) row has as many indices as unique values in search_space.
```

mythrocks: Please read the following review comment, to put this suggestion in context:
https://github.com/NVIDIA/spark-rapids-jni/pull/3490/files#r2243547539

revans2:
I ran some benchmarks against a 16 core CPU and the numbers are not great fabulous. It is about 3.8x faster than a 16 core CPU in my tests.

```
spark.time(spark.range(1, 5000000000L, 1, 512).selectExpr("map(id, id, id+1, id+1) as a1", "map(id, id, id -1, id -1) as a2").selectExpr("map_zip_with(a1, a2, (k, v1, v2) -> coalesce(v1, 0) + coalesce(v2, 0)) as result").selectExpr("SUM(size(result))").show())
```

But that is no where near the 8.7x faster when doing everything but the `map_zip_with`

```
spark.time(spark.range(1, 5000000000L, 1, 512).selectExpr("map(id, id, id+1, id+1) as a1", "map(id, id, id -1, id -1) as a2").selectExpr("SUM(size(a1) + size(a2)) as s").show())
```

At a minimum we should file a follow on issue to see if there is a better way to do this processing. This uses a lot of memory, but even if we end up running a single task the GPU is very close to 100% utilized.

When I do profiling on it `cub::detail::reduce::DeviceSegmentedReduceKernel<cub::detail::reduce::policy_hub<int, int, cudf::detail::cast_functor_fn<int, cudf::detail::cast_functor_fn<int, cudf::DeviceMax>>>::Policy1000, thrust::transform_iterator<cudf::detail::cast_fn<int>, thrust::transform_iterator<cudf::detail::value_accessor<int>, thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, thrust::use_default, thrust::use_default>, thrust::use_default, thrust::use_default>, int *, const int *, const int *, int, cudf::detail::cast_functor_fn<int, cudf::detail::cast_functor_fn<int, cudf::DeviceMax>>, int, int>(T2, T3, T4, T5, T7, T8)` shows up as taking 61.5% of the kernel time. Which I think corresponds to the code that does the cross product to find all of the keys that match in `indices_of`.

I think there has got to be a better way to do this because `map_concat` has to do a similar amount of work in finding duplicate keys, but does it 7x faster than `map_zip_with`. Granted it is not totally apples to apples. But it is compelling.

```
spark.time(spark.range(1, 5000000000L, 1, 512).selectExpr("map(id, id, id+1, id+1) as a1", "map(id, id, id -1, id -1) as a2").selectExpr("map_concat(a1, a2) as result").selectExpr("SUM(size(result))").show())
```

Or I could do something truly horrible like the following

```
spark.time(spark.range(1, 5000000000L, 1, 512).selectExpr("map(id, id, id+1, id+1) as a1", "map(id, id, id -1, id -1) as a2").selectExpr("array_union(map_keys(a1), map_keys(a2)) as all_keys", "a1", "a2").selectExpr("transform(all_keys, key -> struct(key as k, (coalesce(a1[key], 0) + coalesce(a2[key], 0)) as v)) as result").selectExpr("SUM(size(result))").show())
```

which is just missing a `map_from_entries` to have it produce the exact same `result`, that we totally could support, but for some reason do not and it is 4.3x faster than `map_zip_with`

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,368 @@
...
+ * values_idx:  0, 1, 2, 0, 1, 2, 0, 1, 2, 3, 4, 5, 6, 3, 4, 5, 6, 7, 8, 7, 8, 7, 8, 7, 8
+*/
+std::unique_ptr<column> indices_of(
+  lists_column_view const& search_keys,  // Column containing lists of keys to search for
+  lists_column_view const&
+    search_values,  // Column containing lists of values to search through (i.e. search space)
+  rmm::cuda_stream_view stream =
+    cudf::get_default_stream(),               // CUDA stream for asynchronous execution
+  rmm::device_async_resource_ref mr =
+    cudf::get_current_device_resource_ref())  // Memory resource for allocations
```
mythrocks, line 87: I'm not sure `clang-format` has run on this.  The formatting looks completely off.

revans2 (changes requested):
Actually I cannot accept this. It is not just about the performance.

I ran another test because the cross product to find the key locations is really scary to me, so I wrote.

```
val rows_per_gib = 33554432L
val num_tasks = 16
Seq(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024).foreach { keys_per_map =>
  val rows_to_gen = (rows_per_gib * num_tasks) / keys_per_map
  val make_map = "map(" + (0 until keys_per_map).map(i => s"id + $i, id + $i").mkString(",") + ")"
  (0 until 3).foreach { run_num =>
    System.err.println(s"RUN $run_num keys_per_map: $keys_per_map")
    spark.time(spark.range(0, rows_to_gen, 1, num_tasks).selectExpr(s"$make_map as a1", s"$make_map as a2").selectExpr("map_zip_with(a1, a2, (k, v1, v2) -> coalesce(v1, 0) + coalesce(v2, 0)) as result").selectExpr("SUM(size(result))").show())
  }
}
```

And when we get to 64 keys per map we I get a crash. 64 is large, but not so crazy large that it is unlikely to actually happen. Also the crash indicates that if we go even larger would could get some kind of corruption.

```
ai.rapids.cudf.CudfException: CUDF failure at: .../spark-rapids-jni/thirdparty/cudf/cpp/src/column/column_factories.cpp:83: Column size cannot be negative.
	at com.nvidia.spark.rapids.jni.GpuMapZipWithUtils.mapZip(Native Method)
	at com.nvidia.spark.rapids.jni.GpuMapZipWithUtils.mapZip(GpuMapZipWithUtils.java:65)
	at com.nvidia.spark.rapids.GpuMapZipExpression.$anonfun$columnarEval$17(higherOrderFunctions.scala:730)
	at com.nvidia.spark.rapids.Arm$.withResource(Arm.scala:30)
	at com.nvidia.spark.rapids.GpuMapZipExpression.$anonfun$columnarEval$16(higherOrderFunctions.scala:729)
	at com.nvidia.spark.rapids.Arm$.withResource(Arm.scala:30)
	at com.nvidia.spark.rapids.GpuMapZipExpression.columnarEval(higherOrderFunctions.scala:728)
	at com.nvidia.spark.rapids.RapidsPluginImplicits$ReallyAGpuExpression.columnarEval(implicits.scala:34)
	at com.nvidia.spark.rapids.GpuProjectExec$.$anonfun$project$1(basicPhysicalOperators.scala:132)
	at com.nvidia.spark.rapids.RapidsPluginImplicits$MapsSafely.$anonfun$safeMap$1(implicits.scala:166)
	at com.nvidia.spark.rapids.RapidsPluginImplicits$MapsSafely.$anonfun$safeMap$1$adapted(implicits.scala:163)
```

At a minimum we have to prevent the overflow so we crash on large numbers, but we don't have any risk of data corruption. Ideally we find a way to deal with the scale, memory, and performance issues all together.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,368 @@
...
+
+  // Calculate cumulative offsets for the associated list sizes
+  // This gives us the starting position for each key's value comparisons
+  auto values_sizes_offsets = make_numeric_column(
+    data_type(type_to_id<size_type>()), num_keys + 1, cudf::mask_state::UNALLOCATED, stream);
+  auto d_values_sizes_offsets = values_sizes_offsets->mutable_view().template data<size_type>();
+  thrust::exclusive_scan(rmm::exec_policy(stream),
+                         search_key_to_num_search_values,
+                         search_key_to_num_search_values + num_keys + 1,
+                         d_values_sizes_offsets);
```
mythrocks, line 132: Nit: Feel free to ignore, but this is the sort of place an "Immediately Invoked Function Expressions" would be useful.  Note that `values_sizes_offsets` isn't complete in definition until the exclusive-scan has completed.

```suggestion
  auto values_sizes_offsets = [&] {
    auto offsets = make_numeric_column(data_type(type_to_id<size_type>()),
                                       num_keys + 1,
                                       cudf::mask_state::UNALLOCATED, stream);
    auto d_values_sizes_offsets = values_sizes_offsets->mutable_view().template data<size_type>();
    thrust::exclusive_scan(rmm::exec_policy(stream),
                           search_key_to_num_search_values,
                           search_key_to_num_search_values + num_keys + 1,
                           d_values_sizes_offsets);
    return offsets;
  }
```
Note that this allows the outer variable to now be const.

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,368 @@
...
+ *
+ * total_compares_per_row: 9, 8, 8
+ * total_compares_offsets: 0, 9, 17, 25
+ * val_offsets: 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2
+ * values_idx:  0, 1, 2, 0, 1, 2, 0, 1, 2, 3, 4, 5, 6, 3, 4, 5, 6, 7, 8, 7, 8, 7, 8, 7, 8
+*/
+std::unique_ptr<column> indices_of(
+  lists_column_view const& search_keys,  // Column containing lists of keys to search for
+  lists_column_view const&
+    search_values,  // Column containing lists of values to search through (i.e. search space)
```
mythrocks, line 83: Consider renaming to `search_space`.

mythrocks:
> I ran another test because the cross product to find the key locations is really scary to me, so I wrote...

Thanks for testing this out, @revans2.  I hadn't considered that the row-wise cross-product could overflow.

For the moment, we'll put in a `CUDF_EXPECTS` to ensure that we don't cause a corruption, and check that the number of comparisons does not overflow.

We'll also remove the segmented reduce step, under the assumption that the inputs have already been scrubbed of duplicate keys.  That should make things faster, and likely also reduce the memory footprint.

Tackling the overflow in a workable way will involve batching the inputs up to the row preceding the overflow, processing, and then concatenating the results.  In the future might need to do something like the strings-contains kernels, to switch approaches depending on a heuristic.  I don't know what of this we'll be able to tackle in this current PR.

revans2 (approved):
I reran my benchmarks and things look a lot better. The memory usage is down and the performance is up. The overflow was detected and so memory corruption is mitigated.

I still want to see a follow on issue to look at performance and memory usage as I think we can still do a lot better.

As an example I profiled doing a map_zip_with on two columns of maps with 64 entries each.  14.7% of the kernel time was spent doing a `thrust::cuda_cub::_uninitialized_fill` and grew to 21.9% of the kernel time for 1024 by 1024 keys. I also think that we can improve the scaling. map_get and array_union + map_keys all behave much closer to linear time with the number of keys and fixed memory usage

src/main/cpp/src/map_zip_with_utils.cu
```diff
@@ -0,0 +1,420 @@
...
+std::unique_ptr<column> indices_of(
+  lists_column_view const& search_keys,       // Column containing lists of keys to search for
+  lists_column_view const& search_values,     // Column containing lists of values to search through
+  rmm::cuda_stream_view stream =
+    cudf::get_default_stream(),               // CUDA stream for asynchronous execution
+  rmm::device_async_resource_ref mr =
+    cudf::get_current_device_resource_ref())  // Memory resource for allocations
+{
+  /*
+  *
```
mythrocks, line 116:
```suggestion
```

mythrocks:
Build

mythrocks (approved):
If @revans2 finds this version acceptable for now, I'm 👍 on this version.

There are further improvements to be made, for performance and resilience.  Follow-up issues will be filed for those shortly.

mythrocks:
Build

revans2 (approved):

PR #3520, opened 2025-07-11
Ansi multiply

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/13054

#### byte/short check:
```scala
  private def checkOverflow(res: Int, x: Byte, y: Byte, op: String, hint: String): Unit = {
    if (res > Byte.MaxValue || res < Byte.MinValue) {
      throw QueryExecutionErrors.binaryArithmeticCauseOverflowError(x, op, y, hint)
    }
  }

  override def plus(x: Byte, y: Byte): Byte = {
    val tmp = x + y
    checkOverflow(tmp, x, y, "+", "try_add")
    tmp.toByte
  }
```
#### int/long check
```scala
    case _: IntegerType if failOnError =>
      MathUtils.multiplyExact(
        input1.asInstanceOf[Int],
        input2.asInstanceOf[Int],
        getContextOrNull())
    case _: LongType if failOnError =>
      MathUtils.multiplyExact(
        input1.asInstanceOf[Long],
        input2.asInstanceOf[Long],
        getContextOrNull())
```
For more details: 
Refer to [link](https://github.com/apache/spark/blob/v4.0.0/sql/catalyst/src/main/scala/org/apache/spark/sql/types/numerics.scala)

Refer to [link](https://github.com/apache/spark/blob/v4.0.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L558)

## changes
Add ANSI multiplication for types: int8, int16, int32, int64, float, double.
Multipy scalar/column in a unified way.
Use the approach in https://github.com/NVIDIA/spark-rapids-jni/pull/3504 to check nulls info.

### DIFF
--- a/src/main/cpp/src/multiply.cu
+++ b/src/main/cpp/src/multiply.cu
@@ -16,99 +16,314 @@
 
 #include "multiply.hpp"
 #include "row_error_utilities.hpp"
+#include "utilities.hpp"
 
-#include <cudf/column/column_device_view.cuh>
 #include <cudf/column/column_factories.hpp>
 #include <cudf/detail/valid_if.cuh>
 
 #include <rmm/device_uvector.hpp>
 #include <rmm/exec_policy.hpp>
 
+#include <cuda/std/limits>
 #include <thrust/for_each.h>
 
 namespace spark_rapids_jni {
 
 namespace {
 
-// Functor to perform multiplication on two columns with overflow checks if ANSI mode is enabled.
+void check_multiply_inputs(cudf::data_type left_type,
+                           cudf::data_type right_type,
+                           cudf::size_type left_size,
+                           cudf::size_type right_size,
+                           bool is_ansi_mode,
+                           bool is_try_mode)
+{
+  CUDF_EXPECTS(left_type == right_type, "Input columns must have the same data type");
+  CUDF_EXPECTS(is_basic_spark_numeric(left_type), "Unsupported data type for multiplication.");
+  CUDF_EXPECTS(left_size == right_size, "Input columns must have the same size");
+  CUDF_EXPECTS(!(is_ansi_mode && is_try_mode),
+               "Cannot enable both ANSI mode and TRY mode at the same time");
+}
+
+/**
+ * @brief Multiply two int8_t or int16_t values with overflow check.
+ * Set `valid` to false if overflow occurs.
+ */
+template <typename T, CUDF_ENABLE_IF(std::is_same_v<T, int8_t> || std::is_same_v<T, int16_t>)>
+__device__ void multiply(T x, T y, bool check_overflow, bool* valid, T* result)
+{
+  // when int8_t * int8_t or int16_t * int16_t, the result is promoted to int type.
+  int32_t r = x * y;
+  if (check_overflow &&
+      (r < cuda::std::numeric_limits<T>::min() || r > cuda::std::numeric_limits<T>::max())) {
+    *valid = false;  // overflow in ansi mode
+  } else {
+    *valid  = true;
+    *result = static_cast<T>(r);
+  }
+}
+
+/**
+ * @brief Multiply two int32_t values with overflow check.
+ * Set `valid` to false if overflow occurs.
+ */
+template <typename T, CUDF_ENABLE_IF(std::is_same_v<T, int32_t>)>
+__device__ void multiply(T x, T y, bool check_overflow, bool* valid, T* result)
+{
+  if (check_overflow) {
+    int64_t r = static_cast<int64_t>(x) * static_cast<int64_t>(y);
+    if (auto r_cast = static_cast<int>(r); static_cast<long>(r_cast) == r) {
+      *valid  = true;
+      *result = r_cast;
+    } else {
+      *valid = false;  // overflow in ansi mode
+    }
+  } else {
+    *valid  = true;
+    *result = x * y;
+  }
+}
+
+__device__ bool is_multiply_overflow(int64_t a, int64_t b)
+{
+  // constexpr int64_t INT64_MIN = (-9223372036854775807L - 1L); // -2^63
+  // constexpr int64_t INT64_MAX = 9223372036854775807L; // 2^63 - 1
+
+  if (a > 0 && b > 0 && a > 9223372036854775807L / b) return true;          // Positive overflow
+  if (a < 0 && b < 0 && a < 9223372036854775807L / b) return true;          // Negative overflow
+  if (a > 0 && b < 0 && b < (-9223372036854775807L - 1L) / a) return true;  // Mixed sign overflow
+  if (a < 0 && b > 0 && a < (-9223372036854775807L - 1L) / b) return true;  // Mixed sign overflow
+  return false;
+}
+
+/**
+ * @brief Multiply two int64_t values with overflow check.
+ * Set `valid` to false if overflow occurs.
+ */
+template <typename T, CUDF_ENABLE_IF(std::is_same_v<T, int64_t>)>
+__device__ void multiply(T x, T y, bool check_overflow, bool* valid, T* result)
+{
+  if (check_overflow && is_multiply_overflow(x, y)) {
+    *valid = false;  // overflow
+    return;
+  }
+
+  *valid  = true;
+  *result = x * y;
+}
+
+/**
+ * @brief Multiply two float values.
+ * `check_overflow` is ignored for float multiplication.
+ */
+template <typename T, CUDF_ENABLE_IF(std::is_same_v<T, float> || std::is_same_v<T, double>)>
+__device__ void multiply(T x, T y, bool check_overflow, bool* valid, T* result)
+{
+  *valid  = true;
+  *result = x * y;
+}
+
+/**
+ * @brief Functor to perform multiplication on two accessors,
+ * accessor is type of `column_accessor` or `numeric_scalar_accessor`
+ */
+template <typename T, typename LEFT_ACCESSOR, typename RIGHT_ACCESSOR>
 struct multiply_fn {
-  cudf::column_device_view left_cdv;
-  cudf::column_device_view right_cdv;
-  bool is_ansi_mode;
-  int* results;
+  LEFT_ACCESSOR left_accessor;
+  RIGHT_ACCESSOR right_accessor;
+  bool check_overflow;
+  T* results;
   bool* validity;
 
   __device__ void operator()(int row_idx) const
   {
-    if (left_cdv.is_null(row_idx) || right_cdv.is_null(row_idx)) {
+    if (left_accessor.is_null(row_idx) || right_accessor.is_null(row_idx)) {
       validity[row_idx] = false;
       return;
     }
 
-    if (is_ansi_mode) {
-      long a = left_cdv.element<int>(row_idx);
-      long b = right_cdv.element<int>(row_idx);
-      long r = a * b;
-      if (auto r_cast = static_cast<int>(r); static_cast<long>(r_cast) == r) {
-        validity[row_idx] = true;
-        results[row_idx]  = r_cast;
-      } else {
-        // Overflow detected
-        validity[row_idx] = false;
-      }
+    multiply<T>(left_accessor.element(row_idx),
+                right_accessor.element(row_idx),
+                check_overflow,
+                &validity[row_idx],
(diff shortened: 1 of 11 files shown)


res-life:
build

firestarman (commented):
I only reviewed the Java part

res-life:
Follow-up issue: https://github.com/NVIDIA/spark-rapids-jni/issues/3528
Please refer to: https://github.com/NVIDIA/cccl/pull/3755#issuecomment-3072144807

src/main/java/com/nvidia/spark/rapids/jni/Arithmetic.java
```diff
@@ -16,39 +16,130 @@
...
+          leftType.getTypeId() == DType.DTypeEnum.INT16 ||
+          leftType.getTypeId() == DType.DTypeEnum.INT32 ||
+          leftType.getTypeId() == DType.DTypeEnum.INT64 ||
+          leftType.getTypeId() == DType.DTypeEnum.FLOAT32 ||
+          leftType.getTypeId() == DType.DTypeEnum.FLOAT64)) {
+      throw new IllegalArgumentException("Multiplication types must be signed integral or float, " +
+          "but get type: " + leftType);
+    }
+
+    if (leftType != rightType) {
```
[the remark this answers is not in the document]
res-life: Please refer to the Spark `Multiply` in [link](https://github.com/apache/spark/blob/v4.0.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L558):

```scala
  protected override def nullSafeEval(input1: Any, input2: Any): Any = dataType match {
    case DecimalType.Fixed(precision, scale) =>
      checkDecimalOverflow(numeric.times(input1, input2).asInstanceOf[Decimal], precision, scale)
    case _: IntegerType if failOnError =>
      MathUtils.multiplyExact(
        input1.asInstanceOf[Int],
        input2.asInstanceOf[Int],
        getContextOrNull())
    case _: LongType if failOnError =>
      MathUtils.multiplyExact(
        input1.asInstanceOf[Long],
        input2.asInstanceOf[Long],
        getContextOrNull())
    case _ => numeric.times(input1, input2)
  }
```

Spark assumes the types of left and right are the same, e.g.: MathUtils.multiplyExact(long, long), MathUtils.multiplyExact(int, int)

src/main/cpp/src/multiply.cu
```diff
@@ -16,99 +16,314 @@
...
+}
+
+__device__ bool is_multiply_overflow(int64_t a, int64_t b)
+{
+  // constexpr int64_t INT64_MIN = (-9223372036854775807L - 1L); // -2^63
+  // constexpr int64_t INT64_MAX = 9223372036854775807L; // 2^63 - 1
+
+  if (a > 0 && b > 0 && a > 9223372036854775807L / b) return true;          // Positive overflow
+  if (a < 0 && b < 0 && a < 9223372036854775807L / b) return true;          // Negative overflow
+  if (a > 0 && b < 0 && b < (-9223372036854775807L - 1L) / a) return true;  // Mixed sign overflow
```
res-life, line 94: Directly using INT64_MAX is causing building error, so using `-9223372036854775807L - 1L` is a work around.

thirtiseven: 2^63 is not a prime. So is it possible that a * b == -2^63 but the overflow check fails?

res-life: > if (a > 0 && b < 0 && b < (-9223372036854775807L - 1L) / a) return true
2^63 is not a prime. So is it possible that a * b == -2^63 but the overflow check fails?

I tried the following, the issue you described does not exist:
```cpp
#include <cstdio>

int main() {

  long a = 2L;
  long b = -4611686018427387904L;
  long min = -9223372036854775807L - 1L;

  if (a * b == min) {
    printf("a * b == min\n");
  }

  if (a > 0 && b < 0 && b < min / a ) {
    printf("overflow\n");
  } else {
    printf("no overflow\n");
  }
}

prints:

a * b == min
no overflow

```

thirtiseven: > I tried the following, the issue you described does not exist:

Ohh sorry, yes, you are right.

src/main/cpp/src/utilities.hpp
```diff
@@ -26,6 +28,57 @@
...
+ * @brief Element accessor from a column device view.
+ * It has the same interfaces as the `numeric_scalar_accessor`.
+ * This is used to access element from column and scalar with a unified model.
+ */
+template <typename T>
+class column_accessor {
+ public:
+  column_accessor(cudf::column_device_view cdv) : _cdv(cdv) {}
+
+  __device__ bool is_null(cudf::size_type row_index) const { return _cdv.is_null(row_index); }
```
ttnghia, line 41: Do not put device code in `hpp` file. Make a `cuh` file instead.

src/main/cpp/src/multiply.hpp
```diff
@@ -16,30 +16,107 @@
...
+ *
+ * @param left_input The left input.
+ * @param right_input The right input.
+ * @param is_ansi_mode If true, throws exception when overflow occurs.
+ * @param is_try_mode If true, set null when overflow occurs.
+ * @param stream CUDA stream to use for the operation.
+ * @param mr Memory resource to use for allocations.
+ * @return A new column containing the results of the multiplication.
+ */
+std::unique_ptr<cudf::column> multiply(
```
ttnghia, line 115: Instead of having 3 separate public APIs, how about using `std::variant<cudf::column_view, cudf::scalar`?

src/main/cpp/src/utilities.hpp
```diff
@@ -26,6 +28,57 @@
...
+
+  __device__ T element(cudf::size_type row_index) const { return _cdv.element<T>(row_index); }
```
ttnghia, line 43: I see no difference from using `column_device_view` directly. Did I miss anything?

src/main/java/com/nvidia/spark/rapids/jni/Arithmetic.java
```diff
@@ -16,39 +16,130 @@
...
+   * @param isAnsiMode is ANSI mode enabled
+   * @param isTryMode  if true, set null when overflow occurs
+   * @return a new ColumnVector containing the result of multiplying
+   *
+   * @throws ExceptionWithRowIndex if has any overflow in ANSI mode
+   */
+  public static ColumnVector multiply(ColumnVector left, ColumnVector right, boolean isAnsiMode,
+      boolean isTryMode) {
+    checkMultiply(left.getType(), right.getType(), left.getRowCount(),
+        right.getRowCount(), isAnsiMode, isTryMode);
```
ttnghia, line 76: This check is performed at the C++ level, so I think we should not waste time doing it twice. Just checking in C++ is enough.

src/main/cpp/src/multiply.cu
```diff
@@ -16,99 +16,315 @@
...
-                                          mr);
+  // create a validity vector to track valid results
   auto validity =
-    rmm::device_uvector<bool>(left_input.size(), stream, cudf::get_current_device_resource_ref());
-  auto dcv_left  = cudf::column_device_view::create(left_input, stream);
-  auto dcv_right = cudf::column_device_view::create(right_input, stream);
+    rmm::device_uvector<bool>(num_rows, stream, cudf::get_current_device_resource_ref());
 
+  // excute the multiplication
+  bool check_overflow = is_ansi_mode || is_try_mode;
```
ttnghia, line 172: If `ansi_mode` and `try_mode` are not used sepately, we only need to pass in one parameter `bool check_overflow` into this functor.

res-life:
build

src/main/cpp/src/multiply.cu
```diff
@@ -16,99 +16,343 @@
...
 
   __device__ void operator()(int row_idx) const
   {
-    if (left_cdv.is_null(row_idx) || right_cdv.is_null(row_idx)) {
+    auto const& left_pair   = left_accessor[row_idx];
+    auto const& right_pair  = right_accessor[row_idx];
+    auto const& left_value  = left_pair.first;
+    auto const& right_value = right_pair.first;
+    auto const& left_valid  = left_pair.second;
+    auto const& right_valid = right_pair.second;
```
ttnghia, line 149: Store values as the pair accessor may be a temporary.

```suggestion
    auto const left_pair   = left_accessor[row_idx];
    auto const right_pair  = right_accessor[row_idx];
    auto const left_value  = left_pair.first;
    auto const right_value = right_pair.first;
    auto const left_valid  = left_pair.second;
    auto const right_valid = right_pair.second;
```

ttnghia, line 149: Store values as the pair accessor may be a temporary.
```suggestion
    auto const [left_value, left_valid]   = left_accessor[row_idx];
    auto const [right_value, right_valid] = right_accessor[row_idx];
```

src/main/cpp/src/multiply.cu
```diff
@@ -16,99 +16,343 @@
...
+  return result;
+}
+
+/**
+ * @brief Operator for multiply(cv, cv).
+ */
+struct dispatch_multiply {
+  // left operand can be either a column_view or a scalar
+  cudf::column_view const* left_cv;
+  cudf::scalar const* left_scalar;
```
ttnghia, line 200: Similar for `rhs`.
```suggestion
  std::variant<cudf::column_view, cudf::scalar> lhs;
```

src/main/cpp/src/multiply.cu
```diff
@@ -16,99 +16,343 @@
...
+          auto const left_accessor  = cudf::detail::make_pair_iterator<T, true>(*left_cdv);
+          auto const right_accessor = cudf::detail::make_pair_iterator<T, true>(*right_cdv);
+          return multiply_impl<T, decltype(left_accessor), decltype(right_accessor)>(
+            type, num_rows, left_accessor, right_accessor, check_overflow, stream, mr);
+        } else {
+          auto const left_accessor  = cudf::detail::make_pair_iterator<T, true>(*left_cdv);
+          auto const right_accessor = cudf::detail::make_pair_iterator<T, false>(*right_cdv);
+          return multiply_impl<T, decltype(left_accessor), decltype(right_accessor)>(
+            type, num_rows, left_accessor, right_accessor, check_overflow, stream, mr);
+        }
```
ttnghia, line 222:
```suggestion
        auto const left_accessor  = cudf::detail::make_pair_iterator<T, true>(*left_cdv);
        if (right_cv->has_nulls()) {
          auto const right_accessor = cudf::detail::make_pair_iterator<T, true>(*right_cdv);
          return multiply_impl<T, decltype(left_accessor), decltype(right_accessor)>(
            type, num_rows, left_accessor, right_accessor, check_overflow, stream, mr);
        } else {
          auto const right_accessor = cudf::detail::make_pair_iterator<T, false>(*right_cdv);
          return multiply_impl<T, decltype(left_accessor), decltype(right_accessor)>(
            type, num_rows, left_accessor, right_accessor, check_overflow, stream, mr);
        }
```

src/main/cpp/src/multiply.cu
```diff
@@ -16,99 +16,338 @@
...
+          return multiply_impl<T, decltype(left_accessor), decltype(right_accessor)>(
+            type, num_rows, left_accessor, right_accessor, check_overflow, stream, mr);
+        } else {
+          auto const left_accessor  = cudf::detail::make_pair_iterator<T, false>(*left_cdv);
+          auto const right_accessor = cudf::detail::make_pair_iterator<T, false>(*right_cdv);
+          return multiply_impl<T, decltype(left_accessor), decltype(right_accessor)>(
+            type, num_rows, left_accessor, right_accessor, check_overflow, stream, mr);
+        }
+      }
+    } else if (left_cv != nullptr && right_scalar != nullptr) {
```
ttnghia, line 236: In addition, I believe that we can shorten this function can be shorten significantly more using `std::visit`. However, it is optionial as I don't want to push too hard for it 😄

src/main/cpp/src/utilities.hpp
res-life: Now is using cudf::detail::make_pair_iterator, so removed the device code in this hpp

ttnghia (approved):

PR #3521, opened 2025-07-11
Use the acutal digits count for the overflow check when casting strings to decimals

firestarman:
This is a fix to the issue https://github.com/NVIDIA/spark-rapids/issues/10908.

`significant_digits_before_decimal` is the digits count of the stored value, not the actual value. When the scale is negative, the stored digits count is not equal to that of the actual digits. For example, the digits `123456` with DecimalType(4, -2) has `1235` as the stored digits, but the acutal digits are  `123500`.

The overflow check should use the actual digits count instead, which can represented as `decimal_location - leading_zeros`.

Tests will be added in the relevent Plugin PR.

### DIFF
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -537,7 +537,8 @@ CUDF_KERNEL void string_to_decimal_kernel(T* out,
       significant_digits_before_decimal_in_string + zeros_to_decimal + rounding_digits;
 
     // too many digits required to store decimal
-    if (max_digits_before_decimal < significant_digits_before_decimal) { valid = false; }
+    auto const leading_zeros = total_digits - num_precise_digits;
+    if (max_digits_before_decimal < decimal_location - leading_zeros) { valid = false; }
 
     // at this point we have the precise digits we need, but we might need trailing zeros on this
     // value both before and after the decimal


firestarman:
This is a non-breaking PR.

firestarman:
build

res-life:
Could we add a test case in `cast_string.cpp` for this bug?

nartal1:
This fix LGTM. Would like someone else also to take a look.

firestarman:
> Could we add a test case in `cast_string.cpp` for this bug?

Done

res-life (approved):

res-life:
Verified on CPU Spark 400, the GPU get the same result as CPU.

```scala
val df = Seq("123456789", "+987651342", "-1.23e8", "-1256789012").toDF("col")
df.write.mode("OVERWRITE").parquet("TEMP")
val decType = DataTypes.createDecimalType(4, -3)
spark.conf.set("spark.sql.legacy.allowNegativeScaleOfDecimal", "true")
spark.conf.set("spark.sql.ansi.enabled", "false")
spark.read.parquet("TEMP").select(col("col").cast(decType)).show(false)

+----+
|col |
+----+
|NULL|
|NULL|
|NULL|
|NULL|
+----+

```

firestarman:
I am going to merge it to unblock the Plugin PR premerge. If any concern, I can address it by follow-ups.

PR #3535, opened 2025-07-16
Enable shellcheck workflow

YanxuanLiu:
partly fix https://github.com/NVIDIA/spark-rapids-common/issues/41

1. fixed an shell syntax error
2. enable shellcheck workflow

### DIFF
--- a/.github/workflows/shell-check.yml
+++ b/.github/workflows/shell-check.yml
@@ -24,37 +24,8 @@ jobs:
     runs-on: ubuntu-latest
     if: "!contains(github.event.pull_request.title, '[bot]')"
     steps:
-      - name: Get checkout depth
-        run: |
-          echo "PR_FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} + 10 ))" >> $GITHUB_ENV
-
       - name: Checkout code
         uses: actions/checkout@v4
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/shell-check.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Get checkout depth
+        run: |
+          echo "PR_FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} + 10 ))" >> $GITHUB_ENV
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: ${{ env.PR_FETCH_DEPTH }}
+
+      - name: Identify changed Shell files
```
pxLi, line 36: I think we would not only detect changed files, its better to detect all shell scripts within the merged commit if possible

.github/workflows/shell-check.yml
```diff
@@ -0,0 +1,60 @@
...
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  shell-check:
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.pull_request.title, '[bot]')"
+    steps:
+      - name: Get checkout depth
```
pxLi, line 27: IMHO no need for this steps, also please remove the one in common repo

PR #3568, opened 2025-07-25
Upgrade `gcc-toolset` to version 14

ttnghia:
As RAPIDS is moving to gcc-14, we should upgrade our toolset to match them.

There are only a few minor issues in the code that need to be fixed in order to upgrade. They are not bugs, just compiler warnings and changes in the system header.

### DIFF
--- a/build/run-in-docker
+++ b/build/run-in-docker
@@ -20,6 +20,8 @@
 
 set -e
 
+source build/env.sh
+
 # Set this environment variable to ON to build the docker image for local development,
 # with gcc-toolset enabled by default.
 # This is necessary especially if the docker container is used for local development with an IDE.
@@ -115,4 +117,4 @@ $DOCKER_CMD run $DOCKER_GPU_OPTS $DOCKER_RUN_EXTRA_ARGS -u $(id -u):$(id -g) --r
   -e TZ=${TZ} \
   $DOCKER_OPTS \
   $JNI_DOCKER_IMAGE \
-  scl enable gcc-toolset-14 "$RUN_CMD"
+  ${sclCMD} "$RUN_CMD"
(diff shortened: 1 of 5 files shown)


ttnghia:
build

ttnghia:
CI/CD ran successfully without any problem. No error reported. Why "failed"?

ci/Dockerfile
```diff
@@ -32,7 +32,7 @@ FROM --platform=$TARGETPLATFORM nvidia/cuda:$CUDA_VERSION-devel-rockylinux$OS_RE
 ARG DEV_BUILD=OFF
 
 # Dependency versions
-ARG TOOLSET_VERSION=11
+ARG TOOLSET_VERSION=14
```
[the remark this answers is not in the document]
pxLi: Also, please explicitly install, e.g.
line 44:
`dnf ... gcc-toolset-11 gcc-toolset-${TOOLSET_VERSION} ... zip unzip`

since zip&unzip were dependencies of gcc-toolset-11, but have been removed from the dependencies of gcc-toolset-14, lets try install explicitly for future versions, thanks

build/run-in-docker
```diff
@@ -115,4 +115,4 @@ $DOCKER_CMD run $DOCKER_GPU_OPTS $DOCKER_RUN_EXTRA_ARGS -u $(id -u):$(id -g) --r
   -e TZ=${TZ} \
   $DOCKER_OPTS \
   $JNI_DOCKER_IMAGE \
-  scl enable gcc-toolset-11 "$RUN_CMD"
+  scl enable gcc-toolset-14 "$RUN_CMD"
```
pxLi, line 118: Additionally, since version information is currently distributed across several locations (such as rapids, JNI, and internal CI), it would be better to make a single source of truth for each branch.

This would allow different releases to automatically determine the appropriate toolset based on the local HEAD, rather than requiring the version to be hardcoded by any caller

ttnghia: Sure, please move forward with this, either pushing to this branch or creating a new one.

pxLi: sure, let me add some commits to your change

pxLi:
build

pxLi:
Besides pre-merge checks, I will conduct additional verification runs internally.

pxLi (approved):

PR #3617, opened 2025-08-08
Add support for parse_url in ANSI mode

nartal1:
This contributes to https://github.com/NVIDIA/spark-rapids/issues/11193.  This adds ANSI mode support to the ParseURI JNI functions, providing the foundational layer needed to implement Spark's `parse_url` function with ANSI compliance in the spark-rapids plugin.

####  **Added ANSI Mode Support to JNI Layer**
- **Java Layer**: Added `failOnError` parameter to all `ParseURI` JNI methods (`parseURIProtocol`, `parseURIHost`, `parseURIQuery`, `parseURIPath`)
- **C++ Layer**: Implemented `parse_uri_ansi()` function that throws exceptions for invalid URIs in ANSI mode
- **JNI Bindings**: Updated all native method signatures to support ANSI mode parameter

####  **Tests**
- Added ANSI mode tests verifying exception throwing behavior
- Added tests for missing query parameters (should return `NULL`, not throw exceptions)
- Added tests in both cpp and java.

#### ANSI Mode Behavior
- **`failOnError = false`**: Returns `NULL` for invalid URLs (standard behavior)
- **`failOnError = true`**: Throws `exception_with_row_index` for structurally invalid URLs (ANSI mode)
- **Missing query parameters**: Return `NULL` in both modes (not considered invalid URLs)
- **NULL inputs**: Always return `NULL` in both modes

### DIFF
--- a/src/main/cpp/src/parse_uri.cu
+++ b/src/main/cpp/src/parse_uri.cu
@@ -1035,12 +1035,12 @@ std::unique_ptr<cudf::column> parse_uri_to_query(cudf::strings_column_view const
                                                  rmm::device_async_resource_ref mr)
 {
   CUDF_FUNC_RANGE();
+  if (ansi_mode) { detail::validate_input_uris(input, stream); }
 
   // build string_column_view from incoming query_match string
   auto d_scalar = make_string_scalar(query_match, stream);
   auto col      = make_column_from_scalar(*d_scalar, 1);
 
-  if (ansi_mode) { detail::validate_input_uris(input, stream); }
   return detail::parse_uri(input, detail::URI_chunks::QUERY, strings_column_view(*col), stream, mr);
 }
 


nartal1:
build

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
```diff
@@ -30,76 +30,106 @@ public class ParseURI {
...
    */
-  public static ColumnVector parseURIProtocol(ColumnView uriColumn) {
+  public static ColumnVector parseURIProtocol(ColumnView uriColumn, boolean failOnError) {
     assert uriColumn.getType().equals(DType.STRING) : "Input type must be String";
-    return new ColumnVector(parseProtocol(uriColumn.getNativeView()));
+    if (failOnError) {
+      return new ColumnVector(parseProtocol(uriColumn.getNativeView(), true));
+    } else {
+      return new ColumnVector(parseProtocol(uriColumn.getNativeView(), false));
+    }
```
res-life, line 43:
```suggestion
    return new ColumnVector(parseProtocol(uriColumn.getNativeView(), failOnError));
```

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
```diff
@@ -30,76 +30,106 @@ public class ParseURI {
...
    */
-  public static ColumnVector parseURIHost(ColumnView uriColumn) {
+  public static ColumnVector parseURIHost(ColumnView uriColumn, boolean failOnError) {
     assert uriColumn.getType().equals(DType.STRING) : "Input type must be String";
-    return new ColumnVector(parseHost(uriColumn.getNativeView()));
+    if (failOnError) {
+      return new ColumnVector(parseHost(uriColumn.getNativeView(), true));
+    } else {
+      return new ColumnVector(parseHost(uriColumn.getNativeView(), false));
+    }
```
res-life, line 59: Reduce to one line

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
```diff
@@ -30,76 +30,106 @@ public class ParseURI {
...
    */
-  public static ColumnVector parseURIQuery(ColumnView uriColumn) {
+  public static ColumnVector parseURIQuery(ColumnView uriColumn, boolean failOnError) {
     assert uriColumn.getType().equals(DType.STRING) : "Input type must be String";
-    return new ColumnVector(parseQuery(uriColumn.getNativeView()));
+    if (failOnError) {
+      return new ColumnVector(parseQuery(uriColumn.getNativeView(), true));
+    } else {
+      return new ColumnVector(parseQuery(uriColumn.getNativeView(), false));
+    }
```
res-life, line 75: Reduce to one line

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
```diff
@@ -30,76 +30,106 @@ public class ParseURI {
...
    */
-  public static ColumnVector parseURIQueryWithLiteral(ColumnView uriColumn, String query) {
+  public static ColumnVector parseURIQueryWithLiteral(ColumnView uriColumn, String query, boolean failOnError) {
     assert uriColumn.getType().equals(DType.STRING) : "Input type must be String";
-    return new ColumnVector(parseQueryWithLiteral(uriColumn.getNativeView(), query));
+    if (failOnError) {
+      return new ColumnVector(parseQueryWithLiteral(uriColumn.getNativeView(), query, true));
+    } else {
+      return new ColumnVector(parseQueryWithLiteral(uriColumn.getNativeView(), query, false));
+    }
```
res-life, line 92: Reduce to one line

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
```diff
@@ -30,76 +30,106 @@ public class ParseURI {
...
-  public static ColumnVector parseURIQueryWithColumn(ColumnView uriColumn, ColumnView queryColumn) {
+  public static ColumnVector parseURIQueryWithColumn(ColumnView uriColumn, ColumnView queryColumn, boolean failOnError) {
     assert uriColumn.getType().equals(DType.STRING) : "Input type must be String";
     assert queryColumn.getType().equals(DType.STRING) : "Query type must be String";
-    return new ColumnVector(parseQueryWithColumn(uriColumn.getNativeView(), queryColumn.getNativeView()));
+    if (failOnError) {
+      return new ColumnVector(parseQueryWithColumn(uriColumn.getNativeView(), queryColumn.getNativeView(), true));
+    } else {
+      return new ColumnVector(parseQueryWithColumn(uriColumn.getNativeView(), queryColumn.getNativeView(), false));
+    }
```
res-life, line 110: Reduce to one line

nartal1:
build

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +956,128 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
 }
 
+/**
+ * @brief Kernel to check for completely invalid URIs (not just missing components).
+ *
+ * @param input_strings Input string column
+ * @param invalid_uri_flags Output flags indicating which URIs are completely invalid
+ * @param query_match Optional query match parameter
+ */
+CUDF_KERNEL void check_uri_validity_kernel(
```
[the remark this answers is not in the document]
nartal1: Thanks for the suggestion @res-life ! We cannot use just throw_row_error_if_any the parsing result. 
URI parsing differs from typical unary operations: null results can be legitimate (valid URI missing requested component) or errors (malformed URI structure).
So created a validation column that's only null for structurally invalid URIs (uri.valid == 0), then use throw_row_error_if_any on this validation column. This reuses existing kernel while correctly handling URI-specific semantics. PTAL.

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
nartal1: Done here and in all the relevant places. PTAL.

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
```diff
@@ -30,76 +30,106 @@ public class ParseURI {
   /**
    * Parse protocol for each URI from the incoming column.
    *
-   * @param URIColumn The input strings column in which each row contains a URI.
+   * @param uriColumn The input strings column in which each row contains a URI.
+   * @param failOnError If true, use ANSI-aware parsing.
    * @return A string column with protocol data extracted.
    */
-  public static ColumnVector parseURIProtocol(ColumnView uriColumn) {
```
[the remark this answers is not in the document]
nartal1: The plan is to pass ansiMode from the spark-rapids plugin code. I am not sure if I am missing any context here.
Please let me know if this is acceptable or if we need to retain the previous function as well.  
```
def doColumnar(url: GpuColumnVector, partToExtract: GpuScalar): ColumnVector = {
    val part = partToExtract.getValue.asInstanceOf[UTF8String].toString
    part match {
      case PROTOCOL =>
        ParseURI.parseURIProtocol(url.getBase, failOnError)
      ...
     ....
```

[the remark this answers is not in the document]
res-life: If the original function is removed, then the Spark-Rapids nightly building will fail, because the snapshot JNI jar will not contain the original function. So let's keep it and mark it as deprecated, after the new override function gets into the snapshot jar, then it's safe to remove.

nartal1:
build

src/test/java/com/nvidia/spark/rapids/jni/ParseURITest.java
```diff
@@ -175,11 +177,135 @@ void testPath(String[] testData) {
...
+    
+    // Test that ANSI mode still throws for truly invalid URLs
+    String[] invalidData = {"://completely-malformed"};
+    try (ColumnVector v0 = ColumnVector.fromStrings(invalidData)) {
+      try {
+        ColumnVector result = ParseURI.parseURIQueryWithLiteral(v0, "param", true);
+        result.close();
+        throw new AssertionError("Expected exception for invalid URL in ANSI mode");
+      } catch (RuntimeException e) {
+        assert e instanceof com.nvidia.spark.rapids.jni.ExceptionWithRowIndex;
```
res-life, line 277: Nit, we can use `assertThrows`:

```java
import static org.junit.jupiter.api.Assertions.*;

assertThrows(RuntimeException.class, () -> {
  ParseURI.parseURI......
});
```

Please check all the places which can use `assertThrows`

res-life (dismissed):

res-life:
Maybe we can post a follow-up PR to fix the nits in test code.

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +957,105 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
+
+  // In ANSI mode, check for completely invalid URIs
+  if (ansi_mode && input.size() > 0) {
+    auto const d_strings = column_device_view::create(input.parent(), stream);
+
+    // Create boolean vector: true for valid URIs, false for completely invalid URIs
+    auto validity_flags = rmm::device_uvector<bool>(input.size(), stream);
+
+    thrust::transform(
+      rmm::exec_policy(stream),
```
ttnghia, line 978:
```suggestion
      rmm::exec_policy_nosync(stream),
```

nartal1:
build

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +957,105 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
+                                       std::optional<strings_column_view const> query_match,
+                                       bool ansi_mode,
+                                       rmm::cuda_stream_view stream,
+                                       rmm::device_async_resource_ref mr)
+{
+  // Get the result first
+  auto result = parse_uri(input, chunk, query_match, stream, mr);
+
+  // In ANSI mode, check for completely invalid URIs
+  if (ansi_mode && input.size() > 0) {
```
ttnghia, line 971: This block is too big, causing difficulty for future maintenance and code review. Just make it smaller by using the other way.

```suggestion
  if (!ansi_mode || input.size() == 0) { return result; }
```

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +957,105 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
+    auto const d_strings = column_device_view::create(input.parent(), stream);
+
+    // Create boolean vector: true for valid URIs, false for completely invalid URIs
+    auto validity_flags = rmm::device_uvector<bool>(input.size(), stream);
+
+    thrust::transform(
+      rmm::exec_policy_nosync(stream),
+      thrust::counting_iterator<cudf::size_type>(0),
+      thrust::counting_iterator<cudf::size_type>(input.size()),
+      validity_flags.begin(),
```
ttnghia, line 981:
```suggestion
    thrust::tabulate(
      rmm::exec_policy_nosync(stream),
      validity_flags.begin(),
      validity_flags.end(),
```

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +957,105 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
+      cuda::proclaim_return_type<bool>([input = *d_strings] __device__(cudf::size_type row_idx) {
+        // Input nulls remain null
+        if (input.is_null(row_idx)) return false;
```
ttnghia, line 984:
```suggestion
        if (input.is_null(row_idx)) { return false; }
```

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +957,105 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
+        return uri.valid != 0;
+      }));
+
+    auto [validation_mask, null_count] = cudf::detail::valid_if(
+      validity_flags.begin(), validity_flags.end(), cuda::std::identity{}, stream, mr);
+
+    // Create validation column for throw_row_error_if_any
+    auto validation_column = cudf::make_numeric_column(cudf::data_type{cudf::type_id::INT8},
+                                                       input.size(),
+                                                       std::move(validation_mask),
```
ttnghia, line 999:
```suggestion
                                                       null_count > 0 ? std::move(validation_mask) : rmm::device_buffer{0, stream},
```

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +957,105 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
+                                                       null_count,
+                                                       stream,
+                                                       mr);
```
ttnghia, line 1002: This makes a new column which allocates a memory buffer that is unused. Let's reuse the validity flags.
```
std::make_unique<column>(std::move(validity_flags), null_count > 0 ? std::move(validation_mask) : rmm::device_buffer{0, stream}, null_count);
```

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +957,105 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
                                               rmm::device_async_resource_ref mr)
 {
   CUDF_FUNC_RANGE();
-  return detail::parse_uri(input, detail::URI_chunks::PROTOCOL, std::nullopt, stream, mr);
+  if (ansi_mode) {
+    return detail::parse_uri_ansi(
+      input, detail::URI_chunks::PROTOCOL, std::nullopt, ansi_mode, stream, mr);
+  } else {
+    return detail::parse_uri(input, detail::URI_chunks::PROTOCOL, std::nullopt, stream, mr);
+  }
```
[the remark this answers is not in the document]
ttnghia: Similar for all other calls in this file.

src/main/cpp/src/parse_uri.cu
```diff
@@ -34,6 +37,9 @@
 
 #include <cuda/functional>
 #include <cuda/std/optional>
+#include <thrust/for_each.h>
+#include <thrust/logical.h>
+#include <thrust/transform.h>
```
ttnghia, line 42:
```suggestion
#include <thrust/tabulate.h>
```

nartal1:
build

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +954,85 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
                              std::move(null_mask));
 }
 
+void validate_output(strings_column_view const& input,
+                     column const& result,
```
ttnghia, line 958: Output is unused, so this validation is wrong.

ttnghia: In addition, never use `column` directly. Use `column_view` instead.

src/main/cpp/src/parse_uri.cu
```diff
@@ -951,36 +954,85 @@ std::unique_ptr<column> parse_uri(strings_column_view const& input,
...
 
 std::unique_ptr<column> parse_uri_to_protocol(strings_column_view const& input,
+                                              bool ansi_mode,
                                               rmm::cuda_stream_view stream,
                                               rmm::device_async_resource_ref mr)
 {
   CUDF_FUNC_RANGE();
-  return detail::parse_uri(input, detail::URI_chunks::PROTOCOL, std::nullopt, stream, mr);
+  auto output = detail::parse_uri(input, detail::URI_chunks::PROTOCOL, std::nullopt, stream, mr);
+  if (ansi_mode) { detail::validate_output(input, *output, stream, mr); }
```
ttnghia, line 1007: This does not return any result thus we don't need to pass `mr`. The validation function only uses the default memory resource.

src/main/cpp/src/parse_uri.cu
nartal1: Thanks for catching this. The function name is confusing. This function would validate the input urls (non nulls). If the input url  is invalid then the validity vector in this function is set. We actually don't compare the output in this case. I will change the function name and the parameters to this function.

nartal1:
build

src/main/cpp/src/parse_uri.cu
```diff
@@ -990,25 +1040,30 @@ std::unique_ptr<cudf::column> parse_uri_to_query(cudf::strings_column_view const
   auto d_scalar = make_string_scalar(query_match, stream);
   auto col      = make_column_from_scalar(*d_scalar, 1);
 
+  if (ansi_mode) { detail::validate_input_uris(input, stream); }
```
ttnghia, line 1043: Let's move this validation step up, before creating other intermediate results, since that can be constly and we can short circuit other computation if an exception being thrown.

nartal1:
build

src/main/cpp/src/parse_uri.cu
nartal1: Done. PTAL

src/main/java/com/nvidia/spark/rapids/jni/ParseURI.java
```diff
@@ -32,74 +32,154 @@ public class ParseURI {
    *
    * @param URIColumn The input strings column in which each row contains a URI.
    * @return A string column with protocol data extracted.
+   * @deprecated Use {@link #parseURIProtocol(ColumnView, boolean)} instead.
    */
+  @Deprecated
```
revans2, line 37: Please do not check in API changes that deprecate API. This causes a breaking changes just as if the APIs were removed.  These are all internal APIs so we can just file a follow on issue to remove them after the code in the plugin is updated.

```
[INFO] compiling 507 Scala sources and 30 Java sources to .../sql-plugin/target/spark342/classes ...
[ERROR] [Error] .../sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuParseUrl.scala:65: method parseURIProtocol in class ParseURI is deprecated
[ERROR] [Error] .../sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuParseUrl.scala:67: method parseURIHost in class ParseURI is deprecated
[ERROR] [Error] .../sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuParseUrl.scala:69: method parseURIQuery in class ParseURI is deprecated
[ERROR] [Error] .../sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuParseUrl.scala:71: method parseURIPath in class ParseURI is deprecated
[ERROR] [Error] .../sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuParseUrl.scala:87: method parseURIQueryWithLiteral in class ParseURI is deprecated
[ERROR] [Error] .../sql-plugin/src/main/scala/org/apache/spark/sql/rapids/GpuParseUrl.scala:97: method parseURIQueryWithColumn in class ParseURI is deprecated
[ERROR] 6 errors found
```

PR #3629, opened 2025-08-12
Add support for UUID

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/12982

Add support for UUID

### depends on
* https://github.com/rapidsai/cudf/pull/19646

### Spark behavior
```scala
scala> spark.sql("select uuid()").show(false)
+------------------------------------+
|uuid()                              |
+------------------------------------+
|cce89a22-c3d2-40b2-ad8c-30bcfa4dc5a1|
+------------------------------------+
```
The uuid function expects zero argument, MUST NOT specify seed. Each time Spark uses a random seed to guarantee the unique of UUIDs.

### description
Implement Java interface:

```java
  /**
   * Generate a column of UUIDs (String type) with `rowCount` rows.
   * Spark uses `Truly Random or Pseudo-Random` UUID type which is described in
   * the section 4.4 of [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122),
   * The variant in UUID is 2 and the version in UUID is 4. This implementation
   * generates UUIDs in the same format, but does not generate the same UUIDs as
   * Spark. This function is indeterministic, meaning that it will generate
   * different UUIDs each time it is called, even with the same row count.
   * The UUIDs are generated using a seed based on the current time, process name,
   * GPU UUID and running sequence index, ensuring uniqueness across different
   * runs.
   *
   * E.g.: "123e4567-e89b-12d3-a456-426614174000"
   *
   * @param rowCount Number of UUIDs to generate
   * @return ColumnVector containing UUIDs
   */
  public static ColumnVector randomUUIDs(int rowCount)
```

### DIFF
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -210,6 +210,7 @@ add_library(
   src/SparkResourceAdaptorJni.cpp
   src/SubStringIndexJni.cpp
   src/TaskPriorityJni.cpp
+  src/MiscJni.cpp
   src/ZOrderJni.cpp
   src/aggregation64_utils.cu
   src/bloom_filter.cu
@@ -246,6 +247,7 @@ add_library(
   src/substring_index.cu
   src/timezones.cu
   src/utilities.cu
+  src/uuid.cu
   src/xxhash64.cu
   src/zorder.cu
 )
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -208,9 +208,9 @@ add_library(
   src/RegexRewriteUtilsJni.cpp
   src/RowConversionJni.cpp
   src/SparkResourceAdaptorJni.cpp
+  src/StringUtilsJni.cpp
   src/SubStringIndexJni.cpp
   src/TaskPriorityJni.cpp
-  src/MiscJni.cpp
   src/ZOrderJni.cpp
   src/aggregation64_utils.cu
   src/bloom_filter.cu
--- a/src/main/cpp/src/uuid.cu
+++ b/src/main/cpp/src/uuid.cu
@@ -101,19 +101,23 @@ __launch_bounds__(block_size) CUDF_KERNEL void generate_uuids_kernel(
   curandState localState = state[start_idx];
 
   for (cudf::thread_index_type row_idx = start_idx; row_idx < row_count; row_idx += stride) {
-    auto const idx                       = static_cast<cudf::size_type>(row_idx);
-    double d1                            = curand_uniform_double(&localState);
-    double d2                            = curand_uniform_double(&localState);
-    constexpr uint64_t max_unsigned_long = std::numeric_limits<uint64_t>::max();
-    uint64_t const most                  = static_cast<uint64_t>(d1 * max_unsigned_long);
-    uint64_t const least                 = static_cast<uint64_t>(d2 * max_unsigned_long);
-    char* uuid_ptr                       = uuid_chars + idx * CHAR_COUNT_PER_UUID;
+    auto const idx = static_cast<cudf::size_type>(row_idx);
+    uint64_t i1    = curand(&localState);
+    uint64_t i2    = curand(&localState);
+    uint64_t i3    = curand(&localState);
+    uint64_t i4    = curand(&localState);
+
+    uint64_t const most  = i1 << 32 | i2;
+    uint64_t const least = i3 << 32 | i4;
+    char* uuid_ptr       = uuid_chars + idx * CHAR_COUNT_PER_UUID;
 
     // set the version bits to 4 (UUID version 4): Truly Random or Pseudo-Random
     auto most_sig_bits = (most & 0xFFFFFFFFFFFF0FFFL) | 0x0000000000004000L;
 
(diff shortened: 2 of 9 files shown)


res-life:
TODO:
- Use `curand` instead of `curand_uniform_double`
- Has error when row_count is 1

src/main/cpp/CMakeLists.txt
```diff
@@ -210,6 +210,7 @@ add_library(
   src/SparkResourceAdaptorJni.cpp
   src/SubStringIndexJni.cpp
   src/TaskPriorityJni.cpp
+  src/MiscJni.cpp
```
ttnghia, line 213: Let's call `StringUtilsJni.cpp`, since `UUID` is string.

src/main/cpp/src/uuid.cu
```diff
@@ -0,0 +1,184 @@
...
+template <int block_size>
+__launch_bounds__(block_size) CUDF_KERNEL void generate_uuids_kernel(
+  cudf::size_type const row_count, char* uuid_chars, curandState* state, int const num_states)
+{
+  auto const start_idx = cudf::detail::grid_1d::global_thread_id();
+  auto const stride    = cudf::detail::grid_1d::grid_stride();
+
+  assert(start_idx < num_states);
+
+  curandState localState = state[start_idx];
```
ttnghia, line 96:
```suggestion
  curandState local_state = state[start_idx];
```

src/main/cpp/src/uuid.cu
```diff
@@ -0,0 +1,184 @@
...
+  auto const num_sms       = cudf::detail::num_multiprocessors();
+  int num_blocks_per_sm    = -1;
+
+  // Calculate the maximum number of blocks per multiprocessor
+  CUDF_CUDA_TRY(cudaOccupancyMaxActiveBlocksPerMultiprocessor(
+    &num_blocks_per_sm, generate_uuids_kernel<block_size>, block_size, 0));
+  auto num_states = num_sms * num_blocks_per_sm * block_size;
+
+  // Ensure num_states does not exceed row_count
+  if (num_states > row_count) { num_states = cudf::util::round_up_unsafe(row_count, block_size); }
```
ttnghia, line 145: Then should just assign `num_states = row_count`?

src/main/cpp/src/uuid.cu
```diff
@@ -0,0 +1,184 @@
...
+
+  // initialize curand states
+  rmm::device_uvector<curandState> states(
+    num_states, stream, cudf::get_current_device_resource_ref());
+  init_curand_kernel<block_size>
+    <<<num_states / block_size, block_size, 0, stream.value()>>>(states.data(), num_states, seed);
```
ttnghia, line 151: For launching a kernel, use `grid_1d` class to simplify grid calculation and avoid mistake.

ttnghia, line 151: In addition, this can be replaced by a call to `thrust::for_each` since this init kernel doesn't need to stride through multiple steps.

res-life:
build

src/main/cpp/src/uuid.cu
```diff
@@ -0,0 +1,194 @@
...
+
+std::unique_ptr<cudf::column> generate_uuids(cudf::size_type row_count,
+                                             long seed,
+                                             rmm::cuda_stream_view stream,
+                                             rmm::device_async_resource_ref mr)
+
+{
+  // Check if row_count is positive and does not exceed the maximum limit for a column.
+  CUDF_EXPECTS(row_count > 0, "Row count must be positive.");
+  CUDF_EXPECTS(row_count <= std::numeric_limits<cudf::size_type>::max() / CHAR_COUNT_PER_UUID,
```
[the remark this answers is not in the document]
res-life: Done, refer to the case `testUuidInvalidInput`

src/main/cpp/src/uuid.cu
```diff
@@ -0,0 +1,195 @@
...
+
+    // set the variant bits to 2
+    auto least_sig_bits = (least | 0x8000000000000000L) & 0xBFFFFFFFFFFFFFFFL;
+
+    // convert the most and least significant bits to a string format
+    char* uuid_ptr = uuid_chars + idx * CHAR_COUNT_PER_UUID;
+    convert_uuid_to_chars(most_sig_bits, least_sig_bits, uuid_ptr);
+  }
+
+  state[start_idx] = local_state;
```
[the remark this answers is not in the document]
res-life: Copying data from global memory to local memory in CUDA can indeed improve performance for some access pattern. Here the `curandState` is frequently used,  so copy it to local.

I did benchmark test on my laptop with GPU card: RTX 3500 Ada

local memory perf : 723 + 109 + 116 + 115 + 112 + 117 + 113 = 1405ms
global memory perf: 721 + 124 + 115 + 113 + 109 + 118 + 124 = 1424ms

From the above figures, local memory is a little fast than global memory.
I'm fine to remove the local memory copying becasue the speedup is such small.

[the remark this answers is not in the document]
res-life: Which one do you like? keep the local copying or remove.

ttnghia (approved):

Issue #3627, opened 2025-08-12
[BUG]


pxLi:
**Describe the bug**
build 25.10 JNI in cuda13 failed below, it looks like `libcupti_static.a` in cuda13 was compiled without the -fPIC flag, making it incompatible for linking into shared libraries.
```
[2025-08-11T10:40:45.181Z] [INFO]      [exec] [148/197] Linking CXX shared library profiler/libprofilerjni.so
[2025-08-11T10:40:45.181Z] [INFO]      [exec] FAILED: profiler/libprofilerjni.so 
[2025-08-11T10:40:45.181Z] [INFO]      [exec] : && /opt/rh/gcc-toolset-14/root/usr/bin/c++ -fPIC -O3 -DNDEBUG   -shared -Wl,-soname,libprofilerjni.so -o profiler/libprofilerjni.so profiler/CMakeFiles/profilerjni.dir/ProfilerJni.cpp.o profiler/CMakeFiles/profilerjni.dir/profiler_debug.cpp.o profiler/CMakeFiles/profilerjni.dir/profiler_serializer.cpp.o profiler/CMakeFiles/profilerjni.dir/__/generated/src/spark_rapids_jni_version.cpp.o  -Wl,-rpath,"\$ORIGIN"  /usr/local/cuda/lib64/libcupti_static.a  /home/jenkins/agent/workspace/jenkins-peixinl-spark-rapids-jni_nightly-dev-15-cuda13/target/libcudf-install/lib64/libflatbuffers.a  -ldl && :
[2025-08-11T10:40:45.181Z] [INFO]      [exec] /opt/rh/gcc-toolset-14/root/usr/libexec/gcc/x86_64-redhat-linux/14/ld: /usr/local/cuda/lib64/libcupti_static.a(libcupti_static_partial_mangled.o): warning: relocation against `_nv024841cupti' in read-only section `.text'
[2025-08-11T10:40:45.181Z] [INFO]      [exec] /opt/rh/gcc-toolset-14/root/usr/libexec/gcc/x86_64-redhat-linux/14/ld: /usr/local/cuda/lib64/libcupti_static.a(libcupti_static_partial_mangled.o): relocation R_X86_64_PC32 against symbol `_nv024841cupti' can not be used when making a shared object; recompile with -fPIC
[...]

PR #3657, opened 2025-08-20
Disable libprofiler.so build for cuda13 given libcupti_static issues

abellina:
Temporary workaround for https://github.com/NVIDIA/spark-rapids-jni/issues/3627

This disables the profiler and cufaultinj for cuda13 in build-in-docker and also in the ci nightlies. We are doing this to get the cuda13 build going, and once we resolve the libcupti_static issues we'll bring them back.

Note that this will break a plugin that tries to use the profiler for cuda13. A subsequent change in the plugin will be raised that will make the profiler fallback gracefully, and warn the user on the driver about the issue (since libprofiler won't be around to link it).


ttnghia (approved):

abellina:
build

pxLi:
Thanks for the workaround! I will enable cuda13 build after https://github.com/rapidsai/cudf/issues/19710 is resolved

pxLi (approved):

PR #3677, opened 2025-08-28
Make `randomUUIDsWithSeed` method public

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/13324

We want to cache `Cuda.getGpuUuid()` for a Spark executor.
So we need to move the following method to Spark-Rapids repo, and cache `Cuda.getGpuUuid()`:
```Java
private static long randomSeed()
```
And call `randomUUIDsWithSeed` instead of `randomUUIDs(int rowCount)`
So make `randomUUIDsWithSeed` public

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/StringUtils.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/StringUtils.java
@@ -70,9 +70,25 @@ public class StringUtils {
   }
 
   /**
-   * Only for test purpose, please do not use it in production.
+   * Generate a column of UUIDs (String type) with `rowCount` rows.
+   * The same `seed` will generate the same sequence of UUIDs.
+   * Spark uses `Truly Random or Pseudo-Random` UUID type which is described in
+   * the section 4.4 of [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122),
+   * The variant in UUID is 2 and the version in UUID is 4. This implementation
+   * generates UUIDs in the same format, but does not generate the same UUIDs as
+   * Spark. This function is indeterministic, meaning that it will generate
+   * different UUIDs each time it is called, even with the same row count.
+   * The UUIDs are generated using a seed based on the current time, process name,
+   * GPU UUID and running sequence index, ensuring uniqueness across different
+   * runs.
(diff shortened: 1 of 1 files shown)


res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/StringUtils.java
```diff
@@ -70,9 +70,25 @@ public static ColumnVector randomUUIDs(int rowCount) {
...
+   * runs.
+   *
+   * E.g.: "123e4567-e89b-12d3-a456-426614174000"
+   *
+   * @param rowCount Number of UUIDs to generate
+   * @param seed Seed for UUID generation
+   * @return ColumnVector containing UUIDs
    */
-  static ColumnVector randomUUIDsWithSeed(int rowCount, long seed) {
+  public static ColumnVector randomUUIDsWithSeed(int rowCount, long seed) {
```
firestarman, line 91: Any validity check for these two parameters since it becomes public ?

res-life: In `uuid.cu`:

```cpp
  // Check if row_count is positive and does not exceed the maximum limit for a column.
  CUDF_EXPECTS(row_count > 0, "Row count must be positive.");
  CUDF_EXPECTS(row_count <= std::numeric_limits<cudf::size_type>::max() / CHAR_COUNT_PER_UUID,
               "Row count exceeds the maximum limit for UUID generation.",
               std::overflow_error);
```

firestarman (approved):
LGTM

revans2 (commented):
I don't understand why we are making this public. Why not have the following instead?

```
private static byte [] gpuUUID = null;

private static synchronized byte[] getCachedGpuUUID() {
  if (gpuUUID == null) {
    gpuUUID = Cuda.getGpuUuid();
  }
  return gpuUUID;
}
```

res-life:
There can be multiple GPUs,  the `Cuda.getGpuUuid` is to get the UUID of current active GPU. 
So we need to store the cache on the excutor side, since only one GPU is assigned to excecutor.

PR #3690, opened 2025-09-02
add jni interface for collecting gpu metrics with NVML

binmahone:
This PR introduces a JNI interface for collecting GPU metrics using NVIDIA's NVML library, as well as a GPU monitor class for polling these APIs.

This change does not close, but is a step towards resolving https://github.com/NVIDIA/spark-rapids/issues/12470

### DIFF
--- a/src/main/cpp/src/NVMLJni.cpp
+++ b/src/main/cpp/src/NVMLJni.cpp
@@ -567,8 +567,7 @@ Java_com_nvidia_spark_rapids_jni_nvml_NVML_nvmlGetAllGPUInfo(JNIEnv* env, jclass
       cpp_result = populate_gpu_info_from_device(env, device);
     } else {
       // Failed to get device handle
-      cpp_result.return_code = device_result;
-      cpp_result.data        = nullptr;
+      cpp_result = nvml_result{device_result, nullptr};
     }
 
     // Create Java nvml_result object and add to array
--- a/src/main/cpp/tests/nvml_test.cpp
+++ b/src/main/cpp/tests/nvml_test.cpp
@@ -158,8 +158,8 @@ TEST_F(NVMLTest, NVMLDeviceGetNumGpuCores_Succeeds)
   nvmlDevice_t device = getFirstDevice();
   ASSERT_NE(device, nullptr) << "No NVML devices available";
 
-  unsigned int numCores;
-  nvmlReturn_t result = nvmlDeviceGetNumGpuCores(device, &numCores);
+  unsigned int num_cores;
+  nvmlReturn_t result = nvmlDeviceGetNumGpuCores(device, &num_cores);
   EXPECT_TRUE(result == NVML_SUCCESS || result == NVML_ERROR_NOT_SUPPORTED);
 }
 
@@ -168,8 +168,8 @@ TEST_F(NVMLTest, NVMLDeviceGetPerformanceState_Succeeds)
   nvmlDevice_t device = getFirstDevice();
   ASSERT_NE(device, nullptr) << "No NVML devices available";
 
-  nvmlPstates_t pState;
-  nvmlReturn_t result = nvmlDeviceGetPerformanceState(device, &pState);
+  nvmlPstates_t p_state;
+  nvmlReturn_t result = nvmlDeviceGetPerformanceState(device, &p_state);
   EXPECT_TRUE(result == NVML_SUCCESS || result == NVML_ERROR_NOT_SUPPORTED);
 }
 
@@ -208,9 +208,9 @@ TEST_F(NVMLTest, NVMLDeviceGetTotalEccErrors_SingleBit_Succeeds)
   nvmlDevice_t device = getFirstDevice();
   ASSERT_NE(device, nullptr) << "No NVML devices available";
 
-  unsigned long long eccCount;
+  unsigned long long ecc_count;
   nvmlReturn_t result =
-    nvmlDeviceGetTotalEccErrors(device, NVML_SINGLE_BIT_ECC, NVML_VOLATILE_ECC, &eccCount);
+    nvmlDeviceGetTotalEccErrors(device, NVML_SINGLE_BIT_ECC, NVML_VOLATILE_ECC, &ecc_count);
   // ECC error reporting may not be supported on all GPUs
   EXPECT_TRUE(result == NVML_SUCCESS || result == NVML_ERROR_NOT_SUPPORTED);
 }


binmahone:
I'm changing the PR to "ready for review" to receive comments per @zpuller's suggestions, cc @abellina . I have already tested it with https://github.com/NVIDIA/spark-rapids/pull/13398 locally, and the the outcome in logging. But I'm still seeking opportunities to verify it in a real case in real customer environment.

zpuller:
I'm noting a TODO that we discussed offline which is to add some tests here and potentially safeguards against potential compatibility issues

zpuller:
build

src/main/java/com/nvidia/spark/rapids/jni/nvml/NVMLMonitor.java
```diff
@@ -0,0 +1,263 @@
...
+     */
+    public void setCallback(MonitoringCallback callback) {
+        this.callback = callback;
+    }
+
+    /**
+     * Start continuous GPU monitoring
+     */
+    public synchronized void startMonitoring() {
+        if (monitoring.get()) {
```
pmattione-nvidia: how asynchronous of execution is this supposed to support? E.g. can two threads call startMonitoring() at the same time? if so they could both get past this monitoring.get() check, then both set monitoring.set(true) below and create threads, etc.  If this is possible this needs a different sync primitive like a strong CAS.

src/main/java/com/nvidia/spark/rapids/jni/nvml/NVMLMonitor.java
```diff
@@ -0,0 +1,263 @@
...
+        if (callback != null) {
+            callback.onMonitoringStarted();
+        }
+    }
+
+    /**
+     * Stop continuous GPU monitoring
+     */
+    public synchronized void stopMonitoring() {
+        if (!monitoring.get()) {
```
pmattione-nvidia: Same comment here regarding synchronicity.  I.e. can two threads call stopMonitoring() at the same time? Also can one thread call startMonitoring() at about the same time another calls stopMonitoring()? If so we need a lock.

src/main/cpp/src/NVMLJni.cpp
```diff
@@ -0,0 +1,509 @@
...
+  if (powerInfo == nullptr) return nullptr;
+
+  jclass powerInfoClass = env->GetObjectClass(powerInfo);
+
+  jfieldID powerUsageField = env->GetFieldID(powerInfoClass, "powerUsageW", "I");
+  jfieldID powerLimitField = env->GetFieldID(powerInfoClass, "powerLimitW", "I");
+
+  unsigned int power;
+  nvmlReturn_t result = nvmlDeviceGetPowerUsage(device, &power);
+  if (result == NVML_SUCCESS) {
```
pmattione-nvidia: What happens if calls like these fail? If we don't set the field, does it have an uninitialized or zero garbage value, that then gets averaged in with the rest of our good values?  Perhaps we need a way to propagate up failure so that bad information isn't saved.

src/main/java/com/nvidia/spark/rapids/jni/nvml/NVMLMonitor.java
```diff
@@ -0,0 +1,274 @@
...
+    private void monitoringLoop() {
+        while (monitoring.get()) {
+            try {
+                NVMLResult<GPUInfo>[] results = NVML.getAllGPUInfo();
+
+                if (results != null && results.length > 0) {
+                    // Extract successful GPUInfo objects
+                    List<GPUInfo> successfulInfos = new ArrayList<>();
+                    for (NVMLResult<GPUInfo> result : results) {
+                        if (result.isSuccess() && result.getData() != null) {
```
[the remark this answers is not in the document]
zpuller: Yeah I just noticed that too, thinking about the best way to structure it

src/main/java/com/nvidia/spark/rapids/jni/nvml/NVMLMonitor.java
```diff
@@ -0,0 +1,274 @@
...
+        if (!NVML.isAvailable()) {
+            logger.error("NVML not available, cannot start monitoring");
+            return;
+        }
+
+        // Initialize lifecycle stats for all GPUs
+        NVMLResult<GPUInfo>[] initialResults = NVML.getAllGPUInfo();
+        for (int i = 0; i < initialResults.length; i++) {
+            NVMLResult<GPUInfo> result = initialResults[i];
+            if (result.isSuccess() && result.getData() != null && result.getData().deviceInfo != null) {
```
pmattione-nvidia: this check might have the same problem ... requiring global success just to get the device name

zpuller:
build

res-life:
On the Spark-Rapids side, we can use Spark message machenism like Filecache.
Let executor send messages to driver, driver handles the message from executors.
Receive messages: [code](https://github.com/NVIDIA/spark-rapids/blob/v25.08.0/sql-plugin/src/main/scala/com/nvidia/spark/rapids/Plugin.scala#L449)
Send messages:  Executor sends message using `PluginContext.send(message)`.
Executors only monitor the GPUs they are using. If monitoring is enabled, executors use a timer to trigger collecting Gpu usage info via this JNI in periodically. 
The messages should be serializable.
Spark driver collects messages from all the executors.

pmattione-nvidia (approved):

src/main/cpp/src/NVMLJni.cpp
```diff
@@ -0,0 +1,684 @@
...
+struct NVMLResult {
+  nvmlReturn_t returnCode;
+  jobject data;
+};
+
+// Helper functions for individual NVML API groups
+namespace {
+
+// Helper function to create NVMLResult Java object from C++ NVMLResult
+jobject create_nvml_result(JNIEnv* env, NVMLResult cppResult)
```
ttnghia:
```suggestion
jobject create_nvml_result(JNIEnv* env, NVMLResult const& cpp_result)
```

src/main/cpp/src/NVMLJni.cpp
```diff
@@ -0,0 +1,684 @@
...
+#include <stdio.h>
+
+#include <cstdint>
+
+// NVML JNI implementation with comprehensive GPU metrics for Spark Rapids
+
+#define NVML_CLASS_PATH "com/nvidia/spark/rapids/jni/nvml/"
+
+// C++ struct to hold NVML operation results
+struct NVMLResult {
```
ttnghia: C++ convention: `snake_case` for variable/struct/class names.
```suggestion
struct nvml_result {
```

src/main/cpp/src/NVMLJni.cpp
```diff
@@ -0,0 +1,684 @@
...
+  nvmlReturn_t returnCode;
```
ttnghia:
```suggestion
  nvmlReturn_t return_code;
```

src/main/cpp/src/NVMLJni.cpp
```diff
@@ -0,0 +1,684 @@
...
+
+  jclass deviceInfoClass = env->GetObjectClass(deviceInfo);
+
+  jfieldID nameField  = env->GetFieldID(deviceInfoClass, "name", "Ljava/lang/String;");
+  jfieldID brandField = env->GetFieldID(deviceInfoClass, "brand", "Ljava/lang/String;");
+
+  char name[NVML_DEVICE_NAME_BUFFER_SIZE];
+  nvmlReturn_t nvmlResult = nvmlDeviceGetName(device, name, NVML_DEVICE_NAME_BUFFER_SIZE);
+  result.returnCode       = nvmlResult;
+  result.data             = deviceInfo;
```
ttnghia:
```suggestion
  auto return_code = nvmlDeviceGetName(device, name, NVML_DEVICE_NAME_BUFFER_SIZE);
  result = nvml_result{return_code, device_info};
```

src/main/cpp/src/NVMLJni.cpp
```diff
@@ -0,0 +1,684 @@
...
+  return result;
+}
+
+NVMLResult populate_utilization_info(JNIEnv* env, nvmlDevice_t device)
+{
+  NVMLResult result;
+  result.returnCode = NVML_ERROR_UNKNOWN;
+  result.data       = nullptr;
+
+  jobject utilizationInfo = create_object(env, NVML_CLASS_PATH "GPUUtilizationInfo", "()V");
```
ttnghia: C++ name convention please.

zpuller:
build

PR #3693, opened 2025-09-03
Remove prefetching

vyasr:
Spark-RAPIDS does not ever need this call since managed memory is never turned on here and no other prefetching is used in Spark. I'm guessing that this was just copied over from cudf code.


ttnghia:
build

PR #3779, opened 2025-09-26
Convert between non-DST reader/writer timezones for ORC

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/13437

### summary
Implements `org.apache.orc.impl.SerializationUtils.convertBetweenTimezones`
Refer to link: https://github.com/apache/orc/blob/rel/release-1.9.1/java/core/src/java/org/apache/orc/impl/SerializationUtils.java#L1440

This PR only implements for non-DST timezones.
For DST timezone conversions, will come in a following PR.

<details>
<summary>For more details, click to expand</summary>

```java
  public static long convertBetweenTimezones(TimeZone writer, TimeZone reader,
                                             long millis) {
    final long writerOffset = writer.getOffset(millis);
    final long readerOffset = reader.getOffset(millis);
    long adjustedMillis = millis + writerOffset - readerOffset;
    // If the timezone adjustment moves the millis across a DST boundary, we
    // need to reevaluate the offsets.
    long adjustedReader = reader.getOffset(adjustedMillis);
    return writerOffset - adjustedReader;
  }
```
</details>

### Analysis

`java.util.TimeZone.getOffset` and `java.time.ZoneId.getOffset` are not always consistent.
<details>
<summary>For more details, click to expand</summary>

```java
static void testDiffBehaviorBetweenTwoAPIs() {
  // diff in `Africa/Casablanca` timezone at 6424721300000: 0 vs 3600000
  String tzId = "Africa/Casablanca";
  long epochMillis = 6424721300000L;
  int offsetMillis_1 = java.util.TimeZone.getTimeZone(tzId).getOffset(epochMillis);
  int offsetMillis_2 = java.time.ZoneId.of(tzId, ZoneId.SHORT_IDS).getRules().getOffset(Instant.ofEpochMilli(epochMillis)).getTotalSeconds() * 1000;
  if (offsetMillis_1 != offsetMillis_2) {
    // print: get diff!! 0 vs 3600000
    System.out.println("get diff!! " + offsetMillis_1 + " vs " + offsetMillis_2);
  }
}
```
</details>

Although Spark-Rapids has GPU version for `java.time.ZoneId.getOffset`, we can not use here.
ORC is using `java.util.TimeZone.getOffset`, so we need to implement GPU version for `util.TimeZone`

### design

`ZoneInfo` packs transition, offset index and other info into a long value.
Get timezone info from `sun.util.calendar.ZoneInfo` using Java reflection, then pass into kernel to do binary search.

### Others:
- `sun.util.calendar.ZoneInfo` is not public JAVA API. We can not use this class directly for Oracle Java release, so serialize the timezone info into a file.
- `GpuTimeZoneDBTest` spends about 1.5 minutes

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
@@ -52,15 +52,7 @@ class OrcTimezoneInfo {
   // - actual offset = offsets[transition & 0x0FL]
   // For more details, please refer to `sun.util.calendar.ZoneInfo` source code.
 
-  // There are 652 timezones, the file `orc_timezone_info.data` content:
-  // First part: 632 indices(integer) to store the start points in the file for each timezone.
-  // Seconds part: 632 serialized timezone info.
-  // Each timezone info content:
-  // rowOffset: 4 bytes
-  // num of transitions(num_transitions): 4 bytes
-  // transitions: num_transitions * 8 bytes
-  // num of offsets(num_offsets): 4 bytes
-  // offsets: num_offsets * 4 bytes
+  // Refer to `serializeTimezoneInfo` method for how to generate the file.
   private static final String ORC_TIMEZONE_FILE = "orc_timezone_info.data";
 
   // the mapped memory for the file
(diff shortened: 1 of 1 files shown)


res-life:
build

src/main/cpp/src/GpuTimeZoneDBJni.cpp
```diff
@@ -93,4 +93,28 @@ Java_com_nvidia_spark_rapids_jni_GpuTimeZoneDB_convertTimestampColumnToUTCWithTz
...
+    cudf::jni::auto_set_device(env);
+    auto const input              = reinterpret_cast<cudf::column_view const*>(input_handle);
+    auto const writer_tz_info_tab = reinterpret_cast<cudf::table_view const*>(writer_tz_info_table);
+    auto const reader_tz_info_tab = reinterpret_cast<cudf::table_view const*>(reader_tz_info_table);
+    return cudf::jni::ptr_as_jlong(
+      spark_rapids_jni::convert_between_timezones(
+        *input, writer_tz_info_tab, writer_tz_raw_offset, reader_tz_info_tab, reader_tz_raw_offset)
+        .release());
+  }
+  CATCH_STD(env, 0);
```
ttnghia, line 118: Please adopt the new convention `JNI_TRY`/`JNI_CATCH` as introduced in https://github.com/NVIDIA/spark-rapids-jni/pull/3714.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -36,6 +37,77 @@
 import java.util.*;
 import java.util.concurrent.Executors;
 
+/**
+ * Used to save timezone info from `sun.util.calendar.ZoneInfo`
+ */
+class TimeZoneInfoInJavaUtilPackage implements AutoCloseable {
```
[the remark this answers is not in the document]
res-life: Done. Renamed to `TzInfoInJavaUtilForORC`

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -246,4 +247,139 @@ __device__ static timestamp_type convert_timestamp(
...
+
+}  // namespace
+
+/**
+ * @brief Find the relative offset when moving between timezones at a particular point in time.
+ * This is for ORC timezone support.
+ * This function implements `org.apache.orc.impl.SerializationUtils.convertBetweenTimezones`
+ * Refer to link: https://github.com/apache/orc/blob/rel/release-1.9.1/java/core/src/
+ * java/org/apache/orc/impl/SerializationUtils.java#L1440
+ * If the input `trans_begin` == `trans_begin` == 0, it means the timezone is fixed offset.
```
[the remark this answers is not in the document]
mythrocks: Ah, ok.  It should apply to both.  Thanks for changing to `nullptr`.  I did think that would be better, but didn't want to nitpick.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -647,6 +721,103 @@ public static boolean isDST(String timezone) {
...
+    // transitions info.
+    return new TzInfoInJavaUtilForORC(
+        (long[]) FieldUtils.readField(zoneInfo, "transitions"),
+        (int[]) FieldUtils.readField(zoneInfo, "offsets"),
+        (int) FieldUtils.readField(zoneInfo, "rawOffset"));
+  }
+
+  private static ColumnVector getTransitionsForUtilTZ(TzInfoInJavaUtilForORC info) {
+    try (HostColumnVector hcv = HostColumnVector.fromLongs(info.transitions)) {
+      return hcv.copyToDevice();
```
mythrocks: Please correct my understanding of the scale here:  I'm guessing the number of transitions per time-zone id is of the order of a few hundreds?

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -647,6 +721,103 @@ public static boolean isDST(String timezone) {
...
+      return new Table(trans);
+    } catch (Exception e) {
+      throw new IllegalStateException("get timezone info from java.util.TimeZone failed!", e);
+    }
+  }
+
+  /**
+   * Is Daylight Saving Time for the given timezone id
+   */
+  private static boolean isDaylightSavingTime(String timezoneId) {
```
mythrocks: This isn't a check to see if the current time is within DST, right?  It is to check whether the specified timezone-id has any DST transitions at all?

I think `hasDaylightSavingTime` might be a more appropriate name.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -647,6 +721,103 @@ public static boolean isDST(String timezone) {
...
+   * @return timestamp column in microseconds after converting between timezones
+   */
+  public static ColumnVector convertBetweenTimezones(
+      ColumnVector input,
+      String writerTimezone,
+      String readerTimezone) {
+    // Does not support DST timezone now, just throw exception.
+    if (isDaylightSavingTime(writerTimezone) ||
+        isDaylightSavingTime(readerTimezone)) {
+      throw new UnsupportedOperationException("Daylight Saving Time is not supported now.");
```
mythrocks: 👏

src/main/cpp/src/timezones.cu
```diff
@@ -243,6 +243,246 @@ std::unique_ptr<column> convert_to_utc_with_multiple_timezones(
...
+
+/**
+ * @brief Get the transition index for the given time `time_ms` using binary search.
+ * The transition array is sorted, each int64 value is composed of 44 bits
+ * transition time in seconds and 20 bits offset in seconds.
+ * @param begin the beginning of the transition array.
+ * @param end the end of the transition array.
+ * @param time_ms the input time in milliseconds to find the transition index for.
+ */
+__device__ static int get_transition_index(int64_t const* begin,
```
mythrocks: The following is a possibly invasive suggestion.  (If it doesn't fit per current timeline, we can consider it for a follow-up.)

The fact that we have explicit code for binary-search instead of using `thrust::lower_bound()` got me thinking:

Is there value in converting the transition array into milliseconds (after shifting it by 20) first?  Then the rest of the comparison might simply be a `lower_bound` call.  It will save shifting/multiplying log(numTransitions) times per input row.

res-life:
build

src/main/cpp/src/timezones.hpp
```diff
@@ -119,4 +119,37 @@ std::unique_ptr<cudf::column> convert_timestamp_to_utc(
...
+ * @param reader_tz_info_table The reader timezone table which contains a transition column and a
+ * timezone index column both in milliseconds.
+ *
+ * @param reader_raw_offset the raw offset in seconds.
+ *
+ * @param stream CUDA stream used for device memory operations and kernel launches.
+ *
+ * @param mr Device memory resource used to allocate the returned timestamp column's memory
+ *
+ * @return a column of timestamps rebased between writer and reader timezones.
```
ttnghia:
```suggestion
 * @param input The input timestamp column in microseconds.
 * @param writer_tz_info_table The writer timezone table which contains a transition column and a
 *        timezone index column both in milliseconds.
 * @param writer_raw_offset the raw offset in seconds.
 * @param reader_tz_info_table The reader timezone table which contains a transition column and a
 *        timezone index column both in milliseconds.
 * @param reader_raw_offset the raw offset in seconds.
 * @param stream CUDA stream used for device memory operations and kernel launches.
 * @param mr Device memory resource used to allocate the returned timestamp column's memory
 * @return a column of timestamps rebased between writer and reader timezones.
```

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -26,15 +26,127 @@
...
+   * Java public API, e.g.: Oracle JDKs have a different package name.
+   * We cannot use it directly, so we dump the info to a file.
+   * The file is generaged from OpenJDK 8. So for newer JDKs, some timezones
+   * is missing.
+   *
+   * @return a map from timezone ID to OrcTimezoneInfo
+   */
+  static Map<String, OrcTimezoneInfo> readOrcTzInfo() {
+
+    final String fileName = "timezone_info_for_orc.csv";
```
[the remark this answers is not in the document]
ttnghia: Also move this into the class private static variable to easier identify.

src/main/java/com/nvidia/spark/rapids/jni/FieldUtils.java
```diff
@@ -0,0 +1,60 @@
...
+public class FieldUtils {
+
+  /**
+   * Reads a field value via reflection from a target object.
+   * @param target target object to read from
+   * @param fieldName name of the field to read
+   * @return the field value
+   * @throws IllegalStateException if the field cannot be read for any reason
+   */
+  public static Object readField(Object target, String fieldName) {
```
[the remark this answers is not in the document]
mythrocks: This used to be used to read the private fields in `ZoneInfo`.  That info has now been moved into the `timezone_info_for_orc.csv` file.

@res-life can confirm if this is still needed.  It's a useful utility, just not being used here.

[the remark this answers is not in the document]
res-life: Yes, it's not needed. But we can retain it.
Currently Spark-Rapids Iceberg is using a shaded `FieldUtils`: 

```scala
import org.apache.hadoop.shaded.org.apache.commons.lang3.reflect.FieldUtils
```

After this is merged, we can replace the imports in Iceberg code.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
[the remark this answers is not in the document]
res-life: Moved to src/main/resources, it's the conventional path.

res-life:
First getting timezone info spent 15ms.
The second time, it spent 0.3ms.
First time spent more time because it include the reading file to memory.
The 2nd and next times, do not need to read file again.

ttnghia:
> First getting timezone info spent 15ms. The second time, it spent 0.3ms. First time spent more time because it include the reading file to memory. The 2nd and next times, do not need to read file again.

Is it better to map file like this at startup, or load from disk using normal file reader when the buffer is accessed the first time?

src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
```diff
@@ -0,0 +1,133 @@
...
+
+  // There are 652 timezones, the file `orc_timezone_info.data` content:
+  // First part: 632 indices(integer) to store the start points in the file for each timezone.
+  // Seconds part: 632 serialized timezone info.
+  // Each timezone info content:
+  // rowOffset: 4 bytes
+  // num of transitions(num_transitions): 4 bytes
+  // transitions: num_transitions * 8 bytes
+  // num of offsets(num_offsets): 4 bytes
+  // offsets: num_offsets * 4 bytes
```
mythrocks: 👏

src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
```diff
@@ -0,0 +1,133 @@
...
+    long[] transitions = new long[numTransitions];
+    for (int i = 0; i < numTransitions; ++i) {
+      transitions[i] = buf.getLong();
+    }
+
+    int numOffsets = buf.getInt();
+    int[] offsets = new int[numOffsets];
+    for (int i = 0; i < numOffsets; ++i) {
+      offsets[i] = buf.getInt();
+    }
```
mythrocks: Ok, I think I understand how this has been arranged.  I see how this should work.

res-life:
> It would be good to check in the code with which this was generated

Done.

PR #3783, opened 2025-09-26
Fix build due to removing of cudf's  `experimental::` row operators

ttnghia:
The cudf's row operators are migrating out of the `experimental::` namespace in https://github.com/rapidsai/cudf/pull/20097. As such, we also need to adopt such changes.

Depends on: 
 * https://github.com/rapidsai/cudf/pull/20097


PointKernel (commented):
Thanks a lot @ttnghia

PR #3843, opened 2025-10-14
Kudo mixed null fix

nvdbaranec:
Fixes an issue with `shuffle_assemble` where in certain conditions of mixed input partition validity (some partitions for column X containing a validity vector and others not containing a validity vector) would produce bad output. Column partitions that should have been all valid ended up being all null.


zpuller:
build

PR #3851, opened 2025-10-16
Fix cpu kudo overflow when merging table.

liurenjie1024:
Close https://github.com/NVIDIA/spark-rapids/issues/13612

Fix another potential overflow in kudo table merger.

In this pr, we handle a corner case where length of single buffer(data buffer or offset buffer) may exceed integer max value, also added test cases to verify .


liurenjie1024:
This pr is ready for review. cc @abellina @revans2 @res-life  PTAL

pmattione-nvidia (approved):

sameerz:
build

PR #3858, opened 2025-10-20
Add sort_merge joins beyond what cudf supports

revans2:
Please note that this is very much a hack. the code was written by AI over the weekend and it is far from great. It effectively does an inner sort merge join followed by an AST filter in a separate kernel, and then a fixup kernel to change the inner join to left outer, left semi, and left anti. I know there is a lot more that we can do to improve this more, and I plan to work on that. But this is to try and get something working ASAP, so we can test it with a customer.

This addresses some of https://github.com/rapidsai/cudf/issues/20301 and https://github.com/NVIDIA/spark-rapids/issues/13618 but it is far from a final solution.

In profiling it shows that the query that caused the issues (25 million rows on the right table) now takes 2 seconds, when before we didn't even wait for it to finish. When I look at the actual time taken using an nsys trace the join itself appears to be closer to 240ms, with 210ms going to sorting the data.

### DIFF
--- a/src/main/cpp/src/mixed_sort_merge_join.cu
+++ b/src/main/cpp/src/mixed_sort_merge_join.cu
@@ -533,8 +533,8 @@ mixed_sort_merge_inner_join(cudf::table_view const& left_equality,
   }
 
   // Check for nulls in conditional columns (top-level and nested)
-  auto const has_nulls = cudf::has_nested_nulls(left_conditional) ||
-                         cudf::has_nested_nulls(right_conditional);
+  auto const has_nulls =
+    cudf::has_nested_nulls(left_conditional) || cudf::has_nested_nulls(right_conditional);
 
   // Parse the AST expression (intermediate)
   auto const parser = cudf::ast::detail::expression_parser{binary_predicate,
@@ -620,8 +620,8 @@ mixed_sort_merge_left_join(cudf::table_view const& left_equality,
   }
 
   // Check for nulls in conditional columns (top-level and nested)
-  auto const has_nulls = cudf::has_nested_nulls(left_conditional) ||
-                         cudf::has_nested_nulls(right_conditional);
+  auto const has_nulls =
+    cudf::has_nested_nulls(left_conditional) || cudf::has_nested_nulls(right_conditional);
 
   // Parse the AST expression (intermediate)
   auto const parser = cudf::ast::detail::expression_parser{binary_predicate,


revans2:
build

revans2:
build

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+  cudf::size_type const* __restrict__ left_indices,
+  cudf::size_type const* __restrict__ right_indices,
+  cudf::size_type num_pairs,
+  cudf::table_device_view left_table,
+  cudf::table_device_view right_table,
+  cudf::ast::detail::expression_device_view device_expression_data,
+  bool* __restrict__ keep_mask)
+{
+  extern __shared__ char raw_intermediate_storage[];
+  cudf::ast::detail::IntermediateDataType<has_nulls>* intermediate_storage =
```
nvdbaranec: Since we're playing pointer games here with dynamic shared memory, I think it would be a good idea to make `cudf::ast::detail::IntermediateDataType<has_nulls>` a typedef to protect against discrepancies with the kernel call site.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+                                                            right_table,
+                                                            device_expression_data,
+                                                            keep_mask.data());
+
+  // Surface any kernel launch errors immediately
+  CUDF_CUDA_TRY(cudaPeekAtLastError());
+
+  // Count the number of true values in the mask
+  auto const num_matches =
+    thrust::count(rmm::exec_policy(stream), keep_mask.begin(), keep_mask.end(), true);
```
nvdbaranec: Seems like it would be pretty easy for the `filter_join_indices_kernel` to just compute and return this instead of invoking another kernel.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+    keep_mask[i] = output_dest.is_valid() && output_dest.value();
+  }
+}
+
+/**
+ * @brief Filter join results by evaluating conditional expression
+ */
+template <bool has_nulls>
+std::pair<std::unique_ptr<rmm::device_uvector<cudf::size_type>>,
+          std::unique_ptr<rmm::device_uvector<cudf::size_type>>>
```
nvdbaranec: Style-wise these days you'd just normally return the `rmm::device_vectors` directly and let the compiler RVO take care of it. This AI is living in the 2010s.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+                        cudf::size_type left_num_rows,
+                        cudf::size_type right_num_rows,
+                        rmm::cuda_stream_view stream,
+                        rmm::device_async_resource_ref mr)
+{
+  CUDF_FUNC_RANGE();
+
+  // Create a boolean mask to track which left rows have matches
+  auto left_has_match = rmm::device_uvector<bool>(left_num_rows, stream, mr);
+  thrust::fill(rmm::exec_policy(stream), left_has_match.begin(), left_has_match.end(), false);
```
nvdbaranec:
```suggestion
  thrust::fill(rmm::exec_policy_nosync(stream), left_has_match.begin(), left_has_match.end(), false);
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+
+  // Mark left rows that have matches
+  thrust::for_each(rmm::exec_policy(stream),
```
nvdbaranec:
```suggestion
  thrust::for_each(rmm::exec_policy_nosync(stream),
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+  auto out_right_indices =
+    std::make_unique<rmm::device_uvector<cudf::size_type>>(num_matches, stream, mr);
+
+  // Copy indices where condition is true
+  auto input_iter =
+    thrust::make_zip_iterator(thrust::make_tuple(left_indices.begin(), right_indices.begin()));
+  auto output_iter = thrust::make_zip_iterator(
+    thrust::make_tuple(out_left_indices->begin(), out_right_indices->begin()));
+
+  thrust::copy_if(rmm::exec_policy(stream),
```
nvdbaranec:
```suggestion
  thrust::copy_if(rmm::exec_policy_nosync(stream),
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+    auto right_indices =
+      std::make_unique<rmm::device_uvector<cudf::size_type>>(left_keys.num_rows(), stream, mr);
+
+    thrust::sequence(rmm::exec_policy(stream), left_indices->begin(), left_indices->end());
+    thrust::fill(rmm::exec_policy(stream),
+                 right_indices->begin(),
+                 right_indices->end(),
+                 right_keys.num_rows());
+
+    return {std::move(left_indices), std::move(right_indices)};
```
nvdbaranec: Same comment on returning the vectors directly.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+  // Get left semi join result (rows that DO match)
+  auto semi_indices = sort_merge_left_semi_join(
+    left_keys, right_keys, is_left_sorted, is_right_sorted, compare_nulls, stream, mr);
+
+  // Find complement - rows that DON'T match
+  auto left_has_match = rmm::device_uvector<bool>(left_num_rows, stream, mr);
+  thrust::fill(rmm::exec_policy(stream), left_has_match.begin(), left_has_match.end(), false);
+
+  // Mark left rows that have matches
+  thrust::for_each(rmm::exec_policy(stream),
```
nvdbaranec:
```suggestion
  thrust::for_each(rmm::exec_policy_nosync(stream),
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+                               *right_table,
+                               parser.device_expression_data,
+                               parser.device_expression_data.num_intermediates,
+                               has_nulls,
+                               stream,
+                               mr);
+}
+
+std::pair<std::unique_ptr<rmm::device_uvector<cudf::size_type>>,
+          std::unique_ptr<rmm::device_uvector<cudf::size_type>>>
```
nvdbaranec: Same as above.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+
+  // Handle empty right table case
+  if (right_equality.num_rows() == 0) {
+    auto left_indices =
+      std::make_unique<rmm::device_uvector<cudf::size_type>>(left_equality.num_rows(), stream, mr);
+    auto right_indices =
+      std::make_unique<rmm::device_uvector<cudf::size_type>>(left_equality.num_rows(), stream, mr);
+
+    thrust::sequence(rmm::exec_policy(stream), left_indices->begin(), left_indices->end());
+    thrust::fill(rmm::exec_policy(stream),
```
nvdbaranec:
```suggestion
    thrust::fill(rmm::exec_policy_nosync(stream),
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+  // Step 3: Add back left rows with no matches
+  return add_left_unmatched_rows(std::move(*filtered_left_indices),
+                                 std::move(*filtered_right_indices),
+                                 left_equality.num_rows(),
+                                 right_equality.num_rows(),
+                                 stream,
+                                 mr);
+}
+
+std::unique_ptr<rmm::device_uvector<cudf::size_type>> mixed_sort_merge_left_semi_join(
```
nvdbaranec: Same device vector comment.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+                                                      binary_predicate,
+                                                      is_left_sorted,
+                                                      is_right_sorted,
+                                                      compare_nulls,
+                                                      stream,
+                                                      mr);
+
+  // Step 2: Find complement - rows that DON'T match
+  auto left_has_match = rmm::device_uvector<bool>(left_num_rows, stream, mr);
+  thrust::fill(rmm::exec_policy(stream), left_has_match.begin(), left_has_match.end(), false);
```
nvdbaranec:
```suggestion
  thrust::fill(rmm::exec_policy_nosync(stream), left_has_match.begin(), left_has_match.end(), false);
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,759 @@
...
+
+  // Mark left rows that have matches
+  thrust::for_each(rmm::exec_policy(stream),
```
nvdbaranec:
```suggestion
  thrust::for_each(rmm::exec_policy_nosync(stream),
```

nvdbaranec (changes requested):
Reviewed the cuda stuff.  Mostly small comments.  But, there's one oddity that I think is worth addressing.  There's several places where we allocate a vector in function A, then pass it as an rvalue reference to function B, which then uses it.  It is probably safe due to not using `exec_policy_nosync()` in B but it would feel safer to me to 

- just pass these by reference
- use `exec_policy_nosync` everywhere in the callee
- have the caller which allocated the memory in the first place do a manual sync() itself before exit.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
+/*
```
ttnghia: Is this file copied from cudf? If so, can you point out what's the new code here?

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+    int const max_by_shmem = max_shmem_per_block / per_thread_bytes;
+    if (max_by_shmem > 0) {
+      // Prefer multiples of warp size (32)
+      int const rounded = (max_by_shmem / 32) * 32;
+      block_size        = std::max(32, std::min(256, rounded > 0 ? rounded : max_by_shmem));
+    } else {
+      block_size = 32;  // minimal reasonable size
+    }
+  }
+  cudf::size_type const grid_size = (num_pairs + block_size - 1) / block_size;
```
ttnghia: What if `num_pairs` is huge? This will result in a huge grid size. You will not have enough shared memory for all blocks, and the program will crash.

The `grid_size` should be evaluated based on both the amount of shared memory and input size. Please see how we do that here:
 * https://github.com/rapidsai/cudf/blob/b321bb117fdef4c2b17aef3874b68230702edee7/cpp/src/groupby/hash/compute_single_pass_aggs.cuh#L70-L79
 * https://github.com/rapidsai/cudf/blob/b321bb117fdef4c2b17aef3874b68230702edee7/cpp/src/groupby/hash/compute_shared_memory_aggs.cu#L287-L293

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+                   [left_has_match = left_has_match.data()] __device__(cudf::size_type idx) {
+                     left_has_match[idx] = true;
+                   });
+
+  // Count unmatched left rows
+  auto const num_unmatched = thrust::count(
+    rmm::exec_policy_nosync(stream), left_has_match.begin(), left_has_match.end(), false);
+
+  // Allocate output with space for both matched and unmatched
+  auto const total_size  = left_indices.size() + num_unmatched;
```
ttnghia: Is this guaranteed to always be bigger than `right_indices.size()`?

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+    auto left_indices  = rmm::device_uvector<cudf::size_type>(left_keys.num_rows(), stream, mr);
+    auto right_indices = rmm::device_uvector<cudf::size_type>(left_keys.num_rows(), stream, mr);
+
+    thrust::sequence(rmm::exec_policy_nosync(stream), left_indices.begin(), left_indices.end());
+    thrust::fill(rmm::exec_policy_nosync(stream),
+                 right_indices.begin(),
+                 right_indices.end(),
+                 right_keys.num_rows());
+
+    stream.synchronize();
```
ttnghia: Same here, don't need stream sync.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+    join_obj.inner_join(left_keys, is_left_sorted, stream, mr);
+
+  // Add back left rows with no matches
+  auto result = add_left_unmatched_rows(*equality_left_indices,
+                                        *equality_right_indices,
+                                        left_keys.num_rows(),
+                                        right_keys.num_rows(),
+                                        stream,
+                                        mr);
+  stream.synchronize();
```
ttnghia: No need for stream sync if there is no host-to-device memcpy.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+    return {rmm::device_uvector<cudf::size_type>(0, stream, mr),
+            rmm::device_uvector<cudf::size_type>(0, stream, mr)};
+  }
+
+  // Perform sort-merge inner join on equality keys
+  cudf::sort_merge_join join_obj(right_keys, is_right_sorted, compare_nulls, stream);
+  auto [left_indices, right_indices] = join_obj.inner_join(left_keys, is_left_sorted, stream, mr);
+
+  // Return the raw device_uvectors (RVO will handle the move)
+  stream.synchronize();
```
ttnghia: We don't need stream sync here.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+  CUDF_EXPECTS(left_keys.num_columns() > 0, "Left keys table must have at least one column");
+  CUDF_EXPECTS(right_keys.num_columns() > 0, "Right keys table must have at least one column");
+
+  // Handle empty right table case - no matches
+  if (right_keys.num_rows() == 0) { return rmm::device_uvector<cudf::size_type>(0, stream, mr); }
+
+  // Perform sort-merge inner join on equality keys
+  cudf::sort_merge_join join_obj(right_keys, is_right_sorted, compare_nulls, stream);
+  auto [equality_left_indices, equality_right_indices] =
+    join_obj.inner_join(left_keys, is_left_sorted, stream, mr);
```
ttnghia:
```suggestion
    join_obj.inner_join(left_keys, is_left_sorted, stream, cudf::get_current_device_resource_ref());
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+  auto const num_unmatched = thrust::count(
+    rmm::exec_policy_nosync(stream), left_has_match.begin(), left_has_match.end(), false);
+
+  auto result = rmm::device_uvector<cudf::size_type>(num_unmatched, stream, mr);
+  thrust::copy_if(rmm::exec_policy_nosync(stream),
+                  thrust::counting_iterator<cudf::size_type>(0),
+                  thrust::counting_iterator<cudf::size_type>(left_num_rows),
+                  left_has_match.begin(),
+                  result.begin(),
+                  thrust::logical_not<bool>());
```
ttnghia:
```suggestion
                  cuda::std::logical_not<bool>());
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+  }
+
+  // Check for nulls in conditional columns (top-level and nested)
+  auto const has_nulls = cudf::has_nulls(left_conditional) || cudf::has_nulls(right_conditional) ||
+                         cudf::has_nested_nulls(left_conditional) ||
+                         cudf::has_nested_nulls(right_conditional);
+
+  // Parse the AST expression
+  auto const parser = cudf::ast::detail::expression_parser{
+    binary_predicate, left_conditional, right_conditional, has_nulls, stream, mr};
```
ttnghia:
```suggestion
  auto const parser = cudf::ast::detail::expression_parser{
    binary_predicate, left_conditional, right_conditional, has_nulls, stream, cudf::get_current_device_resource_ref()};
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+                 right_equality.num_rows());
+
+    stream.synchronize();
+    return {std::move(left_indices), std::move(right_indices)};
+  }
+
+  // Check for nulls in conditional columns (top-level and nested)
+  auto const has_nulls = cudf::has_nulls(left_conditional) || cudf::has_nulls(right_conditional) ||
+                         cudf::has_nested_nulls(left_conditional) ||
+                         cudf::has_nested_nulls(right_conditional);
```
ttnghia:
```suggestion
  auto const has_nulls =  cudf::has_nested_nulls(left_conditional) ||
                         cudf::has_nested_nulls(right_conditional);
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+
+  // Step 3: Add back left rows with no matches
+  auto result = add_left_unmatched_rows(filtered_left_indices,
+                                        filtered_right_indices,
+                                        left_equality.num_rows(),
+                                        right_equality.num_rows(),
+                                        stream,
+                                        mr);
+  stream.synchronize();
+  return result;
```
ttnghia:
```suggestion
  return add_left_unmatched_rows(filtered_left_indices,
                                        filtered_right_indices,
                                        left_equality.num_rows(),
                                        right_equality.num_rows(),
                                        stream,
                                        mr);
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+                                      *equality_right_indices,
+                                      *left_table,
+                                      *right_table,
+                                      parser.device_expression_data,
+                                      parser.device_expression_data.num_intermediates,
+                                      has_nulls,
+                                      stream,
+                                      mr);
+  stream.synchronize();
+  return result;
```
ttnghia:
```suggestion
  return filter_by_conditional(*equality_left_indices,
                                      *equality_right_indices,
                                      *left_table,
                                      *right_table,
                                      parser.device_expression_data,
                                      parser.device_expression_data.num_intermediates,
                                      has_nulls,
                                      stream,
                                      mr);
```

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+
+  auto result = rmm::device_uvector<cudf::size_type>(num_unmatched, stream, mr);
+  thrust::copy_if(rmm::exec_policy_nosync(stream),
+                  thrust::counting_iterator<cudf::size_type>(0),
+                  thrust::counting_iterator<cudf::size_type>(left_num_rows),
+                  left_has_match.begin(),
+                  result.begin(),
+                  thrust::logical_not<bool>());
+
+  stream.synchronize();
```
ttnghia:
```suggestion
```

src/main/cpp/src/mixed_sort_merge_join.cu
nvdbaranec: We do. `left_indices` and `right_indices` are in use when we exit.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+                           rmm::device_async_resource_ref mr)
+{
+  auto const num_pairs = left_indices.size();
+  if (num_pairs == 0) {
+    return {rmm::device_uvector<cudf::size_type>(0, stream, mr),
+            rmm::device_uvector<cudf::size_type>(0, stream, mr)};
+  }
+
+  // Create a boolean mask for indices to keep
+  auto keep_mask = rmm::device_uvector<bool>(num_pairs, stream, mr);
```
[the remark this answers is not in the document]
ttnghia: We can write this too:
```
auto keep_mask = rmm::device_uvector<bool>(num_pairs, stream);
```
The last parameter is implicitly taken from the current device resource. It is just a matter of preference for clarity.

By convention, we pass the provided `mr` to constructing the output, even for empty output (the constructor of empty output may do anything with memory allocation that we don't know). Other intermediate variables need to be constructed from the current device resource, which could be a "closer" GPU than the GPU storing the output.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+    if (max_by_shmem > 0) {
+      // Prefer multiples of warp size (32)
+      int const rounded = (max_by_shmem / 32) * 32;
+      block_size        = std::max(32, std::min(256, rounded > 0 ? rounded : max_by_shmem));
+    } else {
+      block_size = 32;  // minimal reasonable size
+    }
+  }
+  cudf::size_type const grid_size = (num_pairs + block_size - 1) / block_size;
+  auto const shmem_size = static_cast<size_t>(block_size) * static_cast<size_t>(per_thread_bytes);
```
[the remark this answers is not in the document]
revans2: I have tests that cover the num_intermediates = 0 case. They crash if I put in an assert that per_thread_bytes != 0. These are very basic tests like if A > B. Is a block_size of 256 bad to have for a case like this? is it going to cause a crash?

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+                                                            left_table,
+                                                            right_table,
+                                                            device_expression_data,
+                                                            keep_mask.data());
+
+  // Surface any kernel launch errors immediately
+  CUDF_CUDA_TRY(cudaPeekAtLastError());
+
+  // Count the number of true values in the mask
+  auto const num_matches =
```
pmattione-nvidia: don't we have to wait for the filter_join_indices_kernel kernel launch to finish before we use the results in keep_mask

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+                        cudf::size_type right_num_rows,
+                        rmm::cuda_stream_view stream,
+                        rmm::device_async_resource_ref mr)
+{
+  CUDF_FUNC_RANGE();
+
+  // Create a boolean mask to track which left rows have matches
+  auto left_has_match = rmm::device_uvector<bool>(left_num_rows, stream, mr);
+  thrust::fill(
+    rmm::exec_policy_nosync(stream), left_has_match.begin(), left_has_match.end(), false);
```
pmattione-nvidia: doesn't exec_policy_nosync create a race condition with the below thrust::for_each() call

src/main/cpp/src/mixed_sort_merge_join.cu
revans2: I thought that if it happened in the same stream as the other kernel we will get proper stream order semantics. Also I thought that because it is a reduction that copies the result back to the CPU it will ignore the no_sync policy and do the right thing, but I could be wrong and changed the policy just to be safe.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+  // Mark left rows that have matches
+  thrust::for_each(rmm::exec_policy_nosync(stream),
+                   left_indices.begin(),
+                   left_indices.end(),
+                   [left_has_match = left_has_match.data()] __device__(cudf::size_type idx) {
+                     left_has_match[idx] = true;
+                   });
+
+  // Count unmatched left rows
+  auto const num_unmatched = thrust::count(
```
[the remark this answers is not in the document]
nvdbaranec: It gets ignored in this case because it's forced to synchronize.  In these cases I like to leave the nosync out to indicate that the caller understands this.

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+  auto const total_size  = left_indices.size() + num_unmatched;
+  auto out_left_indices  = rmm::device_uvector<cudf::size_type>(total_size, stream, mr);
+  auto out_right_indices = rmm::device_uvector<cudf::size_type>(total_size, stream, mr);
+
+  // Copy matched pairs
+  thrust::copy(rmm::exec_policy_nosync(stream),
+               left_indices.begin(),
+               left_indices.end(),
+               out_left_indices.begin());
+  thrust::copy(rmm::exec_policy_nosync(stream),
```
pmattione-nvidia: again nosync has race condition with following copy_if calls

src/main/cpp/src/mixed_sort_merge_join.cu
revans2: but doesn't the move make it so that we don't release the memory, so it is okay?

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+
+  // If no equality matches, return empty result
+  if (equality_left_indices->size() == 0) {
+    return rmm::device_uvector<cudf::size_type>(0, stream, mr);
+  }
+
+  // Extract unique left indices
+  auto result = extract_unique_left_indices(*equality_left_indices, stream, mr);
+  stream.synchronize();
+  return result;
```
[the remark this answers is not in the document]
revans2: I thought with RMM on device memory that a free does not actually free the memory device memory until the stream order semantics say it is no longer used.

revans2:
build

src/main/cpp/src/mixed_sort_merge_join.cu
pmattione-nvidia: Nevermind, it uses the same stream

src/main/cpp/src/mixed_sort_merge_join.cu
```diff
@@ -0,0 +1,825 @@
...
+{
+  CUDF_FUNC_RANGE();
+
+  // Create a boolean mask to track which left rows have matches
+  auto left_has_match = rmm::device_uvector<bool>(left_num_rows, stream, mr);
+  thrust::fill(
+    rmm::exec_policy_nosync(stream), left_has_match.begin(), left_has_match.end(), false);
+
+  // Mark left rows that have matches
+  thrust::for_each(rmm::exec_policy_nosync(stream),
```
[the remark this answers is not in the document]
pmattione-nvidia: Nevermind, it uses the same stream

src/main/cpp/src/mixed_sort_merge_join.cu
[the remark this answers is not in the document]
ttnghia: > I thought with RMM on device memory that a free does not actually free the memory device memory until the stream order semantics say it is no longer used.

Yes it is. As long as the intermediate data vectors are stream-order constructed on the same stream as the output, we don't have to worry about them being destroyed when out of scope.

Issue #3891, opened 2025-10-28
[BUG] java.lang.UnsatisfiedLinkError: libnvidia-ml.so.1: cannot open shared object file: No such file or directory


yinqingh:
**Describe the bug**
First seen this error in LHA-xxx-get-json-obj-gpu/255

nvidia driver version: 560.35.05
cuda version: 12.9.1

on test host
```
/usr/lib/x86_64-linux-gnu/libnvidia-ml.so.1 -> libnvidia-ml.so.560.35.05
/usr/lib/x86_64-linux-gnu/libnvidia-ml.so.560.35.05
/usr/lib/x86_64-linux-gnu/stubs/libnvidia-ml.so
```

Error details
```
[2025-10-28T08:09:52.692Z] 2025-10-28 08:09:52,589 INFO datasources.FileSourceStrategy: Output Data Schema: struct<table2_column3: bigint, table2_column7: string, table2_column10: string, table2_column18: string, table2_column20: string ... 13 more fields>
[2025-10-28T08:10:08.171Z] 2025-10-28 08:10:06,348 ERROR cudf.NativeDepsLoader: Could not load cudf jni library...
[2025-10-28T08:10:08.171Z] java.io.IOException: Error loading dependencies
[2025-10-28T08:10:08.171Z] 	at ai.rapids.cudf.NativeDepsLoader.loadNativeDeps(NativeDepsLoader.java:202)
[2025-10-28T08:10:08.171Z] 	at ai.rapids.cudf.NativeDepsLoader.loadNativeDeps(NativeDepsLoader.java:66)
[2025-10-28T08:10:08.171Z] 	at com.nvidia.spark.rapids.jni.JSONUtils.<clinit>(JSONUtils.java:25)
[2025-10-28T08:10:08.171Z] 	at com.nvidia.spark.rapids.JsonPathParser$.fallbackCheck(GpuGetJsonObject.scala:106)
[2025-10-28T08:10:08.171Z] 	at com.nvidia.spark.rapids.GpuGetJsonObjectMeta.$anonfun$tagExprForGpu$3(GpuGetJsonObject.scala:166)
[2025-10-28T08:10:08.171Z] 	at com.nvidia.spark.rapids.GpuGetJsonObjectMeta.$anonfun$tagExprForGpu$3$adapted(GpuGetJsonObject.scala:166)
[2025-10-28T08:10:08.171Z] 	at scala.Option.exists(Option.scala:376)
[...]

PR #3899, opened 2025-10-29
Package nvmljni into a separate .so which can be optionally loaded when available

zpuller:
Fixes https://github.com/NVIDIA/spark-rapids-jni/issues/3891

Package nvmljni into a separate .so which can be optionally loaded when available, otherwise gracefully fallback to no-op behavior. This PR depends on https://github.com/rapidsai/cudf/pull/20414

Tested this by building and running `NVMLTest.java` and then also rerunning with the `libnvidia-ml.so` effectively hidden on the system. Before, this would fail on loading the regular native deps (because libcudfjni.so itself depended on libnvidia-ml.so), but now it passes in both cases.


src/main/cpp/tests/nvml_test.cpp
zpuller: This test is no longer needed, it's superceded by `NVMLTest.java` which exercises the underlying native code paths.

zpuller:
> Are we missing `NVMLJni.cpp` in the diff posted here?

It was merged in in a recent [PR](https://github.com/NVIDIA/spark-rapids-jni/pull/3690), we are just changing the linkage here.

mythrocks:
> > Are we missing `NVMLJni.cpp` in the diff posted here?
> 
> It was merged in in a recent [PR](https://github.com/NVIDIA/spark-rapids-jni/pull/3690), we are just changing the linkage here.

Right, just realized that.  Got it.

src/main/java/com/nvidia/spark/rapids/jni/nvml/NVML.java
```diff
@@ -58,25 +55,36 @@ public class NVML {
...
         try {
             if (nvmlInit()) {
                 nvmlInitialized = true;
-                nativeLibraryLoaded = true;
                 return true;
             } else {
                 logger.error("Failed to initialize NVML");
                 return false;
             }
         } catch (UnsatisfiedLinkError e) {
```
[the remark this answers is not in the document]
zpuller: I think if we ran this code without successfully loading the dependency it could happen, but that really shouldn't happen so I wouldn't say this is strictly necessary, just an extra safeguard. In the usual case if the .so was missing, we would have already returned at line 73. Maybe if there as some corrupted/garbage version of the .so on the host for some reason it could happen

src/main/java/com/nvidia/spark/rapids/jni/nvml/NVML.java
```diff
@@ -58,25 +55,36 @@ public class NVML {
     private static boolean nativeLibraryLoaded = false;
 
     /**
-     * Initialize NVML library
+     * Initialize NVML library and load native dependencies
      */
     public static synchronized boolean initialize() {
```
mythrocks: I don't think I've quite gotten the failure case, when the native library is not available.  Where `NVML::initialize()` is called from, in `spark-rapids`, currently?

In the event that the library isn't loaded, `initialize()` returns `false`, and none of the other methods will/should be called.  Is that right?

zpuller:
build

abellina (dismissed):

Issue #3907, opened 2025-10-31
[BUG] RmmSparkTest.testInsertOOMsGpu fails after cudf submodule update: expected 256 OOM events but got 100 in submodule sync CI


pxLi:
**Describe the bug**
Build: spark-rapids-jni_submodule-sync-dev/5498

The build failed during submodule sync testing. The test RmmSparkTest.testInsertOOMsGpu failed with an assertion error expecting 256 OOM events but only 100 were triggered. This occurred after updating the cudf submodule from e220fa209a to 3ea2e67e. The test failure prevented jar creation.

Error logs:
```
[ERROR] Tests run: 20, Failures: 1, Errors: 0, Skipped: 0
[ERROR] testInsertOOMsGpu  Time elapsed: 1.008 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <256> but was: <100>
	at com.nvidia.spark.rapids.jni.RmmSparkTest.testInsertOOMsGpu(RmmSparkTest.java:372)
[INFO] BUILD FAILURE
```

**Environment details**
- Jenkins Pipeline: spark-rapids-jni_submodule-sync-dev
- Build Number: 5498
- Docker Image: rockylinux8-cuda12.9.1-blossom
- CUDA Version: 12.9.86
- GCC Version: 14.2.1
- Maven: 3.x
- Test Framework: JUnit Jupiter 5.8.1
- cudf submodule: 3ea2e67eaac8d67c4d51d836bd26c457a0fa5535
- rmm commit: 25c52e3f3c928786a5a28ae9f8db8de7bafe6203 (current pinned on main is [a8716a6afb24c0cc7fa8c4fececb1e497db9381b](https://github.com/NVIDIA/spark-rapids-jni/blob/main/thirdparty/cudf-pins/versions.json#L133))

PR #3908, opened 2025-10-31
Fix `RmmSparkTest` test due to breaking change from `rmm`

ttnghia:
Due to a breaking change from `rmm` (https://github.com/rapidsai/rmm/pull/2114) which removes the padding from the `do_allocate` function in memory resource classes, the amount of allocated memory is changed and that fails our test. This fixes it. From now, we get exactly what we ask for (asking for `N` bytes should allocate `N` bytes).

Closes https://github.com/NVIDIA/spark-rapids-jni/issues/3907.

### DIFF
--- a/thirdparty/cudf-pins/versions.json
+++ b/thirdparty/cudf-pins/versions.json
@@ -130,7 +130,7 @@
     {
       "always_download" : true,
       "git_shallow" : false,
-      "git_tag" : "a8716a6afb24c0cc7fa8c4fececb1e497db9381b",
+      "git_tag" : "73d0e66953531fcc4c336745c8e125108dd9123b",
       "git_url" : "https://github.com/rapidsai/rmm.git",
       "source_subdir" : "cpp",
       "version" : "25.12"


src/test/java/com/nvidia/spark/rapids/jni/RmmSparkTest.java
```diff
@@ -369,7 +369,7 @@ public void testInsertOOMsGpu() {
       assertEquals(RmmSparkThreadState.THREAD_RUNNING, RmmSpark.getStateOf(threadId));
       assertEquals(1, RmmSpark.getAndResetNumRetryThrow(taskid));
       assertEquals(0, RmmSpark.getAndResetNumSplitRetryThrow(taskid));
-      assertEquals(ALIGNMENT, RmmSpark.getAndResetGpuMaxMemoryAllocated(taskid));
+      assertEquals(100, RmmSpark.getAndResetGpuMaxMemoryAllocated(taskid));
```
ttnghia: Change to `100` because in this test we always allocate `100` bytes.

abellina (approved):

mythrocks (changes requested):
Minor nit to move the `100` to a `final`.  But looks good otherwise.

I don't know if we need to document here that memory-resources are guaranteed to align allocations at 256-byte boundaries.

ttnghia:
> I don't know if we need to document here that memory-resources are guaranteed to align allocations at 256-byte boundaries.

No it is not. We set the alignment either `256` or `512` depending on cases: https://github.com/rapidsai/cudf/blob/d2a1bcaddc657cfb4da64413555f346ca48525e0/java/src/main/java/ai/rapids/cudf/Rmm.java#L239

pxLi:
Going to merge this to unblock submodule-sync job first, thanks!

PR #3917, opened 2025-11-04
Add join primitives API to let us build other joins from inner join + post processing

revans2:
Any join can be implemented in terms of an inner join followed by post processing. There are cases where this is very helpful, like if we want to switch the build and probe sides of a hash join.  This adds in a sort merge join API and some helper operations that let us create left outer, right outer, full outer, left semi, right semi, left anti, right anti and the spark specific existence join from an inner join.  That said these are just primitive operations and it is likely that this is not going to be ideal in all cases. There is still a lot of testing that needs to be done to come up with the proper join heuristics for all cases.


revans2:
build

revans2:
build

nvauto:
**NOTE: release/25.12 has been created from main. Please retarget your PR to release/25.12 if it should be included in the release.**

pmattione-nvidia (approved):

PR #3927, opened 2025-11-07
Update boost download site from boost.io to sourceforge [skip ci]

pxLi:
Our internal security team has blocked access to `boost.io`. We're currently verifying with internal teams whether this is an intentional security policy change. 

**Note:** This block only affects internal network builds: public developers are not impacted.

In the meantime, we're preparing an update to migrate our download source from boost.io to SourceForge.

Skip CI as verfied internally.


pxLi:
build

PR #3948, opened 2025-11-13
Add Iceberg truncate transform

res-life:
contributes to https://github.com/NVIDIA/spark-rapids/issues/13383

Add Iceberg truncate transform for types: int, long, string, binary, decimal32, decimal64, decimal128

Implements the Iceberge functions:
- `TruncateUtil.truncateInt`
- `TruncateUtil.truncateLong`
- `TruncateUtil.truncateDecimal`
- `BinaryUtil.truncateBinaryUnsafe`
- `UnicodeUtil.truncateString`

### DIFF
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -204,6 +204,7 @@ add_library(
   src/ListSliceJni.cpp
   src/MapJni.cpp
   src/MapZipWithUtilsJNI.cpp
+  src/JoinPrimitivesJni.cpp
   src/NativeParquetJni.cpp
   src/NumberConverterJni.cpp
   src/ParseURIJni.cpp
@@ -244,6 +245,7 @@ add_library(
   src/list_slice.cu
   src/map.cu
   src/map_zip_with_utils.cu
+  src/join_primitives.cu
   src/multiply.cu
   src/murmur_hash.cu
   src/number_converter.cu
--- a/src/main/cpp/src/JoinPrimitivesJni.cpp
+++ b/src/main/cpp/src/JoinPrimitivesJni.cpp
@@ -1,0 +1,407 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cudf_jni_apis.hpp"
+#include "jni_compiled_expr.hpp"
+#include "join_primitives.hpp"
+
+#include <cudf/ast/expressions.hpp>
+#include <cudf/column/column_view.hpp>
+#include <cudf/table/table_view.hpp>
+#include <cudf/types.hpp>
+
+#include <rmm/device_uvector.hpp>
+
+namespace {
+
+/**
+ * @brief Convert pair of device vectors to Java long array
+ * Returns a 5-element array: [size_in_bytes, left_ptr, left_handle, right_ptr, right_handle]
+ */
+jlongArray gather_maps_to_java(
+  JNIEnv* env,
+  std::pair<rmm::device_uvector<cudf::size_type>, rmm::device_uvector<cudf::size_type>> gather_maps)
+{
+  // Both gather maps must have the same size for paired results
+  CUDF_EXPECTS(gather_maps.first.size() == gather_maps.second.size(),
+               "Left and right gather maps must have the same size");
+
+  // Release the underlying device buffers to Java
+  auto left_map_buffer  = std::make_unique<rmm::device_buffer>(gather_maps.first.release());
+  auto right_map_buffer = std::make_unique<rmm::device_buffer>(gather_maps.second.release());
+
+  cudf::jni::native_jlongArray result(env, 5);
+  // Return size in bytes (as expected by DeviceMemoryBuffer.fromRmm)
+  result[0] = static_cast<jlong>(left_map_buffer->size());
+  result[1] = cudf::jni::ptr_as_jlong(left_map_buffer->data());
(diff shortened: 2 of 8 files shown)


res-life:
build

res-life:
build

src/main/cpp/src/iceberg_truncate.hpp
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * The input type can be INT32, INT64, DECIMAL32, DECIMAL64 or DECIMAL128.
+ * Note: For DECIMAL types, the truncation is performed on the integer representation
+ *
+ * For positive values, Iceberg truncation is: value - (value % width)
+ * For negative values, this uses a floored modulo approach:
+ * value - (((value % width) + width) % width)
+ *
+ * Example:
+ * - truncate_integer(10, 5) = 0
```
mythrocks: Is that right?  Per the formula, shouldn't the truncation result be `10`?

res-life: > Just a moment, are the arguments flipped in the example?
Yes. Updated the comments to make the examples clear.

res-life: > Is a negative width even valid, by the way?
I tried to find the constraint (width > 0), but not found.
Let's first assume the width must be positive. Negative width is meaningless.

src/test/java/com/nvidia/spark/rapids/jni/iceberg/IcebergTruncateTest.java
```diff
@@ -0,0 +1,504 @@
...
+    }
+
+    Random rand = new Random(seed);
+    int numRows = 1024;
+    Integer[] inputData = new Integer[numRows];
+    Integer[] expectedData = new Integer[numRows];
+    Integer[] expectedDataNegativeWidth = new Integer[numRows];
+    for (int i = 0; i < numRows; i++) {
+      int val = rand.nextInt();
+      inputData[i] = val;
```
liurenjie1024: One missing check is consistent behavior with null value.

src/test/java/com/nvidia/spark/rapids/jni/iceberg/IcebergTruncateTest.java
```diff
@@ -0,0 +1,504 @@
...
+      // run on CPU to get expected value
+      expectedData[i] = TruncateUtil.truncateInt(width, val);
```
liurenjie1024: nit: We could use public api only, e.g. Transforms.truncate().bind()

src/main/cpp/src/iceberg/iceberg_truncate.cu
```diff
@@ -0,0 +1,273 @@
...
+                     output->mutable_view().end<T>(),
+                     truncate_integral_fn<T>{*d_input, width});
+  } else {
+    CUDF_FAIL("Unsupported type for truncate_integral_impl");
+  }
+  return output;
+}
+
+std::unique_ptr<cudf::column> truncate_string_impl(cudf::column_view const& input,
+                                                   int32_t length,
```
mythrocks: Nit:  The argument name you used elsewhere is better:

```suggestion
                                                   int32_t truncate_length,
```

src/main/cpp/src/iceberg/iceberg_truncate.cu
```diff
@@ -0,0 +1,273 @@
...
+  cudf::size_type* d_sizes;
+  char* d_chars;
+  cudf::detail::input_offsetalator d_offsets;
+
+  __device__ void operator()(cudf::size_type idx)
+  {
+    if (!d_chars) {
+      // first phase
+      int binary_len = input_offsets[idx + 1] - input_offsets[idx];
+      d_sizes[idx]   = (binary_len < truncate_length) ? binary_len : truncate_length;
```
mythrocks: You can use `std::min` here.  No need for even `cuda::std::min`:

```suggestion
      d_sizes[idx]   = std::min(binary_len, truncate_length);
```

src/main/cpp/src/iceberg/iceberg_truncate.cu
```diff
@@ -0,0 +1,273 @@
...
+                          input_offsets[idx + 1] - input_offsets[idx]);
+    if (!d_chars) {
+      // first phase
+      // Note: one character can be multiple(1-4) bytes for UTF8 encoding
+      if (str.length() < truncate_length) {
+        d_sizes[idx] = str.size_bytes();
+      } else {
+        auto tmp     = str.substr(0, truncate_length);
+        d_sizes[idx] = tmp.size_bytes();
+      }
```
mythrocks: Masterful. 👏

src/main/cpp/src/iceberg/iceberg_truncate.cu
```diff
@@ -0,0 +1,273 @@
...
+                                              mr);
+  auto d_input = cudf::column_device_view::create(input, stream);
+
+  if (input.type().id() == cudf::type_id::INT32 || input.type().id() == cudf::type_id::DECIMAL32) {
+    // treat DECIMAL32 column as int32 column
+    using T = int32_t;
+    thrust::tabulate(rmm::exec_policy_nosync(stream),
+                     output->mutable_view().begin<T>(),
+                     output->mutable_view().end<T>(),
+                     truncate_integral_fn<T>{*d_input, width});
```
mythrocks: I was wondering if there would be value in using a utility function to reduce the repetition of commonality:
```c++
template <typename RepT, typename Output>
void truncate_integral_and_fill(Output output, cuda::size_type num_rows, rmm::cuda_stream_view stream)
{
  auto const begin = output->mutable_view().begin<RepT>();
  thrust::fill_n(rmm::exec_policy_nosync(stream), begin, num_rows, truncate_integral_fn<RepT>{...};
}
```
Then the call sites will look a little shorter:
```c++
if (input.type().id() == cudf::type_id::INT32 || input.type().id() == cudf::type_id::DECIMAL32) {
  truncate_integral_and_fill<int32_t>(output, input.size(), stream);
}
```
Less boilerplate.

mythrocks: Oh, right.  In my example, please replace `fill_n` with `tabulate`.  (`fill_n` doesn't take a generator function.)

My main point is to reduce that repetition, if possible.

mythrocks (commented):
Generally looks really good.

I've made a couple of nitpicks, but this is mostly good to go.

I'll take another crack at the Java side of the changes tomorrow.

mythrocks (approved):
👍 
I had to push a CMakeLists change to resolve conflicts.

PR #3971, opened 2025-11-19
Adopt the new rmm interface

ttnghia:
This fixes the breaking in C++ code which uses the deprecated and deleted `rmm` interface.

May depends on:
 * https://github.com/rapidsai/cudf/pull/20661


abellina:
Same question about the cuDF change. Should we target release/25.12 here? Then the automation will cherry-pick to main.

PR #3974, opened 2025-11-19
Optimize multiply operation by avoiding unnecessary validity vector allocation

### LINKED ISSUE #3973: [PERF] Optimize multiply operation by avoiding unnecessary validity vector allocation
### Current Behavior
Currently, the `multiply` operation always allocates a validity vector (`rmm::device_uvector<bool>`) to track null values and overflow results, regardless of whether the input columns have nulls or overflow checking is enabled.

This results in unnecessary memory allocation and computation overhead in cases where:
- Both input columns/scalars have no nulls AND no overflow check is needed

### Proposed Optimization
Optimize the `multiply_impl` function to handle this scenario:

**Scenario**: Both inputs have no nulls + `check_overflow=false`
- Skip validity vector allocation entirely
- Use a simplified kernel that directly computes multiplication
- Return column without null mask (following cudf convention that `nullptr` mask means all valid)

### Expected Benefits
- **Memory savings**: Eliminates `num_rows * sizeof(bool)` allocation 
- **Computation savings**: Avoids `cudf::detail::valid_if` call 

Image

Image

binmahone:
This PR fixes #3973 , and the result nsys loos like:

img_v3_02s6_893838a6-f9d2-4469-86b2-2cd164f6f92g

You can see the yellow bar part reduces significantly. 

This improvement reduces our workload e2e time by 10%.

### DIFF
--- a/src/main/cpp/src/multiply.cu
+++ b/src/main/cpp/src/multiply.cu
@@ -153,7 +153,7 @@ struct multiply_fn {
 };
 
 /**
- * @brief Functor for multiplication when both inputs are valid 
+ * @brief Functor for multiplication when both inputs are valid
  * and no overflow check is needed.
  */
 template <typename T, typename LEFT_ACCESSOR, typename RIGHT_ACCESSOR>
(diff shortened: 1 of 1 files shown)


ttnghia (dismissed):

binmahone:
Let's take a step back and jump out the scope of merely "multiply operator". 

I remember that we previously observed in ClickHouse that having every column in the input schema as `NOT NULL` performs significantly better compared to using `Nullable`, In a broader sense.

However, it seems like our code doesn't handle the `NOT NULL` scenario very well. Do you think this direction has potential? If so, I feel I should first open an issue to investigate how we can quantify the overhead we're currently incurring even if the input is "already not null" . @revans2 @abellina

GaryShen2008 (approved):
approve again after a small format change.

Issue #3905, opened 2025-10-30
[FEA] SparkRapidsAdaptor locking performance issues


abellina:
We have reports of the SparkRapidsAdaptor causing a regression between 25.02 and 25.04 given coarse locks taken while we do some expensive STL and metric operations. We have confirmed that reducing those critical sections improves the performance immensely, removing the regression.

This issue is to help fix this in the proper way, by speeding up the processing within these locks.

More details TBD.

abellina:
Related cuDF change: https://github.com/rapidsai/cudf/issues/20520

PR #3977, opened 2025-11-19
Only check for deadlocks in deadlock busting thread

abellina:
1. This PR changes BUFN deadlock detection so it only happens in the deadlock busting thread, not in every free. 
3. It removes a JNI->Java callback that was done, instead taking in the java values when needed (in the thread) as a parameter.
5. It also removes a static lock in ThreadStateRegistry

This is a performance improvement for the state machine and is part of https://github.com/NVIDIA/spark-rapids-jni/issues/3905


src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -1696,127 +1690,55 @@ class spark_resource_adaptor final : public rmm::mr::device_memory_resource {
...
+   * - Its state is THREAD_BUFN, OR
+   * - Its thread ID is in the java_blocked_thread_ids set if it is provided.
+   *
+   * Threads in THREAD_BLOCKED state are NOT considered BUFN or above.
+   */
+  bool is_thread_bufn_or_above(
+    full_thread_state const& state,
+    std::optional<std::unordered_set<long>> const& java_blocked_thread_ids)
   {
+    LOG_INFO("is_thread_bufn_or_above: state: {}, java_blocked_thread_ids: {}",
```
revans2: nit: Generally we have always had the logging follow a CSV/TSV like structure so that we can parse it and do post processing/analytics on it. Is the needed? Could this be moved to more of that format?

abellina:
build

revans2 (dismissed):

PR #4047, opened 2025-12-05
Fix submodule sync compile failure in `murmur_hash.cu`.

mythrocks:
This commit fixes compile failures in `murmur_hash.cu`, resulting from the changes in https://github.com/rapidsai/cudf/pull/20777.

The CUDF changes modified the seed of the `element_hasher` from being a `uint32_t` to the result type of the hasher-function.  Meanwhile, `spark-rapids-jni`'s `murmur_hash` was using the old `uint32_t`.

It turns out that rapidsai/cudf#20777 was inadvertently a breaking change.

### DIFF
--- a/src/main/cpp/src/murmur_hash.cu
+++ b/src/main/cpp/src/murmur_hash.cu
@@ -65,6 +65,27 @@ class murmur_device_row_hasher {
   friend class cudf::detail::row::hash::row_hasher;  ///< Allow row_hasher to access private
                                                      ///< members.
 
+ public:
+  /**
+   * @brief Return the hash value of a row in the given table.
+   *
+   * @param row_index The row index to compute the hash value of
+   * @return The hash value of the row
+   */
+  __device__ auto operator()(cudf::size_type row_index) const noexcept
+  {
+    return cudf::detail::accumulate(
+      _table.begin(),
+      _table.end(),
+      _seed,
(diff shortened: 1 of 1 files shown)


mythrocks:
The fix requires extracting the `result_type` from the `hash_fn` specified to `murmur_device_row_hasher`.  This necessitated moving the definition of `element_hasher_adapter` up, for visibility.

src/main/cpp/src/murmur_hash.cu
```diff
@@ -142,9 +122,32 @@ class murmur_device_row_hasher {
...
+    return cudf::detail::accumulate(
+      _table.begin(),
+      _table.end(),
+      _seed,
+      cuda::proclaim_return_type<murmur_hash_value_type>(
+        [row_index, nulls = this->_check_nulls] __device__(auto hash, auto column) {
+          return cudf::type_dispatcher(
+            column.type(), element_hasher_adapter<hash_function>{nulls, hash}, column, row_index);
+        }));
+  }
```
mythrocks: I'm checking if I can minimize the change further.  Maybe there isn't a need to move the definition of `element_hasher_adapter`.

mythrocks: Ok, minimized the change, I think.

PR #4095, opened 2025-12-19
Support nanosecond precision in timestamp conversion

### LINKED ISSUE #4089: [FEA] Add TIMESTAMP_NANOSECONDS support to GpuTimeZoneDB timezone conversion
The GpuTimeZoneDB timezone conversion functions (fromTimestampToUtcTimestamp, fromUtcTimestampToTimestamp) currently only support TIMESTAMP_SECONDS, TIMESTAMP_MILLISECONDS, and TIMESTAMP_MICROSECONDS. TIMESTAMP_NANOSECONDS is not supported and throws an error.

This is likely a blocker for our usecase where the timestamp in nanos is used only inside the UDF and never exported to Spark proper as a Timestamp. I will close this issue if we can work around this by using micros, or deprioritize it.

The underlying algorithm (convert_timestamp_tz_functor) is already templated and supports any timestamp type.

The fix might be this simple (other than testing):
```
case cudf::type_id::TIMESTAMP_NANOSECONDS:  
  return convert_timestamp_tz<cudf::timestamp_ns>(
  input, transitions, tz_index, to_utc, stream, mr);
```

No alternatives for timezone conversions, they are in spark-rapids-jni. 

We may close this issue if micros good enough for the usecase.

**Additional context**
Spark's TimestampType only supports microseconds, so this is only for usecases touching the java api directly.

rishic3:
Closes #4089. 

This adds support for nanosecond precision for timestamp conversion for `fromUtcTimestampToTimestamp` and `fromTimestampToUtcTimestamp`. The use case is parsing timestamp strings -> converting between timezones -> storing as strings, preserving nanosecond precision. 

The change is limited to GpuTimeZoneDB since it is hard to replicate outside the JNI, whereas e.g. CastStrings is easier to replicate using plain cuDF parsing. But in the future, supporting nanos in CastStrings `toTimestamp` would be desirable.


src/test/java/com/nvidia/spark/rapids/jni/TimeZoneTest.java
```diff
@@ -418,6 +418,66 @@ void convertFromUtcMicroSecondsTest() {
     }
   }
 
+  @Test
+  void convertToUtcNanoSecondsTest() {
```
abellina: The times here don't look to have nanoseconds populated. It seems to me we'd want that to make sure nothing is getting truncated?

rishic3: Done. FYI micros tests similarly don't populate. If we want, I can also update those in a follow-up

rishic3: Opened https://github.com/NVIDIA/spark-rapids-jni/pull/4098

PR #4143, opened 2026-01-12
clangd workaround:  Explicit types instead of CTAD

mythrocks:
This is a workaround for a crash one sees in clangd when loading the `spark-rapids-jni` project in VSCode / Cursor.

It looks like even the latest version of the
`llvm-vs-code-extensions.vscode-clangd` extension (with Clangd v21.x) seems to crash.  Perusing the logs reveals that it crashes because of our use of Class Template Argument Deduction in a couple of places in code.

This change introduces explicit types, so that code completion works correctly in VSCode /Cursor.

There is no material change to functionality.


mythrocks:
Build

pmattione-nvidia (approved):

PR #4261, opened 2026-02-10
Fix compile break from rapidsai/cudf#21283

mythrocks:
This change is to account for an impending breaking change in rapidsai/cudf#21283, where `cudf::device_span` was replaced with a typedef to `cuda::std::span`.

Surprisingly few modifications were required to accommodate the breakage. The changes mostly involved constructing explicit `device_span`s in a couple of places.


mythrocks:
Thank you for the scrutiny, @pmattione-nvidia.  I've merged this change.  CI hiccups should be sorted now.

PR #4265, opened 2026-02-10
Fix call to libcudf deprecated cudf::detail::copy_if_safe

davidwendt:
Replaces calls to libcudf deprecated `cudf::detail::copy_if_safe` and replaces it with appropriate CUB calls.

This also removes the dependency on the internal `algorithms.cuh` which is being moved in https://github.com/rapidsai/cudf/pull/21387

### DIFF
--- a/src/main/cpp/src/from_json_to_raw_map.cu
+++ b/src/main/cpp/src/from_json_to_raw_map.cu
@@ -141,7 +141,6 @@ OutputIterator copy_if(InputIterator begin,
   return output + num_selected.value(stream);
 }
 
-
 std::unique_ptr<cudf::column> make_empty_map(rmm::cuda_stream_view stream,
                                              rmm::device_async_resource_ref mr)
 {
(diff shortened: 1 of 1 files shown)


src/main/cpp/src/from_json_to_raw_map.cu
```diff
@@ -56,6 +56,92 @@ using namespace cudf::io::json;
...
+                                      begin,
+                                      output,
+                                      num_selected.data(),
+                                      num_items,
+                                      predicate,
+                                      stream.value()));
+
+  // Copy number of selected elements back to host via pinned memory
+  return output + num_selected.value(stream);
+}
```
mythrocks: Asking out of ignorance:  This looks quite similar to the definition of `cudf::detail::copy_if()`.  Is there a difference?

Shouldn't `spark-rapids-jni` use `cudf::detail::copy_if` instead of replicating it?

mythrocks (approved):

PR #4284, opened 2026-02-17
Reduce code complexity, fix minor random issues

pmattione-nvidia:
This PR addresses a series of minor code issues: 

- `NativeParquetJni.cpp`: Fixes a real bug where std::size_t (unsigned) was compared with < 0 — this comparison was always false, meaning encoding errors from std::mbsrtowcs / std::wcsrtombs would never be caught. Now correctly compares against std::numeric_limits<std::size_t>::max(). Also adds explicit to a single-argument constructor, refactors a while-with-done-flag to a cleaner do-while, and condenses an if-else to a ternary.
- `SparkResourceAdaptorJni.cpp`: Removes a redundant copy assignment operator (compiler-generated one is identical for POD members), flattens nested if-else via early return, and removes a duplicate is_retry_alloc_before_bufn check inside an outer if that already guards on the same condition.
- `KudoTableMerger.java`: Inverts a guard condition to early-return/continue, reducing nesting depth in deserializeOffsetBuffer.
- `NVMLMonitor.java`: Inverts guard conditions with early continue to reduce nesting in the monitoring loop. Duplicates Thread.sleep in early-return paths to preserve the original one-sleep-per-iteration behavior.

Overall functionality should not change.  The loop changes are attempts to reduce code complexity.  When reviewing, turn off whitespace diffs to reduce the noise in the diff.

### DIFF
--- a/src/main/cpp/src/NativeParquetJni.cpp
+++ b/src/main/cpp/src/NativeParquetJni.cpp
@@ -15,11 +15,11 @@
  */
 
 #include <cwctype>
+#include <limits>
 #include <sstream>
 #include <stdexcept>
 #include <string>
 #include <vector>
-#include <limits>
 
 // TCompactProtocol requires some #defines to work right.
 // This came from the parquet code itself...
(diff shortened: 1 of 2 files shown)


pmattione-nvidia:
build

src/main/java/com/nvidia/spark/rapids/jni/kudo/KudoTableMerger.java
```diff
@@ -224,77 +224,80 @@ private long deserializeValidityBuffer(ColumnOffsetInfo curColOffset) {
   private void deserializeOffsetBuffer(ColumnOffsetInfo curColOffset) {
     Arrays.fill(sliceInfoBuf, EMPTY_SLICE);
 
-    if (curColOffset.getOffset() != INVALID_OFFSET) {
-      long outputOffset = curColOffset.getOffset();
-      HostMemoryBuffer offsetBuf = buffer;
+    if (curColOffset.getOffset() == INVALID_OFFSET) {
+      return;
+    }
```
mythrocks: 👏

PR #4285, opened 2026-02-18
Add hostCrc32 function

jihoonson:
This PR adds a new `hostCrc32` function that computes a crc32 checksum of the data in a `HostMemoryBuffer`. Expected use cases are computing the checksum for relatively small data, such as computing checksums for Delta Lake deletion vectors where the deletion vector size is usually within a range from tens of bytes to a couple of MB.


mythrocks (approved):
I'm generally 👍.  This should be good to ship, once the error condition checking is consolidated.

jihoonson:
build

jihoonson:
build

PR #4298, opened 2026-02-20
Minor code fixes

pmattione-nvidia:
This addresses a few more minor code issues.  It removes a "using namespace" from a cpp header, it disentangles a class definition cycle, and it simplifies another function. 

The change to `RmmSpark` is to fix a hole in the exception handling. It uses a local variable before assigning to the static `sra` field, ensuring proper exception handling.


pmattione-nvidia:
build

PR #4301, opened 2026-02-20
Replace `cudf::detail::valid_if` with `cudf::bools_to_mask`

mythrocks:
This commit is part of the continuing effort to reduce the dependency of spark-rapids-jni on `cudf::detail` APIs.  In this commit, some of the references to `cudf::detail::valid_if` with `cudf::bools_to_mask`.

The functionality should not be altered.  Existing tests ought to cover the changes.


mythrocks:
Build

src/main/cpp/src/from_json_to_raw_map.cu
```diff
@@ -769,10 +769,17 @@ std::pair<rmm::device_buffer, cudf::size_type> create_null_mask(
                      });
   }
 
-  auto const valid_it          = should_be_nullified->view().begin<bool>();
-  auto [null_mask, null_count] = cudf::detail::valid_if(
-    valid_it, valid_it + should_be_nullified->size(), thrust::logical_not<bool>{}, stream, mr);
-  return {null_count > 0 ? std::move(null_mask) : rmm::device_buffer{0, stream, mr}, null_count};
+  rmm::device_uvector<bool> valids(should_be_nullified->size(), stream);
```
ttnghia: In case we are no longer be able to use `cudf::detail::valid_if`, now I am OK to make a copy of `valid_if` inside `spark-rapids-jni` so we can just call it, similar to what we have done with `make_counting_transform_iterator`.

ttnghia: If cudf continue to bar downstream libraries/applications from using its detail utilities, this will be the trend that we are unavoidable to follow, unfortunately. There are many more things that would be copied very soon.

mythrocks:
I'm closing this for now.  I might reopen this for `main`, with a copy of the `valid_if` iterator-based kernel.

Issue #4378, opened 2026-03-16
[FEA] Get rid of relaxed constexpr use cases


PointKernel:
**Is your feature request related to a problem? Please describe.**
https://github.com/rapidsai/cudf/pull/21703 removes all the use cases of relaxed constexpr in cudf and it breaks spark JNI according to CI in several places. See https://github.com/rapidsai/cudf/actions/runs/23165149643/job/67303140891?pr=21703

**Describe the solution you'd like**
The fix should be fairly straightforward, for places using STL utilities in device code, we should replace it with `cuda::std::` equivalents, e.g. 
https://github.com/NVIDIA/spark-rapids-jni/blob/e1d2df16ef534ed0c76eeb37650e797de619a88a/src/main/cpp/src/decimal_utils.cu#L1266
it should be 
```cuda
 bool const round_up = unsigned_floating > cuda::std::numeric_limits<std::int64_t>::max(); 
```

and https://github.com/NVIDIA/spark-rapids-jni/blob/e1d2df16ef534ed0c76eeb37650e797de619a88a/src/main/cpp/src/parse_uri.cu#L93-L96
should be marked as host-device, e.g. 
```cuda
 __host__ __device__ constexpr bool is_hex(char c) 
```
to fix the build error:
```bash
[...]

PR #4396, opened 2026-03-18
Replaced STL utilities used in device code with the corresponding `cuda::std::*` equivalents

zpuller:
Fixes https://github.com/NVIDIA/spark-rapids-jni/issues/4378

This PR removes all the use cases of relaxed constexpr in spark-rapids-jni to match the corresponding changes in cudf (https://github.com/rapidsai/cudf/pull/21703). It is needed to fix the jni build following that PR, but it doesn't depend on that PR so we could merge it anytime.

This change should not introduce any behavioral change, only compile time non-functional changes, and the behavior should thus be covered by existing tests.

### DIFF
--- a/src/main/cpp/src/hash/hive_hash.cu
+++ b/src/main/cpp/src/hash/hive_hash.cu
@@ -26,6 +26,7 @@
 #include <rmm/exec_policy.hpp>
 
 #include <cuda/functional>
+#include <cuda/std/type_traits>
 #include <thrust/iterator/counting_iterator.h>
 #include <thrust/tabulate.h>
 
@@ -165,7 +166,7 @@ class hive_device_row_hasher {
     : _check_nulls{check_nulls}, _table{t}
   {
     // Error out if passed an unsupported hash_function
-    static_assert(std::is_base_of_v<hive_hash_function<int>, hash_function<int>>,
+    static_assert(cuda::std::is_base_of_v<hive_hash_function<int>, hash_function<int>>,
                   "hive_device_row_hasher only supports the 'hive_hash_function' hash function");
   }
 
--- a/src/main/cpp/src/hash/murmur_hash.cu
+++ b/src/main/cpp/src/hash/murmur_hash.cu
@@ -25,6 +25,7 @@
 #include <rmm/exec_policy.hpp>
 
 #include <cuda/functional>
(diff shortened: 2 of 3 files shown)


zpuller:
build

src/main/cpp/src/cast_string.cu
ttnghia: Need `cuda::std::` here, but `CUDF_ENABLE_IF` is better and it also covers `cuda::std::`.
```suggestion
template <typename T, CUDF_ENABLE_IF(cuda::std::is_signed_v<T>)>
```

ttnghia: Same here.

src/main/cpp/src/parse_uri.cu
ttnghia:
```suggestion
    iter                = cuda::std::move(iter_);
```

src/main/cpp/src/hash/hive_hash.cu
ttnghia: Need `#include <cuda/std/utility>` and:
```suggestion
        : _column(cuda::std::move(col)), _idx_to_process(0), _cur_hash(HIVE_INIT_HASH)
```

src/main/cpp/src/row_conversion.cu
```diff
@@ -348,9 +350,9 @@ CUDF_KERNEL void copy_from_rows_fixed_width_optimized(const size_type num_rows,
 
     auto const shared_output_index  = threadIdx.x + (threadIdx.y * blockDim.x);
     auto const shared_output_stride = blockDim.x * blockDim.y;
-    auto const row_index_end        = std::min(num_rows, ((row_group_index + 1) * rows_per_group));
-    auto const num_rows_in_group    = row_index_end - (row_group_index * rows_per_group);
-    auto const shared_length        = row_size * num_rows_in_group;
+    auto const row_index_end = cuda::std::min(num_rows, ((row_group_index + 1) * rows_per_group));
```
ttnghia: This needs `#include <cuda/std/algorithm>`.

src/main/cpp/src/row_conversion.cu
ttnghia: Redundant after this transition.

zpuller:
build

PR #4431, opened 2026-04-03
Add GBK charset decoding kernel for StringDecode

thirtiseven:
Add a GPU kernel that decodes GBK-encoded binary data to UTF-8 strings, matching Java's GBK CharsetDecoder behavior for use with Spark's StringDecode expression.

- Lookup table generated from Java's GBK CharsetDecoder (REPORT mode), with generator script included for reproducibility
- Two-pass kernel using cudf::strings::detail::make_strings_children
- Invalid byte sequences replaced with U+FFFD
- Second byte 0x7F excluded from pair consumption (matches Java)
- Second byte 0xFF consumed as pair but mapped to U+FFFD (matches Java)
- Java unit test with full 24066 byte-pair end-to-end verification

### DIFF
--- a/src/main/cpp/src/charset_decode.cu
+++ b/src/main/cpp/src/charset_decode.cu
@@ -113,8 +113,8 @@ struct gbk_decode_fn {
   cudf::size_type const* input_offsets;
   uint16_t const* gbk_table;
 
-  cudf::size_type* d_sizes;
-  char* d_chars;
+  cudf::size_type* d_sizes{};
+  char* d_chars{};
   cudf::detail::input_offsetalator d_offsets;
 
   __device__ void operator()(cudf::size_type idx)
@@ -178,9 +178,6 @@ std::unique_ptr<cudf::column> decode_gbk(cudf::column_view const& input,
                                          rmm::cuda_stream_view stream,
                                          rmm::device_async_resource_ref mr)
 {
-  auto const num_rows = input.size();
-  if (num_rows == 0) { return cudf::make_empty_column(cudf::type_id::STRING); }
-
   CUDF_EXPECTS(input.type().id() == cudf::type_id::LIST,
                "Input must be LIST type (BinaryType)",
                std::invalid_argument);
@@ -192,19 +189,17 @@ std::unique_ptr<cudf::column> decode_gbk(cudf::column_view const& input,
   CUDF_EXPECTS(
     !child.nullable(), "Child column of binary column must be non-nullable", std::invalid_argument);
 
+  auto const num_rows = input.size();
+  if (num_rows == 0) { return cudf::make_empty_column(cudf::type_id::STRING); }
+
   auto const* gbk_table = get_gbk_table(stream);
 
   // offsets_begin() accounts for the parent list offset on sliced columns
-  auto [new_offsets, new_chars] =
-    cudf::strings::detail::make_strings_children(gbk_decode_fn{child.data<uint8_t>(),
-                                                               list_col.offsets_begin(),
-                                                               gbk_table,
-                                                               nullptr,
-                                                               nullptr,
-                                                               cudf::detail::input_offsetalator{}},
-                                                 num_rows,
-                                                 stream,
-                                                 mr);
+  auto [new_offsets, new_chars] = cudf::strings::detail::make_strings_children(
+    gbk_decode_fn{child.data<uint8_t>(), list_col.offsets_begin(), gbk_table},
+    num_rows,
+    stream,
+    mr);
 
   return cudf::make_strings_column(num_rows,
                                    std::move(new_offsets),
--- a/src/test/java/com/nvidia/spark/rapids/jni/CharsetDecodeTest.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/CharsetDecodeTest.java
@@ -215,8 +215,9 @@ public class CharsetDecodeTest {
     byte[][] empty = {};
 
     try (ColumnVector cv = binaryColumn(empty);
-         ColumnVector result = CharsetDecode.decode(cv, CharsetDecode.GBK)) {
-      AssertUtils.assertColumnsAreEqual(ColumnVector.fromStrings(), result);
+         ColumnVector result = CharsetDecode.decode(cv, CharsetDecode.GBK);
+         ColumnVector expected = ColumnVector.fromStrings()) {
+      AssertUtils.assertColumnsAreEqual(expected, result);
     }
   }
 


src/main/cpp/src/charset_decode.cu
```diff
@@ -0,0 +1,220 @@
...
+  std::call_once(gbk_table_init_flag, [stream]() {
+    gbk_table_device = std::make_unique<rmm::device_uvector<uint16_t>>(GBK_TABLE_SIZE, stream);
+    cudf::detail::cuda_memcpy_async(
+      cudf::device_span<uint16_t>(gbk_table_device->data(), GBK_TABLE_SIZE),
+      cudf::host_span<uint16_t const>(gbk_to_unicode_host, GBK_TABLE_SIZE),
+      stream);
+    stream.synchronize();
+  });
+  return gbk_table_device->data();
+}
```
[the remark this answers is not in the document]
thirtiseven: Acknowledged. This is a known pattern in the project — other static singletons like the bloom filter hash seed table (`bloom_filter.cu`) follow the same approach. In spark-rapids-jni, the CUDA context and RMM allocator are not torn down and re-initialized within an executor process lifetime.

If this becomes a project-wide concern, it should be addressed holistically rather than in this PR alone.

res-life (commented):
P1: This looks incorrect for sliced `LIST<UINT8>` inputs. `lists_column_view::offsets()` returns the backing offsets column, and passing `offsets.data<cudf::size_type>()` into `gbk_decode_fn` does not apply the parent list offset. For a sliced binary column, `idx == 0` will therefore use the unsliced row-0 boundaries rather than the first row in the slice, so `StringDecode(GBK)` can silently return wrong strings. Please use a slice-aware offset accessor (for example `offsets_begin()`, or otherwise account for `input.offset()`) and add a regression test that decodes a sliced binary column.

res-life:
Could we test the perf and give some numbers?

thirtiseven:
> Could we test the perf and give some numbers?

I added some e2e perf numbers in https://github.com/NVIDIA/spark-rapids/pull/14545

thirtiseven:
> P1: This looks incorrect for sliced `LIST<UINT8>` inputs. `lists_column_view::offsets()` returns the backing offsets column, and passing `offsets.data<cudf::size_type>()` into `gbk_decode_fn` does not apply the parent list offset. For a sliced binary column, `idx == 0` will therefore use the unsliced row-0 boundaries rather than the first row in the slice, so `StringDecode(GBK)` can silently return wrong strings. Please use a slice-aware offset accessor (for example `offsets_begin()`, or otherwise account for `input.offset()`) and add a regression test that decodes a sliced binary column.

nice catch! updated.

src/test/java/com/nvidia/spark/rapids/jni/CharsetDecodeTest.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+  }
+
+  @Test
+  void testZeroRowColumn() {
+    byte[][] empty = {};
+
+    try (ColumnVector cv = binaryColumn(empty);
+         ColumnVector result = CharsetDecode.decode(cv, CharsetDecode.GBK)) {
+      AssertUtils.assertColumnsAreEqual(ColumnVector.fromStrings(), result);
```
res-life: leak: ColumnVector.fromStrings()

```suggestion
@Test
void testZeroRowColumn() {
  byte[][] empty = {};
  try (ColumnVector cv = binaryColumn(empty);
       ColumnVector result = CharsetDecode.decode(cv, CharsetDecode.GBK);
       ColumnVector expected = ColumnVector.fromStrings()) {
    AssertUtils.assertColumnsAreEqual(expected, result);
  }
}
```

src/main/cpp/src/charset_decode.cu
```diff
@@ -0,0 +1,230 @@
...
+    if (!d_chars) { d_sizes[idx] = output_size; }
+  }
+};
+
+std::unique_ptr<cudf::column> decode_gbk(cudf::column_view const& input,
+                                         rmm::cuda_stream_view stream,
+                                         rmm::device_async_resource_ref mr)
+{
+  auto const num_rows = input.size();
+  if (num_rows == 0) { return cudf::make_empty_column(cudf::type_id::STRING); }
```
res-life: Zero-row path skips type validation

src/main/cpp/src/charset_decode.cu
```diff
@@ -0,0 +1,230 @@
...
+               "Input must be LIST<UINT8> (BinaryType)",
+               std::invalid_argument);
+  CUDF_EXPECTS(
+    !child.nullable(), "Child column of binary column must be non-nullable", std::invalid_argument);
+
+  auto const* gbk_table = get_gbk_table(stream);
+
+  // offsets_begin() accounts for the parent list offset on sliced columns
+  auto [new_offsets, new_chars] =
+    cudf::strings::detail::make_strings_children(gbk_decode_fn{child.data<uint8_t>(),
```
res-life: Verbose make_strings_children invocation. The functor is constructed with explicit nullptr, nullptr, cudf::detail::input_offsetalator{} for the three members that make_strings_children always overwrites internally. Other call sites in this repo omit them (see spark-rapids-jni/src/main/cpp/src/format_float.cu:90). Consider letting those default-initialize to keep the code consistent with peers:
```cpp
struct gbk_decode_fn {
  uint8_t const* input_data;
  cudf::size_type const* input_offsets;
  uint16_t const* gbk_table;
  cudf::size_type* d_sizes{};
  char* d_chars{};
  cudf::detail::input_offsetalator d_offsets;
  ...
};
auto [new_offsets, new_chars] = cudf::strings::detail::make_strings_children(
  gbk_decode_fn{child.data<uint8_t>(), list_col.offsets_begin(), gbk_table},
  num_rows, stream, mr);
```

res-life:
Only some small NITs, LGTM.

src/main/cpp/src/charset_decode.cu
thirtiseven: thanks, done.

src/main/cpp/src/charset_decode.cu
thirtiseven: nice catch

src/test/java/com/nvidia/spark/rapids/jni/CharsetDecodeTest.java
thirtiseven: done

thirtiseven:
build

PR #4107, opened 2025-12-23
Add a protocol buffer decode kernel

thirtiseven:
This PR completes the native GPU protobuf decoder used by the Spark `from_protobuf` expression. It converts a `LIST<INT8/UINT8>` column containing one serialized protobuf message per row into a nested cuDF `STRUCT` column.

The public API, schema validation, and JNI framework were introduced in [part 0](https://github.com/NVIDIA/cudf-spark-jni/pull/4373). The broader JNI and Spark integration plan is tracked in [NVIDIA/cudf-spark#14069](https://github.com/NVIDIA/cudf-spark/issues/14069).

### Supported protobuf features

- All protobuf scalar wire representations used by Spark: signed and unsigned varints, zigzag integers, fixed32/fixed64, float, double, bool, string, and bytes
- Nested messages up to `MAX_NESTING_DEPTH = 10`
- Repeated scalar and message fields, including packed and unpacked encodings
- Repeated fields inside nested and repeated messages
- Singular message merge semantics when the same message field appears multiple times
- Numeric enums and enum-as-string output
- Proto2 required fields and explicit or implicit default values
- `PERMISSIVE` and `FAILFAST` malformed-input handling
- Schema projection through per-field `isOutput` flags; hidden required fields are still validated
- Input slices and null input rows

### Protobuf compatibility details

- A known field encoded with the wrong wire type is skipped as an unknown occurrence.
- Raw 32-bit tags and length prefixes accept protobuf's ten-byte non-canonical varint representation.
- Unknown proto2 enum occurrences at the root follow Spark's same-number `UnknownFieldSet` check: `PERMISSIVE` nulls the top-level row and `FAILFAST` throws. Inside nested messages, unknown singular occurrences are ignored and unknown repeated enum elements are dropped because Spark does not recursively apply that root check.
- Invalid UTF-8 in `string` fields is repaired with replacement characters; `bytes` fields preserve the original byte sequence.
- Duplicate singular scalar fields use last-one-wins semantics, while duplicate singular message fields are merged.
- Schemas with more than 32 repeated fields are supported by batching occurrence-scan kernel launches.

### C++ layout

```
src/main/cpp/src/protobuf/
├── protobuf_types.cuh           Device-facing views, descriptors, and error types
├── protobuf_device_helpers.cuh  Wire parsing and UTF-8 device helpers
├── protobuf_host_helpers.hpp    Host-side work bundles, lookup tables, and launch helpers
├── protobuf_kernels.cuh         Templated extraction and construction kernels
[...]

PR #4435, opened 2026-04-07
from_protobuf kernel (part1): scalar field decoding

thirtiseven:
## `from_protobuf` kernel (part1): scalar field decoding [2/4]

Part 1 of https://github.com/NVIDIA/spark-rapids-jni/pull/4107

### Summary

Second PR in the series. Replaces the all-null stub from part0 with actual GPU decoding of flat (non-nested, non-repeated) scalar protobuf fields. After this PR, `decode_protobuf_to_struct` can decode messages with scalar fields end-to-end; repeated and nested fields still produce null columns.

### What is included

**Scan kernel** (`protobuf_kernels.cu`):
- `scan_all_fields_kernel`: one-thread-per-row parser that scans each protobuf message and records `(offset, length)` for all top-level fields. Supports "last one wins" semantics, unknown field skipping, and malformed-row detection.
- Real `check_required_fields_kernel` replacing the part0 stub.
- `validate_enum_values_kernel` with binary-search validation.
- `compute_enum_string_lengths_kernel` / `copy_enum_string_chars_kernel` for enum-as-string decoding.

**Decode pipeline** (`protobuf.cu`):
- Field classification into scalar / repeated / nested categories.
- Batched scalar extraction in 12 type groups (varint int32/uint32/int64/uint64/bool/zigzag32/zigzag64, fixed32/fixed64, float32/float64, fallback).
- Per-field extraction for STRING, BYTES (LIST\<UINT8\>), and enum-as-string fields.
- Required field validation (proto2 semantics).
- Input null mask propagation: null input rows produce null output struct rows (combined with PERMISSIVE-mode row invalidation).
- Repeated / nested fields fall through to `make_null_column_with_schema`.

**Enum builders** (`protobuf_builders.cu`):
- `build_enum_string_column`: validates enum values and converts to UTF-8 name strings.

### Test coverage (98 tests)

- Scalar types: varint, zigzag, fixed32/64, float/double special values, bool, string, bytes
- Default values: int, float, bool, string, mixed present/absent
- Required fields: present, missing (permissive + failfast), multiple fields, null input rows
- Enum: as-int, as-string, unknown values, mixed valid/unknown, row nulling in PERMISSIVE mode
- Error handling: malformed varint, truncated fields, wrong wire type, field number validation
- Unknown field skipping: varint, fixed64, length-delimited, fixed32
- Schema projection, last-one-wins, empty message, zero-row, null propagation

### Follow-up PRs

3. **Repeated + nested** — count/scan/build for repeated fields and recursive nested message decode
4. **Benchmarks + cleanup**

### DIFF
--- a/src/main/cpp/src/protobuf/protobuf.cu
+++ b/src/main/cpp/src/protobuf/protobuf.cu
@@ -862,18 +862,25 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   }
   if (fail_on_errors && h_error != 0) throw cudf::logic_error(error_message(h_error));
 
-  // Build final struct null mask from PERMISSIVE-mode row invalidation.
-  // Note: input null mask is NOT merged here — the Scala/Java caller handles input-row-level
-  // nulls via mergeAndSetValidity after the JNI call returns.
+  // Build final struct null mask by combining input nulls with PERMISSIVE-mode row invalidation.
   cudf::size_type struct_null_count = 0;
   rmm::device_buffer struct_mask{0, stream, mr};
+  auto const input_null_count = binary_input.null_count();
 
-  if (track_permissive_null_rows) {
+  if (track_permissive_null_rows || input_null_count > 0) {
+    auto const* input_mask = binary_input.null_mask();
+    auto input_offset      = binary_input.offset();
     auto [mask, null_count] = cudf::detail::valid_if(
       thrust::make_counting_iterator<cudf::size_type>(0),
       thrust::make_counting_iterator<cudf::size_type>(num_rows),
-      [row_invalid = d_row_force_null.data()] __device__(cudf::size_type row) {
-        return !row_invalid[row];
+      [row_invalid = track_permissive_null_rows ? d_row_force_null.data() : nullptr,
+       input_mask,
+       input_offset] __device__(cudf::size_type row) {
+        if (input_mask != nullptr && !cudf::bit_is_set(input_mask, input_offset + row)) {
+          return false;
+        }
+        if (row_invalid != nullptr && row_invalid[row]) return false;
+        return true;
       },
       stream,
       mr);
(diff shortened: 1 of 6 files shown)


src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -28,6 +31,112 @@ namespace detail {
...
+  auto const child_size     = child.size();
+  if (child_size == 0) { return; }
+
+  CUDF_EXPECTS(list_view.offset() == 0,
+               "decoder list null propagation expects unsliced list columns");
+  auto const* offsets_begin = list_view.offsets_begin();
+  auto const* offsets_end   = list_view.offsets_end();
+  // LIST children are not row-aligned with their parent. Expand the list-row null mask across
+  // every covered child element so direct access to the backing child column also observes nulls.
+  auto [element_mask, element_null_count] = cudf::detail::valid_if(
```
ttnghia, line 89: I'm not sure how long we still can use the `cudf::detail::` APIs. @mythrocks?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -31,37 +32,405 @@ namespace spark_rapids_jni::protobuf::detail {
...
+ * For "last one wins" semantics (protobuf standard for repeated scalars),
+ * we continue scanning even after finding a field.
+ *
+ * If a row hits a parse error that leaves the cursor in an unsafe state (for example, malformed
+ * varint bytes or a schema-matching field with the wrong wire type), the scan aborts for that row
+ * instead of guessing where the next field begins. In permissive mode the caller may also supply a
+ * row-level invalidity buffer so the full struct row can be nulled to match Spark CPU semantics for
+ * malformed messages.
+ */
+CUDF_KERNEL void scan_all_fields_kernel(
```
ttnghia, line 57: Side comment: We should add a macro JNI_KERNEL instead.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -31,37 +32,405 @@ namespace spark_rapids_jni::protobuf::detail {
...
+  cudf::column_device_view const d_in,
+  field_descriptor const* field_descs,  // [num_fields]
+  int num_fields,
+  int const* field_lookup,              // direct-mapped lookup table (nullable)
+  int field_lookup_size,                // size of lookup table (0 if null)
+  field_location* locations,            // [num_rows * num_fields] row-major
+  int* error_flag,
+  bool* row_has_invalid_data)
 {
+  auto row = static_cast<cudf::size_type>(blockIdx.x * blockDim.x + threadIdx.x);
```
ttnghia, line 67: Again, we are losing the ability to use `global_thread_id` from `cudf::detail::`. @mythrocks.

ttnghia (approved):
Assuming that we still can use some `cudf::detail::` API until some later time, let's move on with this. We will migrate later if needed.

thirtiseven:
build

PR #4448, opened 2026-04-09
Add bytes to hex conversion kernel

thirtiseven:
Contributes to https://github.com/NVIDIA/spark-rapids/issues/12550

This PR:

Add `bytesToHex` kernel that converts each byte of a STRING/BINARY column to its 2-character uppercase hex representation (e.g., `"AB"` → `"4142"`). This is used by `GpuHex` in spark-rapids for `hex(string)` and `hex(binary)` inputs.

- New CUDA kernel in `hex.cu` using the standard two-pass `make_strings_children` pattern
- JNI binding in `CastStringJni.cpp` with LIST<INT8> → STRING reinterpretation for BinaryType support
- Java API `CastStrings.bytesToHex(ColumnView)`

The `hex(long)` path reuses the existing `CastStrings.fromIntegersWithBase(col, 16)`.

### DIFF
--- a/src/main/cpp/src/CastStringJni.cpp
+++ b/src/main/cpp/src/CastStringJni.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2025, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2026, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -296,17 +296,22 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_bytesToHex(
     // STRING layout: data=chars bytes, children=[offsets]  (1 child)
     // LIST<INT8> layout: data=nullptr, children=[offsets, child_int8]  (2 children)
     auto const str_col = [&]() -> cudf::column_view {
-      if (col.type().id() == cudf::type_id::LIST) {
-        auto const lv = cudf::lists_column_view(col);
-        return cudf::column_view(cudf::data_type{cudf::type_id::STRING},
-                                 col.size(),
-                                 lv.child().head<char>(),
-                                 col.null_mask(),
-                                 col.null_count(),
-                                 col.offset(),
-                                 {lv.offsets()});
+      switch (col.type().id()) {
+        case cudf::type_id::LIST: {
+          auto const lv = cudf::lists_column_view(col);
+          CUDF_EXPECTS(lv.child().type().id() == cudf::type_id::INT8,
+                       "bytesToHex: LIST child must be INT8 (BinaryType)");
+          return cudf::column_view(cudf::data_type{cudf::type_id::STRING},
+                                   col.size(),
+                                   lv.child().head<char>(),
+                                   col.null_mask(),
+                                   col.null_count(),
+                                   col.offset(),
+                                   {lv.offsets()});
+        }
+        case cudf::type_id::STRING: return col;
+        default: CUDF_FAIL("bytesToHex: unsupported input type, expected STRING or LIST<INT8>");
       }
-      return col;
     }();
     auto const input = cudf::strings_column_view{str_col};
     auto result      = spark_rapids_jni::bytes_to_hex(input, cudf::get_default_stream());
--- a/src/main/cpp/src/CastStringJni.cpp
+++ b/src/main/cpp/src/CastStringJni.cpp
@@ -22,6 +22,7 @@
 #include <cudf/binaryop.hpp>
 #include <cudf/column/column_factories.hpp>
 #include <cudf/copying.hpp>
+#include <cudf/lists/lists_column_view.hpp>
 #include <cudf/replace.hpp>
 #include <cudf/scalar/scalar_factories.hpp>
 #include <cudf/strings/contains.hpp>
@@ -278,6 +279,38 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromInteger
       }
     }();
     return jni::release_as_jlong(result);
+  }
+  CATCH_CAST_EXCEPTION(env, 0);
+}
+
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_bytesToHex(JNIEnv* env,
+                                                                                jclass,
+                                                                                jlong input_column)
+{
+  JNI_NULL_CHECK(env, input_column, "input column is null", 0);
+  JNI_TRY
+  {
+    cudf::jni::auto_set_device(env);
+    auto const col = *reinterpret_cast<cudf::column_view const*>(input_column);
+    // BinaryType arrives as LIST<INT8>. Reinterpret as STRING for the kernel.
+    // STRING layout: data=chars bytes, children=[offsets]  (1 child)
+    // LIST<INT8> layout: data=nullptr, children=[offsets, child_int8]  (2 children)
+    auto const str_col = [&]() -> cudf::column_view {
+      if (col.type().id() == cudf::type_id::LIST) {
+        auto const lv = cudf::lists_column_view(col);
+        return cudf::column_view(cudf::data_type{cudf::type_id::STRING},
+                                 col.size(),
+                                 lv.child().head<char>(),
+                                 col.null_mask(),
+                                 col.null_count(),
+                                 col.offset(),
+                                 {lv.offsets()});
+      }
+      return col;
+    }();
+    auto const input = cudf::strings_column_view{str_col};
+    auto result      = spark_rapids_jni::bytes_to_hex(input, cudf::get_default_stream());
+    return cudf::jni::release_as_jlong(result);
   }
   CATCH_CAST_EXCEPTION(env, 0);
 }
--- a/src/main/cpp/src/cast_string.hpp
+++ b/src/main/cpp/src/cast_string.hpp
@@ -194,4 +194,20 @@ std::unique_ptr<cudf::column> parse_strings_to_date(
   rmm::cuda_stream_view stream      = cudf::get_default_stream(),
   rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
 
+/**
+ * @brief Convert a strings column to its hexadecimal representation.
+ *
+ * Each byte of each string is converted to a 2-character hex string (uppercase).
+ * For example, "AB" (bytes 0x41, 0x42) becomes "4142".
+ *
+ * @param input The input strings column
+ * @param stream Stream on which to operate.
+ * @param mr Memory resource for returned column
+ * @return A new strings column with hex representation
+ */
+std::unique_ptr<cudf::column> bytes_to_hex(
+  cudf::strings_column_view const& input,
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
+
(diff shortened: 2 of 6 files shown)


src/main/cpp/src/CastStringJni.cpp
```diff
@@ -282,6 +283,38 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromInteger
...
+      if (col.type().id() == cudf::type_id::LIST) {
+        auto const lv = cudf::lists_column_view(col);
+        return cudf::column_view(cudf::data_type{cudf::type_id::STRING},
+                                 col.size(),
+                                 lv.child().head<char>(),
+                                 col.null_mask(),
+                                 col.null_count(),
+                                 col.offset(),
+                                 {lv.offsets()});
+      }
```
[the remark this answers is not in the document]
thirtiseven: Thanks for the review. This is not correct — `lists_column_view::offsets()` does NOT incorporate the parent's offset. It simply returns the raw child column:

```cpp
// cudf/cpp/src/lists/lists_column_view.cu
column_view lists_column_view::offsets() const {
  return column_view::child(offsets_column_index);  // no slice, no offset adjustment
}
```

So `lv.offsets()` is equivalent to `col.child(0)`, and the `col.offset()` on the new STRING column_view correctly positions within it — same as how any non-sliced STRING column works internally. No double-offset.

src/main/cpp/src/hex.cu
```diff
@@ -0,0 +1,99 @@
...
+      for (cudf::size_type i = 0; i < nbytes; ++i) {
+        uint8_t const byte = in[i];
+        uint8_t const hi   = byte >> 4;
+        uint8_t const lo   = byte & 0x0F;
+        out[i * 2]         = hi < 10 ? '0' + hi : 'A' + (hi - 10);
+        out[i * 2 + 1]     = lo < 10 ? '0' + lo : 'A' + (lo - 10);
+      }
+    } else {
+      d_sizes[idx] = nbytes * 2;
+    }
```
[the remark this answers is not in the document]
thirtiseven: Valid in theory, but a single string >1GB is not possible in Spark. Also, the latest version of this PR uses a transform iterator to compute sizes from input offsets (no sizing kernel), and `make_offsets_child_column` handles the int32→int64 offset switch internally when the total byte count exceeds the threshold.

src/main/cpp/src/CastStringJni.cpp
res-life: 2026 Copyright

src/main/cpp/src/cast_string.hpp
```diff
@@ -194,4 +194,20 @@ std::unique_ptr<cudf::column> parse_strings_to_date(
...
+ *
+ * @param input The input strings column
+ * @param stream Stream on which to operate.
+ * @param mr Memory resource for returned column
+ * @return A new strings column with hex representation
+ */
+std::unique_ptr<cudf::column> bytes_to_hex(
+  cudf::strings_column_view const& input,
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
```
res-life: Use cudf::get_current_device_resource_ref()

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -282,6 +283,38 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromInteger
...
+  JNI_NULL_CHECK(env, input_column, "input column is null", 0);
+  JNI_TRY
+  {
+    cudf::jni::auto_set_device(env);
+    auto const col = *reinterpret_cast<cudf::column_view const*>(input_column);
+    // BinaryType arrives as LIST<INT8>. Reinterpret as STRING for the kernel.
+    // STRING layout: data=chars bytes, children=[offsets]  (1 child)
+    // LIST<INT8> layout: data=nullptr, children=[offsets, child_int8]  (2 children)
+    auto const str_col = [&]() -> cudf::column_view {
+      if (col.type().id() == cudf::type_id::LIST) {
```
res-life: A small switch with a clear error for unsupported type_ids would help operators and tests.

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -282,6 +283,38 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_CastStrings_fromInteger
...
+        auto const lv = cudf::lists_column_view(col);
+        return cudf::column_view(cudf::data_type{cudf::type_id::STRING},
+                                 col.size(),
+                                 lv.child().head<char>(),
```
res-life: For LIST, assert (or check and throw CastException) that lv.child().type().id() == INT8 (and ideally one list level). Mis-typed columns would otherwise be read as raw bytes incorrectly.

src/main/cpp/src/hex.cu
```diff
@@ -0,0 +1,126 @@
...
+                                           rmm::device_async_resource_ref mr)
+{
+  if (input.is_empty()) { return cudf::make_empty_column(cudf::type_id::STRING); }
+
+  // Build output offsets directly from input offsets — no sizing kernel needed.
+  auto const d_input_offsets =
+    cudf::detail::offsetalator_factory::make_input_iterator(input.offsets());
+  auto sizes_itr = cudf::detail::make_counting_transform_iterator(
+    cudf::size_type{0}, output_size_fn{d_input_offsets, input.offset()});
+  auto [offsets_column, total_bytes] = cudf::strings::detail::make_offsets_child_column(
```
res-life: Better to avoid using APIs in detail namespace, it's not intended to use.
Recently, Mithun filed several PRs to remove the using of detail.

res-life:
From AI:
output_size_fn and col_offset (Greptile P2, still worth fixing)
If offsetalator_factory::make_input_iterator(input.offsets()) already indexes the logical offset range for the strings_column_view, then adding input.offset() again when indexing (col_offset + idx) can double-apply the slice offset for non-zero-offset columns. JNI paths often use offset 0, so this may never show up in production today but is a footgun if bytes_to_hex is ever called from C++ on a sliced view. Safer options: confirm against cudf’s contract for that iterator and either drop col_offset from the index or document that the API requires an unsliced column.

src/main/cpp/src/hex.cu
```diff
@@ -0,0 +1,126 @@
...
+}  // namespace
+
+std::unique_ptr<cudf::column> bytes_to_hex(cudf::strings_column_view const& input,
+                                           rmm::cuda_stream_view stream,
+                                           rmm::device_async_resource_ref mr)
+{
+  if (input.is_empty()) { return cudf::make_empty_column(cudf::type_id::STRING); }
+
+  // Build output offsets directly from input offsets — no sizing kernel needed.
+  auto const d_input_offsets =
```
res-life: If we do not consider the sliced column.
The target offsets is the result of input offsets multipling 2.
So we may use like: thrust::transform(item -> item * 2)

src/main/cpp/src/hex.cu
thirtiseven: Good point. Replaced `make_offsets_child_column` with a direct `thrust::transform` on input offsets (multiply by 2). This also removes the need for `cudf/detail/iterator.cuh` and `cudf/strings/detail/strings_children.cuh`.

`offsetalator_factory` is still used as it has no public alternative, but that's consistent with the rest of the JNI codebase.

src/main/cpp/src/hex.cu
thirtiseven: Done — switched to `thrust::transform` with `offset * 2`. Total bytes computed via `input.chars_size(stream) * 2`. This is simpler and avoids the detail `make_offsets_child_column` API entirely.

Note: this assumes a non-sliced column (offset = 0), which is always the case for JNI columns.


(sorry ai just replied directly, i'm confirming...

src/main/cpp/src/hex.cu
```diff
@@ -0,0 +1,126 @@
...
+      (d_input_offsets[col_offset + idx + 1] - d_input_offsets[col_offset + idx]) * 2);
+  }
+};
+
+/**
+ * @brief Functor to convert each byte of a string to its 2-character hex representation.
+ */
+struct write_hex_fn {
+  cudf::column_device_view d_strings;
+  cudf::detail::input_offsetalator d_offsets;
```
res-life: Could we find a way to remove this `d_offsets` array? then we can touch less memories to speed up.

src/main/cpp/src/hex.cu
thirtiseven: I don't think there is a public api that can replace `make_offsets_child_column`. It handles the int32/int64 case, but `thrust::transform` doesn't.

src/main/cpp/src/cast_string.hpp
thirtiseven: done thanks.

src/main/cpp/src/CastStringJni.cpp
thirtiseven: done.

src/main/cpp/src/CastStringJni.cpp
thirtiseven: done.

src/main/cpp/src/hex.cu
thirtiseven: We need `make_offsets_child_column` to handles the INT32/INT64 switch for large strings. Added a comment here.

src/main/cpp/src/hex.cu
thirtiseven: Now updated to just keep a start pointer instead of an offsets array.

src/main/cpp/src/hex.cu
```diff
@@ -0,0 +1,127 @@
...
+  if (input.is_empty()) { return cudf::make_empty_column(cudf::type_id::STRING); }
+
+  // Build output offsets from input offsets. Each byte produces 2 hex chars,
+  // so output_size[i] = (input_offset[i+1] - input_offset[i]) * 2.
+  // make_offsets_child_column handles the INT32/INT64 switch for large strings.
+  auto const d_in_offsets =
+    cudf::detail::offsetalator_factory::make_input_iterator(input.offsets());
+  auto sizes_itr = thrust::make_transform_iterator(
+    thrust::make_counting_iterator(cudf::size_type{0}), output_size_fn{d_in_offsets});
+  auto [offsets_column, total_bytes] = cudf::strings::detail::make_offsets_child_column(
```
res-life, line 97: Could we avoid using APIs in the `detail` namespace?

thirtiseven: I can't find a public API to handle int32/64 cases. The only option I can see is to copy the 'make_offsets_child_column' out to the JNI repo, but that doesn't look ideal. We already use it in multiple places, so I think it can be done in a separate PR. What do you think?

https://github.com/NVIDIA/spark-rapids-jni/pull/4448#discussion_r3062857956

src/main/cpp/src/hex.cu
```diff
@@ -0,0 +1,127 @@
...
+
+namespace detail {
+namespace {
+
+/**
+ * @brief Functor to compute output sizes from input offsets.
+ *
+ * Each input byte produces 2 hex characters, so output size = input byte count * 2.
+ */
+struct output_size_fn {
```
res-life, line 45: If we do not allow sliced list, then we can avoid this fn.
IMO, we can disable sliced list, add check in JNI interface.

res-life (approved):
LGTM

thirtiseven:
build

PR #4453, opened 2026-04-10
Follow-up improvements to ParquetFooter from PR #4445 review

jihoonson:
This is a follow-up to #4445 to address review comments.

- Use `cudf::jni::native_jlongArray` in `NativeParquetJni.getRowIndexOffsets()` to safely handle OOM from `NewLongArray`
- Add `checkClosed()` to `ParquetFooter` methods to prevent use after close
- Remove dead null check on Java side
- Misc feedbacks: rename `singleIntColumnSchema` to `makeReadSchema`, remove unused import, fix misleading comment


src/main/java/com/nvidia/spark/rapids/jni/ParquetFooter.java
```diff
@@ -98,12 +98,19 @@ private ParquetFooter(long handle) {
     nativeHandle = handle;
   }
 
+  private void checkClosed() {
+    if (nativeHandle == 0) {
```
gerashegalov: nit: I think in this context 
```suggestion
    if (!nativeHandle) {
```
is more readable because I read it "if no nativeHandle"

src/main/java/com/nvidia/spark/rapids/jni/ParquetFooter.java
```diff
@@ -98,12 +98,19 @@ private ParquetFooter(long handle) {
     nativeHandle = handle;
   }
 
+  private void checkClosed() {
```
gerashegalov, line 101: I wonder if it's worth a method. It would not be necessarily the right exception (IllegalArgumentException) but we could just use 
```Scala
require(nativeHandle != 0, "ParquetFooter has already been closed")
```

src/main/java/com/nvidia/spark/rapids/jni/ParquetFooter.java
jihoonson: This is Java. It doesn't support implicit conversion between boolean and int.

src/main/java/com/nvidia/spark/rapids/jni/ParquetFooter.java
jihoonson: It's important to use the right exception type. Users can be confused otherwise. I realized that we have the `Preconditions.ensure()` for this use case which throws the right exception. But I still prefer keeping `checkClosed()` to avoid duplicating the same `Preconditions.ensure()` call everywhere. I just changed `checkClosed()` to relying on `Preconditions.ensure()` instead.

jihoonson:
build

jihoonson:
build

gerashegalov (approved):
LGTM

gerashegalov:
build

PR #4454, opened 2026-04-10
Append new authorized user to blossom-ci whitelist [skip ci]

jihoonson:
Add myself as an authorized blossom user.


gerashegalov:
@pxLi do we need separate lists per repo. It sounds like this list can be the same for all and should just live in spark-rapids-common

gerashegalov (approved):

jihoonson:
build

pxLi:
> @pxLi do we need separate lists per repo. It sounds like this list can be the same for all and should just live in spark-rapids-common

@gerashegalov This is a limitation of Blossom's ACL management. We did file a ticket requesting support for a single list shared across multiple repos, but it was marked as lower priority.

pxLi:
build

gerashegalov:
> > @pxLi do we need separate lists per repo. It sounds like this list can be the same for all and should just live in spark-rapids-common
> 
> @gerashegalov This is a limitation of Blossom's ACL management. We did file a ticket requesting support for a single list shared across multiple repos, but it was marked as lower priority.

Can we not import spark-rapid-common as a git submodule and symlink to .github?

pxLi:
> > > @pxLi do we need separate lists per repo. It sounds like this list can be the same for all and should just live in spark-rapids-common
> > 
> > 
> > @gerashegalov This is a limitation of Blossom's ACL management. We did file a ticket requesting support for a single list shared across multiple repos, but it was marked as lower priority.
> 
> Can we not import spark-rapid-common as a git submodule and symlink to .github?

The blossom-ci workflow file is monitored directly by the blossom-ci runner (workflow file content integrity check). 

We cannot move it into a submodule, because the submodule is only fetched after blossom-ci is triggered, and triggering blossom-ci requires this file to already be present in this repo. The feature requires blossom team to update their internal process first

PR #4475, opened 2026-04-16
[MapFromEntries] Native JNI implementation with Spark null-struct semantics

wjxiz1992:
## Summary

Adds `MapUtils.mapFromEntries` as a native JNI function in spark-rapids-jni to handle Spark's `map_from_entries` null-struct semantics correctly and efficiently.

### Motivation

The current fix in spark-rapids ([NVIDIA/spark-rapids#14612](https://github.com/NVIDIA/spark-rapids/pull/14612)) orchestrates ~10 cuDF column operations from Scala to handle the case where an input array contains null struct entries. Moving this logic into JNI:
- Reduces GPU kernel launches from ~10 → ~5
- Reduces GPU-CPU sync points from 2 → 1
- Simplifies spark-rapids Scala code to a single JNI call + `dedupByPolicy`

### Spark semantics implemented

- If a row's array contains any **null struct entry** (the whole struct is null), the output row is **null** — even if another entry in that same row has a null key inside a valid struct (CPU short-circuits on the null struct entry).
- If a row has **no null struct entry** but a valid struct's key is null → throw `"Cannot use null as map key."` (when `throwOnNullKey=true`).

### Implementation

**Fast path** (no null struct entries anywhere): uses `segmented_reduce(max)` over per-entry `key_is_null` to check for null keys within the visible slice range; AND with `is_valid(input)` to guard outer-null rows from triggering a throw.

**Slow path** (≥1 row has a null struct entry):
1. `contains_nulls` → per-row `has_null_entry` (1 kernel)
2. `is_null(keys) AND is_valid(structs)` → per-entry `null_key_in_valid` (2 kernels)
3. `segmented_reduce(max)` on flat bool with list offsets → per-row `row_has_null_key` (1 kernel)
4. `reduce(any, !has_null_entry AND row_has_null_key)` → throw check (1 kernel + **1 GPU-CPU sync**)
5. Build null mask via `bools_to_mask(!has_null_entry)` → AND with input's existing null mask using the offset-aware `bitmask_and(masks, begin_bits, mask_size)` raw overload (correctly handles sliced columns by consuming `input.null_mask()` from `input.offset()`); result_view uses `offset=0` with a `sliced_offsets` child view → `purge_nonempty_nulls` to satisfy cudf's invariant that null LIST rows have empty offset spans.

### New files

| File | Description |
|---|---|
| `src/main/cpp/src/map_utils.hpp` | C++ API declaration |
| `src/main/cpp/src/map_utils.cu` | Implementation |
| `src/main/cpp/src/MapUtilsJni.cpp` | JNI wrapper |
| `src/main/java/.../jni/MapUtils.java` | Java API |
| `src/test/java/.../jni/MapUtilsTest.java` | 13 unit tests |

### Local Validation

```
Tests run: 13, Failures: 0, Errors: 0, Skipped: 0
```

All 13 `MapUtilsTest` cases pass:
- No-null fast path (unchanged behavior)
- Null key in valid struct → throws / no-throw per policy
- Null struct entry → row masked to null
- All-null struct entries → all rows null
- Mixed edge case: `[null_struct, {null_key, 20}]` → null (not throw)
- Null struct in row 0 does not suppress throw for null key in row 1
- Empty list row handled correctly
- Outer null row preserved as null
- All outer-null rows remain null (throw policy)
- All outer-null rows remain null (no-throw policy) — verifies `is_valid` guard on reduce scalar
- Single outer-null row — boundary check for `is_valid` guard
- Outer-null row + null-struct-entry row both null — exercises `bitmask_and` with pre-existing null mask

### Follow-up in spark-rapids

Once this PR merges and a new spark-rapids-jni version is released, [NVIDIA/spark-rapids#14612](https://github.com/NVIDIA/spark-rapids/pull/14612) will be updated to replace the Scala slow path with a single call:

```scala
withResource(MapUtils.mapFromEntries(inputBase, throwOnNullKey)) { validated =>
  dedupByPolicy(validated)
}
```

Relates to NVIDIA/spark-rapids#14128.

---

Documentation
- [ ] Updated for new or modified user-facing features or behaviors
- [x] No user-facing change

Testing
- [x] Added or modified tests to cover new code paths
- [ ] Covered by existing tests
- [ ] Not required

Performance
- [ ] Tests ran and results are added in the PR description
- [ ] Issue filed with a link in the PR description
- [x] Not required


thirtiseven:
pre-commit.ci autofix

src/main/cpp/src/MapUtilsJni.cpp
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
thirtiseven, line 2: **[MUST FIX] Copyright year**

New files should use the current year only (`2026`), not `2025`. Convention: `.review/conventions.md` "Year enforcement (MUST FIX): New files: current year only". Applies to all five new files in this PR (`MapUtilsJni.cpp`, `map_utils.cu`, `map_utils.hpp`, `MapUtils.java`, `MapUtilsTest.java`).

```suggestion
 * Copyright (c) 2026, NVIDIA CORPORATION.
```

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,176 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
thirtiseven, line 2: **[MUST FIX] Copyright year**

```suggestion
 * Copyright (c) 2026, NVIDIA CORPORATION.
```

src/main/cpp/src/map_utils.hpp
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
thirtiseven, line 2: **[MUST FIX] Copyright year**

```suggestion
 * Copyright (c) 2026, NVIDIA CORPORATION.
```

src/main/java/com/nvidia/spark/rapids/jni/MapUtils.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
thirtiseven, line 2: **[MUST FIX] Copyright year**

```suggestion
 * Copyright (c) 2026, NVIDIA CORPORATION.
```

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
```diff
@@ -0,0 +1,192 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
thirtiseven, line 2: **[MUST FIX] Copyright year**

```suggestion
 * Copyright (c) 2026, NVIDIA CORPORATION.
```

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,176 @@
...
+#include <cudf/utilities/memory_resource.hpp>
+#include <cudf/utilities/span.hpp>
+
+namespace spark_rapids_jni {
+
+std::unique_ptr<cudf::column> map_from_entries(cudf::column_view const& input,
+                                               bool throw_on_null_key,
+                                               rmm::cuda_stream_view stream,
+                                               rmm::device_async_resource_ref mr)
+{
```
thirtiseven, line 42: **[MUST FIX] Missing `SRJ_FUNC_RANGE()`**

The spark-rapids-jni convention (`.review/conventions.md` §"NVTX Ranges (MUST FIX — overrides cudf)") requires `SRJ_FUNC_RANGE()` as the first statement of every main implementation function. Without it, this new public entry point is invisible in NVTX profiling traces.

Also need to include `nvtx_ranges.hpp`:

```diff
 #include "map_utils.hpp"
+#include "nvtx_ranges.hpp"
 #include <cudf/binaryop.hpp>
```

And add the macro as the first statement of `map_from_entries`:

```diff
 {
+  SRJ_FUNC_RANGE();
   CUDF_EXPECTS(input.type().id() == cudf::type_id::LIST,
                "map_from_entries: input must be a LIST column");
```

Reference pattern: `src/main/cpp/src/list_slice.cu:358`, `src/main/cpp/src/bloom_filter.cu:305`.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,176 @@
...
+               "map_from_entries: list child must be a STRUCT column");
+  CUDF_EXPECTS(structs.num_children() >= 1,
+               "map_from_entries: struct must have at least one child column (KEY)");
+
+  // Step 1: Per-row flag — does row i contain any null struct entry?
+  // contains_nulls returns BOOL8, size = input.size().
+  // A null outer row itself yields null in has_null_entry; bools_to_mask treats null as false
+  // (bit=0), so outer-null rows are correctly kept null in the final mask.
+  auto has_null_entry = cudf::lists::contains_nulls(lists_cv, stream, mr);
+
```
thirtiseven, line 60: **[SHOULD FIX] Scratch allocations should use `cudf::get_current_device_resource_ref()`, not the caller's `mr`**

Per cudf `DEVELOPER_GUIDE` ("Always use the default resource obtained from `cudf::get_current_device_resource_ref()` for temporary memory allocations") and spark-rapids-jni's memory-stream convention: only memory returned to the caller should use the passed-in `mr`; scratch/temporary buffers should use the default resource.

Every intermediate allocation in this function (`has_null_entry`, `any_null_entry_scalar`, `key_is_null`, `struct_is_null`, `struct_is_valid`, `null_key_in_valid`, `row_has_null_key`, `no_null_entry`, `should_throw`, `any_throw_scalar`, `keep_valid_col`, `entry_mask_uptr`, `combined_mask`) currently passes `mr`. Threading the caller's MR into short-lived scratch can fragment specialized pools the caller chose for output buffers only.

Suggested pattern — introduce a local `scratch_mr` and reserve `mr` only for the returned column and the final `purge_nonempty_nulls`:

```cpp
auto const scratch_mr = cudf::get_current_device_resource_ref();
auto has_null_entry = cudf::lists::contains_nulls(lists_cv, stream, scratch_mr);
// ... all intermediates use scratch_mr ...
// Fast-path returned column: mr
return std::make_unique<cudf::column>(input, stream, mr);
// Final output: mr
return cudf::purge_nonempty_nulls(result->view(), stream, mr);
```

thirtiseven (commented):
👍  Findings from Nghia's skill.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
```diff
@@ -0,0 +1,192 @@
...
+  // Helpers
+  private static HostColumnVector.StructData entry(Integer key, Integer value) {
+    return new HostColumnVector.StructData(Arrays.asList(key, value));
+  }
+
+  // --------------------------------------------------------------------------
+  // Fast-path tests (no null struct entries)
+  // --------------------------------------------------------------------------
+
+  @Test
```
thirtiseven, line 56: **[SHOULD FIX] Missing test for zero-row (empty column) early-return**

The C++ implementation at `src/main/cpp/src/map_utils.cu:46` has an explicit early return:

```cpp
if (input.size() == 0) { return cudf::make_empty_column(input.type()); }
```

This branch is entirely untested. If a future refactor removed the guard, no test would catch it.

Suggested test:

```java
@Test
void emptyColumnReturnsEmptyColumn() {
  try (ColumnVector input    = ColumnVector.fromLists(LIST_TYPE);
       ColumnVector resultT  = MapUtils.mapFromEntries(input, true);
       ColumnVector resultF  = MapUtils.mapFromEntries(input, false);
       ColumnVector expected = ColumnVector.fromLists(LIST_TYPE)) {
    assertColumnsAreEqual(expected, resultT);
    assertColumnsAreEqual(expected, resultF);
  }
}
```

src/main/cpp/src/MapUtilsJni.cpp
wjxiz1992: Fixed in commit 485dabc5 — all five files now use `Copyright (c) 2026`.

src/main/cpp/src/map_utils.cu
wjxiz1992: Fixed in commit 485dabc5 — all five files now use `Copyright (c) 2026`.

src/main/cpp/src/map_utils.hpp
wjxiz1992: Fixed in commit 485dabc5 — all five files now use `Copyright (c) 2026`.

src/main/java/com/nvidia/spark/rapids/jni/MapUtils.java
wjxiz1992: Fixed in commit 485dabc5 — all five files now use `Copyright (c) 2026`.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
wjxiz1992: Fixed in commit 485dabc5 — all five files now use `Copyright (c) 2026`.

wjxiz1992:
build

src/main/cpp/src/map_utils.cu
wjxiz1992: Fixed — `SRJ_FUNC_RANGE()` is present at `map_utils.cu:77` as the first statement of `map_from_entries`, and `nvtx_ranges.hpp` is included at line 18.

src/main/cpp/src/map_utils.cu
wjxiz1992: Fixed — all intermediate/scratch allocations now use `temp_mr = cudf::get_current_device_resource_ref()` (set at `map_utils.cu:91`). Only the final returned column uses the caller's `mr`. This matches the cudf DEVELOPER_GUIDE convention and spark-rapids-jni's memory-stream contract.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
wjxiz1992: Added in commit 0db14859 — `emptyColumnReturnsEmptyColumn()` covers the `if (input.size() == 0)` early-return path at `map_utils.cu:81`.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
```diff
@@ -0,0 +1,174 @@
...
+  void nullStructRowDoesNotSuppressThrowInOtherRow() {
+    // Row 0: [null_struct]      →  null (masked)
+    // Row 1: [{null_key, 20}]   →  must throw (valid struct, null key, no null struct entry)
+    List<HostColumnVector.StructData> row0 =
+        Arrays.asList((HostColumnVector.StructData) null);
+    List<HostColumnVector.StructData> row1 = Arrays.asList(entry(null, 20));
+    try (ColumnVector input = ColumnVector.fromLists(LIST_TYPE, row0, row1)) {
+      assertThrows(Exception.class, () -> MapUtils.mapFromEntries(input, true).close());
+    }
```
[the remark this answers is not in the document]
wjxiz1992: Fixed — both `assertThrows` calls now use `CudfException.class` (latest commit).

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,152 @@
...
+               "map_from_entries: input must be a LIST column");
+
+  if (input.size() == 0) { return cudf::make_empty_column(input.type()); }
+
+  auto const lists_cv = cudf::lists_column_view(input);
+  auto const structs  = lists_cv.child();
+  CUDF_EXPECTS(structs.type().id() == cudf::type_id::STRUCT,
+               "map_from_entries: list child must be a STRUCT column");
+  CUDF_EXPECTS(structs.num_children() >= 1,
+               "map_from_entries: struct must have at least one child column (KEY)");
```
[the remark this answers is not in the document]
wjxiz1992: The LIST type check (`CUDF_EXPECTS(input.type().id() == cudf::type_id::LIST)`) fires before the early return, so a non-LIST input is rejected. The STRUCT/arity checks are intentionally after the early return: a zero-row column has no entries to validate structure for, and `cudf::empty_like` preserves the full column type faithfully.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,152 @@
...
+      static_cast<cudf::numeric_scalar<bool>*>(any_throw_scalar.get())->value(stream);
+    if (any_throw) { throw cudf::logic_error("Cannot use null as map key."); }
+  }
+
+  // Null-mask rows that contain null struct entries.
+  // copy_if_else(lhs=null_scalar, rhs=input, mask=has_null_entry):
+  //   mask[i] = true  → null_scalar   (row had a null struct entry → output null row)
+  //   mask[i] = false → input[i]      (row was fine → keep original)
+  //   mask[i] = null  → input[i]      (outer null row stays null via input[i])
+  auto null_scalar = cudf::make_default_constructed_scalar(input.type(), stream, mr);
```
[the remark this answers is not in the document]
wjxiz1992: This code path no longer exists. The current slow-path implementation does not use `make_default_constructed_scalar` or `copy_if_else` at all — it builds the null mask directly via `bools_to_mask` + `bitmask_and` + `purge_nonempty_nulls`.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,152 @@
...
+    *has_null_entry,
+    *cudf::make_any_aggregation<cudf::reduce_aggregation>(),
+    cudf::data_type{cudf::type_id::BOOL8},
+    stream,
+    mr);
+  bool const any_null_entry =
+    any_null_entry_scalar->is_valid(stream) &&
+    static_cast<cudf::numeric_scalar<bool>*>(any_null_entry_scalar.get())->value(stream);
+
+  if (!any_null_entry) {
```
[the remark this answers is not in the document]
wjxiz1992: The `structs.null_count() == 0` short-circuit is a valid micro-optimisation. For now the current design (`contains_nulls` + `reduce(any)`) is one extra kernel on the fast path and readable. Happy to add the `null_count` guard if preferred — please let us know.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  @Test
+  void nullKeyInValidStructThrows() {
+    // [{null_key, 10}]  →  must throw because the struct is valid but key is null
+    List<HostColumnVector.StructData> row0 = Arrays.asList(entry(null, 10));
+    try (ColumnVector input = ColumnVector.fromLists(LIST_TYPE, row0)) {
+      assertThrows(Exception.class, () -> MapUtils.mapFromEntries(input, true).close());
+    }
```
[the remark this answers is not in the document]
wjxiz1992: Fixed — both `assertThrows` calls now use `CudfException.class` (latest commit).

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,152 @@
...
+  auto struct_is_null     = cudf::is_null(structs, stream, mr);  // flat BOOL8
+  auto struct_is_valid    = cudf::unary_operation(
+    *struct_is_null, cudf::unary_operator::NOT, stream, mr);     // flat BOOL8
+  auto null_key_in_valid  = cudf::binary_operation(
+    *key_is_null,
+    *struct_is_valid,
+    cudf::binary_operator::BITWISE_AND,
+    cudf::data_type{cudf::type_id::BOOL8},
+    stream,
+    mr);
```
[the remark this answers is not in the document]
wjxiz1992: Fixed — changed to `cudf::binary_operator::LOGICAL_AND` throughout.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,152 @@
...
+    any_null_entry_scalar->is_valid(stream) &&
+    static_cast<cudf::numeric_scalar<bool>*>(any_null_entry_scalar.get())->value(stream);
+
+  if (!any_null_entry) {
+    // All struct entries are valid.  Any null key in the flat key column is a real null key.
+    auto const keys = structs.child(0);
+    if (throw_on_null_key && keys.null_count(stream) > 0) {
+      throw cudf::logic_error("Cannot use null as map key.");
+    }
+    return std::make_unique<cudf::column>(input, stream, mr);
```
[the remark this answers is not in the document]
wjxiz1992: Addressed — added a comment at `map_utils.cu:107-110` explaining that outer-null LIST rows satisfy cudf's invariant of empty child segments (`offsets[i] == offsets[i+1]`), so they contribute zero keys and cannot inflate `null_count()`.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,152 @@
...
+    mr);
+
+  // Reduce per-list: does this row contain any entry where null_key_in_valid = true?
+  // segmented_reduce(max) over the flat boolean values using the list offsets as boundaries.
+  // is_null() always returns a fully-valid boolean column, so null_key_in_valid has no nulls;
+  // EXCLUDE null_policy only affects empty-list rows (which yield a null result, safely
+  // treated as false in the AND below).
+  auto const offsets_col  = lists_cv.offsets();
+  auto const offsets_span = cudf::device_span<cudf::size_type const>(
+    offsets_col.data<cudf::size_type>(), offsets_col.size());
```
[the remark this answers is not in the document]
wjxiz1992: Addressed — the code now uses `lists_cv.offsets_begin()` (offset-aware) to build `offsets_span`, covering exactly the visible rows of a sliced input. A comment at `map_utils.cu:97-100` explains this. The output-path offsets are also explicitly sliced with `raw_offsets.offset() + input.offset()`.

src/main/cpp/src/map_utils.cu
```diff
@@ -0,0 +1,239 @@
...
+std::unique_ptr<cudf::column> map_from_entries(cudf::column_view const& input,
+                                               bool throw_on_null_key,
+                                               rmm::cuda_stream_view stream,
+                                               rmm::device_async_resource_ref mr)
+{
+  SRJ_FUNC_RANGE();
+  CUDF_EXPECTS(input.type().id() == cudf::type_id::LIST,
+               "map_from_entries: input must be a LIST column");
+
+  if (input.size() == 0) { return cudf::empty_like(input); }
```
firestarman, line 267: This zero-row early return happens before the nested child validation below. That means an empty `LIST<non-STRUCT>` or empty `LIST<STRUCT<wrong arity>>` input would be accepted, even though the header docs say this API throws on non-`LIST(STRUCT(KEY,...))` inputs. If that contract matters, I think the STRUCT/arity checks need to move above the early return; otherwise the docs should be loosened to match the implementation.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
```diff
@@ -0,0 +1,236 @@
...
+    // Result must be an empty LIST column of the same type — no error thrown.
+    try (ColumnVector input    = ColumnVector.fromLists(LIST_TYPE);
+         ColumnVector result   = MapUtils.mapFromEntries(input, true);
+         ColumnVector expected = ColumnVector.fromLists(LIST_TYPE)) {
+      assertColumnsAreEqual(expected, result);
+    }
+  }
+
+  @Test
+  void outerNullRowAndNullStructEntryRowBothNull() {
```
firestarman, line 243: Given that the native implementation now has slice-aware offset/null-mask handling, can we add at least one sliced-input regression here too? The current tests all build whole columns with `ColumnVector.fromLists(...)`, so they do not exercise the new `input.offset()` / rebuilt-offsets path. A case where a null key or null struct entry exists outside the visible slice would help guard against future regressions in that logic.

src/main/cpp/src/map_utils.cu
res-life: Also get: Medium: map_from_entries() still returns success for zero-row inputs before it verifies the child shape. That means an empty LIST<non-STRUCT> or wrong-arity LIST<STRUCT<...>> will be accepted and returned unchanged instead of failing the documented contract. Because MapUtils.mapFromEntries() is a public Java API and forwards any ColumnView without prevalidation, this is user-reachable for empty inputs.

res-life:
- Could we test the performance in Spark-Rapids
- I asked AI: could we combin slow path and fast path to reduce code lines without get perf degrade? it said no, but it gives me a refactor to reduce the code. Please ask AI do to some refactor.

src/main/cpp/src/map_utils.cpp
```diff
@@ -0,0 +1,239 @@
...
+    cudf::column_view(input.type(),
+                      input.size(),
+                      nullptr,
+                      mask_ptr,
+                      combined_nc,
+                      /*offset=*/0,
+                      {sliced_offsets, input.child(cudf::lists_column_view::child_column_index)});
+  // purge_nonempty_nulls adjusts list offsets so each null outer row has an empty span,
+  // satisfying cudf's invariant that null rows in nested columns are empty.
+  return cudf::purge_nonempty_nulls(result_view, stream, mr);
```
res-life: purge_nonempty_nulls means a waste of memory, because the intermediate result may contain dirty entries, and intermediate result may be unnecessary if we can find a way to generate a clean result directly.

src/main/cpp/src/map_utils.cpp
```diff
@@ -0,0 +1,152 @@
...
+    any_null_entry_scalar->is_valid(stream) &&
+    static_cast<cudf::numeric_scalar<bool>*>(any_null_entry_scalar.get())->value(stream);
+
+  if (!any_null_entry) {
+    // All struct entries are valid.  Any null key in the flat key column is a real null key.
+    auto const keys = structs.child(0);
+    if (throw_on_null_key && keys.null_count(stream) > 0) {
+      throw cudf::logic_error("Cannot use null as map key.");
+    }
+    return std::make_unique<cudf::column>(input, stream, mr);
```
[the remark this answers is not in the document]
wjxiz1992: The `keys.null_count() > 0` check on line 144 is only a cheap early-exit filter — the actual throw decision uses `segmented_reduce` over `offsets_span` (built from `offsets_begin()`, slice-correct). A null key outside the visible slice would pass the `null_count()` guard but be excluded by the segmented reduce, so `reduce_any` never fires on it. No spurious throw possible. The comment block above the check documents this explicitly.

src/main/cpp/src/map_utils.cu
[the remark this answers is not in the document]
wjxiz1992: Addressed — `offsets_span` is now built from `lists_cv.offsets_begin()` with an explicit `input.size() + 1` length (map_utils.cu:130-131). `offsets_begin()` already accounts for `input.offset()`, so the slice-aware precondition is enforced rather than implicit, and the sliced-input regression tests added in e1edb87 exercise the path.

src/main/cpp/src/map_utils.cu
wjxiz1992: Fixed in b9eae67 — the `structs.type().id() == STRUCT` and `num_children() == 2` CUDF_EXPECTS are now above the `input.size() == 0` early return, so empty `LIST<non-STRUCT>` and empty `LIST<STRUCT<wrong arity>>` inputs fail with the documented error rather than passing through.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
wjxiz1992: Added in e1edb87 — four sliced-input regression tests (`slicedInputSkipsNullKeyOutsideSlice`, `slicedInputThrowsOnNullKeyInsideSlice`, `slicedInputSkipsNullStructOutsideSlice`, `slicedInputWithOuterNullAndNullStructInsideSlice`) use `subVector(...)` so `input.offset() != 0`. They cover both the fast path (`offsets_begin()`) and the slow path (`bitmask_and` with `begin_bits = input.offset()`), including a null key / null struct entry outside the visible slice to verify the slice is respected. Local run: `Tests run: 18, Failures: 0, Errors: 0, Skipped: 0`.

src/main/cpp/src/map_utils.cpp
wjxiz1992: `purge_nonempty_nulls` is needed because `bitmask_and` leaves non-empty child segments for the newly-nulled rows, and cudf's nested-null invariant requires null outer rows to have empty child ranges. The alternative — rebuilding the offsets child so null rows already have `offsets[i] == offsets[i+1]` — would replace one scan (`purge_nonempty_nulls`) with a conditional offsets rewrite plus a gather over the flat child, so net kernel count is about the same. I'll benchmark the two against representative inputs once the spark-rapids side integration (NVIDIA/spark-rapids#14612) is in place and switch if the direct-build path is materially faster; tracking as a follow-up so we don't block this PR on it.

wjxiz1992:
Thanks — addressing both:

**Refactor** — done in b9eae67. Extracted `any_null_key_in_guarded_row()` so the fast path and slow path now share the `is_null(keys)` → `segmented_reduce(MAX)` → `AND(guard)` → `reduce_any(...)` pipeline. Also dropped the redundant `key_is_null AND struct_is_valid` pre-mask on the slow path — rows with null struct entries are already short-circuited by the `no_null_entry` guard, so the extra binary_op kernel was unnecessary. Net: 55 lines removed, 51 added in the C++ (one shared helper replacing two duplicated blocks). All 18 `MapUtilsTest` cases still pass locally.

**Performance in spark-rapids** — will run the benchmark on the spark-rapids side via [NVIDIA/spark-rapids#14612](https://github.com/NVIDIA/spark-rapids/pull/14612) once a spark-rapids-jni snapshot that includes these changes is published, and post the numbers on that PR. Not blocking this merge.

thirtiseven:
> **Performance in spark-rapids** — will run the benchmark on the spark-rapids side via [NVIDIA/spark-rapids#14612](https://github.com/NVIDIA/spark-rapids/pull/14612) once a spark-rapids-jni snapshot that includes these changes is published, and post the numbers on that PR. Not blocking this merge.

FYI to test locally with jni and plugin please refer to https://github.com/NVIDIA/spark-rapids-jni/blob/main/CONTRIBUTING.md#local-testing-of-cross-repo-contributions-cudf-spark-rapids-jni-and-spark-rapids

In my humble opinion it's a better practice to identify performance issues earlier, so that we won't have to file another follow-up performance improvement PR a couple of days later after we found that in spark-rapids side.

wjxiz1992:
Ran a before/after microbenchmark comparing pre-refactor (at 0db1485) vs post-refactor (at b9eae67). Same machine, same inputs, only `map_utils.cu` swapped between runs. GPU: RTX PRO 6000 Blackwell (sm_120), CUDA 13.1, `throw_on_null_key=false`.

| Scenario | Pre (ms) | Post (ms) | Ratio |
|---|---:|---:|---:|
| fast / 4k rows × 4 / clean | 0.102 | 0.100 | 1.02× |
| fast / 64k rows × 4 / clean | 0.245 | 0.244 | 1.00× |
| fast / 1M rows × 4 / clean | 0.681 | 0.686 | 0.99× |
| fast / 64k rows × 16 / clean | 0.347 | 0.349 | 0.99× |
| fast / 1M rows × 16 / clean | 1.095 | 1.153 | 0.95× |
| fast / 1M rows × 4 / 1% null keys | 0.621 | 0.636 | 0.98× |
| **slow / 64k rows × 4 / 1% null struct** | **0.814** | **0.592** | **1.38×** |
| **slow / 1M rows × 4 / 1% null struct** | **4.703** | **1.697** | **2.77×** |
| **slow / 1M rows × 16 / 1% null struct** | **6.588** | **3.402** | **1.94×** |
| **slow / 1M rows × 4 / 10% null struct** | **4.370** | **1.456** | **3.00×** |

**Observations:**
- **Fast path unchanged** — within ±5% noise (same kernels, same work).
- **Slow path 1.4–3.0× faster**. Source of the speedup:
  - Dropped the redundant `key_is_null AND struct_is_valid` pre-mask (2 kernels removed).
  - Moved `is_null(keys)` + `segmented_reduce(MAX)` **inside** the `if (throw_on_null_key)` block. With `throw=false` (the common Spark path when policy allows null keys), the entire null-key check is now skipped — 4 kernels saved.
  - With `throw=true`, only the first optimisation applies (≈1 kernel saved); still faster but not by 3×.
- Slow path 10% nulls vs 1% nulls: post-refactor is flat — the extra work in the 1%/10% variants is in `purge_nonempty_nulls`, dominated by row count, not null rate.

Net: the refactor is not perf-neutral — it's a clear win on the slow path while leaving the fast path unchanged.

The spark-rapids end-to-end perf test (per earlier comment) will still run separately on [NVIDIA/spark-rapids#14612](https://github.com/NVIDIA/spark-rapids/pull/14612) once this PR merges and a snapshot is published.

wjxiz1992:
End-to-end CPU vs GPU benchmark in spark-rapids. Setup:

1. Packaged this PR's spark-rapids-jni as local SNAPSHOT JAR (`spark-rapids-jni-26.06.0-SNAPSHOT-cuda12.jar`).
2. Locally modified [NVIDIA/spark-rapids#14612](https://github.com/NVIDIA/spark-rapids/pull/14612)'s `GpuMapFromEntries.doColumnar` to replace the Scala slow-path orchestration with a single `MapUtils.mapFromEntries(inputBase, /*throwOnNullKey=*/true)` JNI call, as planned in the PR description.
3. Built the rapids-4-spark dist jar against the local JNI SNAPSHOT.
4. Ran spark-shell (Spark 3.3.0, `local[4]`, same input DataFrame cached, toggled `spark.rapids.sql.enabled` between CPU and GPU runs).

GPU: RTX PRO 6000 Blackwell, CUDA 13.1. Each scenario: 1 warm-up + 3 measurements, median reported. Timing covers the full `select(map_from_entries(...)).selectExpr("sum(size(m))").collect()` path (Spark scheduling + expression evaluation + JNI call), so numbers include non-trivial per-query overhead.

| Scenario (rows × entries/row × null-struct rate) | CPU (ms) | GPU (ms) | Speedup |
|---|---:|---:|---:|
| 1M × 4 × 0% (fast path) | 253 | 301 | 0.84× |
| 1M × 16 × 0% (fast path) | 545 | 285 | **1.91×** |
| 10M × 4 × 0% (fast path) | 1,371 | 776 | **1.77×** |
| 1M × 4 × 1% (slow path) | 174 | 233 | 0.75× |
| 1M × 16 × 1% (slow path) | 411 | 308 | **1.33×** |
| 10M × 4 × 1% (slow path) | 1,347 | 803 | **1.68×** |
| 1M × 4 × 10% (slow path) | 121 | 233 | 0.52× |
| 1M × 16 × 10% (slow path) | 84 | 291 | 0.29× † |
| 10M × 4 × 10% (slow path) | 891 | 829 | 1.08× |

† With 16 entries × 10% null-struct rate, virtually every input row contains ≥1 null struct entry, so the entire map is null for all rows — CPU short-circuits this early while GPU runs the full slow-path pipeline. Not a representative Spark workload.

**Takeaways:**

- **Where GPU wins: large inputs and wider rows** — 10M rows (1.7× fast path, 1.7× slow path) and 1M × 16 entries (1.9× fast path, 1.3× slow path). GPU amortises JNI + kernel-launch + Spark executor overhead at this scale.
- **Where CPU wins: small inputs with narrow rows** — 1M × 4 entries runs in 120–250 ms on CPU; the fixed per-query GPU overhead dominates at that size.
- The JNI slow path itself is 2–3× faster than the previous Scala orchestration at the kernel level (see [earlier microbenchmark comment](https://github.com/NVIDIA/spark-rapids-jni/pull/4475#issuecomment-4302472501)), but Spark pipeline overhead flattens that advantage in end-to-end measurements at small scale.

These numbers validate that the JNI path is not a regression at any scale and is a meaningful win (1.3–1.9×) once the row count grows past ~5M or entries-per-row grows past ~8.

wjxiz1992:
> End-to-end CPU vs GPU benchmark in spark-rapids. Setup:
> 
> 1. Packaged this PR's spark-rapids-jni as local SNAPSHOT JAR (`spark-rapids-jni-26.06.0-SNAPSHOT-cuda12.jar`).
> 2. Locally modified [NVIDIA/spark-rapids#14612](https://github.com/NVIDIA/spark-rapids/pull/14612)'s `GpuMapFromEntries.doColumnar` to replace the Scala slow-path orchestration with a single `MapUtils.mapFromEntries(inputBase, /*throwOnNullKey=*/true)` JNI call, as planned in the PR description.
> 3. Built the rapids-4-spark dist jar against the local JNI SNAPSHOT.
> 4. Ran spark-shell (Spark 3.3.0, `local[4]`, same input DataFrame cached, toggled `spark.rapids.sql.enabled` between CPU and GPU runs).
> 
> GPU: RTX PRO 6000 Blackwell, CUDA 13.1. Each scenario: 1 warm-up + 3 measurements, median reported. Timing covers the full `select(map_from_entries(...)).selectExpr("sum(size(m))").collect()` path (Spark scheduling + expression evaluation + JNI call), so numbers include non-trivial per-query overhead.
> 
> Scenario (rows × entries/row × null-struct rate)	CPU (ms)	GPU (ms)	Speedup
> 1M × 4 × 0% (fast path)	253	301	0.84×
> 1M × 16 × 0% (fast path)	545	285	**1.91×**
> 10M × 4 × 0% (fast path)	1,371	776	**1.77×**
> 1M × 4 × 1% (slow path)	174	233	0.75×
> 1M × 16 × 1% (slow path)	411	308	**1.33×**
> 10M × 4 × 1% (slow path)	1,347	803	**1.68×**
> 1M × 4 × 10% (slow path)	121	233	0.52×
> 1M × 16 × 10% (slow path)	84	291	0.29× †
> 10M × 4 × 10% (slow path)	891	829	1.08×
> † With 16 entries × 10% null-struct rate, virtually every input row contains ≥1 null struct entry, so the entire map is null for all rows — CPU short-circuits this early while GPU runs the full slow-path pipeline. Not a representative Spark workload.
> 
> **Takeaways:**
> 
> * **Where GPU wins: large inputs and wider rows** — 10M rows (1.7× fast path, 1.7× slow path) and 1M × 16 entries (1.9× fast path, 1.3× slow path). GPU amortises JNI + kernel-launch + Spark executor overhead at this scale.
> * **Where CPU wins: small inputs with narrow rows** — 1M × 4 entries runs in 120–250 ms on CPU; the fixed per-query GPU overhead dominates at that size.
> * The JNI slow path itself is 2–3× faster than the previous Scala orchestration at the kernel level (see [earlier microbenchmark comment](https://github.com/NVIDIA/spark-rapids-jni/pull/4475#issuecomment-4302472501)), but Spark pipeline overhead flattens that advantage in end-to-end measurements at small scale.
> 
> These numbers validate that the JNI path is not a regression at any scale and is a meaningful win (1.3–1.9×) once the row count grows past ~5M or entries-per-row grows past ~8.

@res-life @thirtiseven this is the e2e benchmark on spark-rapids, please re-evaluate this JNI implementation. GPU lost in some scenarios.

res-life:
About the perf of fast path, a Spark MAP<K,V> and cuDF LIST<STRUCT<K,V>> are the same physical layout — identical offsets, child struct, null mask. When every row is valid,  the input IS the output. The plugin just needs to:    
```scala
  // In GpuMapFromEntries.doColumnar:                    
  val inputBase = input.getBase
  val result = MapUtils.mapFromEntries(inputBase, throwOnNullKey)
  if (result == null) {                                                                                                                                                                       
    // Fast path — input unchanged, reinterpret LIST<STRUCT> as MAP.
    inputBase.incRefCount()                                                                                                                                                                   
  } else {                                                                                                                                                                                    
    result
  }                                                                                                                                                                                           
```

```cpp
// use a byte to indicates:
  constexpr std::uint8_t STATE_NULL     = 0;  // output row is null                                                                                                                           
  constexpr std::uint8_t STATE_VALID    = 1;  // output row is valid                                                                                                                          
  constexpr std::uint8_t STATE_NULL_KEY = 2;  // throw-worthy (valid struct + null key + policy on)

// For each row i:
//   - Outer LIST row is null           → STATE_NULL, size 0
//   - Otherwise scan entries in [offsets[i], offsets[i+1]):
//       * any null struct entry        → STATE_NULL, size 0
//       * else null key (throw policy) → STATE_NULL_KEY, size = end-start
//       * else                         → STATE_VALID, size = end-start
// ---------------------------------------------------------------------------
implement kernel to output a byte array and a size array.

if (cub::DeviceReduce::Max)  <= 1, return null column, then Spark-Rapids incRefCount
```

For the slow path, we can file a follow-up issue, or fix in this PR, from AI:
```
 Proposed redesign

  1. Reject sliced input. CUDF_EXPECTS(input.offset() == 0, ...). Callers with a slice materialize it first. Removes all bit-offset arithmetic on masks — the kernel only ever reads from bit 
  0 / index 0.
                                                                                                                                                                                              
  2. Single unified path — delete fast/slow branching. One code path with two custom kernels.                                                                                                 
   
  3. Phase 1 — byte-per-row state collection (no atomics, branchless per thread).                                                                                                             
                                                         
  byte row_state[i]:                                                                                                                                                                          
    0 = STATE_NULL      : outer-null OR row has a null struct entry
    1 = STATE_VALID     : valid row, no null key OR throw policy off                                                                                                                          
    2 = STATE_NULL_KEY  : valid struct has null key AND throw policy on                                                                                                                       
  size_type row_size[i]:                                                                                                                                                                      
    0 if STATE_NULL, else (offsets[i+1] - offsets[i])                                                                                                                                         
                                                                                                                                                                                              
  One thread per row. Each thread writes its own byte + size. No atomicOr, no atomicAdd, no shared memory, no __syncthreads. The state ordering (0 < 1 < 2) lets a single MAX reduction answer
   "does anyone need to throw?" and a single MIN reduction answer "are all rows already valid?".                                                                                              
                                                                                                                                                                                              
  4. Between phases — reduce, scan, then one bundled sync.                                                                                                                                    
   
  - cub::DeviceReduce::Max(row_state) → on-device max_state                                                                                                                                   
  - cub::DeviceReduce::Min(row_state) → on-device min_state
  - thrust::inclusive_scan(row_size) → output offsets written directly into the final offsets column                                                                                          
  - One bundled cudaMemcpy pulls {max_state, min_state, total_entries} across to the host — a single blocking sync for the whole function                                                     
  - max_state ≥ 2 → throw                                                                                                                                                                     
  - min_state ≥ 1 → fast-path signal: return nullptr, no further work                                                                                                                         
                                                                                                                                                                                              
  5. Fast-path cooperation with the Spark-RAPIDS plugin. A Spark MAP<K,V> is already stored in cuDF as LIST<STRUCT<K,V>> — identical offsets, identical struct child, identical null mask.    
  When every row is STATE_VALID, the input IS the output. The C++ function returns nullptr; the Java binding returns null; the plugin:                                                        
                                                                                                                                                                                              
  val result = MapUtils.mapFromEntries(inputBase, throwOnNullKey)
  if (result == null) inputBase.incRefCount()   // reinterpret LIST<STRUCT> as MAP — zero copies                                                                                              
  else                result                                                                                                                                                                  
                                                                                                                                                                                              
  No device allocation, no kernel past phase 1, no data movement. This directly kills the 0%-null regression.                                                                                 
                                                         
  6. Phase 2 — only runs on genuine slow path (some row is STATE_NULL, no throw).                                                                                                             
                                                         
  - cudf::bools_to_mask(row_state_view) — the standard cuDF utility turns the byte array into the output bitmask; BOOL8 treats state 0 as false, states 1/2 as true. Returns the null_count as
   a bonus.                                              
  - One-kernel gather-map build: each valid row writes its entries' source indices into [output_offsets[i], output_offsets[i+1]).                                                             
  - cudf::gather on the struct child — one call, handles arbitrary nested key/value types.                                                                                                    
  - cudf::make_lists_column assembles the result.                                                                                                                                             
                                                                                                                                                                                              
  No purge_nonempty_nulls, no bitmask_and, no segmented_reduce, no unary_operation, no binary_operation, no contains_nulls, no is_null, no is_valid, no reduce.      
```

src/main/cpp/tests/CMakeLists.txt
thirtiseven, line 1: 2026 copyright

wjxiz1992: Fixed in commit 938de3dc — `# Copyright (c) 2022-2026, NVIDIA CORPORATION.`. Thanks!

wjxiz1992:
Implemented the redesign proposed in https://github.com/NVIDIA/spark-rapids-jni/pull/4475#issuecomment-4310831247 — commit 6633fb21 (with copyright fix in 938de3dc on top).

## Implementation

**C++ (`src/main/cpp/src/map_utils.cu`)** — back from `.cpp` because the file now contains device kernels and CUB / Thrust device algorithms.

- **Phase 1 kernel** — one thread per row. No atomics, no shared memory, no `__syncthreads`. Emits `row_state ∈ {NULL=0, VALID=1, NULL_KEY=2}` and `row_size`. The state ordering encodes both "must throw" (`max == 2`) and "all valid" (`min == 1`) in a single byte.
- **Phase 1.5** — `cub::DeviceReduce::Max/Min(row_state)` + `thrust::inclusive_scan(row_size)` directly into the final offsets column. One bundled `cudaMemcpyAsync` pulls `{max_state, min_state, total_entries}` to host followed by a single `stream.synchronize()` — the only host sync in the whole function.
- **Decision tree**: `max_state == NULL_KEY` ⇒ throw; `min_state == VALID` ⇒ `return nullptr` (fast-path signal); otherwise ⇒ Phase 2.
- **Phase 2** — `cudf::bools_to_mask(row_state)` for the final null mask, a single one-thread-per-row gather-map kernel, one `cudf::gather` over the struct child (handles arbitrary nested key/value types), then `make_lists_column`.
- Drops `contains_nulls`, `is_null`, `is_valid`, `binary_operation`, `unary_operation`, `segmented_reduce`, `reduce`, `bitmask_and`, and `purge_nonempty_nulls`. **Sliced input is rejected outright** — removes all bit-offset arithmetic from the kernel.

**JNI / Java contract** — native returns `0` to signal the fast path; Java method returns `null`. A Spark `MAP<K,V>` and a cuDF `LIST<STRUCT<K,V>>` share the same physical layout, so the caller is expected to `incRefCount` the input and reinterpret it as the result — no allocation, no kernel past Phase 1, no copy.

**Tests** — kept the C++ gtest suite (renamed `StringKeyNonNullSucceeds` → `StringKeyNonNullFastPath`, expects `nullptr`); the 4 sliced-input pass-through tests are obsolete under the new contract (sliced is rejected) and were replaced by a single `slicedInputThrows` regression. The slow-path-no-throw and 70-row bitmask-boundary tests from b2e69df are preserved — they exercise the false branch of `throw_on_null_key` on the slow path and the gather-map kernel's bit alignment, respectively.

## Local validation (sm_89, dockerized build)

| Suite | Result |
|---|---|
| Java `MapUtilsTest` | 17 / 0 / 0 |
| C++ `MAP_UTILS` gtest | 6 / 0 / 0 |
| `ColumnViewNonEmptyNullsTest` | 4 / 0 / 0 |
| `CudaFatalTest` | 1 / 0 / 0 |

Plus, against an out-of-PR plugin patch in NVIDIA/spark-rapids#14612 that wires `MapUtils.mapFromEntries` into `GpuMapFromEntries.doColumnar` with the new `incRefCount` fast path:

| Plugin Suite | Result |
|---|---|
| `RapidsCollectionExpressionsSuite` (`MapFromEntries` GPU offload coverage) | 47 / 0 / 0 |

## End-to-end benchmark

Same setup as https://github.com/NVIDIA/spark-rapids-jni/pull/4475#issuecomment-4302885098 (Spark 3.3.0, `local[4]`, cached input, `select(map_from_entries(...)).selectExpr("sum(size(m))").collect()`, 1 warm-up + 3 measurements, median). **HW differs** — RTX 5880 Ada + CUDA 12.9.1 here vs. RTX PRO 6000 Blackwell + CUDA 13.1 in the prior comment — so absolute numbers aren't directly comparable; the speedup column is.

| Scenario (rows × entries × null-rate) | CPU (ms) | GPU (ms) | Speedup | Prior speedup |
|---|---:|---:|---:|---:|
| 1M × 4 × 0% (fast) | 382 | 225 | **1.70×** | 0.84× |
| 1M × 16 × 0% (fast) | 895 | 264 | **3.39×** | 1.91× |
| 10M × 4 × 0% (fast) | 2,394 | 731 | **3.27×** | 1.77× |
| 1M × 4 × 1% (slow) | 265 | 233 | **1.14×** | 0.75× |
| 1M × 16 × 1% (slow) | 774 | 289 | **2.68×** | 1.33× |
| 10M × 4 × 1% (slow) | 2,146 | 833 | **2.58×** | 1.68× |
| 1M × 4 × 10% (slow) | 207 | 182 | **1.14×** | 0.52× |
| 1M × 16 × 10% (slow) | 208 | 320 | 0.65× | 0.29× † |
| 10M × 4 × 10% (slow) | 1,616 | 876 | **1.84×** | 1.08× |

† CPU short-circuits on null struct entries and at 16 entries × 10% rate every input row contains ≥1 null struct, so CPU finishes early while GPU runs the full slow-path pipeline. Not a representative Spark workload.

The 0%-null fast path went from 0.84-1.91× to 1.70-3.39× — exactly the regression target the redesign aimed at — by returning `nullptr` from `map_from_entries` so the plugin can reinterpret the input as the result without a deep copy. The slow path is also faster across the board (1.14-2.68× on the rest) thanks to the single-sync, two-kernel pipeline replacing the `contains_nulls` + `bitmask_and` + `purge_nonempty_nulls` chain.

## Plugin-side follow-up

The plugin patch above is bench-only (kept in a local worktree) and not part of this PR. A spark-rapids plugin PR will follow once this lands, replacing the Scala `listContainsNulls` orchestration with the `MapUtils.mapFromEntries` + `incRefCount` fast path. The end-to-end perf gain only materialises after the plugin PR; this PR alone changes the JNI contract (`null` = fast path, sliced input rejected).

Ready for re-review — cc @res-life @thirtiseven.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
wjxiz1992: Update for the redesign in 6633fb21: the new C++ implementation **rejects sliced input outright** (`CUDF_EXPECTS(input.offset() == 0, ...)` plus the same on the struct child) instead of doing slice-aware offset/null-mask handling. This was a deliberate trade in the redesign — dropping the bit-offset arithmetic from the kernel keeps Phase 1 a single, branchless thread-per-row pass.

Consequences for these tests:
- The 4 sliced-pass-through tests (`slicedInputSkipsNullKeyOutsideSlice` / `slicedInputThrowsOnNullKeyInsideSlice` / `slicedInputSkipsNullStructOutsideSlice` / `slicedInputWithOuterNullAndNullStructInsideSlice`) are no longer applicable and were dropped.
- A single `slicedInputThrows` regression replaces them, pinning the new contract: `MapUtils.mapFromEntries(slicedView, …)` ⇒ `CudfException`.

Callers must materialize a slice (e.g. via `ColumnVector.copyToColumnVector`) before calling. The spark-rapids plugin patch in NVIDIA/spark-rapids#14612 already only sees offset==0 columns from `input.getBase`, so this is a non-issue end-to-end.

res-life:
From the desc in the [comment](https://github.com/NVIDIA/spark-rapids-jni/pull/4475#issuecomment-4341437234).
Seems we can swith some steps for fast path.
It's not need to do `thrust::inclusive_scan` and `sync` for fast path.

src/main/cpp/src/map_utils.hpp
```diff
@@ -0,0 +1,68 @@
...
+ *  - If a row's array contains any null struct entry (the whole struct is null), the output row
+ *    is null — even if another entry in that same row has a null key.
+ *  - If a row's array contains no null struct entry but does contain a null key inside a valid
+ *    struct, behavior depends on throw_on_null_key:
+ *      - true  → throws a logic_error ("Cannot use null as map key.")
+ *      - false → returns the row as-is (caller is responsible for deduplication policy)
+ *
+ * Fast-path contract: when every row would be returned unchanged (no null struct entries, and
+ * either no null key in a valid struct, or @p throw_on_null_key is false), this function
+ * returns @c nullptr.  A Spark MAP<K,V> and a cuDF LIST<STRUCT<K,V>> share the same physical
```
ttnghia: So if there is no changes needed: return nullptr. This can be error-prone if there is something wrong in the execution path and the function is returning nullptr due to error. The more explicit way, which is cleaner, is to add a function to check if the input is already a valid map:
```
bool is_valid_map(...);
``` 

This way will inspect the input by the proper way.

wjxiz1992:
## API split + fast-path optimisation — re-benchmark

Per @ttnghia's review on https://github.com/NVIDIA/spark-rapids-jni/pull/4475#discussion_r3163894107 the API has been split:

```cpp
bool is_valid_map(input, throw_on_null_key, stream, mr);                         // new
std::unique_ptr<cudf::column> map_from_entries(input, throw_on_null_key, ...);   // ALWAYS non-null
```

Plugin-side (NVIDIA/spark-rapids#14612 + local patch):
```scala
val effective = if (MapUtils.isValidMap(inputBase, true)) inputBase.incRefCount()
                else MapUtils.mapFromEntries(inputBase, true)
```

Per @res-life's note on https://github.com/NVIDIA/spark-rapids-jni/pull/4475#issuecomment-4342716359, the `thrust::inclusive_scan` and `total_entries` host pull are now done only on the slow path of `map_from_entries`.  `is_valid_map` and the deep-copy fast path of `map_from_entries` skip them entirely.

Sync count after the optimisation:

| Call | scan kernel | host syncs |
|---|---:|---:|
| `is_valid_map(*)` | 0 | 1 |
| `map_from_entries(valid input)` | 0 | 1 |
| `map_from_entries(invalid input)` | 1 | 2 |

### Benchmark — same 9 scenarios as https://github.com/NVIDIA/spark-rapids-jni/pull/4475#issuecomment-4341437234

Same hardware (RTX 5880 Ada, CUDA 12.9.1), same input shapes, same method (1 warm-up + 3 runs, median).  The "Prev" column is the original `nullptr`-signal design from commit 6633fb21; the "New" column is the current `is_valid_map` + `map_from_entries` design at HEAD.

| Scenario | CPU (ms) | **New GPU (ms) / speedup** | Prev GPU (ms) / speedup | Δ speedup |
|---|---:|---:|---:|---:|
| 1M × 4 × 0% (fast) | 419 | **258 / 1.62×** | 225 / 1.70× | -0.08× |
| 1M × 16 × 0% (fast) | 1,011 | **303 / 3.34×** | 264 / 3.39× | -0.05× |
| 10M × 4 × 0% (fast) | 3,263 | **794 / 4.11×** | 731 / 3.27× | **+0.84×** |
| 1M × 4 × 1% (slow) | 314 | **206 / 1.52×** | 233 / 1.14× | **+0.38×** |
| 1M × 16 × 1% (slow) | 826 | **313 / 2.64×** | 289 / 2.68× | -0.04× |
| 10M × 4 × 1% (slow) | 2,482 | **867 / 2.86×** | 833 / 2.58× | **+0.28×** |
| 1M × 4 × 10% (slow) | 234 | **195 / 1.20×** | 182 / 1.14× | **+0.06×** |
| 1M × 16 × 10% (slow) | 214 | **324 / 0.66×** | 320 / 0.65× | +0.01× |
| 10M × 4 × 10% (slow) | 1,824 | **900 / 2.03×** | 876 / 1.84× | **+0.19×** |

### Takeaway

Slow path does NOT regress — it actually improves on most scenarios.  Splitting the API means slow-path callers run Phase 1 twice (once in `is_valid_map`, once in `map_from_entries`), but on the same call site eliding the `inclusive_scan` / `total_entries` pull from the fast portion of `map_from_entries` more than offsets the second Phase 1.  Five scenarios improved by 0.06×–0.84×, three are within ±0.1× noise, none degraded meaningfully.

The 10M × 4 × 0% jump (3.27× → 4.11×) is the clearest signal that res-life's `inclusive_scan` elision was real cost on large-N fast-path inputs.

I think this addresses both review threads — happy to discuss further if either of you want a different shape.  cc @res-life @ttnghia.

res-life (approved):
LGTM

wjxiz1992:
build

wjxiz1992:
We will be off-work until 05/05, so keep it open until we come back from leave.

PR #4485, opened 2026-04-20
Drop _V2 suffix from URM/Artifactory symbols [skip ci]

pxLi:
After the artifactory migrations,
Drop _V2 suffix from URM/Artifactory symbols, use default VARs.


pxLi:
build

YanxuanLiu (approved):

PR #4493, opened 2026-04-23
Fix convert_to_rows dropping the trailing column at a tile boundary

thirtiseven:
## Summary

- `detail::determine_tiles` had an off-by-one in how it accounts for the first column of a new tile after the running tile is closed mid-loop. When the shmem-budget threshold happened to fire on the very last column of the table, the tail-close at the end of the loop was skipped and that column was never emitted to the kernel. `rmm::device_buffer` is not zero-initialized, so the dropped column read back whatever the pool allocator had there — typically 0, occasionally stale — yielding the non-deterministic `b = 0` failures tracked in NVIDIA/spark-rapids#10062.
- Fix: set `current_tile_width = 1` after the reset. `col` is the first (and so far only) column of the new tile.
- Add `ColumnToRowTests.PivotLikeLayout` that reproduces the failing schema (191 INT64 + 1 INT32). It reads the trailing INT32 from the raw JCUDF bytes directly rather than round-tripping through `convert_from_rows`, which would mask the bug by using the same layout code on the read side.

The failing pivot schema packs to exactly 192 columns; with the default 48 KB shmem budget and `tile_height = 32`, the 191 Longs fit in one tile (`1528 * 32 = 48896 ≤ 49136`) and adding the trailing Int tips the estimate to `1536 * 32 = 49152 > 49136`, firing the threshold on the last column — the shape required to hit this bug. The existing `convert_to_rows` tests all use uniform-size schemas, so the threshold either fired early (leaving later columns to re-increment the width) or never fired; none reached the "threshold fires on the last column" path.

Related to NVIDIA/spark-rapids#10062.

## Test plan

- [x] `ROW_CONVERSION` gtest, including the new `ColumnToRowTests.PivotLikeLayout`, passes
- [x] With this fix applied and the `isAcceleratedTransposeSupported` workaround reverted in spark-rapids, the `test_hash_multiple_grpby_pivot` integration test passes under `DATAGEN_SEED=1702610203`

### DIFF
--- a/src/main/cpp/src/row_conversion.cu
+++ b/src/main/cpp/src/row_conversion.cu
@@ -1733,12 +1733,7 @@ void determine_tiles(std::vector<size_type> const& column_sizes,
       row_size += col_size;  // alignment required for shared memory tile boundary to match
                              // alignment of output row
       current_tile_start_col = col;
-      // `col` becomes the first (and so far only) column of the new tile, so
-      // width starts at 1. Leaving it at 0 caused the tail-close below to be
-      // skipped whenever the threshold fired on the final column, silently
-      // dropping that column from the output; see
-      // https://github.com/NVIDIA/spark-rapids/issues/10062.
-      current_tile_width = 1;
+      current_tile_width     = 1;
     } else {
       row_size = row_size_with_this_col;
       current_tile_width++;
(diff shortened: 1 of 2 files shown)


ttnghia (approved):

ttnghia:
build

PR #4494, opened 2026-04-24
Refactor: add more check for RapidsInputFile.readVectored

res-life:
## Summary
- Add an output-buffer capacity check to the default `readVectored` on `RapidsInputFile`: throw `IllegalArgumentException` when any `CopyRange`'s `outputOffset + length` exceeds `output.getLength()`, before opening the stream.
- `CopyRange`'s constructor already rejects negative offsets and non-positive lengths, so those guards don't need to be duplicated.
- Motivation: spark-rapids subclasses (`HadoopInputFile`, `IcebergInputFile`) override `readVectored` almost entirely to add this one check. With the check here, those overrides can be deleted.

## Test plan
- [ ] `mvn test -Dtest=RapidsInputFileTest` — new `readVectoredRejectsRangeExceedingOutputBeforeOpeningStream` asserts `IllegalArgumentException` and `getOpenCount() == 0` (fail-fast before `open()`).
- [ ] Existing `RapidsInputFileTest` cases continue to pass.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFile.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFile.java
@@ -72,8 +72,14 @@ public interface RapidsInputFile {
     if (copyRanges.isEmpty()) {
       return;
     }
+    long outputLength = output.getLength();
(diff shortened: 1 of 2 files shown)


res-life:
build

liurenjie1024 (approved):

PR #4499, opened 2026-04-28
Add REPORT error action to GBK charset decode

thirtiseven:
### Description

Follow-up to #4431. Adds `java.nio.charset.CodingErrorAction.REPORT` semantics to the GBK decode kernel so the spark-rapids plugin can honor Spark 4.0+'s `spark.sql.legacy.codingErrorAction=false` without falling back to CPU.

#### Kernel

- `charset_decode` accepts a new `error_action { REPLACE, REPORT }`.
- In REPORT mode the functor receives a device-side `malformed_flag`; the three error branches atomically set it: invalid lead byte (incl. truncated lead and 0x80 / 0xFF), lead followed by invalid trail (`<0x40` or `==0x7F`), and table-mapped U+FFFD entries (Java's unmappable bucket, including `second == 0xFF`).
- Output is still emitted (U+FFFD) so the strings column is well-formed; the caller is expected to discard it on `malformed == true`.

#### API

- `decode_charset` now returns `decode_result { unique_ptr<column> output, bool malformed }`.
- JNI: extra `jint error_action` parameter; on `malformed == true` we throw `CharsetDecode$MalformedInputException` (unchecked).
- Java: new `REPLACE` / `REPORT` constants and a three-arg `decode` overload. The original two-arg `decode(cv, charset)` keeps REPLACE semantics, so existing callers are unaffected.

#### Tests

- Seven new `CharsetDecodeTest` cases: clean input, null rows mixed with valid rows, four individual malformed branches, and one batched cross-check that partitions every `(0x81..0xFE) x (0x40..0xFE)` byte pair by Java's REPORT decision and verifies the GPU agrees on both buckets.


src/main/cpp/src/charset_decode.cu
```diff
@@ -113,10 +114,21 @@ struct gbk_decode_fn {
...
   cudf::size_type* d_sizes{};
   char* d_chars{};
   cudf::detail::input_offsetalator d_offsets;
 
+  __device__ __forceinline__ void flag_malformed() const
+  {
+    // Any non-zero write is fine; use a plain store guarded by a load to avoid
+    // unnecessary atomic traffic once someone else has already flagged.
+    if (malformed_flag != nullptr && *malformed_flag == 0) { atomicExch(malformed_flag, 1); }
+  }
```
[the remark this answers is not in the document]
thirtiseven: Good catch, switched to unconditional atomicOr in 6c58e230.

src/main/cpp/src/charset_decode.cu
```diff
@@ -190,34 +211,49 @@ std::unique_ptr<cudf::column> decode_gbk(cudf::column_view const& input,
...
+    flag_ptr = flag->data();
+  }
+
   // offsets_begin() accounts for the parent list offset on sliced columns
   auto [new_offsets, new_chars] = cudf::strings::detail::make_strings_children(
-    gbk_decode_fn{child.data<uint8_t>(), list_col.offsets_begin(), gbk_table},
+    gbk_decode_fn{child.data<uint8_t>(), list_col.offsets_begin(), gbk_table, flag_ptr},
     num_rows,
     stream,
     mr);
```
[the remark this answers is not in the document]
thirtiseven: Done in 6c58e230 — gated flag_malformed on d_chars == nullptr so only the size pass fires the atomic.

wjxiz1992:
I found one REPORT-mode correctness issue around null parent LIST rows in `src/main/cpp/src/charset_decode.cu`.

`gbk_decode_fn::operator()` scans every row from `input_offsets[idx]` to `input_offsets[idx + 1]` without first checking whether the parent LIST row is valid. For a non-empty null row, malformed child bytes can set `malformed_flag`, and `decode_gbk` returns a malformed-input error before the output strings column is built with the copied parent null mask.

I reproduced this on this PR with a manually constructed `LIST<UINT8>`:

```text
child bytes = [0x81, 'A']
offsets     = [0, 1, 2]
validity    = row0 null, row1 valid
```

Expected REPORT result is `[null, "A"]`, because the malformed `0x81` is inside a null parent row. Actual result from `CharsetDecode.decode(input, CharsetDecode.GBK, CharsetDecode.REPORT)` is:

```text
CharsetDecode$MalformedInputException: malformed or unmappable input for charset decode
```

I ran the focused test with `-da:ai.rapids.cudf.AssertEmptyNulls`, so this reached `decodeNative` rather than being rejected by the Java-side non-empty-null assertion.

thirtiseven:
> I found one REPORT-mode correctness issue around null parent LIST rows in `src/main/cpp/src/charset_decode.cu`.

Good catch! Fixed.

src/main/cpp/src/charset_decode.cu
```diff
@@ -156,27 +183,34 @@ struct gbk_decode_fn {
...
+        int utf8_len    = codepoint_to_utf8(UNICODE_REPLACEMENT, out_ptr);
         if (out_ptr) { out_ptr += utf8_len; }
         output_size += utf8_len;
         i += 1;
       }
     }
 
     if (!d_chars) { d_sizes[idx] = output_size; }
+    if (report && local_malformed) {
+      cuda::atomic_ref<int32_t, cuda::thread_scope_device> ref(*malformed_flag);
```
res-life, line 204: suggest to use load-then-store pattern:
```cpp
  cuda::atomic_ref<int32_t, cuda::thread_scope_device> ref(*malformed_flag);
  if (ref.load(cuda::memory_order_relaxed) == 0) {
    ref.store(1, cuda::memory_order_relaxed);
  }
```

src/test/java/com/nvidia/spark/rapids/jni/CharsetDecodeTest.java
```diff
@@ -258,6 +296,143 @@ void testAllGbkDoubleBytePairs() {
...
+      if (reportsGbkJava(input)) {
+        assertThrows(CharsetDecode.MalformedInputException.class,
+            () -> CharsetDecode.decode(cv, CharsetDecode.GBK, CharsetDecode.REPORT));
+      } else {
+        try (ColumnVector result = CharsetDecode.decode(cv, CharsetDecode.GBK, CharsetDecode.REPORT);
+             ColumnVector expected = ColumnVector.fromStrings(decodeGbkJava(input))) {
+          AssertUtils.assertColumnsAreEqual(expected, result);
+        }
+      }
+    }
```
res-life: Confusing: behaviors vary by JDKs,  seems GPU can detect and adjust according to JDK version? But GPU code does not know JDK version. 
Could we double confirm all JDKs, java8, 11, 17, 21, and check GPU can vary according to java8, 11, 17, 21. If it does vary, then we should pass java version into GPU kernel to do shims.

res-life:
LGTM in general, just have 2 comments.

src/test/java/com/nvidia/spark/rapids/jni/CharsetDecodeTest.java
thirtiseven: Thanks, that was confusing as written. I checked the available local JDKs here (8, 11, and 17), and they all reject `0x80` in GBK REPORT mode. I updated the test to make this explicit: it now first asserts that the current Java GBK decoder reports `0x80` as malformed, and the failure message calls out that we need a GPU compatibility shim if a supported JDK accepts it.

src/main/cpp/src/charset_decode.cu
thirtiseven: Good point, done.

res-life (approved):
LGTM

thirtiseven:
build

PR #4495, opened 2026-04-27
from_protobuf kernel (part 2): Repeated fields + nested messages

thirtiseven:
## `from_protobuf` kernel (part 2): repeated fields + nested messages [3/4]

Part 2 of https://github.com/NVIDIA/spark-rapids-jni/pull/4107

### Summary

Third PR in the series. Replaces the null-column fallback for repeated and nested fields with actual GPU decoding. Adds the three-phase repeated-field pipeline (count → prefix-sum offsets → combined scan → build LIST columns) and recursive nested-message decoding up to 10 levels deep, including repeated-in-nested and repeated-in-repeated. After this PR, `decode_protobuf_to_struct` covers the full protobuf type system; only benchmarks and cleanup remain.

This is a pure insertion into the orchestrator: the new pipeline sits between the scalar section from part 1 and the final assembly. Existing scalar / enum-as-string / required-field / PERMISSIVE-mode logic is unchanged.

### What is included

**Repeated / nested kernels** (`protobuf_kernels.cu`):
- `count_repeated_fields_kernel`: per-row count of repeated occurrences with packed/unpacked auto-detection.
- `scan_all_repeated_occurrences_kernel`: combined-scan kernel that records `(offset, length, row_idx)` for every top-level repeated field in one launch.
- `scan_nested_message_fields_kernel`, `scan_repeated_message_children_kernel`: location scans inside nested messages and across all occurrences of a repeated message field.
- `count_repeated_in_nested_kernel`, `scan_repeated_in_nested_kernel`: repeated fields inside nested messages, packed and unpacked.
- `compute_nested_struct_locations_kernel`, `compute_grandchild_parent_locations_kernel`, `compute_virtual_parents_for_nested_repeated_kernel`, `compute_msg_locations_from_occurrences_kernel`, `extract_strided_locations_kernel`: GPU location-arithmetic helpers that replace D2H + CPU loop + H2D round-trips.

**Three-phase repeated pipeline** (`protobuf.cu`):
- Phase A — count repeated occurrences per row, exclusive-scan into per-row offsets.
- Phase B — allocate occurrence buffers and launch one combined scan kernel covering all top-level repeated fields.
- Phase C — type-dispatched LIST-column build (varint, fixed, zigzag, float/double, bool, string, bytes, enum-as-string, repeated message).

**Recursive nested decode** (`protobuf.cu` + `protobuf_builders.cu`):
[...]

PR #4504, opened 2026-04-29
from_protobuf kernel (part2): top-level repeated scalar / string / enum-as-string

thirtiseven:
Part 2 of https://github.com/NVIDIA/spark-rapids-jni/pull/4107.

This PR was originally attempted as #4495 but reviewers found it too large. Following the updated split plan (see #4107 for details), it has been broken into three smaller PRs (3a / 3b / 3c). This is **3a** — the smallest, focused on top-level repeated fields with non-message element types.

## What this adds

The three-phase top-level repeated pipeline (count → exclusive-scan offsets → combined occurrence scan → per-type LIST builders), restricted to non-message elements. After this PR, the following decode correctly at the top level:

- `repeated <int32 / int64 / uint32 / uint64 / float / double / bool>`
- `repeated string`, `repeated bytes`
- `repeated <enum-as-string>`

Repeated `MessageType` and nested messages still produce typed null columns; those are covered by 3b and 3c.

## Components

**Kernels** (`protobuf_kernels.cu`):
- `count_repeated_element` / `scan_repeated_element` device helpers (handle both packed and unpacked encoding per row)
- `count_repeated_fields_kernel` (called with `num_nested = 0` for now)
- `scan_all_repeated_occurrences_kernel` — single combined scan covering all top-level repeated fields per row
- launch wrappers for the above

**Builders** (`protobuf_builders.cu`):
- `build_repeated_string_column` (STRING and BYTES)
- `build_repeated_enum_string_column` — varint → validated enum → UTF-8 with PERMISSIVE row invalidation
- shared helpers: `make_list_offsets_from_counts`, `make_list_column_with_input_nulls`

**Orchestrator** (`protobuf.cu`):
- Field classification + per-row count buffers
- Phase A/B/C dispatch for non-STRUCT element types

## Tests

10 new tests in `ProtobufTest.java` (108 total pass):

- `testUnpackedRepeatedInt32`, `testPackedRepeatedDoubleWithMultipleFields`, `testRepeatedUint32`, `testRepeatedUint64`
- `testLargeRepeatedField`, `testMixedPackedUnpacked`
- `testPackedFixedMisaligned`, `testPackedFixedMisaligned64`
- `testRepeatedEnumAsString`
- `testPermissiveRepeatedWrongWireTypeDoesNotCorruptFollowingRow`

## Size

~1,700 lines net (5 files; +1,719 / −84). Pure addition to the orchestrator between the scalar section (part1) and the final assembly — no existing logic is modified.

### DIFF
--- a/src/main/cpp/src/protobuf/protobuf.cu
+++ b/src/main/cpp/src/protobuf/protobuf.cu
@@ -879,22 +879,20 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
                                                        has_def,
                                                        def_int);
 
-            if (schema_idx < static_cast<int>(enum_valid_values.size()) &&
-                schema_idx < static_cast<int>(enum_names.size())) {
-              auto const& valid_enums     = enum_valid_values[schema_idx];
-              auto const& enum_name_bytes = enum_names[schema_idx];
-              if (!valid_enums.empty() && valid_enums.size() == enum_name_bytes.size()) {
-                column_map[schema_idx] = build_enum_string_column(
-                  out, valid, valid_enums, enum_name_bytes, d_row_force_null, num_rows, stream, mr);
-              } else {
-                // Missing enum metadata for enum-as-string field; mark as decode error.
-                set_error_once_async(d_error.data(), ERR_MISSING_ENUM_META, stream);
-                column_map[schema_idx] = make_null_column(dt, num_rows, stream, mr);
-              }
-            } else {
-              set_error_once_async(d_error.data(), ERR_MISSING_ENUM_META, stream);
-              column_map[schema_idx] = make_null_column(dt, num_rows, stream, mr);
-            }
+            // Missing or mismatched enum metadata is a schema/config bug (host-side check),
+            // not a row-level decode error — fail loudly rather than silently produce a null
+            // column that downstream can't distinguish from a real all-null result.
+            CUDF_EXPECTS(schema_idx < static_cast<int>(enum_valid_values.size()) &&
+                           schema_idx < static_cast<int>(enum_names.size()),
+                         "Protobuf decode error: missing or mismatched enum metadata for "
+                         "enum-as-string field");
+            auto const& valid_enums     = enum_valid_values[schema_idx];
+            auto const& enum_name_bytes = enum_names[schema_idx];
+            CUDF_EXPECTS(!valid_enums.empty() && valid_enums.size() == enum_name_bytes.size(),
+                         "Protobuf decode error: missing or mismatched enum metadata for "
+                         "enum-as-string field");
+            column_map[schema_idx] = build_enum_string_column(
+              out, valid, valid_enums, enum_name_bytes, d_row_force_null, num_rows, stream, mr);
           } else {
             // Regular protobuf STRING (length-delimited)
             bool has_def_str    = has_def;
@@ -955,9 +953,9 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
           break;
         }
         default:
-          // For LIST (bytes) and other unsupported types, create placeholder columns
-          column_map[schema_idx] = make_null_column(dt, num_rows, stream, mr);
-          break;
+          // Unreachable: schema validation only admits the scalar element types enumerated
+          // above (STRUCT is dispatched through the nested path, not this scalar switch).
+          CUDF_FAIL("Protobuf decode internal error: unsupported scalar element type");
       }
     }
   }
@@ -979,7 +977,10 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
         extract_strided_count{d_repeated_info.data(), ri, num_repeated});
       auto counts_end = counts_begin + num_rows;
 
-      // Reduce in int64 to detect overflow before truncating to int32 LIST offsets.
+      // Reduce the per-row counts into a single int64 total. Using int64 lets us detect
+      // overflow before truncating into the int32 LIST offsets buffer (cudf list offsets are
+      // int32). The same total is reused as the `offsets[num_rows]` sentinel below, so this
+      // single reduce serves both the overflow guard and the offsets construction.
       int64_t total_64 =
         thrust::reduce(rmm::exec_policy_nosync(stream), counts_begin, counts_end, int64_t{0});
       CUDF_EXPECTS(total_64 <= std::numeric_limits<int32_t>::max(),
@@ -1053,6 +1054,13 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       auto element_type   = cudf::data_type{schema[schema_idx].output_type};
       int32_t total_count = w.total_count;
 
+      // Repeated MessageType is not supported in this part; full implementation lands in
+      // parts 3b/3c. Fail fast rather than silently returning a null LIST<STRUCT>, which
+      // would be indistinguishable from a real all-null result downstream.
+      CUDF_EXPECTS(
+        element_type.id() != cudf::type_id::STRUCT,
+        "Protobuf decode: repeated MessageType is not yet supported (covered by parts 3b/3c)");
+
       if (total_count <= 0) {
         // All rows empty: w.offsets is already a zero-filled buffer from Phase A.
         auto offsets_col = std::make_unique<cudf::column>(cudf::data_type{cudf::type_id::INT32},
@@ -1061,16 +1069,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
                                                           rmm::device_buffer{},
                                                           0);
 
-        std::unique_ptr<cudf::column> child_col;
-        auto const child_type_id =
-          static_cast<cudf::type_id>(h_device_schema[schema_idx].output_type_id);
-        if (child_type_id == cudf::type_id::STRUCT) {
-          child_col =
-            make_empty_struct_column_with_schema(schema, schema_idx, num_fields, stream, mr);
-        } else {
-          child_col = make_empty_column_safe(element_type, stream, mr);
-        }
-
+        auto child_col         = make_empty_column_safe(element_type, stream, mr);
         column_map[schema_idx] = make_list_column_with_input_nulls(
           num_rows, std::move(offsets_col), std::move(child_col), binary_input, stream, mr);
         continue;
@@ -1078,10 +1077,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
       auto& d_occurrences = *w.occurrences;
 
-      // For repeated fields, output_type holds the element type (not the outer LIST type).
-      auto child_type_id = static_cast<cudf::type_id>(h_device_schema[schema_idx].output_type_id);
-
-      switch (child_type_id) {
+      // For repeated fields, schema[].output_type holds the element type (not the outer LIST).
+      switch (element_type.id()) {
         case cudf::type_id::INT32:
           column_map[schema_idx] =
             build_repeated_scalar_column<int32_t>(binary_input,
@@ -1189,28 +1186,26 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
           auto const field_meta = make_field_meta_view(context, schema_idx);
           auto enc              = field_meta.schema.encoding;
           if (enc == proto_encoding::ENUM_STRING) {
-            if (!field_meta.enum_valid_values.empty() &&
-                field_meta.enum_valid_values.size() == field_meta.enum_names.size()) {
-              column_map[schema_idx] =
-                build_repeated_enum_string_column(binary_input,
-                                                  message_data,
-                                                  list_offsets,
-                                                  base_offset,
-                                                  std::move(w.offsets),
-                                                  d_occurrences,
-                                                  total_count,
-                                                  num_rows,
-                                                  field_meta.enum_valid_values,
-                                                  field_meta.enum_names,
-                                                  d_row_force_null,
-                                                  d_error,
-                                                  stream,
-                                                  mr);
-            } else {
-              set_error_once_async(d_error.data(), ERR_MISSING_ENUM_META, stream);
-              column_map[schema_idx] = make_null_column(
-                cudf::data_type{schema[schema_idx].output_type}, num_rows, stream, mr);
-            }
+            // Same host-side schema check as the scalar enum path — fail loudly instead of
+            // silently emitting a null column.
+            CUDF_EXPECTS(!field_meta.enum_valid_values.empty() &&
+                           field_meta.enum_valid_values.size() == field_meta.enum_names.size(),
+                         "Protobuf decode error: missing or mismatched enum metadata for "
+                         "enum-as-string field");
+            column_map[schema_idx] = build_repeated_enum_string_column(binary_input,
+                                                                       message_data,
+                                                                       list_offsets,
+                                                                       base_offset,
+                                                                       std::move(w.offsets),
+                                                                       d_occurrences,
+                                                                       total_count,
+                                                                       num_rows,
+                                                                       field_meta.enum_valid_values,
+                                                                       field_meta.enum_names,
+                                                                       d_row_force_null,
+                                                                       d_error,
+                                                                       stream,
+                                                                       mr);
           } else {
             column_map[schema_idx] = build_repeated_string_column(binary_input,
                                                                   message_data,
@@ -1244,10 +1239,10 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
                                                                 mr);
           break;
         default:
-          // Composite element types (STRUCT etc.) are placeholders for parts 3b/3c.
-          column_map[schema_idx] = make_null_list_column_with_child(
-            make_empty_column_safe(element_type, stream, mr), num_rows, stream, mr);
-          break;
+          // Unreachable: schema validation admits the types enumerated above; STRUCT is
+          // rejected at the loop entry above (parts 3b/3c). Reaching this branch means a
+          // schema slipped past validation.
+          CUDF_FAIL("Protobuf decode internal error: unsupported repeated element type");
       }
     }  // for (ri)
   }
--- a/src/main/cpp/src/protobuf/protobuf.cu
+++ b/src/main/cpp/src/protobuf/protobuf.cu
@@ -21,6 +21,7 @@
 
 #include <thrust/binary_search.h>
 
+#include <limits>
 #include <set>
 #include <string>
 #include <unordered_set>
@@ -392,10 +393,33 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   auto const* message_data = reinterpret_cast<uint8_t const*>(in_list_view.child().data<int8_t>());
   auto const* list_offsets = in_list_view.offsets().data<cudf::size_type>();
 
-  cudf::size_type base_offset = 0;
-  CUDF_CUDA_TRY(cudaMemcpyAsync(
-    &base_offset, list_offsets, sizeof(cudf::size_type), cudaMemcpyDeviceToHost, stream.value()));
+  // Stage list_offsets[0] through pinned host memory so the D2H stays truly async (pageable
+  // destinations turn small cudaMemcpyAsync calls into synchronous bounce-buffer copies).
+  auto h_base_offset = cudf::detail::make_pinned_vector_async<cudf::size_type>(1, stream);
+  CUDF_CUDA_TRY(cudaMemcpyAsync(h_base_offset.data(),
+                                list_offsets,
+                                sizeof(cudf::size_type),
+                                cudaMemcpyDeviceToHost,
+                                stream.value()));
   stream.synchronize();
+  cudf::size_type base_offset = h_base_offset[0];
+
+  // Scratch allocations consumed inside this function go through the current device resource;
+  // only buffers that flow into the returned column should use the caller-supplied `mr`.
+  auto const scratch_mr = cudf::get_current_device_resource_ref();
+
+  // Copy schema to device
+  std::vector<device_nested_field_descriptor> h_device_schema(num_fields);
+  for (int i = 0; i < num_fields; i++) {
+    h_device_schema[i] = device_nested_field_descriptor{schema[i]};
+  }
+
+  rmm::device_uvector<device_nested_field_descriptor> d_schema(num_fields, stream, scratch_mr);
+  CUDF_CUDA_TRY(cudaMemcpyAsync(d_schema.data(),
+                                h_device_schema.data(),
+                                num_fields * sizeof(device_nested_field_descriptor),
+                                cudaMemcpyHostToDevice,
+                                stream.value()));
 
   auto d_in = cudf::column_device_view::create(binary_input, stream);
   // Identify repeated and nested fields at depth 0
@@ -415,7 +439,9 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     }
   }
 
-  int num_scalar = static_cast<int>(scalar_field_indices.size());
+  int num_repeated = static_cast<int>(repeated_field_indices.size());
+  int num_nested   = static_cast<int>(nested_field_indices.size());
+  int num_scalar   = static_cast<int>(scalar_field_indices.size());
 
   // Error flag
   rmm::device_uvector<int> d_error(1, stream, cudf::get_current_device_resource_ref());
@@ -434,7 +460,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       case ERR_REQUIRED: return "Protobuf decode error: missing required field";
       case ERR_SCHEMA_TOO_LARGE:
         return "Protobuf decode error: schema exceeds maximum supported repeated fields per kernel "
-               "(128)";
+               "(32)";
       case ERR_MISSING_ENUM_META:
         return "Protobuf decode error: missing or mismatched enum metadata for enum-as-string "
                "field";
@@ -457,6 +483,77 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
   auto const threads = THREADS_PER_BLOCK;
   auto const blocks  = static_cast<int>((num_rows + threads - 1u) / threads);
+
+  // Allocate for counting repeated fields
+  rmm::device_uvector<repeated_field_info> d_repeated_info(
+    num_repeated > 0 ? static_cast<size_t>(num_rows) * num_repeated : 1, stream, scratch_mr);
+  rmm::device_uvector<field_location> d_nested_locations(
+    num_nested > 0 ? static_cast<size_t>(num_rows) * num_nested : 1, stream, scratch_mr);
+
+  rmm::device_uvector<int> d_repeated_indices(
+    num_repeated > 0 ? num_repeated : 1, stream, scratch_mr);
+  rmm::device_uvector<int> d_nested_indices(num_nested > 0 ? num_nested : 1, stream, scratch_mr);
+
+  if (num_repeated > 0) {
+    CUDF_CUDA_TRY(cudaMemcpyAsync(d_repeated_indices.data(),
+                                  repeated_field_indices.data(),
+                                  num_repeated * sizeof(int),
+                                  cudaMemcpyHostToDevice,
+                                  stream.value()));
+  }
+  if (num_nested > 0) {
+    CUDF_CUDA_TRY(cudaMemcpyAsync(d_nested_indices.data(),
+                                  nested_field_indices.data(),
+                                  num_nested * sizeof(int),
+                                  cudaMemcpyHostToDevice,
+                                  stream.value()));
+  }
+
+  // Count repeated fields at depth 0 (with O(1) field_number lookup tables)
+  rmm::device_uvector<int> d_fn_to_rep(0, stream, scratch_mr);
+  rmm::device_uvector<int> d_fn_to_nested(0, stream, scratch_mr);
+
+  if (num_repeated > 0 || num_nested > 0) {
+    auto h_fn_to_rep =
+      build_index_lookup_table(schema.data(), repeated_field_indices.data(), num_repeated);
+    auto h_fn_to_nested =
+      build_index_lookup_table(schema.data(), nested_field_indices.data(), num_nested);
+
+    if (!h_fn_to_rep.empty()) {
+      d_fn_to_rep = rmm::device_uvector<int>(h_fn_to_rep.size(), stream, scratch_mr);
+      CUDF_CUDA_TRY(cudaMemcpyAsync(d_fn_to_rep.data(),
+                                    h_fn_to_rep.data(),
+                                    h_fn_to_rep.size() * sizeof(int),
+                                    cudaMemcpyHostToDevice,
+                                    stream.value()));
+    }
+    if (!h_fn_to_nested.empty()) {
+      d_fn_to_nested = rmm::device_uvector<int>(h_fn_to_nested.size(), stream, scratch_mr);
+      CUDF_CUDA_TRY(cudaMemcpyAsync(d_fn_to_nested.data(),
+                                    h_fn_to_nested.data(),
+                                    h_fn_to_nested.size() * sizeof(int),
+                                    cudaMemcpyHostToDevice,
+                                    stream.value()));
+    }
+
+    launch_count_repeated_fields(*d_in,
+                                 d_schema.data(),
+                                 num_fields,
+                                 0,
+                                 d_repeated_info.data(),
+                                 num_repeated,
+                                 d_repeated_indices.data(),
+                                 d_nested_locations.data(),
+                                 num_nested,
+                                 d_nested_indices.data(),
+                                 d_error.data(),
+                                 d_fn_to_rep.data(),
+                                 static_cast<int>(d_fn_to_rep.size()),
+                                 d_fn_to_nested.data(),
+                                 static_cast<int>(d_fn_to_nested.size()),
+                                 num_rows,
+                                 stream);
+  }
 
   // Store decoded columns by schema index for ordered assembly at the end.
   std::vector<std::unique_ptr<cudf::column>> column_map(num_fields);
@@ -615,7 +712,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
           int si                  = scalar_field_indices[idxs[j]];
           auto dt                 = cudf::data_type{schema[si].output_type};
           auto& bp                = *bufs[j];
-          auto [mask, null_count] = make_null_mask_from_valid(bp.valid, stream, mr);
+          auto [mask, null_count] = make_null_mask_from_valid(bp.valid, num_rows, stream, mr);
           column_map[si]          = std::make_unique<cudf::column>(
             dt, num_rows, bp.out_bytes.release(), std::move(mask), null_count);
         }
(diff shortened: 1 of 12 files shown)


src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     }
   }
 
+  // Process repeated fields (three-phase: offsets → combined scan → build columns)
+  if (num_repeated > 0) {
+    // Phase A: Compute per-row offsets for each repeated field.
+    std::vector<std::unique_ptr<repeated_field_work>> rep_work;
```
igorpeshansky: Is there any reason to have a `unique_ptr` indirection here, rather than just inline objects (i.e., `std::vector<repeated_field_work>`)?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -458,6 +518,77 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   auto const threads = THREADS_PER_BLOCK;
   auto const blocks  = static_cast<int>((num_rows + threads - 1u) / threads);
 
+  // Allocate for counting repeated fields
+  rmm::device_uvector<repeated_field_info> d_repeated_info(
+    num_repeated > 0 ? static_cast<size_t>(num_rows) * num_repeated : 1, stream, scratch_mr);
```
igorpeshansky: Is `std::max` available here? If so, this could be `std::max<size_t>(static_cast<size_t>(num_rows) * num_repeated, 1)` instead. Also below.

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -434,7 +496,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       case ERR_REQUIRED: return "Protobuf decode error: missing required field";
       case ERR_SCHEMA_TOO_LARGE:
         return "Protobuf decode error: schema exceeds maximum supported repeated fields per kernel "
-               "(128)";
+               "(32)";
```
igorpeshansky: Is `MAX_REPEATED_FIELDS_PER_KERNEL` available here? Nicer to embed it instead of using a magic constant… Then `error_message` could return a computed `std::string` (rather than a C string), since `cudf::logic_error` can be constructed from one…

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+                                    stream.value()));
+
+      // Build field_number -> scan_desc_index lookup for the combined kernel
+      int max_scan_fn = 0;
+      for (auto const& sd : h_scan_descs) {
+        max_scan_fn = std::max(max_scan_fn, sd.field_number);
+      }
+      rmm::device_uvector<int> d_fn_to_scan(0, stream, scratch_mr);
+      int fn_to_scan_size = 0;
+      if (max_scan_fn <= FIELD_LOOKUP_TABLE_MAX) {
```
igorpeshansky: Should there be an error (or some sort of indication) when this is false? Silently ignoring it feels wrong…

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+      rmm::device_uvector<repeated_field_scan_desc> d_scan_descs(
+        h_scan_descs.size(), stream, scratch_mr);
+      CUDF_CUDA_TRY(cudaMemcpyAsync(d_scan_descs.data(),
+                                    h_scan_descs.data(),
+                                    h_scan_descs.size() * sizeof(h_scan_descs[0]),
+                                    cudaMemcpyHostToDevice,
+                                    stream.value()));
+
+      // Build field_number -> scan_desc_index lookup for the combined kernel
+      int max_scan_fn = 0;
```
igorpeshansky: [Optional] This computation could be piggybacked on the above `rep_work` loop.

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+
+    // Phase C: Build columns per field.
+    for (int ri = 0; ri < num_repeated; ri++) {
+      auto& w              = *rep_work[ri];
+      int schema_idx       = w.schema_idx;
+      auto element_type    = cudf::data_type{schema[schema_idx].output_type};
+      int32_t total_count  = w.total_count;
+      auto& d_field_counts = w.counts;
+
+      if (total_count > 0) {
```
igorpeshansky: This applies to other loops above as well, but especially for longer loops like this, making this a guard clause could save some indentation levels, i.e.:

```c++
      if (total_count <= 0) {
        // All rows have count=0 — produce an all-zeros offsets column wrapped in a LIST whose
        // null mask comes from the input column. Reuses the same helpers the inner builders use.
        …
        continue;
      }
```

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+        auto& d_occurrences = *w.occurrences;
+
+        // Build the appropriate column type based on element type
+        auto child_type_id = static_cast<cudf::type_id>(h_device_schema[schema_idx].output_type_id);
+
+        // The output_type in schema is the LIST type, but we need element type
+        // For repeated int32, output_type should indicate the element is INT32
```
igorpeshansky: Nit: these are two sentences in what looks like one contiguous comment. I'd put a `.` after each one…

Actually, is the first sentence even correct? Looks like the output type is already the element type. Am I missing something here?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+                                                                  d_occurrences,
+                                                                  total_count,
+                                                                  num_rows,
+                                                                  true,
+                                                                  d_error,
+                                                                  stream,
+                                                                  mr);
+            break;
+          default:
+            // Unsupported element type - create null column
```
igorpeshansky: [Optional] Silently?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -21,6 +21,43 @@
...
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr)
+{
+  rmm::device_uvector<int32_t> offsets(num_rows + 1, stream, mr);
+  thrust::exclusive_scan(
+    rmm::exec_policy_nosync(stream), counts.begin(), counts.end(), offsets.begin(), 0);
+  thrust::fill_n(rmm::exec_policy_nosync(stream),
+                 offsets.data() + num_rows,
+                 1,
+                 static_cast<int32_t>(total_count));
```
igorpeshansky: [Optional] My initial knee-jerk reaction was that you could have avoided the `total_count` computation and used effectively `offsets[0] = 0` plus `thrust::inclusive_scan(…, counts.begin(), counts.end(), std::next(offsets.begin()))`, but then I realized that `total_count` was computed with overflow detection. This point is subtle enough to potentially merit a comment.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -198,6 +201,27 @@ std::unique_ptr<cudf::column> build_enum_string_column(
   int32_t const* top_row_indices = nullptr,
   bool propagate_invalid_rows    = true);
 
+// LIST<...> offsets buffer of size `num_rows + 1` from per-row counts. The last entry is
+// `total_count` (caller-provided to avoid an extra reduction). Both the prefix scan and the
```
igorpeshansky: See my other message — it's not just avoiding an extra reduction, but also guards against 32-bit overflow… Worth mentioning?

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -198,6 +201,27 @@ std::unique_ptr<cudf::column> build_enum_string_column(
...
+rmm::device_uvector<int32_t> make_list_offsets_from_counts(
+  rmm::device_uvector<int32_t> const& counts,
+  int total_count,
+  int num_rows,
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr);
+
+// Wrap an offsets / child column pair into a LIST column, propagating the input column's null
+// mask when it has nulls. Used by every repeated-* builder so the input-null shape stays
+// consistent across paths.
```
igorpeshansky: [Optional] Is the fact that `mr` is ignored when there are no nulls significant and worth mentioning?

src/main/cpp/src/protobuf/protobuf_kernels.cuh
```diff
@@ -403,18 +385,56 @@ CUDF_KERNEL void extract_lengths_kernel(LocationProvider loc_provider,
...
 // ============================================================================
 
+// Build a row-aligned null mask from `valid[row]` boolean flags. `num_rows` is the logical row
+// count and must be <= `valid.size()` — this matters because some callers pad `valid` to
+// `max(1, num_rows)` to avoid a 0-sized device_uvector, and feeding that padded size into
+// `valid_if` would produce a 1-bit mask for a 0-row column.
 template <typename T>
 inline std::pair<rmm::device_buffer, cudf::size_type> make_null_mask_from_valid(
   rmm::device_uvector<T> const& valid,
+  cudf::size_type num_rows,
```
igorpeshansky: [Optional] This seems like a general bug fix unrelated to protobuf processing. Am I missing something, or could this have been a separate PR?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+      while (p < packed_end) {
+        uint64_t dummy;
+        int vbytes;
+        if (!read_varint(p, packed_end, dummy, vbytes)) {
+          set_error_once(error_flag, ERR_VARINT);
+          return false;
+        }
+        p += vbytes;
+        count++;
+      }
```
igorpeshansky: [Optional] A little cleaner:

```suggestion
      int vbytes = INT_MAX;
      for (uint8_t const* p = packed_start; p < packed_end; p += vbytes) {
        uint64_t dummy;
        if (!read_varint(p, packed_end, dummy, vbytes)) {
          set_error_once(error_flag, ERR_VARINT);
          return false;
        }
        count++;
      }
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+    }
+    uint8_t const* packed_start = cur + len_bytes;
+    if (packed_len > static_cast<uint64_t>(msg_end - packed_start)) {
+      set_error_once(error_flag, ERR_OVERFLOW);
+      return false;
+    }
+    uint8_t const* packed_end = packed_start + packed_len;
+
+    int count = 0;
+    if (expected_wt == wire_type_value(proto_wire_type::VARINT)) {
```
igorpeshansky: [Optional] Can this be a `switch`?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+        return false;
+      }
+      count = static_cast<int>(packed_len / 4);
+    } else if (expected_wt == wire_type_value(proto_wire_type::I64BIT)) {
+      if ((packed_len % 8) != 0) {
+        set_error_once(error_flag, ERR_FIXED_LEN);
+        return false;
+      }
+      count = static_cast<int>(packed_len / 8);
+    }
```
igorpeshansky: [Optional] These two are the same, so they can become something like:

```suggestion
    } else if (expected_wt == wire_type_value(proto_wire_type::I32BIT) || expected_wt == wire_type_value(proto_wire_type::I64BIT)) {
      int width = expected_wt == wire_type_value(proto_wire_type::I32BIT) ? 4 : 8;
      if ((packed_len % width) != 0) {
        set_error_once(error_flag, ERR_FIXED_LEN);
        return false;
      }
      count = static_cast<int>(packed_len / width);
    }
```

It'd be even cleaner with a `switch`…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+    uint8_t const* packed_end = packed_start + packed_len;
+
+    int count = 0;
+    if (expected_wt == wire_type_value(proto_wire_type::VARINT)) {
+      uint8_t const* p = packed_start;
+      while (p < packed_end) {
+        uint64_t dummy;
+        int vbytes;
+        if (!read_varint(p, packed_end, dummy, vbytes)) {
```
igorpeshansky: I'm trying to understand whether there's some sort of implicit data validation going on here, because if not, this seems like it could be more efficiently implemented via `skip_field` or some helper extracted from `skip_field`…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+        set_error_once(error_flag, ERR_FIXED_LEN);
+        return false;
+      }
+      count = static_cast<int>(packed_len / 8);
+    }
+
+    info.count += count;
+  } else {
+    int32_t data_offset, data_length;
+    if (!get_field_data_location(cur, msg_end, wt, data_offset, data_length)) {
```
igorpeshansky: `skip_field` seems equivalent and cheaper…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+                                      uint8_t const* msg_base,
+                                      int wt,
+                                      int expected_wt,
+                                      int32_t row,
+                                      repeated_occurrence* occurrences,
+                                      int& write_idx,
+                                      int write_end,
+                                      int* error_flag)
+{
+  bool is_packed = (wt == wire_type_value(proto_wire_type::LEN) &&
```
igorpeshansky, line 192: I will note that, except for recording the offsets, this kernel is nearly identical to `count_repeated_element`, so all of the same comments apply.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+      return false;
+    }
+    uint8_t const* packed_start = cur + len_bytes;
+    if (packed_len > static_cast<uint64_t>(msg_end - packed_start)) {
+      set_error_once(error_flag, ERR_OVERFLOW);
+      return false;
+    }
+    uint8_t const* packed_end = packed_start + packed_len;
+
+    if (expected_wt == wire_type_value(proto_wire_type::VARINT)) {
```
igorpeshansky: [Optional] I'm wondering if this whole `if` can be collapsed into a loop that just repeatedly calls `skip_field` or `get_field_data_location`… I'll admit that I don't have a good intuition on how that would affect on-GPU performance, though, since this implementation pulls out the conditional outside of the loops.

[Really optional] One nice C++-y way to handle this would be to create templatized versions of `skip_field`, `get_field_data_location`, `get_wire_type_size`, etc, with the wire_type values as NTTPs, and then create one generic loop over those that is dispatched via a switch or a nested if… This would take up the same amount of room as the inline switches used in those functions, but would relegate most dispatch to compile time (except for the outermost layer), making it (presumably) more efficient. This is a much, much larger change, however, and way outside the scope of this PR. Just noting it here for completeness…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+
+  while (cur < msg_end) {
+    proto_tag tag;
+    if (!decode_tag(cur, msg_end, tag, error_flag)) return;
+    int fn = tag.field_number;
+    int wt = tag.wire_type;
+
+    if (fn_to_rep_idx != nullptr && fn > 0 && fn < fn_to_rep_size) {
+      int i = fn_to_rep_idx[fn];
+      if (i >= 0) {
```
igorpeshansky: Do we need an error if we encounter an unknown field number during lookup?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+  uint8_t const* cur     = bytes + start;
+  uint8_t const* msg_end = bytes + end;
+
+  while (cur < msg_end) {
+    proto_tag tag;
+    if (!decode_tag(cur, msg_end, tag, error_flag)) return;
+    int fn = tag.field_number;
+    int wt = tag.wire_type;
+
+    if (fn_to_rep_idx != nullptr && fn > 0 && fn < fn_to_rep_size) {
```
igorpeshansky: The two branches are identical. Instead of repeating the code, would it be just as efficient to compute `i` first, and then run `count_repeated_element`? You'd then be able to reuse the lookup for handling nested fields, too. We could use inline lambdas to make the lookup cleaner:

```c++
    auto lookup_function = [&](int const* fn_to_idx, int fn_tbl_size, int const* field_indices) {
      return [&](int fn) {
        auto check_schema = [&](int fn, int i) {
          int schema_idx = field_indices[i];
          if (schema[schema_idx].field_number == fn && schema[schema_idx].depth == depth_level) {
            return i;
          }
          return -1;
        };
        if (fn_to_idx != nullptr && fn > 0 && fn < fn_tbl_size) {
          int i = fn_to_rep_idx[fn];
          // Defensive: also verify field_number matches, mirroring the linear-scan fallback below.
          // A correct lookup table guarantees this, but failing closed on a buggy table is cheap
          // and avoids dispatching to the wrong field index silently.
          return check_schema(fn, i);
        }
        for (int i = 0; i < num_repeated_fields; ++i) {
          if (int r = check_schema(fn, i); r >= 0) {
            return r;  // Each tag dispatches to at most one repeated index (mirrors lookup-table path)
          }
        }
        return -1;
      };
    };
    auto rep_lookup = lookup_function(fn_to_rep_idx, fn_to_rep_size, repeated_field_indices);
    if (int i = rep_lookup(fn); i >= 0) {
      int schema_idx = repeated_field_indices[i];
      if (!count_repeated_element(
            cur,
            msg_end,
            wt,
            schema[schema_idx].wire_type,
            repeated_info[flat_index(static_cast<size_t>(row),
                                     static_cast<size_t>(num_repeated_fields),
                                     static_cast<size_t>(i))],
            error_flag)) {
        return;
      }
    }
    auto nested_lookup = lookup_function(fn_to_nested_idx, fn_to_nested_size, nested_field_indices);
    if (int i = nested_lookup(fn); i >= 0) {
      if (!handle_nested(i)) return;  // This is now the only use and could be inlined…
    }
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+        set_error_once(error_flag, ERR_OVERFLOW);
+        return false;
+      }
+      nested_locations[flat_index(
+        static_cast<size_t>(row), static_cast<size_t>(num_nested_fields), static_cast<size_t>(i))] =
+        {msg_offset, static_cast<int32_t>(len)};
+      return true;
+    };
+
+    if (fn_to_nested_idx != nullptr && fn > 0 && fn < fn_to_nested_size) {
```
igorpeshansky: [Optional] You could reuse the lookup function here…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+ * repeated fields simultaneously, scanning each message only once.
+ */
+CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view const d_in,
+                                                      repeated_field_scan_desc const* scan_descs,
+                                                      int num_scan_fields,
+                                                      int* error_flag,
+                                                      int const* fn_to_desc_idx,
+                                                      int fn_to_desc_size)
+{
+  auto row = static_cast<cudf::size_type>(blockIdx.x * blockDim.x + threadIdx.x);
```
igorpeshansky, line 422: The body of this kernel is almost identical to `count_repeated_fields_kernel` — I wonder how much can (or should) be reused…

igorpeshansky (commented):
None of the comments are technically blocking, so I won't "request changes", though there _are_ a couple of correctness-related questions.

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -21,6 +21,43 @@
 
 namespace spark_rapids_jni::protobuf::detail {
 
+rmm::device_uvector<int32_t> make_list_offsets_from_counts(
+  rmm::device_uvector<int32_t> const& counts,
```
igorpeshansky: Defensively, we should ensure that `counts.size()` is exactly `num_rows` (e.g., `CUDF_EXPECTS(counts.size() == static_cast<size_t>(num_rows), …)`)…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2731,4 +3188,150 @@ void testNullSchemaThrows() {
           Protobuf.decodeToStruct(input.getColumn(0), null, true));
     }
   }
+
+  @Test
+  void testRepeatedString() {
```
igorpeshansky, line 3279: Do we need a `testRepeatedBytes` as well?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2289,10 +2544,166 @@ void testEnumValidWithOtherFields() {
...
+            new int[]{Protobuf.ENC_FIXED},
+            new boolean[]{true},
+            new boolean[]{false},
+            new boolean[]{false},
+            new long[]{0},
+            new double[]{0.0},
+            new boolean[]{false},
+            new byte[][]{null},
+            new int[][]{null},
+            true)) {
```
igorpeshansky, line 2633: Do we want to test with `false` too? Ditto for the 64-bit version…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -163,6 +163,447 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+  if (!check_message_bounds(start, end, child.size(), error_flag)) return;
+
+  uint8_t const* cur     = bytes + start;
+  uint8_t const* msg_end = bytes + end;
+
+  // The host validates `num_repeated <= MAX_REPEATED_FIELDS_PER_KERNEL` in
+  // `validate_decode_context`, so this device-side check is a defense-in-depth assert. Keeping
+  // the cap small keeps `write_idx` in registers / L1 instead of spilling into local memory.
+  if (num_scan_fields > MAX_REPEATED_FIELDS_PER_KERNEL) {
+    set_error_once(error_flag, ERR_SCHEMA_TOO_LARGE);
```
igorpeshansky, line 444: Should this be `cuda::std::assert`, then?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+      int schema_idx       = w.schema_idx;
+      auto element_type    = cudf::data_type{schema[schema_idx].output_type};
+      int32_t total_count  = w.total_count;
+      auto& d_field_counts = w.counts;
+
+      if (total_count > 0) {
+        auto& d_occurrences = *w.occurrences;
+
+        // Build the appropriate column type based on element type
+        auto child_type_id = static_cast<cudf::type_id>(h_device_schema[schema_idx].output_type_id);
```
igorpeshansky: Is this identical to the definition in the `else` branch? Can it be factored out of the conditional? Also, is this `element_type.id() `?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -1747,6 +1747,147 @@ void testDefaultStringMultipleRows() {
...
+          new double[]{0.0, 0.0, 0.0, 0.0},
+          new boolean[]{false, false, false, false},
+          new byte[][]{null, null, null, null},
+          new int[][]{null, null, null, null},
+          false)) {
+        assertNotNull(result);
+        assertEquals(DType.STRUCT, result.getType());
+        assertEquals(3, result.getRowCount());
+
+        // Verify double_values child column has correct total count: 2 + 2 + 1 = 5
```
igorpeshansky: Do we need to verify other types? If we don't, e.g. because we can infer their correctness from just the double tests, can we comment about this explicitly?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -252,4 +289,197 @@ std::unique_ptr<cudf::column> build_enum_string_column(
...
+  int total_count,
+  int num_rows,
+  bool is_bytes,
+  rmm::device_uvector<int>& d_error,
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr)
+{
+  auto const input_null_count = binary_input.null_count();
+
+  if (total_count == 0) {
```
igorpeshansky: Is this function ever called with `total_count == 0`? If so, where? If not, should this be an assertion instead?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+
+        std::unique_ptr<cudf::column> child_col;
+        auto const child_type_id =
+          static_cast<cudf::type_id>(h_device_schema[schema_idx].output_type_id);
+        if (child_type_id == cudf::type_id::STRUCT) {
+          child_col =
+            make_empty_struct_column_with_schema(schema, schema_idx, num_fields, stream, mr);
+        } else {
+          child_col =
+            make_empty_column_safe(cudf::data_type{schema[schema_idx].output_type}, stream, mr);
```
igorpeshansky: Isn't this `element_type`?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+        }
+
+        column_map[schema_idx] = make_list_column_with_input_nulls(
+          num_rows, std::move(offsets_col), std::move(child_col), binary_input, stream, mr);
+      }
+    }
```
igorpeshansky: [Optional] With really long code blocks like these, it helps to have a trailing comment on a closing brace indicating what block it closes (e.g., `// for` or `// total_count <= 0` or `if total_count`)…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -494,14 +964,22 @@ void propagate_invalid_enum_flags_to_rows(rmm::device_uvector<bool> const& item_
...
 
-  thrust::for_each(rmm::exec_policy_nosync(stream),
-                   thrust::make_counting_iterator(0),
-                   thrust::make_counting_iterator(num_items),
-                   [item_invalid = item_invalid.data(),
-                    top_row_indices,
-                    row_invalid = row_invalid.data()] __device__(int idx) {
-                     if (item_invalid[idx]) row_invalid[top_row_indices[idx]] = true;
-                   });
+  // Multiple items may share the same `top_row_indices[idx]` (e.g. several occurrences of a
```
igorpeshansky, line 867: [Optional] This also seems like a general bug fix unrelated to protobuf processing. Am I missing something, or could this have been a separate PR?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+    for (int ri = 0; ri < num_repeated; ri++) {
+      int schema_idx = repeated_field_indices[ri];
+      auto& w        = *rep_work.emplace_back(
+        std::make_unique<repeated_field_work>(schema_idx, num_rows, stream, scratch_mr));
+
+      thrust::transform(rmm::exec_policy_nosync(stream),
+                        thrust::make_counting_iterator(0),
+                        thrust::make_counting_iterator(num_rows),
+                        w.counts.data(),
+                        extract_strided_count{d_repeated_info.data(), ri, num_repeated});
```
igorpeshansky: _Caveat: this one was fully suggested by Claude, and I don't have the full context to validate it, though it sounds plausible._

[Optional] This `transform` materializes a `num_rows`-sized `int32_t` scratch buffer per repeated field, plus an extra kernel launch, just to surface a strided count out of `d_repeated_info`. The reduce + scan that consume `w.counts` could read the strided counts on demand via a fancy iterator instead — eliminating both the buffer and the kernel.

#### A

The simplest form is a `transform_iterator` over a counting iterator, reusing the existing functor:

```c++
auto counts_begin = thrust::make_transform_iterator(
    thrust::make_counting_iterator<int>(0),
    extract_strided_count{d_repeated_info.data(), ri, num_repeated});
auto counts_end = counts_begin + num_rows;

int64_t total_64 = thrust::reduce(
    rmm::exec_policy_nosync(stream), counts_begin, counts_end, int64_t{0});
// ... overflow check, w.total_count assignment ...

if (num_rows > 0) {
  rmm::device_uvector<int32_t> offsets(num_rows + 1, stream, scratch_mr);
  thrust::exclusive_scan(rmm::exec_policy_nosync(stream),
                         counts_begin, counts_end,
                         offsets.begin(), 0);
  thrust::fill_n(rmm::exec_policy_nosync(stream),
                 offsets.data() + num_rows, 1,
                 static_cast<int32_t>(total_64));
  w.offsets = std::move(offsets);
}
```

#### B

The equivalent `permutation_iterator` form is more "canonical thrust" — it cleanly separates the field projection from the strided gather, which would matter if `repeated_field_info` ever grew more fields:

```c++
auto count_proj = thrust::make_transform_iterator(
    d_repeated_info.data(),
    cuda::proclaim_return_type<int32_t>(
      [] __device__ (repeated_field_info const& rfi) { return rfi.count; }));

auto strided_idx = thrust::make_transform_iterator(
    thrust::make_counting_iterator<int>(0),
    cuda::proclaim_return_type<int>(
      [num_repeated, ri] __device__ (int row) { return row * num_repeated + ri; }));

auto counts_begin = thrust::make_permutation_iterator(count_proj, strided_idx);
auto counts_end   = counts_begin + num_rows;
```

Either way, `w.counts` (and the whole `extract_strided_count` `transform`) goes away. To make it work, however, `make_list_offsets_from_counts` would need to be templated on the input iterator type rather than taking `rmm::device_uvector<int32_t> const&`; the body (`exclusive_scan` + `fill_n`) is unchanged, and the other callers (`build_repeated_string_column`, `build_repeated_enum_string_column`) continue to work because `device_uvector::begin()/end()` deduces correctly.

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -830,6 +961,318 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+                   "Top-level repeated field total element count exceeds 2^31-1");
+      w.total_count = static_cast<int32_t>(total_64);
+
+      if (num_rows > 0) {
+        // Reuse the shared LIST offsets helper. Exclusive scan starting at 0 is equivalent to
+        // the previous inclusive-scan-into-offsets+1 pattern when offsets[0] is forced to 0.
+        // Empty input (num_rows == 0) is handled by skipping this branch — the offsets buffer
+        // goes unused on the no-row path.
+        w.offsets =
+          make_list_offsets_from_counts(w.counts, w.total_count, num_rows, stream, scratch_mr);
```
igorpeshansky: _Caveat: another one fully suggested by Claude, again with me not having the full context to validate, and again sounds plausible._

[Optional] Per-repeated-field, the LIST offsets buffer is computed twice: once here into `w.offsets` (with `scratch_mr`) so the Phase B scan kernel can seed each row's `write_idx[f] = scan_descs[f].row_offsets[row]`, and again inside each Phase C builder via `make_list_offsets_from_counts(d_field_counts, total_count, num_rows, stream, mr)`. The two buffers are bit-for-bit identical — same counts, same `total_count`, same `num_rows`, same `exclusive_scan` + `fill_n`. The only thing that differs is the resource (`scratch_mr` vs. `mr`).

The cheapest fix is to compute it once with `mr` and hand the same buffer to both phases — Phase B reads it (`scan_descs[f].row_offsets` is only ever indexed, never written), and Phase C moves it into the output column. Roughly:

```c++
// Phase A: build offsets directly into the caller's resource, since they will
// become the output column's offsets buffer.
w.offsets = make_list_offsets_from_counts(/* iterator from the other comment */,
                                          w.total_count, num_rows, stream, mr);

// Phase B is unchanged: it only reads w.offsets via scan_descs[f].row_offsets.

// Phase C: hand w.offsets to the builder pre-wrapped as an INT32 column,
// instead of d_field_counts.
auto offsets_col = std::make_unique<cudf::column>(
    cudf::data_type{cudf::type_id::INT32}, num_rows + 1,
    w.offsets.release(), rmm::device_buffer{}, 0);
column_map[schema_idx] = build_repeated_string_column(
    binary_input, message_data, list_offsets, base_offset,
    h_device_schema[schema_idx],
    std::move(offsets_col),  // <-- replaces d_field_counts + total_count
    d_occurrences, total_count, num_rows, /* is_bytes */ false,
    d_error, stream, mr);
```

Builder side, the signature change is just `rmm::device_uvector<int32_t> const& d_field_counts` → `std::unique_ptr<cudf::column> offsets_col` (or a `column_view` plus moved buffer if you want to keep the option of reusing the offsets a third time). The builder body's `make_list_offsets_from_counts(...)` call goes away, replaced by `auto list_offs_col = std::move(offsets_col);`. Same change applies to `build_repeated_enum_string_column` and `build_repeated_scalar_column`.

Trade-off worth flagging: this moves the offsets buffer from `scratch_mr` to `mr` for the duration of Phase B. For typical schemas (single-digit top-level repeated fields, num_rows on the order of a batch) this is `O(num_repeated * num_rows * 4)` bytes of `mr` held a little longer than today, in exchange for skipping `num_repeated` Thrust kernel launches and freeing one `mr`-backed buffer per field at the end. Worth it unless `mr` is a tightly-budgeted output-only pool.

igorpeshansky (commented):
A few more, now with help from Claude.

It also flagged that `enum_binary_search` is not mentioned in the PR description — could be useful…

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Yes. Updated to `MAX_REPEATED_FIELDS_PER_KERNEL` and `std::string`.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Updated all to `std::max`

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Good catch, updated to `std::vector<repeated_field_work>`.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Claude's suggestion looks correct to me. My Claude implemented it, along with the one below.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: ditto

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: The fallback is intentional — build_lookup_table returns {} when the max field_number exceeds `FIELD_LOOKUP_TABLE_MAX` (4096) so we don't over-allocate host memory on sparse high field numbers. The kernel detects this via `fn_to_desc_idx == nullptr` (or `fn_to_desc_size == 0`) and falls back to a linear scan over scan_descs. 

Added a comment at the host site to make the contract obvious; sparse-schema fallback isn't an error and shouldn't surface one.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Did a small refactor to build the lookup table with a helper function `build_lookup_table`, and `rep_work` is no longer necessary.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: It's not correct. I did some comments clean up on the whole PR.

src/main/cpp/src/protobuf/protobuf.cu

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: 👍 Yes, yes, and yes. Updated to `element_type.id()` now.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Split the default into two: `STRUCT` and `default`. For structs, it returns nulls as placeholders. For the default, I added a CUDF_FAIL.

thirtiseven: Wait, returning nulls silently is not a good practice itself. Updated to make it fail-fast.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Yes, updated.

src/main/cpp/src/protobuf/protobuf.cu

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: `make_list_offsets_from_counts` is removed after the claude refactor.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: ditto

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Yes it's dead code. Updated to a `CUDF_EXPECTS`.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: The function disappeared after refactoring. But I added the 32-bit overflow comment, in protobuf.cu#L980

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf_kernels.cuh
thirtiseven: Filed PR https://github.com/NVIDIA/spark-rapids-jni/pull/4521

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Added comments.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: done.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Yes there are validations. Comments added.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Ditto.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Combined to one.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: > [Optional] I'm wondering if this whole `if` can be collapsed into a loop that just repeatedly calls `skip_field` or `get_field_data_location`… I'll admit that I don't have a good intuition on how that would affect on-GPU performance, though, since this implementation pulls out the conditional outside of the loops.

Ditto, there are validations.

> [Really optional] One nice C++-y way to handle this would be to create templatized versions of `skip_field`, `get_field_data_location`, `get_wire_type_size`, etc, with the wire_type values as NTTPs, and then create one generic loop over those that is dispatched via a switch or a nested if… This would take up the same amount of room as the inline switches used in those functions, but would relegate most dispatch to compile time (except for the outermost layer), making it (presumably) more efficient. This is a much, much larger change, however, and way outside the scope of this PR. Just noting it here for completeness…

Noted and will try to do it in another PR.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Done. Thanks!

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: We don't need one. protobuf parsers must silently skip unknown fields to preserve forward compatibility
  (wire format spec (https://protobuf.dev/programming-guides/encoding))

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Did a refactor on them and removed some duplicated code.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: My AI convinced me:
>Kept as set_error_once + return. The failure mode (overrunning the fixed-size write_idx[MAX_REPEATED_FIELDS_PER_KERNEL] stack array on the next line) is silent UB, not a clean abort — assert would be a no-op under NDEBUG and leave the OOB write live in release builds. Host-side validate_decode_context makes the check unreachable on a correct config, so this is purely defense-in-depth, but cheap enough to keep active everywhere. Added a comment explaining the choice.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: The race only became reachable with this PR. In part 1 (scalar enum), `propagate_invalid_enum_flags_to_rows` had a 1:1 item→row mapping, so distinct threads never wrote to the same `row_invalid[]` byte. This PR introduces the N:1 case (multiple packed enum items per row), which is what makes the concurrent same-byte write reachable. A standalone PR would have no triggering caller and no test that fails without it, so I'd rather keep the fix here. Happy to revisit if you'd prefer otherwise.

thirtiseven (commented):
@igorpeshansky Thank you for the really insightful review! Please take another look.

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -294,7 +277,7 @@ std::unique_ptr<cudf::column> build_repeated_enum_string_column(
   uint8_t const* message_data,
   cudf::size_type const* list_offsets,
   cudf::size_type base_offset,
-  rmm::device_uvector<int32_t> const& d_field_counts,
+  rmm::device_uvector<int32_t> list_offs,
```
igorpeshansky: This new naming clashes with `list_offsets`. We could rename `list_offsets` to either `message_offsets` or `input_offsets`, or rename `list_offs` to, e.g., `d_field_offsets`…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -252,4 +272,177 @@ std::unique_ptr<cudf::column> build_enum_string_column(
...
+  return make_list_column_with_input_nulls(
+    num_rows, std::move(list_offs_col), std::move(child_col), binary_input, stream, mr);
+}
+
+std::unique_ptr<cudf::column> build_repeated_string_column(
+  cudf::column_view const& binary_input,
+  uint8_t const* message_data,
+  cudf::size_type const* list_offsets,
+  cudf::size_type base_offset,
+  device_nested_field_descriptor const& field_desc,
```
igorpeshansky: `field_desc` seems unused in all `build_repeated_*` functions except `build_repeated_scalar_column`. Yet it's propagated into all functions except `build_repeated_enum_string_column`. Any reason you omitted it only from that one?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -168,93 +168,21 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
                                       uint8_t const* msg_base,
                                       int wt,
                                       int expected_wt,
-                                      int32_t row,
-                                      repeated_occurrence* occurrences,
-                                      int& write_idx,
-                                      int write_end,
-                                      int* error_flag)
+                                      int* error_flag,
+                                      Action&& act)
```
igorpeshansky: [Optional] The traditional name for such parameters is `f`…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -278,64 +206,53 @@ __device__ bool scan_repeated_element(uint8_t const* cur,
...
-      if ((packed_len % 8) != 0) {
-        set_error_once(error_flag, ERR_FIXED_LEN);
-        return false;
-      }
-      for (uint64_t i = 0; i < packed_len; i += 8) {
-        if (write_idx >= write_end) {
-          set_error_once(error_flag, ERR_REPEATED_COUNT_MISMATCH);
-          return false;
+        for (uint64_t i = 0; i < packed_len; i += width) {
+          int32_t elem_offset = static_cast<int32_t>(packed_start - msg_base + i);
```
igorpeshansky: This can also be a pointer loop, symmetric with the VARINT case:

```suggestion
        for (uint8_t const* p = packed_start; p < packed_end; p += width) {
          int32_t elem_offset = static_cast<int32_t>(p - msg_base);
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -278,64 +206,53 @@ __device__ bool scan_repeated_element(uint8_t const* cur,
...
+          if (!act(elem_offset, width)) return false;
         }
-        occurrences[write_idx] = {row, static_cast<int32_t>(packed_start - msg_base + i), 8};
-        write_idx++;
+        break;
       }
+      default: break;  // LEN is filtered out above by the !is_packed path.
```
igorpeshansky: [Optional] This `default` clause is vacuous — have you added it because of a compiler warning for an incomplete `switch`? Also, if this case should not occur in practice, do we want to add a defensive failure here?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -394,106 +311,87 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
-          break;  // Each tag dispatches to at most one repeated index (mirrors lookup-table path)
-        }
+    if (int i = lookup_field_idx(
+          fn, fn_to_rep_idx, fn_to_rep_size, repeated_field_indices, num_repeated_fields);
+        i >= 0) {
+      int schema_idx    = repeated_field_indices[i];
+      auto& info        = repeated_info[flat_index(static_cast<size_t>(row),
+                                            static_cast<size_t>(num_repeated_fields),
+                                            static_cast<size_t>(i))];
+      auto count_action = [&info](int32_t /*off*/, int32_t /*len*/) {
```
igorpeshansky: You could just name the parameters:

```suggestion
      auto count_action = [&info](int32_t unused_off, int32_t unused_len) {
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -168,93 +168,21 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
- * Returns false on error (error_flag set), true on success.
+ * Visit each occurrence of a repeated field (packed or unpacked) and invoke `act` for it.
+ *
+ * `act(int32_t elem_offset, int32_t elem_len) -> bool` runs once per occurrence with the
+ * element's offset relative to `msg_base` and its length. Returning false aborts the walk.
+ * The walker handles wire-type validation, packed-vs-unpacked dispatch, varint/fixed-width
+ * length decoding, and packed-buffer bounds checking.
  */
-__device__ bool scan_repeated_element(uint8_t const* cur,
+template <typename Action>
```
igorpeshansky: [Optional] Since this code is C++20 compatible, you could make the error messages clearer with:

```suggestion
template <typename Action>
  requires std::is_invocable_r_v<bool, Action, int32_t /*elem_offset*/, int32_t /*elem_len*/>
```

or

```suggestion
template <typename Action>
  requires std::invocable<Action, int32_t, int32_t> && 
           std::convertible_to<std::invoke_result_t<Action, int32_t, int32_t>, bool>
```

or even:

```c++
__device__ bool walk_repeated_element(
    …,
    std::invocable<int32_t /*elem_offset*/, int32_t /*elem_len*/> auto&& act) {
  …
}
```

The first one could also be a `static_assert` in pre-C++20 code.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -394,106 +311,87 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
+  // table when supplied, else linear scan. Both paths verify (field_number, depth) so a
+  // buggy table can't silently dispatch to the wrong index.
+  auto lookup_field_idx = [&](int fn,
+                              int const* fn_to_idx,
+                              int fn_tbl_size,
+                              int const* field_indices,
+                              int num_fields_at_depth) -> int {
+    auto matches = [&](int local_i) {
+      int schema_idx = field_indices[local_i];
+      return schema[schema_idx].field_number == fn && schema[schema_idx].depth == depth_level;
```
igorpeshansky: [Optional] Fewer lookups:

```suggestion
      const auto& field_schema = schema[field_indices[local_i]];
      return field_schema.field_number == fn && field_schema.depth == depth_level;
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -394,106 +311,87 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
+        info.count++;
+        return true;
+      };
+      if (!walk_repeated_element(
+            cur, msg_end, msg_base, wt, schema[schema_idx].wire_type, error_flag, count_action)) {
+        return;
       }
     }
 
-    // Check nested message fields at this depth
```
igorpeshansky: This comment was useful — otherwise the `lookup_field_idx` calls look very similar and it's not immediately clear why there are two.

src/main/cpp/src/protobuf/protobuf_builders.cu
igorpeshansky: Right, and now `counts_end - counts_begin == num_rows` by construction.

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -386,26 +370,8 @@ std::unique_ptr<cudf::column> build_repeated_string_column(
...
-                                : cudf::make_empty_column(cudf::data_type{cudf::type_id::STRING});
-
-    return make_list_column_with_input_nulls(
-      num_rows, std::move(offsets_col), std::move(child_col), binary_input, stream, mr);
-  }
-
-  auto list_offs = make_list_offsets_from_counts(d_field_counts, total_count, num_rows, stream, mr);
+  CUDF_EXPECTS(total_count > 0, "build_repeated_string_column: total_count must be > 0");
 
-  // Extract string lengths from occurrences
```
igorpeshansky: Any reason you're removing this comment? Seems useful…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -459,8 +425,6 @@ std::unique_ptr<cudf::column> build_repeated_string_column(
 
   std::unique_ptr<cudf::column> child_col;
   if (is_bytes) {
-    // Transfer ownership of the chars buffer instead of copying — the strings path below uses
-    // `chars.release()` for the same reason.
```
igorpeshansky: Any reason you're removing this comment? Seems useful…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -545,46 +445,47 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
     write_idx[f] = scan_descs[f].row_offsets[row];
   }
 
+  // Returns descriptor index for `fn`, or -1. Same lookup-table-or-linear pattern + buggy-
+  // table guard as count_repeated_fields_kernel above.
+  auto lookup_desc_idx = [&](int fn) -> int {
```
igorpeshansky, line 454: This is the same as `lookup_field` in `protobuf_device_helpers.cuh`, but with an additional buggy lookup table guard. Does it make sense to propagate the sanity check to `lookup_field` instead, and use that here?

src/main/cpp/src/protobuf/protobuf_kernels.cu
igorpeshansky: No, your explanation makes sense, let's leave it here.

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -878,22 +879,20 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-              }
-            } else {
-              set_error_once_async(d_error.data(), ERR_MISSING_ENUM_META, stream);
-              column_map[schema_idx] = make_null_column(dt, num_rows, stream, mr);
-            }
+            // Missing or mismatched enum metadata is a schema/config bug (host-side check),
+            // not a row-level decode error — fail loudly rather than silently produce a null
+            // column that downstream can't distinguish from a real all-null result.
+            CUDF_EXPECTS(schema_idx < static_cast<int>(enum_valid_values.size()) &&
+                           schema_idx < static_cast<int>(enum_names.size()),
```
igorpeshansky: [Optional] `validate_decode_context` already checks that `enum_valid_values.size() == enum_names.size() == num_fields`, so one check might be enough here…

[Really optional] Technically, even the `schema_idx < enum_valid_values.size()` check is redundant, since all `scalar_field_indices` are under `num_fields` [by construction](https://github.com/NVIDIA/spark-rapids-jni/pull/4504/changes#diff-9569500dd68b3a114790c052ad11b4181483794aa9e43bddb3f068b4d552c404R465-R475), but it should be fine as a sanity check…

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -954,68 +953,61 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
           break;
         }
         default:
-          // For LIST (bytes) and other unsupported types, create placeholder columns
-          column_map[schema_idx] = make_null_column(dt, num_rows, stream, mr);
-          break;
+          // Unreachable: schema validation only admits the scalar element types enumerated
+          // above (STRUCT is dispatched through the nested path, not this scalar switch).
+          CUDF_FAIL("Protobuf decode internal error: unsupported scalar element type");
```
igorpeshansky: [Optional] Is there any way to include the offending type in the error?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1061,216 +1049,202 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+                                                                true,
+                                                                d_error,
+                                                                stream,
+                                                                mr);
+          break;
+        default:
+          // Unreachable: schema validation admits the types enumerated above; STRUCT is
+          // rejected at the loop entry above (parts 3b/3c). Reaching this branch means a
+          // schema slipped past validation.
+          CUDF_FAIL("Protobuf decode internal error: unsupported repeated element type");
```
igorpeshansky: [Optional] Any way to include the offending type in the error?

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: You are right, removed both.

src/main/cpp/src/protobuf/protobuf.cu

src/main/cpp/src/protobuf/protobuf.cu

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: renamed `list_offs` to `d_field_offsets`.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Thanks! Updated to the first one.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Added a defensive failure.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Updated to `[[maybe_unused]]`

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Done. Extracted `lookup_field_generic<Match>` in `protobuf_device_helpers.cuh`

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Added back.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: added back.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
```diff
@@ -286,23 +286,49 @@ __device__ inline uint64_t load_le<uint64_t>(uint8_t const* p)
...
+ *
+ * `match(int candidate) -> bool` decides whether the candidate index actually corresponds
+ * to `field_number` (and any other criteria the caller wants to enforce, such as schema
+ * depth). The lookup-table fast path applies the same `match` predicate, so a buggy
+ * lookup table can't silently dispatch to the wrong index. Falls back to linear scan when
+ * the table is empty (e.g., field numbers too sparse to build a direct-mapped table).
+ *
+ * Returns the matching candidate index in `[0, num_candidates)`, or `-1` if not found.
+ */
+template <typename Match>
```
igorpeshansky, line 301: You can also add an `std::is_invocable_r_v` constraint here.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
```diff
@@ -286,23 +286,49 @@ __device__ inline uint64_t load_le<uint64_t>(uint8_t const* p)
...
+                                                    int const* lookup_table,
+                                                    int lookup_table_size,
+                                                    int num_candidates,
+                                                    Match&& match)
+{
+  if (lookup_table != nullptr && field_number > 0 && field_number < lookup_table_size) {
+    int const f = lookup_table[field_number];
+    // Bound `f` against `num_candidates` before invoking `match`, so a buggy table can't
+    // cause an out-of-range read inside the predicate.
+    return (f >= 0 && f < num_candidates && match(f)) ? f : -1;
```
igorpeshansky: [Really optional] Does it make sense to also pass `field_number` into the predicate? I know it can always be captured, but this way we can ensure that it's the same one that's passed into `lookup_field_generic`…

src/main/cpp/src/protobuf/protobuf_builders.cu
igorpeshansky: [Optional] It's still a parameter in one other `build_repeated_*` function (`build_repeated_struct_column`), which seems to be declared but not defined, so I assume it'll be implemented later. If that will eventually end up not using this parameter, it probably makes sense to remove it from `build_repeated_struct_column` as well…

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
```diff
@@ -286,23 +286,49 @@ __device__ inline uint64_t load_le<uint64_t>(uint8_t const* p)
 }
 
 /**
- * O(1) lookup of field_number -> field_index using a direct-mapped table.
- * Falls back to linear search when the table is empty (field numbers too large).
+ * Generic field-number -> field-index lookup.
```
igorpeshansky: This implementation is also O(1) when the table exists, so it's worth keeping the original comment here:

```suggestion
 * O(1) lookup of field_number -> field_index using a direct-mapped table.
 * Falls back to linear search when the table is empty (field numbers too large).
```

Also, nit: `field_number`, not `field-number` (fixed by my suggestion).

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
```diff
@@ -286,23 +286,49 @@ __device__ inline uint64_t load_le<uint64_t>(uint8_t const* p)
...
+  }
+  return -1;
+}
+
+/**
+ * O(1) lookup of field_number -> field_index in a flat `field_descs[]` array.
+ * Thin wrapper over `lookup_field_generic` — the predicate is just the field-number match.
  */
-// Keep this definition in the header so all CUDA translation units can inline it.
 __device__ __forceinline__ int lookup_field(int field_number,
```
igorpeshansky: [Optional] This is only called once, from `scan_all_fields_kernel` — does it still make sense to keep the wrapper? Let's just inline it there?

```c++
    int f = lookup_field_generic(fn, field_lookup, field_lookup_size, num_fields, [&](int i) {
      return field_descs[i].field_number == fn;
    });
```

Also, if we remove this wrapper, we can just rename `lookup_field_generic` to `lookup_field`, right?

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
thirtiseven: Applied, thanks.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
thirtiseven: Yes, done.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Removed it too.

igorpeshansky (approved):
LGTM :shipit:

thirtiseven:
build

PR #4507, opened 2026-04-30
Add CastStrings.parseTimestampWithFormat fused parser kernel

thirtiseven:
## Summary

Adds `CastStrings.parseTimestampWithFormat`, a generic Spark-style date/timestamp parser that replaces the cuDF regex-validate + regex-rewrite + `asTimestamp` chain used by `GpuToTimestamp` for the 24 documented `LEGACY_COMPATIBLE_FORMATS` / `CORRECTED_COMPATIBLE_FORMATS` patterns on the spark-rapids side.

**The kernel itself is format-agnostic.** Mirroring how Spark's `DateTimeFormatter` / `SimpleDateFormat` represent a parser internally, the format string is compiled host-side into a sequence of tokens, and a generic device-side walker consumes the tokens against each row. The `(sparkFormat, policy)` pair flows in as a JNI argument, not as an enum index — so the kernel has no per-format `if`/`switch` arms and the spark-rapids gating set (today fixed at 24 formats by `*_COMPATIBLE_FORMATS`) lives entirely in the caller. Extending the supported set in the future is a one-line caller-side change with no JNI rebuild.

The compile rules follow JDK conventions:
- `y` → year (run length sets digit width, e.g. `yyyy` = 4); `M` → month, `m` → minute (NOT month); `d` → day; `H` → hour; `s` → second.
- "Packed" digit runs (two digit fields abutting without a literal between them, e.g. `yyyyMMdd` or `MMyyyy`) compile to exact-width digit tokens — otherwise the boundary is ambiguous. Other digit fields are exact-width in CORRECTED and `[1, 2]` digits in LEGACY.
- A literal space compiles to `TOK_T_OR_SPACE` (Spark accepts both ' ' and 'T' as the date/time separator).
- In LEGACY, `-` and `/` literals are followed by `TOK_SKIP_HT_WS`, folding the legacy `REMOVE_WHITESPACE_FROM_MONTH_DAY` regex rewrite into the state machine instead of materializing an intermediate string column.
- The terminating token is `TOK_TRAIL_EOF` for CORRECTED and `TOK_TRAIL_NON_DIGIT` for LEGACY.

Iterates `cudf::string_view` directly via `thrust::for_each_n`. No regex engine, no `stringReplaceWithBackrefs`, no `fixedUp` chars buffer on the hot path. The closest precedent is `cast_string_to_datetime.cu`'s `parse_string_to_date_fn` for the dispatch shape and `validity` buffer → `cudf::bools_to_mask` flow.

In addition to the perf benefit, the fused kernel makes this code path's GPU memory footprint predictable — today the regex chain can produce a large single allocation under cuDF whose size is not visible to the spark-rapids estimator, which has caused OOMs on large batches.

## Notable behaviour

- **Lowercase `mm` is minute, not month.** `yyyymmdd` and `yyyy-mm-dd` LEGACY parse the middle 2 digits as minute-of-hour and store a `00:mm:00` time on day 1 of January, matching `SimpleDateFormat` (and matching what `spark-rapids/DateUtils.toStrf` previously routed to cuDF as `%M`).
- **`yyyyMMdd HH:mm:ss` LEGACY**: the original `\d{4}\d{1,2}\d{1,2}` regex relied on backtracking to split 6/7/8 date digits across month/day. The streaming walker has no backtracking and only accepts the 4+2+2 = 8-digit form. This matches what Spark CPU's non-lenient `SimpleDateFormat` does in practice; the regex's flexibility was unreachable in the production chain because cuDF's `%Y%m%d` never accepted partial dates either.

## Test plan

- [x] New `CastStringsTest#parseTimestampWithFormat_*` cases cover corrected/legacy date-only and date-time formats, the LEGACY whitespace fold, the lowercase-`mm` minute interpretation, packed `yyyyMMdd`, and invalid calendar dates (`2023-02-29`, `2024-04-31`, `2024-13-01`).
- [x] Existing `CastStringsTest` and cpp `CAST_STRING` tests still pass.
- [ ] spark-rapids-side wiring and integration tests in NVIDIA/spark-rapids#14706.

Related to NVIDIA/spark-rapids#14689.


res-life:
do we have perf number compared to legacy spark-rapids solution, or compared to cuDF API. I know cuDF is incompatible to Spark, but we can compare perf.

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -0,0 +1,390 @@
...
+  bool legacy,
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr)
+{
+  SRJ_FUNC_RANGE();
+  auto const host_tokens = compile_format(format, legacy);
+  auto const num_rows    = input.size();
+  if (num_rows == 0) {
+    return cudf::make_empty_column(cudf::data_type{cudf::type_to_id<cudf::timestamp_us>()});
+  }
```
res-life, line 351: nit: check num_rows == 0 before `compile_format`

thirtiseven:
> do we have perf number compared to legacy spark-rapids solution, or compared to cuDF API. I know cuDF is incompatible to Spark, but we can compare perf.

Yes I did an e2e perf test on https://github.com/NVIDIA/spark-rapids/pull/14706

res-life:
nit: letter_to_field rejects everything except yMdHms. Worth a line in the Javadoc clarifying that lowercase h (12-hour), S (fraction-of-second), and timezone letters.

res-life:
In general, LGTM, only some nits.

src/main/cpp/src/parse_timestamp_with_format.cu
thirtiseven: spark-rapids PR already gates this JNI call to the compatible format sets, so this does not affect the spark-rapids fast path.

For the JNI API itself, I think we should keep format validation before the empty-input short-circuit. Otherwise invalid patterns would be silently accepted when the input has zero rows, making the API behavior data-dependent.

thirtiseven:
> nit: letter_to_field rejects everything except yMdHms. Worth a line in the Javadoc clarifying that lowercase h (12-hour), S (fraction-of-second), and timezone letters.

Done. I added a Javadoc sentence to `parseTimestampWithFormat` clarifying that unsupported letters, including lowercase `h`, `S`, and timezone pattern letters, are rejected.

nvauto:

res-life (approved):
LGTM

thirtiseven:
build

PR #4521, opened 2026-05-06
Protobuf: harden make_null_mask_from_valid against padded valid buffers

thirtiseven:
Fixes a latent issue in the `from_protobuf` decoder helpers.

`make_null_mask_from_valid` (in `src/main/cpp/src/protobuf/protobuf_kernels.cuh`) used `valid.size()` as the row count. That works as long as the caller's `valid` buffer is exactly `num_rows` long, but the decoder builders pad `valid` to `max(1, num_rows)` to sidestep 0-sized `device_uvector`s. With the padded shape, `valid.size()` no longer equals the logical row count, so the helper would either over-read or hand `valid_if` a counting iterator that doesn't match the column it ends up backing.

## Fix

- Add an explicit `num_rows` parameter to the helper.
- Guard misuse with two `CUDF_EXPECTS`: one for negative `num_rows` (so a wraparound on `static_cast<size_t>` doesn't surface as a misleading "buffer smaller" message), and one for `valid.size() < num_rows`.
- All current call sites already have the right value in scope; updated each to pass it through. No behavior change on existing paths.

## Tests

No dedicated unit test added. The reachable behavior is exercised by every nullable-column protobuf decode test on the Java side; the new `CUDF_EXPECTS` guards cover code paths that no current caller can reach, so a unit test for them would only restate what code review already establishes.

## Context

Surfaced during review of PR #4504 (from_protobuf part 2). That PR adds the first caller that triggers the padded-buffer pattern; landing the helper hardening separately keeps it small and self-contained.


src/main/cpp/src/protobuf/protobuf_kernels.cuh
```diff
@@ -407,14 +407,22 @@ CUDF_KERNEL void extract_lengths_kernel(LocationProvider loc_provider,
 // Host-side template helpers that launch CUDA kernels
 // ============================================================================
 
+// Build a row-aligned null mask from `valid[row]` boolean flags. The caller must ensure
+// `valid.size() >= num_rows`; some builders pad `valid` to `max(1, num_rows)` to avoid a
+// 0-sized device_uvector and pass `num_rows` separately so the resulting mask matches the
+// logical row count rather than the padded buffer length.
```
igorpeshansky: Are there any such builders in the code today that call `make_null_mask_from_valid` (including the pending PRs)? If not, I'd rephrase as "some builders pad \`valid\` to \`max(1, num_rows)\` to avoid a 0-sized device_uvector and _would need to_ pass `num_rows` separately…" instead, and frame this as a defensive change.

That said, a new unit test that passes in a `valid` buffer of size 1 (or more) and `num_rows`==0 (or smaller than `valid` size) could verify that the returned `rmm::device_buffer`'s size is proportional to `num_rows`, and not to the size of `valid`. Alternatively, the test could initialize `valid` to `{…, false}`, for which the old code would return a larger `null_count` than your new code (e.g., for `valid`==`{false}` and `num_rows`==0, the old code would return a `null_count` of 1, and your new code of 0).

thirtiseven: Thanks for the review. Updated the comment and added a unit test.

igorpeshansky: I don't see any new changes. Did you need to push your branch?

thirtiseven: Sorry, pushed now.

igorpeshansky (approved):
LGTM :shipit:

thirtiseven:
build

PR #4536, opened 2026-05-11
from_json: null depth-1 ancestor on nested schema mismatch (Spark parity)

### LINKED ISSUE #4535: [BUG] from_json: nested schema-category mismatch should null depth-1 ancestor (SPARK-33134 parity)
## Description

`JSONUtils.fromJSONToStructs` (the GPU implementation behind Spark's `from_json` for struct schemas) does not match Spark's CPU behavior when a JSON record contains a nested value whose type category disagrees with the requested schema. Spark's `JacksonParser.convertObject` runs with `isRoot=false` for nested fields and propagates the conversion exception up, nulling the entire depth-1 ancestor field for that row; the GPU path silently returns the partial result.

Concrete repro from the SPARK-33134 test (`org.apache.spark.sql.JsonFunctionsSuite`):

```scala
val st = new StructType().add("c1", LongType).add("c2",
            ArrayType(new StructType().add("c3", LongType).add("c4", StringType)))
val df2 = Seq("""{"data": {"c2": [19], "c1": 123456}}""").toDF("c0")
withSQLConf(SQLConf.JSON_ENABLE_PARTIAL_RESULTS.key -> "false") {
  checkAnswer(df2.select(from_json($"c0", new StructType().add("data", st))),
              Row(Row(null)))  // CPU: data is NULL
}
// GPU returns Row(Row(Row(123456, null))): "data" stays populated, "data.c2" is nulled.
```

In `RapidsJsonFunctionsSuite` this is exposed as an `.exclude(...)` entry in `RapidsTestSettings.scala`, pointing at this issue. The mismatch is on `c2` (array of struct elements that are scalars in the JSON), which according to Spark should null the whole `data` field.

## Root cause

cuDF's `read_json` detects the category mismatch in `host_tree_algorithms::mark_is_pruned` and prunes the offending subtree, but it does not surface that information to the caller. spark-rapids-jni's `from_json_to_structs.cu` therefore has no signal that the parent should be nulled and returns a partial result.

## Proposed fix

A two-layer split, so cuDF stays Spark-agnostic:

1. **Upstream (cuDF)**: expose a per-top-level-column diagnostic `column_name_info::had_schema_mismatch` (`std::optional<bool>`) set whenever the JSON tree under that column was pruned for category mismatch against the user-supplied schema. Tracked at rapidsai/cudf#22450 (closes rapidsai/cudf#22423).
2. **spark-rapids-jni**: consume the diagnostic. In `from_json_to_structs.cu`, after `read_json` returns and before `convert_data_type` runs, replace each flagged top-level `parsed_columns[i]` with an ALL_NULL mask. The downstream conversion path propagates the NULL through descendants, so Spark consumers observe every field of the depth-1 column as NULL — matching `JacksonParser.convertObject` with `isRoot=false`.

The jni-side patch is a self-contained 11-line addition; PR to follow.

## Validation

Local `mvn package -pl tests -am -Dbuildver=330 -DwildcardSuites=org.apache.spark.sql.rapids.suites.RapidsJsonFunctionsSuite` against a worktree that consumes the cuDF #22450 branch: `Tests: succeeded 66, failed 0, canceled 0, ignored 0, pending 0` (66/66 incl. SPARK-33134, both PARTIAL_RESULTS=true and false branches).

wjxiz1992:
Closes #4535.

## Summary

`JSONUtils.fromJSONToStructs` did not match Spark CPU when a nested JSON value's category disagreed with the requested struct schema. Spark's `JacksonParser.convertObject` runs with `isRoot=false` for nested fields and propagates conversion errors up by nulling the entire depth-1 ancestor field; the GPU path silently returned a partial result. Concrete repro from `SPARK-33134` in `RapidsJsonFunctionsSuite`:

```
input:  {"data": {"c2": [19], "c1": 123456}}
schema: data: struct<c1: long, c2: array<struct<c3, c4>>>
Spark CPU:  Row(Row(null))                  -- data is NULL
GPU before: Row(Row(Row(123456, null)))     -- partial result
```

## What changed

`src/main/cpp/src/from_json_to_structs.cu` (~20 lines): instead of `cudf::io::read_json`, call the new `cudf::io::read_json_with_diagnostics`. The returned `json_reader_result` carries a `diagnostics.top_level_columns_with_schema_mismatch: std::vector<std::string>` side channel from rapidsai/cudf#22450. We hoist that list into an `unordered_set<std::string>` once before the per-column loop and, for each top-level column whose name is in the set, replace its null mask with `ALL_NULL` *before* `convert_data_type` runs. `convert_data_type` then propagates the NULL through descendants naturally, so Spark sees every field of the depth-1 column as NULL.

The diagnostic-only design keeps cuDF Spark-agnostic — cuDF only reports what it observed; the parent-NULL policy is applied here on the spark-rapids-jni side.

### History note on the cuDF dependency

The first revision of this PR consumed `column_name_info::had_schema_mismatch`, which was the original Plan-A shape of rapidsai/cudf#22450. That ABI extension to a public IO struct broke prebuilt `librapidsmpf-cu13` wheels (inlined `std::vector<column_name_info>::~vector()` ran with a stale 72-byte stride against new 80-byte elements, segfaulting at `address 0x38`). The merged cuDF design moved the diagnostic to a side-channel API (`read_json_with_diagnostics` / `json_reader_result` / `json_reader_diagnostics`) so the public `column_name_info` ABI stays unchanged. This PR was rewritten in commit f35ba775 to consume that final API; rapidsai/cudf#22450 is now merged (dc9061491e) and the submodule is bumped accordingly.

### Why mask before `convert_data_type` rather than after

Spark consumers read struct child columns directly (no `push_down_nulls` in the GPU read path). Masking the converted struct's parent leaves leaf children's data addressable and the scalar `c1 = 123456` reads back regardless. Setting the mask on `parsed_columns[i]` before conversion lets cuDF's existing conversion path AND parent/child masks during column construction, which produces the correct all-NULL result for downstream consumers.

## Validation

Local `mvn package -pl tests -am -Dbuildver=330 -Dmaven.repo.local=./.mvn-repo -DwildcardSuites=org.apache.spark.sql.rapids.suites.RapidsJsonFunctionsSuite -Drapids.test.gpu.allocFraction=0.3 -Drapids.test.gpu.maxAllocFraction=0.3 -Drapids.test.gpu.minAllocFraction=0` from a spark-rapids worktree consuming a snapshot of this branch built against cuDF main with #22450:

```
Tests: succeeded 66, failed 0, canceled 0, ignored 0, pending 0
```

All 66 tests in `RapidsJsonFunctionsSuite` pass, including the previously-failing SPARK-33134 case with both `JSON_ENABLE_PARTIAL_RESULTS=true` and `false` branches.

The follow-up spark-rapids PR (NVIDIA/spark-rapids#14773) that removes the `.exclude(...)` line for SPARK-33134 in `RapidsTestSettings.scala` and re-enables the test will become merge-ready once this PR lands and a JNI snapshot ships.

## Checklists

- [x] This PR has added documentation for new or modified features or behaviors.


nvauto:

src/main/cpp/src/from_json_to_structs.cu
```diff
@@ -855,6 +863,19 @@ std::unique_ptr<cudf::column> from_json_to_structs(cudf::strings_column_view con
...
     auto const& [col_name, col_schema] = schema_with_precision.child_types[i];
     CUDF_EXPECTS(parsed_meta.schema_info[i].name == col_name, "Mismatched column name.");
+
+    // Apply Spark's parent-NULL policy: JacksonParser.convertObject with isRoot=false propagates
+    // nested conversion errors up and nulls the depth-1 ancestor. cuDF reports schema-category
+    // mismatches via the side-channel `json_reader_diagnostics` returned by
+    // `read_json_with_diagnostics`; replicate the policy here so it doesn't leak into cuDF's
+    // public ABI. We null the parsed column BEFORE `convert_data_type` so the conversion path
+    // naturally propagates nulls through descendants and Spark consumers see every field NULL.
+    if (mismatched_columns.contains(col_name)) {
```
thirtiseven: Does this null the entire top-level column across all rows once any row in the batch has a nested schema mismatch? The issue and cuDF diagnostic docs both describe Spark's behavior as row-level ("for such rows"), but `top_level_columns_with_schema_mismatch` only carries column names, not row positions.

Do we need a row-level diagnostic/mask here, or at least a mixed-row test before accepting the all-column mask? For example, what should happen for one valid row and one mismatched row in the same input column?

```scala
Seq(
  """{"data":{"c2":[{"c3":19,"c4":"x"}],"c1":1}}""",
  """{"data":{"c2":[19],"c1":2}}"""
)
```

Would Spark preserve the first row's data while this code makes both rows' data NULL?

src/main/cpp/src/from_json_to_structs.cu
```diff
@@ -855,6 +863,19 @@ std::unique_ptr<cudf::column> from_json_to_structs(cudf::strings_column_view con
 
     auto const& [col_name, col_schema] = schema_with_precision.child_types[i];
     CUDF_EXPECTS(parsed_meta.schema_info[i].name == col_name, "Mismatched column name.");
+
+    // Apply Spark's parent-NULL policy: JacksonParser.convertObject with isRoot=false propagates
+    // nested conversion errors up and nulls the depth-1 ancestor. cuDF reports schema-category
+    // mismatches via the side-channel `json_reader_diagnostics` returned by
+    // `read_json_with_diagnostics`; replicate the policy here so it doesn't leak into cuDF's
+    // public ABI. We null the parsed column BEFORE `convert_data_type` so the conversion path
+    // naturally propagates nulls through descendants and Spark consumers see every field NULL.
```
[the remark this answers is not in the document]
wjxiz1992: Good catch — addressed in `d955de89`. Rewrote the comment to attribute the "every descendant field NULL" effect to Spark, not cuDF:

> Setting the columns null mask to ALL_NULL here marks the depth-1 ancestor invalid in the validity buffer; downstream Spark consumers then see every descendant field as NULL because Sparks expression evaluator returns NULL when reading fields off a null parent struct (not because cuDF rewrites the child columns).

You are right that cuDF still holds the child data; only the parent struct columns validity flips. Spark catalysts struct-field accessor is what observes the NULL downstream.

src/main/cpp/src/from_json_to_structs.cu
```diff
@@ -855,6 +863,19 @@ std::unique_ptr<cudf::column> from_json_to_structs(cudf::strings_column_view con
...
+    if (mismatched_columns.contains(col_name)) {
+      auto const num_rows = parsed_columns[i]->size();
```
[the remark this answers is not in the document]
wjxiz1992: Verified, you are right. cuDFs `top_level_columns_with_schema_mismatch` diagnostic is column-scoped — it flags column X when SOME row in X has a schema mismatch, but does not return WHICH rows. The current `ALL_NULL` mask therefore over-nulls valid rows in a mixed-row batch (e.g. row 1 fine, row 2 has `c2` type mismatch → both rows ALL_NULL on the GPU, vs Spark which keeps row 1 intact and nulls row 2 only).

I documented the caveat inline in `d955de89`:

> Granularity caveat: the cuDF diagnostic is column-level, not row-level, so we null ALL rows of `col_name` even when only some rows in that column actually had the mismatch. This over-nulls mixed-row batches relative to Sparks per-row depth-1 nulling. Per-row diagnostics from cuDF would let us null only the affected rows; tracked in #4645.

Filed the per-row diagnostic request as [#4645](https://github.com/NVIDIA/spark-rapids-jni/issues/4645). Once cuDF exposes `top_level_columns_with_schema_mismatch_rows`, the loop here can switch from a global `cudf::create_null_mask(num_rows, ALL_NULL)` to a `cudf::detail::valid_if` driven by a per-row boolean mask, matching Spark exactly.

For this PR I am keeping the column-wide mask because (a) it preserves Sparks behavior in the common case where all mismatched rows share a column (all-mismatch batches), (b) it matches the column-level granularity of the diagnostic cuDF currently exposes, and (c) the alternative — per-row JSON re-parsing on the host — would defeat the purpose of the GPU JSON reader. Happy to gate it behind a kill switch (e.g. `spark.rapids.sql.fromJson.parentNullPolicy.enable`) if you would prefer that until #4645 lands.

src/main/cpp/src/from_json_to_structs.cu
wjxiz1992: Good question — and the answer is that the current ALL_NULL mask **does** null both rows for the mixed-row case you described, but for a different reason than Spark:

- Java/Spark: row 1 keeps `data = Row(c1=1, c2=Array(Row(19,"x")))`, row 2 has `data = null` (JacksonParsers per-row depth-1 nulling).
- cuDF + this PRs workaround: cuDFs `top_level_columns_with_schema_mismatch` flags `data` as the mismatched top-level column, the loop applies ALL_NULL to the `data` column → both rows have `data = null`.

So for your specific 2-row input the GPU happens to converge on "both null", but for a more skewed mix (e.g. 999 valid + 1 mismatch) the GPU over-nulls all 1000 rows where Spark would null only the offending row.

I documented the caveat in the inline comment in `d955de89` and filed [#4645](https://github.com/NVIDIA/spark-rapids-jni/issues/4645) for the cuDF API enhancement that would let us null per-row. The end-to-end test belongs in spark-rapids Python ITs (where the JNI surface is consumed); I will add (a) an all-mismatch test case asserting the current depth-1 NULL, and (b) the mixed-row case from your example as an `xfail` documenting the over-nulling until cuDF #4645 lands. Will link the spark-rapids PR back to this thread when it is up.

If you would rather we **block this PR** on the per-row diagnostic (waiting for #4645 to land first), let me know — I can pull the column-wide mask and instead gate the policy behind a feature flag that defaults off until the per-row variant is wired up.

wjxiz1992:
build

src/main/cpp/src/from_json_to_structs.cu
wjxiz1992: not blocking on the per-row diagnostic. I added a per-batch mismatch flag in JNI (a314a89a); on the spark-rapids side we read it and fall the whole batch back to CPU JsonToStructs when any row mismatches, so those batches are computed on CPU and match it. the column-wide null stays as the default when fallback is off. spark-rapids PR coming, will link here. per-row diagnostic still tracked in #4645 to drop the fallback later.

wjxiz1992:
build

src/main/cpp/src/from_json_to_structs.cu
wjxiz1992: spark-rapids side is up: NVIDIA/spark-rapids#14773 (folded into the SPARK-33134 re-enable PR). the existing IT test_from_json_struct_of_list_with_mixed_nested_types_input mixes valid + mismatched rows, so it covers the mixed-row case. both stay DO-NOT-MERGE until this lands + a JNI nightly publishes the flag.

wjxiz1992:
Kindly ping @thirtiseven for another look. Major decision is to block this PR or process with a follow up https://github.com/NVIDIA/spark-rapids-jni/issues/4645.

src/main/cpp/src/from_json_to_structs.cu
```diff
@@ -855,6 +868,26 @@ std::unique_ptr<cudf::column> from_json_to_structs(cudf::strings_column_view con
...
+    //
+    // Granularity caveat: the cuDF diagnostic is column-level, not row-level, so we null ALL
+    // rows of `col_name` even when only some rows in that column actually had the mismatch.
+    // This over-nulls mixed-row batches relative to Spark's per-row depth-1 nulling. Per-row
+    // diagnostics from cuDF would let us null only the affected rows; tracked in #4645.
+    if (mismatched_columns.contains(col_name)) {
+      auto const num_rows = parsed_columns[i]->size();
+      auto all_null_mask = cudf::create_null_mask(num_rows, cudf::mask_state::ALL_NULL, stream, mr);
+      parsed_columns[i]->set_null_mask(std::move(all_null_mask), num_rows);
+    }
```
thirtiseven: This PR ships two independent fix mechanisms that appear to be mutually redundant, and one of them is documented as incorrect.

1. **GPU-side masking** (this block): `ALL_NULL` is applied to each mismatched column before `convert_data_type`, producing a null depth-1 ancestor on the GPU.
2. **Flag + CPU fallback**: `had_schema_mismatch` flows up to `FromJSONResult.hasSchemaMismatch`, and per the PR description spark-rapids uses it to fall back the *whole batch* to CPU `JsonToStructs`.

If the batch always falls back to CPU when the flag is set, this masked GPU result is computed and then discarded -- wasted work. And the masking is documented as not matching Spark anyway:

> the cuDF diagnostic is column-level, not row-level, so we null ALL rows of `col_name` even when only some rows in that column actually had the mismatch. This over-nulls mixed-row batches relative to Spark's per-row depth-1 nulling.

So the column-granularity masking can produce results that diverge from Spark on mixed-row batches -- which is precisely why the CPU fallback exists. If masking is ever actually consumed (fallback not taken), it is a latent correctness bug.

**Suggestion:** pick one mechanism. If the design is flag -> full CPU fallback for parity, then this masking block (plus the `<unordered_set>` include and the `mismatched_columns` set) is dead, over-nulling code -- remove it and derive the flag directly from emptiness of `parsed_result.diagnostics.top_level_columns_with_schema_mismatch`. That keeps the change minimal and removes the latent hazard. If masking is meant to stay live in some non-fallback config, please state that explicitly and add a test covering a mixed-row batch (row 1 matches, row 2 mismatches) to pin down the over-null behavior.

Note: the current `FromJsonToStructsTest` only asserts the flag, not the masked data, so neither the masking correctness nor this interaction is covered here.

wjxiz1992: good catch. since #14773 always falls the whole batch back to CPU when the flag is set, the GPU-side masking was dead work — removed it plus the unordered_set, and the flag now comes straight from top_level_columns_with_schema_mismatch.empty(). FromJsonToStructsTest only asserts the flag so it still passes.

src/main/cpp/src/from_json_to_structs.cu
wjxiz1992: removed the masking entirely, so this column-level over-nulling is gone too.

wjxiz1992:
build

src/main/cpp/src/from_json_to_structs.cu
```diff
@@ -809,6 +809,7 @@ std::unique_ptr<cudf::column> from_json_to_structs(cudf::strings_column_view con
                                                    bool allow_nonnumeric_numbers,
                                                    bool allow_unquoted_control,
                                                    bool is_us_locale,
+                                                   bool* had_schema_mismatch,
```
thirtiseven, line 812: [Optional] Should this return a small result struct instead of adding `bool* had_schema_mismatch` as an output parameter? The [libcudf developer guide](https://github.com/rapidsai/cudf/blob/main/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md) says outputs should generally be returned rather than passed as output parameters, and suggests `struct`/`std::pair` for multiple return values. I realize this is JNI-adjacent internal code rather than a libcudf public API, so maybe the pointer is fine here, but a result struct would keep ownership and diagnostics together.

```c++
struct from_json_to_structs_result {
  std::unique_ptr<cudf::column> data;
  bool had_schema_mismatch{};
};

from_json_to_structs_result from_json_to_structs(...);
```

thirtiseven (approved):
LGTM, only a style nit. And also please update PR description before merging.

wjxiz1992:
build

src/main/cpp/src/from_json_to_structs.cu
wjxiz1992: kept the bool* out-param — it's the internal JNI path and matches the existing style in this file. would switch to a result struct if we add more outputs here.

PR #4539, opened 2026-05-12
[Part1] [ORC-timezone]: Build ORC timezone metadata at runtime, drop pregenerated binary

res-life:
First in a stack splitting #4432 into reviewable chunks.

### Description

Previously `OrcTimezoneInfo` loaded its transition tables from a pregenerated 544KB binary (`src/main/resources/orc_timezone_info.data`, derived from `sun.util.calendar.ZoneInfo` on OpenJDK 8). This froze ORC timezone behavior to that snapshot and required regenerating and committing the binary whenever the data needed to change.

This PR replaces the binary with a runtime build path:

- `buildHistoricalTransitions` walks `ZoneRules.getTransitions()` from year 0001 (the lower bound ORC supports) and probes `TimeZone.getOffset()` to capture the offset before each known transition.
- `collectTimeZoneTransitionsByScanning` covers any gap between two known transitions by stepping forward at 1-day granularity and binary-searching the exact transition millisecond when an offset change is detected.
- Per-id results are cached in `RUNTIME_TIMEZONE_INFOS`.
- `getAllTimezoneIds()` now returns the live `TimeZone.getAvailableIDs()` instead of a hard-coded array.

`OrcTimezoneInfo` keeps its `(rawOffset, transitions, offsets)` shape so the CUDA kernel is unchanged. DST timezones remain blocked by the existing `hasDaylightSavingTime` guard in `convertOrcTimezones` — follow-up PRs lift this guard:

1. **(this PR)** Build OrcTimezoneInfo at runtime
2. DST rule extraction (Java side, CPU verified)
3. GPU DST kernel + new `convertOrcTimezones` signature
4. `OrcTimezoneContext` + remaining tests

`isSupportedTimeZone` is broadened from `catch (ZoneRulesException)` to `catch (DateTimeException)` so IDs like `+05:30` are rejected/accepted consistently with the runtime build path (`ZoneId.of` may throw either, depending on whether the id has a malformed offset vs. an unknown name).

### Checklists

- [ ] Documentation: no user-facing change.
- [x] Tests: behavior preserved for non-DST timezones; covered by existing tests in this repo and the companion spark-rapids PR.
- [x] Performance: no expected change for the existing non-DST path. The first call per timezone pays a one-time runtime-build cost (measured ~4 ms typical, ~12 ms worst on the previous full-DST iteration); subsequent calls hit the cache.


nvauto:

res-life:
build

res-life:
build

res-life:
@greptile-apps review

res-life:
build

firestarman (approved):
LGTM

PR #4548, opened 2026-05-13
ORC tz split 4/6: CUDA civil calendar helpers and DST math

res-life:
Part of the split of #4432.
Companion: https://github.com/NVIDIA/spark-rapids/pull/14544
Previous: #4547

Adds the GPU-side machinery needed for runtime DST handling. Old `get_transition_index` / `convert_timestamp_between_timezones` / kernel are left in place; they are replaced by the new versions in #orc-tz-5.

Changes in `timezones.cu`:

- Add `#include <cudf/utilities/bit.hpp>` and `<cuda/std/array>`.
- MS_PER_SECOND/MINUTE/HOUR/DAY constants.
- Branchless civil calendar primitives: DAYS_BEFORE_MONTH lookup, is_leap_year, days_in_month, days_before_month, days_from_epoch_to_year, millis_to_year.
- DST rule math: compute_rule_day (SimpleTimeZone-style start-mode evaluation), compute_transition_utc_ms.
- `compute_dst_offset`: GPU-side evaluation of the DST rule.
- New `get_transition_index` overload that consults the DST rule and historical initial offset on out-of-range lookups.
- `is_fixed_offset_tz` helper.

Signed-off-by: Chong Gao <chongg@nvidia.com>

PR #4549, opened 2026-05-13
ORC tz split 5/6: CUDA kernel rewrite with shared-memory and DST integration

res-life:
Part of the split of #4432.
Companion: https://github.com/NVIDIA/spark-rapids/pull/14544
Previous: #4548

Replaces the old ORC timezone conversion path with the DST-aware version:

- Remove the duplicate old `get_transition_index` overload.
- Rewrite `convert_timestamp_between_timezones` to call the new `get_transition_index` that consults DST rules and to short-circuit on fixed-offset writer/reader timezones.
- Replace `convert_timezones_functor` with `convert_timezones_kernel`: staging writer/reader transition tables into shared memory with int64 alignment padding, looping over rows per block.
- Refactor `convert_timezones` host wrapper to accept the per-table OrcTimezoneContext (initial offset, raw offset, dst_rule, optional transition tables) and pick between the kernel and a fast-path fixed-offset transform when both sides are fixed.
- Update `spark_rapids_jni::convert_orc_timezones` to dispatch through the new path.

After this PR, `timezones.cu` matches #4432.

### DIFF
--- a/src/main/cpp/src/timezones.cu
+++ b/src/main/cpp/src/timezones.cu
@@ -518,103 +518,201 @@ __device__ static int32_t get_transition_index(int64_t const* begin,
 }
 
 /**
- * @brief Find the relative offset when moving between timezones at a particular point in time.
- * This is for ORC timezone support.
+ * @brief Convert a timestamp between ORC writer and reader timezones.
  *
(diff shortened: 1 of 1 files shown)


res-life:
useless now.

PR #4560, opened 2026-05-15
Add correctly-rounded fallback for digits > 2^53 in string_to_float

### LINKED ISSUE #4559: [BUG] string_to_float loses 1 ULP when decimal mantissa > 2^53
## Bug

`spark_rapids_jni::string_to_float` (and the same algorithm in `cudf::strings::detail::stod`) is not correctly rounded when the parsed decimal mantissa exceeds 2^53. The kernel does

```cpp
double digitsf = static_cast<double>(digits);           // lossy when digits > 2^53
double exponent = exp10(static_cast<double>(abs(q)));
double result = q < 0 ? digitsf / exponent : digitsf * exponent;
```

The `static_cast<double>(uint64_t)` rounds to the nearest representable double, losing up to 1 bit when `digits > 2^53 ≈ 9.0e15`. The subsequent multiply / divide cannot recover the IEEE-754-nearest result, so the GPU output drifts up to 1 ULP from Java's `Double.parseDouble` (which uses `BigDecimal`-mediated correctly-rounded conversion under the hood).

## Repro

Parse `"1.7976931348623157"` as a double in any path that reaches `spark_rapids_jni::string_to_float` (e.g. JSON float read in the legacy schema path, ANSI `String → DoubleType` cast):

| Side | Bits | Value |
|---|---|---|
| Java `Double.parseDouble` (CPU) | `0x3FFCC359E067A348` | `1.7976931348623157` |
| Current GPU | `0x3FFCC359E067A346` | `1.7976931348623155` (one ULP low) |

This was originally surfaced by the Spark UT framework on `RapidsJsonSuite` in NVIDIA/spark-rapids#10773; the 5 `SPARK-18352` / "Loading a JSON dataset" / "Applying schemas" sub-bullets all fail on this single literal.

## Source pointers

- `src/main/cpp/src/cast_string_to_float.cu:158-201` — final-value compute in `string_to_float` warp-lane-0 finalizer
- `thirdparty/cudf/cpp/include/cudf/strings/detail/convert/string_to_float.cuh:112-128` — same algorithm in cuDF
- Diagnostic post on NVIDIA/spark-rapids#10773: https://github.com/NVIDIA/spark-rapids/issues/10773#issuecomment-4449053254

## Proposed fix

When `digits > 2^53`, route through a `__uint128_t`-based correctly-rounded path:

1. Form `quotient128 * 2^binary_scale` exactly equal to `digits * 10^q`
   (q<0: `(digits << 64) / 10^|q|` with division remainder as sticky;
    q>0: `digits * 10^q` in 128-bit; binary_scale picks up the `<<64`)
2. Find the MSB of `quotient128`, extract 53 mantissa bits + round bit + sticky
3. Round-to-nearest-ties-to-even, handle mantissa overflow into the exponent
4. Compose the IEEE 754 double (including subnormal & overflow-to-infinity paths)

Within `|q| <= 19` (where `10^|q|` fits in `uint64`) this matches `std::stod` bit-for-bit. PR forthcoming.

## Local validation done

- Host-side prototype against `std::stod`: 200,451 cases bit-exact (200k random with `digits > 2^53` and `|q| in [-19, 19]`, 429-case 2^53 boundary sweep at q in [-19, 19], 22 fixed cases including `Double.MAX_VALUE`, `Double.MIN_NORMAL`, halfway values, 19-digit mantissas, negative sign).
- New `castToDoubleHighPrecisionTest` in `CastStringsTest.java` on real GPU: PASS.
- Downstream: `RapidsJsonSuite` (V1 / V2 / LegacyTimeParser) went from `succeeded 122, failed 5` → `succeeded 127, failed 0` after removing the matching exclusions in NVIDIA/spark-rapids.

wjxiz1992:
## Summary

Adds a `__device__` correctly-rounded fallback in `spark_rapids_jni::string_to_float` for the case where the parsed decimal mantissa exceeds 2^53. The previous algorithm — `static_cast<double>(uint64_t digits) * exp10(exp_ten)` — loses up to 1 bit in the cast for `digits > 2^53 ≈ 9.0e15`, and the subsequent multiply / divide can't recover the IEEE-754-nearest answer. After the patch, GPU float parsing matches Java's `Double.parseDouble` (which uses a `BigDecimal`-mediated correctly-rounded conversion) bit-for-bit.

Closes #4559.
Recovers the 5 `RapidsJsonSuite` test exclusions tracked under NVIDIA/spark-rapids#10773 (the `SPARK-18352` / `Applying schemas` / `Loading a JSON dataset …` sub-bullets). The matching `RapidsTestSettings.scala` change is in NVIDIA/spark-rapids#14802 and depends on this PR being merged + a JNI snapshot bump.

## Algorithm

The helper `correctly_rounded_uint64_times_pow10` (in `src/main/cpp/src/cast_string_to_float.cu`) is gated by the caller's `slow_path_eligible = (digits > 2^53) && (|exp_ten| <= 19)`. Inside that window:

1. Form `quotient128 * 2^binary_scale` exactly equal to `digits * 10^q`:
   - `q < 0`: `quotient128 = (digits << 64) / 10^|q|`, with the division remainder feeding the sticky bit; `binary_scale = -64`.
   - `q > 0`: `quotient128 = digits * 10^q` (exact in 128 bits); `binary_scale = 0`.
   - `q == 0`: identity.
2. Find the MSB position of `quotient128`, extract the top 53 bits as the mantissa, the next-lower bit as the round bit, and OR the remaining low bits with the division remainder for the sticky.
3. Apply round-to-nearest-ties-to-even; handle mantissa overflow into one extra exponent bit.
4. Compose IEEE 754 double via `cuda::std::bit_cast`.

Subnormal range, overflow to infinity, and degenerate inputs (`digits == 0`, `q_lo == q_hi == 0`, `abs_q > 19`) are all unreachable under the caller's precondition and are documented with `assert`s rather than runtime branches — keeps the hot path tight.

Outside the window (`digits <= 2^53` OR `|q| > 19`), the legacy `digits * exp10(exp_ten)` path runs. For `digits <= 2^53` the operands are exact as doubles so that path is already correctly rounded; the rare `|q| > 19` + large-digits case is the legacy 1-ULP path (would need an int128 power-of-10 table to fix — out of scope here).

The slow path is the warp-lane-0 finalizer block in `src/main/cpp/src/cast_string_to_float.cu` inside the `if (_warp_lane == 0)` finalizer. Other warp lanes are unaffected; the kernel's compute is concentrated on a single thread per row so the new path adds no warp-sync constraints.

### Unit tests

`castToDoubleHighPrecisionTest` in `src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java` exercises:

- happy path (helper, q <= 0): 17–19-digit mantissas across normal magnitudes
- helper q > 0 path: `9007199254740993e10`, `12345678901234567e7`, `-9007199254740993e15`
- helper `|q| = 19` boundary (both signs): `9007199254740993e19`, `9007199254740993e-19`
- 2^53 trigger boundary (`9007199254740992`, `9007199254740993`, `9.007199254740993`)
- tie-to-even round UP via `(mantissa & 1ULL)` arm: `9007199254740995`
- mantissa rounding rolls into exponent: `9999999999999999.5`
- overflow to ±infinity, subnormal magnitudes (caller's legacy path)
- `|q| > 19` legacy fallback

Comparison is **bit-exact** via `Double.doubleToLongBits` (the cudf default `assertColumnsAreEqual` for FLOAT64 uses 0.01% relative tolerance, ~13 orders of magnitude looser than 1 ULP — would not trip a regression).

Local result: **26/26 CastStringsTest methods pass on sm_89 / CUDA 13.0.**

### End-to-end

With this snapshot installed and the matching 5 `.exclude(...)` lines removed in spark-rapids, `RapidsJsonSuite` × {default, V1, V2, LegacyTimeParser} reports `succeeded 127, failed 0` (was `succeeded 122, failed 5`).

### Benchmark — `src/main/cpp/benchmarks/cast_string_to_float.cpp`

Three axes, each prints `[helper-trigger] N/M = X%` before the timed run so trigger rate is host-verified:

```
[helper-trigger] string_to_double_helper_off: 0/1048576    rows triggered (by construction)
[helper-trigger] string_to_double_helper_off: 0/104857600  rows triggered
[helper-trigger] string_to_double_helper_on:  1000000/1000000 = 100.00% triggered
```

Numbers (NVIDIA RTX 5880 Ada Generation, SM 89, CUDA 13.0):

| Axis | num_rows | GPU Time | Noise |
|---|---:|---:|---:|
| FP32 baseline | 100M | 161.6 ms | 0.07% |
| FP64 helper **OFF** (legacy path) | 100M | 151.9 ms | 0.11% |
| FP64 helper **ON** (100% trigger) | 10M | 11.64 ms | 0.67% |

The helper-OFF axis at 100M (~1.45 us per 1k rows) is the regression-check baseline: cost on the FP64 path without the new helper. helper-ON (~1.11 us per 1k rows) is the worst-case helper cost. The helper path is slightly faster per row than the legacy path because the caller takes an early return on `slow_path_eligible`, skipping the legacy `exp10()` + subnormal-shift branch.

The helper-ON axis uses `{1M, 10M}` instead of `{1M, 100M}` because the handcrafted strings average ~20 chars/row, and `100M × 20 > INT32_MAX` would overflow the strings column's int32 offsets. A host-side `assert(pos <= INT32_MAX)` in `pack_host_strings` guards against silent overflow if the axis is bumped later.

Documentation
- [ ] Updated for new or modified user-facing features or behaviors
- [x] No user-facing change

Testing
- [x] Added or modified tests to cover new code paths
- [ ] Covered by existing tests
      (Please provide the names of the existing tests in the PR description.)
- [ ] Not required

Performance
- [x] Tests ran and results are added in the PR description
- [ ] Issue filed with a link in the PR description
- [ ] Not required

### DIFF
--- a/src/main/cpp/benchmarks/cast_string_to_float.cpp
+++ b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -21,6 +21,7 @@
 #include <cudf/strings/convert/convert_floats.hpp>
 #include <cudf/strings/strings_column_view.hpp>
 #include <cudf/utilities/default_stream.hpp>
+#include <cudf/utilities/error.hpp>
 
 #include <rmm/cuda_stream_view.hpp>
 #include <rmm/device_buffer.hpp>
@@ -29,9 +30,11 @@
 #include <cast_string.hpp>
 #include <nvbench/nvbench.cuh>
 
+#include <cassert>
 #include <cstdint>
 #include <cstdio>
 #include <cstring>
+#include <limits>
 #include <random>
 #include <string>
 #include <utility>
@@ -76,7 +79,10 @@ parsed_decimal parse_for_trigger_gate(std::string const& s)
     if (c == 'e' || c == 'E') { break; }
     if (c < '0' || c > '9') { return out; }
     digits = digits * 10 + static_cast<uint64_t>(c - '0');
-    if (digits > UINT64_MAX) { return out; }
+    // Reject inputs whose accumulator no longer fits in uint64_t. `digits` is
+    // __uint128_t so this is a direct high-half check rather than a
+    // cross-width comparison.
+    if ((digits >> 64) != 0) { return out; }
     any_digit = true;
     if (seen_dot) { ++frac_digits; }
   }
@@ -186,6 +192,10 @@ packed_host_strings pack_host_strings(std::vector<std::string> const& strings)
   for (auto const& s : strings) {
     std::memcpy(out.chars.data() + pos, s.data(), s.size());
     pos += s.size();
+    // Guard the int32 cast — cudf strings columns currently use int32 offsets.
+    // If a future change bumps `num_rows` or string length past INT32_MAX,
+    // fail loudly here instead of silently truncating to a negative offset.
+    assert(pos <= static_cast<size_t>(std::numeric_limits<int32_t>::max()));
     out.offsets.push_back(static_cast<int32_t>(pos));
   }
   return out;
@@ -204,16 +214,16 @@ std::unique_ptr<cudf::column> string_column_from_host(std::vector<std::string> c
   auto const mr = cudf::get_current_device_resource_ref();
   rmm::device_uvector<char> d_chars(packed.chars.size(), stream, mr);
   rmm::device_uvector<int32_t> d_offsets(packed.offsets.size(), stream, mr);
-  cudaMemcpyAsync(d_chars.data(),
-                  packed.chars.data(),
-                  packed.chars.size(),
-                  cudaMemcpyHostToDevice,
-                  stream.value());
-  cudaMemcpyAsync(d_offsets.data(),
-                  packed.offsets.data(),
-                  packed.offsets.size() * sizeof(int32_t),
-                  cudaMemcpyHostToDevice,
-                  stream.value());
+  CUDF_CUDA_TRY(cudaMemcpyAsync(d_chars.data(),
+                                packed.chars.data(),
+                                packed.chars.size(),
+                                cudaMemcpyHostToDevice,
+                                stream.value()));
+  CUDF_CUDA_TRY(cudaMemcpyAsync(d_offsets.data(),
+                                packed.offsets.data(),
+                                packed.offsets.size() * sizeof(int32_t),
+                                cudaMemcpyHostToDevice,
+                                stream.value()));
   stream.synchronize();
 
   auto offsets_col = std::make_unique<cudf::column>(std::move(d_offsets), rmm::device_buffer{}, 0);
--- a/src/main/cpp/benchmarks/cast_string_to_float.cpp
+++ b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -32,7 +32,6 @@
 
 #include <cassert>
 #include <cstdint>
-#include <cstdio>
 #include <cstring>
 #include <limits>
 #include <random>
@@ -41,109 +40,6 @@
 #include <vector>
 
 namespace {
-
-// Caller-side helper trigger gate (mirrors the device-side gate in
-// cast_string_to_float.cu): `slow_path_eligible = (digits > 2^53) AND
-// (|exp_ten| <= 19)`. Re-computed on the host from the same input strings so
-// each benchmark axis can print a verifiable trigger rate before timing,
-// instead of inferring it from input shape.
-struct parsed_decimal {
-  uint64_t digits;
-  int exp_ten;
-  bool parse_ok;
-};
-
-// Lightweight host-side decimal parser sufficient to evaluate the
-// `slow_path_eligible` predicate for our benchmark inputs (sign / leading
-// whitespace / NaN / Inf / hex etc. are not relevant for the inputs we
-// construct here and are not handled — that matches what the device parser
-// peels off before reaching the slow-path gate).
-parsed_decimal parse_for_trigger_gate(std::string const& s)
-{
-  parsed_decimal out{0, 0, false};
-  size_t i           = 0;
-  bool seen_dot      = false;
-  int frac_digits    = 0;
-  bool any_digit     = false;
-  __uint128_t digits = 0;
-
-  if (i < s.size() && (s[i] == '+' || s[i] == '-')) { ++i; }
-
-  for (; i < s.size(); ++i) {
-    char c = s[i];
-    if (c == '.') {
-      if (seen_dot) { return out; }
-      seen_dot = true;
-      continue;
-    }
-    if (c == 'e' || c == 'E') { break; }
-    if (c < '0' || c > '9') { return out; }
-    digits = digits * 10 + static_cast<uint64_t>(c - '0');
-    // Reject inputs whose accumulator no longer fits in uint64_t. `digits` is
-    // __uint128_t so this is a direct high-half check rather than a
-    // cross-width comparison.
-    if ((digits >> 64) != 0) { return out; }
-    any_digit = true;
-    if (seen_dot) { ++frac_digits; }
-  }
-  if (!any_digit) { return out; }
-
-  int manual_exp = 0;
-  if (i < s.size() && (s[i] == 'e' || s[i] == 'E')) {
-    ++i;
-    bool exp_neg = false;
-    if (i < s.size() && (s[i] == '+' || s[i] == '-')) {
-      exp_neg = (s[i] == '-');
-      ++i;
-    }
-    if (i >= s.size()) { return out; }
-    int e = 0;
-    for (; i < s.size(); ++i) {
-      char c = s[i];
-      if (c < '0' || c > '9') { return out; }
-      e = e * 10 + (c - '0');
-    }
-    manual_exp = exp_neg ? -e : e;
-  }
-
-  out.digits   = static_cast<uint64_t>(digits);
-  out.exp_ten  = manual_exp - frac_digits;
-  out.parse_ok = true;
-  return out;
-}
-
-// Print the share of rows that satisfy the helper's caller-side gate.
-// Run once per axis right before timing so the trigger rate is part of the
-// benchmark transcript and the perf numbers cannot be misread. `sample_cap`
-// is the maximum number of rows to inspect (the host parser is O(N) and we
-// don't want to pay it for 100M-row axes).
-void print_trigger_rate(std::vector<std::string> const& strings,
-                        char const* tag,
-                        size_t sample_cap = 1'000'000)
-{
-  size_t const inspect = std::min(strings.size(), sample_cap);
-  size_t triggered     = 0;
-  size_t parse_err     = 0;
-  for (size_t i = 0; i < inspect; ++i) {
-    auto const p = parse_for_trigger_gate(strings[i]);
-    if (!p.parse_ok) {
-      ++parse_err;
-      continue;
-    }
-    int const abs_q = p.exp_ten < 0 ? -p.exp_ten : p.exp_ten;
-    if (p.digits > (1ULL << 53) && abs_q <= 19) { ++triggered; }
-  }
-  double const pct =
-    inspect == 0 ? 0.0 : 100.0 * static_cast<double>(triggered) / static_cast<double>(inspect);
-  double const errpct =
-    inspect == 0 ? 0.0 : 100.0 * static_cast<double>(parse_err) / static_cast<double>(inspect);
-  std::printf("[helper-trigger] %s: %zu/%zu inspected = %.2f%% triggered (parse-err: %.2f%%)\n",
-              tag,
-              triggered,
-              inspect,
-              pct,
-              errpct);
-}
 
 // Construct host strings that all satisfy the helper's caller-side gate:
 // `digits > 2^53 AND |exp_ten| <= 19`. Random uint64 in [2^53+1, UINT64_MAX]
@@ -259,23 +155,17 @@ NVBENCH_BENCH(string_to_float)
 // produced by `cudf::strings::from_floats`, which caps significant digits at
 // 10 via `nine_digits = 10^9` in `dissect_value` (cudf
 // `convert_floats.cu:134`), so the parsed mantissa is at most ~10^10 — well
-// below 2^53 (~9.007e15). The caller's `slow_path_eligible` is `false` for
-// every row and the legacy `digits * exp10(exp_ten)` path runs. The trigger
-// rate is not host-verified here (avoids reading the device output back to
-// the host); instead it is asserted statically by `from_floats`'s contract.
-// This is the regression-check baseline: cost on the FP64 path WITHOUT the
-// new helper.
+// below 2^53 (~9.007e15). The caller's `use_high_precision_path` is `false` for
+// every row and the default `digits * exp10(exp_ten)` path runs. The trigger
+// rate is asserted statically by `from_floats`'s contract (no host-side
+// re-parse needed). This is the regression-check baseline: cost on the FP64
+// path WITHOUT the new helper.
 void string_to_double_helper_off(nvbench::state& state)
 {
   cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
   auto const float_tbl  = create_random_table({cudf::type_id::FLOAT64}, row_count{n_rows});
   auto const float_col  = float_tbl->get_column(0);
   auto const string_col = cudf::strings::from_floats(float_col.view());
-
-  std::printf(
-    "[helper-trigger] string_to_double_helper_off: 0/%d rows triggered "
-    "(by construction — from_floats caps digits at ~10^10 < 2^53)\n",
-    static_cast<int>(n_rows));
 
   state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
     auto rows = spark_rapids_jni::string_to_float(cudf::data_type{cudf::type_id::FLOAT64},
@@ -292,17 +182,14 @@ NVBENCH_BENCH(string_to_double_helper_off)
 // FP64 cast where the correctly-rounded helper IS entered for every row.
 // Strings are hand-built on the host so the parsed mantissa is always
 // `> 2^53` and the explicit exponent is always in `[-19, 19]`, guaranteeing
-// the caller's `slow_path_eligible` predicate is `true` for every row. A
-// host-side re-parse over a sample prints the actual trigger rate before
-// timing. This measures the worst case: every row pays the int128 helper
-// cost.
+// the caller's `use_high_precision_path` predicate is `true` for every row.
+// This measures the worst case: every row pays the int128 helper cost.
 void string_to_double_helper_on(nvbench::state& state)
 {
   cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
   auto const host_strings = make_helper_triggering_strings(n_rows);
-  print_trigger_rate(host_strings, "string_to_double_helper_on");
-  auto const stream     = cudf::get_default_stream();
-  auto const string_col = string_column_from_host(host_strings, stream);
+  auto const stream       = cudf::get_default_stream();
+  auto const string_col   = string_column_from_host(host_strings, stream);
 
   state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
     auto rows = spark_rapids_jni::string_to_float(
--- a/src/main/cpp/benchmarks/cast_string_to_float.cpp
+++ b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -42,14 +42,15 @@
 namespace {
 
 // Construct host strings that all satisfy the helper's caller-side gate:
-// `digits > 2^53 AND |exp_ten| <= 19`. Random uint64 in [2^53+1, UINT64_MAX]
-// gets printed via std::to_string, then an optional `e{q}` with
-// q in [-19, 19] is appended. By construction every row exercises the
-// correctly-rounded helper.
+// `digits > 2^53 AND |exp_ten| <= 19`. The mantissa is capped at 19 digits
+// (max 9999999999999999999) so the parser's `max_safe_digits == 19` truncation
+// is never triggered — otherwise a 20-digit uint64 would lose its trailing
+// digit, bump `exp_base` by 1, and push `exp_ten` to 20, missing the gate.
 std::vector<std::string> make_helper_triggering_strings(cudf::size_type n_rows)
 {
   std::mt19937_64 rng{0xC0FFEEULL};
-  std::uniform_int_distribution<uint64_t> digit_dist{(1ULL << 53) + 1, UINT64_MAX};
+  // Inclusive upper bound = 10^19 - 1 (the largest 19-digit value).
+  std::uniform_int_distribution<uint64_t> digit_dist{(1ULL << 53) + 1, 9999999999999999999ULL};
   std::uniform_int_distribution<int> q_dist{-19, 19};
 
   std::vector<std::string> out;
@@ -159,11 +160,15 @@ NVBENCH_BENCH(string_to_float)
 // every row and the default `digits * exp10(exp_ten)` path runs. The trigger
 // rate is asserted statically by `from_floats`'s contract (no host-side
 // re-parse needed). This is the regression-check baseline: cost on the FP64
-// path WITHOUT the new helper.
+// path WITHOUT the new helper. Input column is non-nullable to match
+// helper_on's column nullability and keep the comparison apples-to-apples
+// (the default `create_random_table` profile injects nulls that would early
+// out of the parser before the default path runs).
 void string_to_double_helper_off(nvbench::state& state)
 {
   cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
-  auto const float_tbl  = create_random_table({cudf::type_id::FLOAT64}, row_count{n_rows});
+  data_profile const profile = data_profile_builder().no_validity();
+  auto const float_tbl  = create_random_table({cudf::type_id::FLOAT64}, row_count{n_rows}, profile);
   auto const float_col  = float_tbl->get_column(0);
   auto const string_col = cudf::strings::from_floats(float_col.view());
 
@@ -188,8 +193,8 @@ void string_to_double_helper_on(nvbench::state& state)
 {
   cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
   auto const host_strings = make_helper_triggering_strings(n_rows);
-  auto const stream       = cudf::get_default_stream();
-  auto const string_col   = string_column_from_host(host_strings, stream);
+  auto const stream     = cudf::get_default_stream();
+  auto const string_col = string_column_from_host(host_strings, stream);
 
   state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
     auto rows = spark_rapids_jni::string_to_float(
--- a/src/main/cpp/benchmarks/cast_string_to_float.cpp
+++ b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -178,7 +178,9 @@ packed_host_strings pack_host_strings(std::vector<std::string> const& strings)
   out.offsets.reserve(strings.size() + 1);
   out.offsets.push_back(0);
   size_t total = 0;
-  for (auto const& s : strings) { total += s.size(); }
+  for (auto const& s : strings) {
+    total += s.size();
+  }
   out.chars.resize(total);
   size_t pos = 0;
   for (auto const& s : strings) {
@@ -193,7 +195,7 @@ packed_host_strings pack_host_strings(std::vector<std::string> const& strings)
 // pulling in cudf::cudftestutil (which isn't linked into benchmarks unless
 // BUILD_CUDF_TESTS=ON).
 std::unique_ptr<cudf::column> string_column_from_host(std::vector<std::string> const& strings,
-                                                     rmm::cuda_stream_view stream)
+                                                      rmm::cuda_stream_view stream)
 {
   auto const n = static_cast<cudf::size_type>(strings.size());
   if (n == 0) { return cudf::make_empty_column(cudf::type_id::STRING); }
@@ -214,8 +216,7 @@ std::unique_ptr<cudf::column> string_column_from_host(std::vector<std::string> c
                   stream.value());
   stream.synchronize();
 
-  auto offsets_col =
-    std::make_unique<cudf::column>(std::move(d_offsets), rmm::device_buffer{}, 0);
+  auto offsets_col = std::make_unique<cudf::column>(std::move(d_offsets), rmm::device_buffer{}, 0);
   return cudf::make_strings_column(
     n, std::move(offsets_col), d_chars.release(), 0, rmm::device_buffer{});
 }
@@ -261,9 +262,10 @@ void string_to_double_helper_off(nvbench::state& state)
   auto const float_col  = float_tbl->get_column(0);
   auto const string_col = cudf::strings::from_floats(float_col.view());
 
-  std::printf("[helper-trigger] string_to_double_helper_off: 0/%d rows triggered "
-              "(by construction — from_floats caps digits at ~10^10 < 2^53)\n",
-              static_cast<int>(n_rows));
+  std::printf(
+    "[helper-trigger] string_to_double_helper_off: 0/%d rows triggered "
+    "(by construction — from_floats caps digits at ~10^10 < 2^53)\n",
+    static_cast<int>(n_rows));
 
   state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
     auto rows = spark_rapids_jni::string_to_float(cudf::data_type{cudf::type_id::FLOAT64},
--- a/src/main/cpp/benchmarks/cast_string_to_float.cpp
+++ b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -16,24 +16,16 @@
 
 #include <benchmarks/common/generate_input.hpp>
 
-#include <cudf/column/column.hpp>
-#include <cudf/column/column_factories.hpp>
+#include <cudf_test/column_wrapper.hpp>
+
 #include <cudf/strings/convert/convert_floats.hpp>
 #include <cudf/strings/strings_column_view.hpp>
 #include <cudf/utilities/default_stream.hpp>
-#include <cudf/utilities/error.hpp>
-
-#include <rmm/cuda_stream_view.hpp>
-#include <rmm/device_buffer.hpp>
-#include <rmm/device_uvector.hpp>
 
 #include <cast_string.hpp>
 #include <nvbench/nvbench.cuh>
 
-#include <cassert>
 #include <cstdint>
-#include <cstring>
-#include <limits>
 #include <random>
 #include <string>
 #include <utility>
@@ -66,66 +58,6 @@ std::vector<std::string> make_helper_triggering_strings(cudf::size_type n_rows)
     out.push_back(std::move(s));
   }
   return out;
-}
-
-// Pack a host vector of strings into a single contiguous chars buffer plus
-// matching offsets. Used to upload to device for cudf::make_strings_column.
-struct packed_host_strings {
-  std::vector<char> chars;
-  std::vector<int32_t> offsets;
-};
-
-packed_host_strings pack_host_strings(std::vector<std::string> const& strings)
-{
-  packed_host_strings out;
-  out.offsets.reserve(strings.size() + 1);
-  out.offsets.push_back(0);
-  size_t total = 0;
-  for (auto const& s : strings) {
-    total += s.size();
-  }
-  out.chars.resize(total);
-  size_t pos = 0;
-  for (auto const& s : strings) {
-    std::memcpy(out.chars.data() + pos, s.data(), s.size());
-    pos += s.size();
-    // Guard the int32 cast — cudf strings columns currently use int32 offsets.
-    // If a future change bumps `num_rows` or string length past INT32_MAX,
-    // fail loudly here instead of silently truncating to a negative offset.
-    assert(pos <= static_cast<size_t>(std::numeric_limits<int32_t>::max()));
-    out.offsets.push_back(static_cast<int32_t>(pos));
-  }
-  return out;
-}
-
-// Build a non-nullable cudf STRING column from a host string vector without
-// pulling in cudf::cudftestutil (which isn't linked into benchmarks unless
-// BUILD_CUDF_TESTS=ON).
-std::unique_ptr<cudf::column> string_column_from_host(std::vector<std::string> const& strings,
-                                                      rmm::cuda_stream_view stream)
-{
-  auto const n = static_cast<cudf::size_type>(strings.size());
-  if (n == 0) { return cudf::make_empty_column(cudf::type_id::STRING); }
-  auto packed = pack_host_strings(strings);
-
-  auto const mr = cudf::get_current_device_resource_ref();
-  rmm::device_uvector<char> d_chars(packed.chars.size(), stream, mr);
-  rmm::device_uvector<int32_t> d_offsets(packed.offsets.size(), stream, mr);
-  CUDF_CUDA_TRY(cudaMemcpyAsync(d_chars.data(),
-                                packed.chars.data(),
-                                packed.chars.size(),
-                                cudaMemcpyHostToDevice,
-                                stream.value()));
-  CUDF_CUDA_TRY(cudaMemcpyAsync(d_offsets.data(),
-                                packed.offsets.data(),
-                                packed.offsets.size() * sizeof(int32_t),
-                                cudaMemcpyHostToDevice,
-                                stream.value()));
-  stream.synchronize();
-
-  auto offsets_col = std::make_unique<cudf::column>(std::move(d_offsets), rmm::device_buffer{}, 0);
-  return cudf::make_strings_column(
-    n, std::move(offsets_col), d_chars.release(), 0, rmm::device_buffer{});
 }
 
 }  // namespace
@@ -193,12 +125,13 @@ void string_to_double_helper_on(nvbench::state& state)
 {
   cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
   auto const host_strings = make_helper_triggering_strings(n_rows);
-  auto const stream       = cudf::get_default_stream();
-  auto const string_col   = string_column_from_host(host_strings, stream);
+  cudf::test::strings_column_wrapper string_col(host_strings.begin(), host_strings.end());
+  cudf::strings_column_view const scv{static_cast<cudf::column_view>(string_col)};
+  auto const stream = cudf::get_default_stream();
 
   state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
     auto rows = spark_rapids_jni::string_to_float(
-      cudf::data_type{cudf::type_id::FLOAT64}, string_col->view(), false, stream);
+      cudf::data_type{cudf::type_id::FLOAT64}, scv, false, stream);
   });
 }
 
--- a/src/main/cpp/benchmarks/cast_string_to_float.cpp
+++ b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -63,8 +63,8 @@ std::vector<std::string> make_helper_triggering_strings(cudf::size_type n_rows)
 }  // namespace
 
 // FP32 cast — the correctly-rounded helper does not run on this code path
-// (the `if constexpr (is_same_v<T, double>)` gate skips it for float). Kept
-// as the original FP32 baseline.
+// (the `cuda::std::is_same_v<T, double>` term in the `helper_eligible` gate
+// skips it for float). Kept as the original FP32 baseline.
 void string_to_float(nvbench::state& state)
 {
   cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
@@ -88,14 +88,14 @@ NVBENCH_BENCH(string_to_float)
 // produced by `cudf::strings::from_floats`, which caps significant digits at
 // 10 via `nine_digits = 10^9` in `dissect_value` (cudf
 // `convert_floats.cu:134`), so the parsed mantissa is at most ~10^10 — well
-// below 2^53 (~9.007e15). The caller's `use_high_precision_path` is `false` for
-// every row and the default `digits * exp10(exp_ten)` path runs. The trigger
-// rate is asserted statically by `from_floats`'s contract (no host-side
-// re-parse needed). This is the regression-check baseline: cost on the FP64
-// path WITHOUT the new helper. Input column is non-nullable to match
-// helper_on's column nullability and keep the comparison apples-to-apples
-// (the default `create_random_table` profile injects nulls that would early
-// out of the parser before the default path runs).
+// below 2^53 (~9.007e15). The caller's `helper_eligible` predicate is `false`
+// for every row and `default_double_path<T>` runs. The trigger rate is
+// asserted statically by `from_floats`'s contract (no host-side re-parse
+// needed). This is the regression-check baseline: cost on the FP64 path
+// WITHOUT the new helper. Input column is non-nullable to match helper_on's
+// column nullability and keep the comparison apples-to-apples (the default
+// `create_random_table` profile injects nulls that would early out of the
+// parser before the default path runs).
 void string_to_double_helper_off(nvbench::state& state)
 {
   cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
@@ -119,7 +119,7 @@ NVBENCH_BENCH(string_to_double_helper_off)
 // FP64 cast where the correctly-rounded helper IS entered for every row.
 // Strings are hand-built on the host so the parsed mantissa is always
 // `> 2^53` and the explicit exponent is always in `[-19, 19]`, guaranteeing
-// the caller's `use_high_precision_path` predicate is `true` for every row.
+// the caller's `helper_eligible` predicate is `true` for every row.
 // This measures the worst case: every row pays the int128 helper cost.
 void string_to_double_helper_on(nvbench::state& state)
 {
--- a/src/main/cpp/benchmarks/cast_string_to_float.cpp
+++ b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -193,8 +193,8 @@ void string_to_double_helper_on(nvbench::state& state)
 {
   cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
   auto const host_strings = make_helper_triggering_strings(n_rows);
-  auto const stream     = cudf::get_default_stream();
-  auto const string_col = string_column_from_host(host_strings, stream);
+  auto const stream       = cudf::get_default_stream();
+  auto const string_col   = string_column_from_host(host_strings, stream);
 
   state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
     auto rows = spark_rapids_jni::string_to_float(
--- a/src/main/cpp/benchmarks/cast_string_to_float.cpp
+++ b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2024, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2026, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +41,148 @@ namespace detail {
...
+                               10000000000ULL,
+                               100000000000ULL,
+                               1000000000000ULL,
+                               10000000000000ULL,
+                               100000000000000ULL,
+                               1000000000000000ULL,
+                               10000000000000000ULL,
+                               100000000000000000ULL,
+                               1000000000000000000ULL,
+                               10000000000000000000ULL};
```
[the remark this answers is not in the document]
wjxiz1992: Done in 412b951a — `kPow10` is now `static constexpr`, so the table lives in constant memory rather than on the kernel's local stack per call.

On the `_ipow` table reuse suggestion: I considered it, but `_ipow` is a per-block shared-memory copy that the kernel's existing flow populates on warp-lane-0 inside `string_to_float_kernel`, then reads under `__syncthreads()`. The helper sits inside the per-thread `operator()` and doesn't take an `ipow` pointer, so reusing it would mean adding a parameter and assuming the existing sync ordering holds. The `static constexpr` route gets the same constant-memory residency with no API/sync changes, so I went with that.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +41,148 @@ namespace detail {
 
 __device__ __inline__ bool is_digit(char c) { return c >= '0' && c <= '9'; }
 
+/**
+ * @brief Correctly-rounded conversion of `digits * 10^q` to a double for the
+ *        case where `digits` exceeds the safe-cast range of double (2^53) and
+ *        the existing `static_cast<double>(digits) * exp10(q)` path therefore
+ *        loses up to 1 ULP. See NVIDIA/spark-rapids#10773.
+ *
+ *        The fast path (digits <= 2^53 AND |q| <= 22) is still handled by the
```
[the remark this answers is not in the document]
wjxiz1992: Done in 412b951a. The docstring now states the actual trigger condition (`digits > 2^53 AND |q| <= 19`) and explains why 19 is the cap (10^19 is the largest power of ten that fits in `uint64_t`, which the 128-bit arithmetic relies on).

Note: I checked the other two locations you flagged (lines 143 and 166 of the prior version). They were plain code lines (no docstring text), so there was nothing to update there — only the line-50 doc had the stale `<= 22` claim.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -215,6 +215,38 @@ void castToFloatsInfTest(){
...
+        "1.0000000000000001",
+        "1.0000000000000002",
+        "3.1415926535897932",
+        "2.7182818284590452",
+        "2.2250738585072014",        // ~Double.MIN_NORMAL
+        "1.234567890123456789",      // 19-digit mantissa
+        "9.999999999999999999",
+        "-1.7976931348623157",
+        "1.5",
+        "0.1"
```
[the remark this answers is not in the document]
wjxiz1992: Extended in 412b951a — `castToDoubleHighPrecisionTest` now covers:

- **2^53 trigger boundary**: `9007199254740992` (==, fast path), `9007199254740993` (helper's first trigger), `9.007199254740993` (boundary with q<0)
- **Mantissa overflow rounds into the exponent**: `9999999999999999.5` rounds to `1e16`, exercising the `rounded >= (1ULL << 53)` branch that bumps `unbiased_exp`
- **Overflow to ±infinity**: `1.8e308`, `-1.8e308` (just past `Double.MAX_VALUE`)
- **Subnormal magnitudes**: `5e-324` (`Double.MIN_VALUE`), `1.5e-310` (mid-subnormal)
- **Helper-returns-NaN fallback path** (|q| > 19, caller falls back to legacy): `1e20`, `1e-100`, `9.99e25` — digits chosen to stay under 2^53 so the legacy path is also correctly rounded and the test still passes

All 22 inputs match `Double.parseDouble` bit-for-bit on real GPU (sm_89). Output: `Tests run: 1, Failures: 0, Errors: 0, Skipped: 0`.

wjxiz1992:
build

thirtiseven:
Could we get a quick result from the microbenchmark at  https://github.com/NVIDIA/spark-rapids-jni/blob/main/src/main/cpp/benchmarks/cast_string_to_float.cpp to assess the performance impact? Also see https://github.com/NVIDIA/spark-rapids-jni/issues/4559#issuecomment-4457063846.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +41,148 @@ namespace detail {
...
+    bool const sub_round_bit = ((rounded >> (subnormal_shift - 1)) & 1ULL) != 0;
+    uint64_t const sub_low_mask =
+      subnormal_shift > 1 ? ((1ULL << (subnormal_shift - 1)) - 1ULL) : 0ULL;
+    bool const sub_sticky = ((rounded & sub_low_mask) != 0) || sticky;
+    uint64_t sub_mantissa = rounded >> subnormal_shift;
+    if (sub_round_bit && (sub_sticky || (sub_mantissa & 1ULL))) { sub_mantissa += 1; }
+    bits = sub_mantissa & ((1ULL << 52) - 1);
+    if (sign < 0) { bits |= (1ULL << 63); }
+    return __longlong_as_double(static_cast<long long>(bits));
+  }
```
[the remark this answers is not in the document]
wjxiz1992: Done in f4ddbdac — removed the unreachable subnormal branch. The caller's `digits > 2^53` guard makes `msb_pos >= 53` in every reachable path, so combined with `binary_scale` in `{0, -64}` the smallest reachable `unbiased_exp` is `53 - 64 = -11`, giving `biased_exp >= 1012`. Replaced the dead branch with `assert(biased_exp > 0)` plus a comment that points back to the caller's precondition. That also drops the latent double-rounding bug you flagged — it can no longer be reached, so no need to fix it in dead code.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -215,6 +215,71 @@ void castToFloatsInfTest(){
...
+        "9.007199254740993",
+        // --- mantissa rounding rolls into the exponent ---
+        // After rounding, the 53-bit mantissa hits 2^53 and the helper
+        // increments unbiased_exp.
+        "9999999999999999.5",        // rounds to 1e16
+        // --- overflow to infinity ---
+        "1.8e308",                   // > Double.MAX_VALUE
+        "-1.8e308",                  // negative infinity
+        // --- subnormal magnitudes ---
+        "5e-324",                    // Double.MIN_VALUE (smallest subnormal)
```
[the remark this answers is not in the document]
wjxiz1992: Done in f4ddbdac — added three inputs with `digits > 2^53 AND |exp_ten| > 19` that actually trigger the helper's NaN return and the caller's `\!cuda::std::isnan(cr)` fall-through:

```
"12345678901234567e21",   // q = +21, large magnitude
"12345678901234567e-23",  // q = -23, tiny magnitude
"99999999999999999e30"    // q = +30, near overflow-to-infinity
```

`12345678901234567 ≈ 1.234e16 > 2^53 (9.007e15)`, so `slow_path_eligible` is true and `correctly_rounded_uint64_times_pow10` is called and returns NaN. Kept the original `1e20` / `1e-100` / `9.99e25` cases but relabeled their comment — those still exercise the `slow_path_eligible=false` branch in the caller, which is also worth covering.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +41,148 @@ namespace detail {
...
+  // initialized on the kernel's local stack every invocation.
+  static constexpr uint64_t kPow10[20] = {1ULL,
+                                          10ULL,
+                                          100ULL,
+                                          1000ULL,
+                                          10000ULL,
+                                          100000ULL,
+                                          1000000ULL,
+                                          10000000ULL,
+                                          100000000ULL,
```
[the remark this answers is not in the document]
wjxiz1992: Good catch — done in f4ddbdac. Reworded the comment to drop the CUDA `__constant__` claim. New wording:

```
// 10^abs_q for abs_q in [0, 19] — exact as uint64. `static constexpr` marks
// the table as a compile-time constant so the compiler is free to fold each
// load as an immediate, place it in the binary's read-only data, or keep it
// in registers — anywhere but re-initialized on the kernel's local stack on
// every invocation. (This is not CUDA `__constant__` memory; that requires a
// file-scope `__constant__` qualifier and is not applicable to a local
// table inside a `__device__` helper.)
```

wjxiz1992:
build

wjxiz1992:
@thirtiseven benchmark is in progress. The existing `benchmarks/cast_string_to_float.cpp` only covers FP32, which goes through the `if constexpr (is_same_v<T, double>)`-gated code and therefore does not exercise the new correctly-rounded path. To get a meaningful read I added two FP64 axes to the benchmark file locally:

- `string_to_double_short_mantissa` — FP32-random doubles printed with `from_floats` (~7 sig figs), `slow_path_eligible` is false in the caller so the new helper is NOT entered. Regression check.
- `string_to_double_long_mantissa` — FP64-random doubles printed with `from_floats` (15–17 sig figs), so most rows hit `digits > 2^53` and the helper IS entered. Worst-case for the new code path.

Both builds (PR head + a fresh `origin/main` baseline at the same cudf submodule SHA) are compiling locally now. I will post the FP32 / FP64-short / FP64-long throughput numbers in this thread once both runs finish, together with the benchmark patch so it can be reproduced or upstreamed.

wjxiz1992:
@thirtiseven benchmark results below. **Net: no measurable perf regression — the correctly-rounded helper costs <1% of total kernel time even when 100% of rows trigger it.**

## Setup

- GPU: NVIDIA RTX 5880 Ada Generation (SM 89), CUDA 13.0, 110 SMs, 48 GiB.
- Built with `BUILD_TESTS=ON BUILD_BENCHMARKS=ON CPP_PARALLEL_LEVEL=8`, default `CMAKE_CUDA_ARCHITECTURES=RAPIDS` (sm 75/80/86/90a/100f/120a/120).
- Three nvbench cases at 1M and 100M rows each (patch attached at the bottom of this comment so anyone can drop it in and re-measure):

| Case | Input | Helper triggered? |
|---|---|---|
| FP32 baseline | random FP32 → `from_floats` (~7 sig figs), parsed as FP32 | n/a (gated by `is_same_v<T, double>`) |
| FP64 short mantissa | random FP32 → `from_floats` (~7 sig figs), parsed as FP64 | NO — `slow_path_eligible=false` in caller (digits ≤ 2^53) |
| FP64 long mantissa | random FP64 → `from_floats` (15–17 sig figs), parsed as FP64 | YES — most rows hit `digits > 2^53` |

## Results

### 100M rows (low noise: 0.03–0.08 %)

| Case | GPU Time | Δ vs FP32 | Δ vs FP64-short |
|---|---:|---:|---:|
| FP32 baseline                                                | 161.901 ms | – | – |
| FP64 short mantissa (helper NOT triggered)                   | 159.121 ms | **−1.7 %** | – |
| FP64 long mantissa (helper IS triggered on most rows)        | 160.365 ms | **−0.9 %** | **+0.8 %** |

### 1M rows (higher noise: 2.3–4.8 %)

| Case | GPU Time | Δ vs FP32 | Δ vs FP64-short |
|---|---:|---:|---:|
| FP32 baseline               | 1.743 ms | – | – |
| FP64 short mantissa         | 1.753 ms | +0.6 % | – |
| FP64 long mantissa          | 1.763 ms | +1.1 % | +0.6 % |

## Reading

- **FP32 path is unchanged.** The new code is wrapped in `if constexpr (is_same_v<T, double>)`, so the FP32 kernel codegen is bit-identical to `main`. Confirmed by the FP32 row.
- **FP64 short mantissa ≈ FP32.** The FP64 wrapper adds ~0% overhead when the helper is not entered (it short-circuits on `digits ≤ 2^53` before doing any of the 128-bit work). This is the **regression-check** row and it passes.
- **FP64 long mantissa shows ≤1.2 % overhead vs FP64 short.** That is the actual cost of the int128 helper, measured at the worst case where *every* row triggers it. In real workloads the trigger rate is much lower (Spark’s `Double.parseDouble` is rare on user data with 17+ digit mantissas), so the realistic cost is well below 1 %.

The 128-bit branch-divergence concern you raised in #4559 is real in theory, but in practice the helper is dwarfed by the rest of the parsing pipeline (digit accumulation, exponent parsing, whitespace, etc.). So this PR doesn’t cause the big perf regression — Eisel-Lemire is still worth exploring later as a broader rewrite, but it's independent of merging this fix.

## Reproduce

Patch to `src/main/cpp/benchmarks/cast_string_to_float.cpp` used for these numbers (kept out of the merge to keep this PR focused on the fix; happy to spin it into a follow-up PR if useful):

```diff
diff --git a/src/main/cpp/benchmarks/cast_string_to_float.cpp b/src/main/cpp/benchmarks/cast_string_to_float.cpp
@@ -42,3 +42,51 @@ void string_to_float(nvbench::state& state)
 NVBENCH_BENCH(string_to_float)
   .set_name("Strings to Float Cast")
   .add_int64_axis("num_rows", {1 * 1024 * 1024, 100 * 1024 * 1024});
+
+// FP64 path, "short mantissa" inputs (digits <= 2^53). The helper added in
+// #4560 is NOT entered for these — the caller short-circuits on
+// digits <= 2^53. FP32-precision random doubles formatted with from_floats
+// keep mantissa width at ~7 sig figs (well below 2^53).
+void string_to_double_short_mantissa(nvbench::state& state)
+{
+  cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
+  auto const float_tbl  = create_random_table({cudf::type_id::FLOAT32}, row_count{n_rows});
+  auto const float_col  = float_tbl->get_column(0);
+  auto const string_col = cudf::strings::from_floats(float_col.view());
+
+  state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
+    auto rows = spark_rapids_jni::string_to_float(cudf::data_type{cudf::type_id::FLOAT64},
+                                                  string_col->view(),
+                                                  false,
+                                                  cudf::get_default_stream());
+  });
+}
+
+NVBENCH_BENCH(string_to_double_short_mantissa)
+  .set_name("Strings to Double Cast (short mantissa, fallback NOT triggered)")
+  .add_int64_axis("num_rows", {1 * 1024 * 1024, 100 * 1024 * 1024});
+
+// FP64 path, "long mantissa" inputs (digits > 2^53 in practice). Random FP64
+// values printed with from_floats produce 15-17 sig figs, so most rows hit
+// digits > 2^53 and the new correctly-rounded helper IS invoked.
+void string_to_double_long_mantissa(nvbench::state& state)
+{
+  cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
+  auto const float_tbl  = create_random_table({cudf::type_id::FLOAT64}, row_count{n_rows});
+  auto const float_col  = float_tbl->get_column(0);
+  auto const string_col = cudf::strings::from_floats(float_col.view());
+
+  state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
+    auto rows = spark_rapids_jni::string_to_float(cudf::data_type{cudf::type_id::FLOAT64},
+                                                  string_col->view(),
+                                                  false,
+                                                  cudf::get_default_stream());
+  });
+}
+
+NVBENCH_BENCH(string_to_double_long_mantissa)
+  .set_name("Strings to Double Cast (long mantissa, correctly-rounded fallback)")
+  .add_int64_axis("num_rows", {1 * 1024 * 1024, 100 * 1024 * 1024});
```

nvauto:

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -215,6 +215,81 @@ void castToFloatsInfTest(){
...
+        "1.5e-310",                  // mid-subnormal range
+        // --- |q| > 19, helper NOT triggered because digits <= 2^53 ---
+        // slow_path_eligible is already false in the caller, so the helper
+        // is never called for these. They keep the legacy `digits * exp10(q)`
+        // path honest (both operands exact, so already correctly rounded).
+        "1e20",
+        "1e-100",
+        "9.99e25",
+        // --- |q| > 19 AND digits > 2^53 -> helper IS called and returns NaN ---
```
thirtiseven: This comment does not look correct. I don't think the helper is called as it's guarded by:
```
bool const slow_path_eligible = (digits > (1ULL << 53)) && (cuda::std::abs(exp_ten) <= 19);
```

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,145 @@ namespace detail {
...
+ *        largest power of ten that fits in `uint64_t` (10^19), which the
+ *        128-bit arithmetic below relies on.
+ */
+__device__ __inline__ double correctly_rounded_uint64_times_pow10(uint64_t digits, int q, int sign)
+{
+  if (digits == 0) { return sign >= 0 ? 0.0 : -0.0; }
+
+  int const abs_q = q < 0 ? -q : q;
+  if (abs_q > 19) {
+    // Out of the helper's range; caller falls back.
```
thirtiseven: It looks unreachable in the current code. Should we either update the guard condition to handle the fallback here, or assert that it's unreachable and raise an exception?

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -215,6 +215,81 @@ void castToFloatsInfTest(){
...
+  void castToDoubleHighPrecisionTest() {
+    String[] inputs = new String[] {
+        // --- happy path (helper) ---
+        "1.7976931348623157",        // the literal that broke RapidsJsonSuite
+        "9.9999999999999999",
+        "1.0000000000000001",
+        "1.0000000000000002",
+        "3.1415926535897932",
+        "2.7182818284590452",
+        "2.2250738585072014",        // ~Double.MIN_NORMAL
```
thirtiseven: Is this really `Double.MIN_NORMAL`?

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,145 @@ namespace detail {
...
+  if (round_bit && (sticky || (mantissa & 1ULL))) { rounded += 1; }
+
+  int unbiased_exp = msb_pos + binary_scale;
+  if (rounded >= (1ULL << 53)) {
+    rounded >>= 1;
+    unbiased_exp += 1;
+  }
+
+  int const biased_exp = unbiased_exp + 1023;
+  if (biased_exp >= 0x7FF) {
```
thirtiseven: This branch looks to be dead code under the helper's eligibility preconditions. 

From AI:
The maximum reachable unbiased_exp in the helper is msb_pos + binary_scale. Since `msb_pos <= 127` (128-bit value) and binary_scale in $\\{0, -64\\}$, the largest unbiased_exp is 127, giving `biased_exp = 127 + 1023 = 1150`, which is well below `0x7FF = 2047`. The test comment at CastStringsTest.java:261-262 describes "1.8e308" and "-1.8e308" as overflow cases — these are tested but go through the legacy path (`digits = 18 <= 2^53`), not the helper. Either the dead branch should be removed with a comment explaining why overflow is impossible here, or an assert(`biased_exp < 0x7FF`) added to document the invariant, consistent with how the subnormal branch was handled (assert at line 177).

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,145 @@ namespace detail {
...
+  bool sticky;
+
+  if (shift > 0) {
+    // Shift in [1, 127]. All bit ops use __uint128_t to avoid undefined
+    // 64-bit shifts at the shift==64 boundary.
+    mantissa                   = static_cast<uint64_t>(quotient128 >> shift);
+    round_bit                  = (static_cast<uint64_t>(quotient128 >> (shift - 1)) & 1ULL) != 0;
+    __uint128_t const low_mask = (static_cast<__uint128_t>(1) << (shift - 1)) - 1;
+    sticky                     = ((quotient128 & low_mask) != 0) || (division_rem != 0);
+  } else {
```
thirtiseven: Is it also dead code? ditto.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,145 @@ namespace detail {
...
+ *        multiplier — for digits <= 2^53 that path is already correctly
+ *        rounded (both operands are exact as doubles), and for the rare
+ *        |q| > 19 case the helper returns NaN to signal the caller to fall
+ *        back to the old (less accurate) path. The 19 cap matches the
+ *        largest power of ten that fits in `uint64_t` (10^19), which the
+ *        128-bit arithmetic below relies on.
+ */
+__device__ __inline__ double correctly_rounded_uint64_times_pow10(uint64_t digits, int q, int sign)
+{
+  if (digits == 0) { return sign >= 0 ? 0.0 : -0.0; }
```
thirtiseven: Is it also dead code? ditto.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,145 @@ namespace detail {
...
+  }
+
+  // 10^abs_q for abs_q in [0, 19] — exact as uint64. `static constexpr` marks
+  // the table as a compile-time constant so the compiler is free to fold each
+  // load as an immediate, place it in the binary's read-only data, or keep it
+  // in registers — anywhere but re-initialized on the kernel's local stack on
+  // every invocation. (This is not CUDA `__constant__` memory; that requires a
+  // file-scope `__constant__` qualifier and is not applicable to a local
+  // table inside a `__device__` helper.)
+  static constexpr uint64_t kPow10[20] = {1ULL,
```
thirtiseven: Please use snake_case for `kPow10`.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,145 @@ namespace detail {
...
+  //   q == 0: quotient128 = digits, binary_scale = 0.
+  __uint128_t quotient128;
+  uint64_t division_rem;
+  int binary_scale;
+  if (q < 0) {
+    __uint128_t const numerator = static_cast<__uint128_t>(digits) << 64;
+    quotient128                 = numerator / pow10_abs_q;
+    division_rem                = static_cast<uint64_t>(numerator % pow10_abs_q);
+    binary_scale                = -64;
+  } else if (q > 0) {
```
thirtiseven: AI told me that all the test inputs going into the helper have q ≤ 0, so this branch is not covered by any tests. Nice to have one if that's true.

thirtiseven (commented):
Some comments on testing and style, looks good in general.

> @thirtiseven benchmark results below. **Net: no measurable perf regression — the correctly-rounded helper costs <1% of total kernel time even when 100% of rows trigger it.**

That's great. Could you submit the benchmark code since you've already written it?

thirtiseven:
Oh, sorry, wait! It seems that there are similar issues with the benchmarks, so the helper isn't actually being tested. Following from AI:

> Thanks for the numbers, but I think the benchmark doesn't actually exercise the helper — both the short-mantissa and long-mantissa cases short-circuit before reaching it.
> 
> `cudf::strings::from_floats` is capped at 10 significant digits (convert_floats.cu:134):
> 
> ```
> static constexpr unsigned int significant_digits = 10;
> ```
> 
> `dissect_value` stores the integer and decimal parts in `unsigned int` with `nine_digits = 10^9` as the max, so
>   the output is at most 10 sig figs regardless of FP32 vs FP64 input. The "Random FP64 values printed with
>   from_floats produce 15-17 sig figs" comment in the patch isn't what cudf's implementation does (that's
> what Ryu in `ftos_converter.cuh` would do, but the bench calls `cudf::strings::from_floats`).
> 
> Consequence:
> - Max `digits` from `from_floats` is ~10^10
> - Caller's gate is `digits > 2^53 ≈ 9.007e15`
> - 10^10 ≪ 2^53, so `slow_path_eligible` is `false` for every row in all three cases
> - The helper is never called
> 
> The ~0.8% gap between FP64-short and FP64-long is most likely the trailing-zero strip in `dissect_value`
> (line 219-222): FP32 upcast to double has zeros in the low decimal digits that get stripped, producing a
> slightly shorter string and slightly less parsing work. It's a string-length effect, not a helper cost.
> 
> Same root cause as the comment issue at `CastStringsTest.java:273-280`: the inputs `12345678901234567e21`
> etc. also fail `slow_path_eligible` because `|exp_ten| > 19`, so the helper isn't called there either despite
>   the comment claiming otherwise.
> 
> To actually measure the helper, the bench needs inputs with `digits > 2^53` AND `|exp_ten| ≤ 19`. Options:
> - Generate strings on the host with `std::to_chars(..., chars_format::scientific, 17)` (or `printf("%.17g",
> ...)`)
> - Use `spark_rapids_jni::ftos_converter` (Ryu, 17 digits) instead of `cudf::strings::from_floats`
> - Hand-craft strings like "1.234567890123456789", "9007199254740993", etc.
> 
> It would also help to drop an atomicAdd counter at the helper entry (or a `printf` on a small input set) to
>   confirm the trigger rate before trusting the timing. Right now the "<1% overhead" conclusion only shows
> that the cheap short-circuit path is cheap — which we already knew from the `is_same_v<T, double>` guard
> and the `digits > 2^53` check.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,145 @@ namespace detail {
...
+ *        Trigger condition (see caller): `digits > 2^53 AND |q| <= 19`.
+ *        Outside that window the caller keeps using the existing exp10
+ *        multiplier — for digits <= 2^53 that path is already correctly
+ *        rounded (both operands are exact as doubles), and for the rare
+ *        |q| > 19 case the helper returns NaN to signal the caller to fall
+ *        back to the old (less accurate) path. The 19 cap matches the
+ *        largest power of ten that fits in `uint64_t` (10^19), which the
+ *        128-bit arithmetic below relies on.
+ */
+__device__ __inline__ double correctly_rounded_uint64_times_pow10(uint64_t digits, int q, int sign)
```
thirtiseven: (perf comment from AI): `__noinline__` have a chance of being faster than `__inline__` here. It's a trade-off between call overhead and register pressure.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,145 @@ namespace detail {
...
+  //          stays correct.
+  //   q > 0: quotient128 = digits * 10^q (exact in 128 bits), binary_scale = 0.
+  //   q == 0: quotient128 = digits, binary_scale = 0.
+  __uint128_t quotient128;
+  uint64_t division_rem;
+  int binary_scale;
+  if (q < 0) {
+    __uint128_t const numerator = static_cast<__uint128_t>(digits) << 64;
+    quotient128                 = numerator / pow10_abs_q;
+    division_rem                = static_cast<uint64_t>(numerator % pow10_abs_q);
```
thirtiseven: (perf comment from AI): To save an int128 division
```suggestion
    division_rem                = static_cast<uint64_t>(numerator - quotient128 * static_cast<__uint128_t>(pow10_abs_q));
```

src/main/cpp/src/cast_string_to_float.cu
thirtiseven: It's a nit and I'm not sure about it and it depends on the benchmark results. Feel free to ignore.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
wjxiz1992: Right — my earlier comment was wrong. `slow_path_eligible = (digits > 2^53) && (|exp_ten| <= 19)`, so when `|exp_ten| > 19` the helper is never called regardless of `digits`. Fixed the comment block in 07ead3a0 — those inputs are now labeled as exercising the caller's legacy path (the `!slow_path_eligible` branch), not the helper-NaN fallback path.

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Agreed — replaced with `assert(abs_q <= 19)` plus a doc comment pointing back to the caller's precondition. 07ead3a0. Widening the guard would mean extending the 128-bit pipeline past 10^19 (which already overflows `uint64_t`), so the legacy path stays as the documented fallback for the rare `|q|>19 AND digits>2^53` inputs.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
wjxiz1992: No, you're right — that's just the mantissa digits of `Double.MIN_NORMAL` (`2.2250738585072014e-308`) without the `e-308`. The actual numeric value is ~2.225 (normal range). Dropped the MIN_NORMAL label in 07ead3a0.

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Confirmed dead under the precondition: `msb_pos <= 127`, `binary_scale in {0, -64}`, so `biased_exp <= 127 + 1023 = 1150` — well below `0x7FF`. Removed the explicit infinity branch and folded both tails into `assert(biased_exp > 0 && biased_exp < 0x7FF)`. 07ead3a0.

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Yes — `digits > 2^53` ⇒ `msb_pos(quotient128) >= 53` ⇒ `shift = msb_pos - 52 > 0` in every (q<0, q==0, q>0) branch. Removed the `shift <= 0` `else` and kept `assert(shift > 0)` with a one-line proof. 07ead3a0.

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Yes — the caller's parser already returns early for `digits == 0` (the `0 / -0` block in `operator()`), and the `digits > 2^53` gate makes it doubly unreachable. Replaced the `if (digits == 0)` early return with `assert(digits > (1ULL << 53))` so the helper's precondition is visible at the top. 07ead3a0.

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Renamed to `k_pow10` in 07ead3a0.

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Good catch — added three `q > 0` inputs to `castToDoubleHighPrecisionTest` so the `quotient128 = digits * 10^q` branch is now covered:

- `9007199254740993e10` (boundary: digits = 2^53 + 1, q = +10)
- `12345678901234567e7`  (17-digit mantissa, q = +7)
- `-9007199254740993e15` (negative sign, q near upper bound)

07ead3a0. All 26 CastStringsTest methods pass locally on sm_89 / CUDA 13.0.

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Keeping `__inline__` for this round per your follow-up nit ("feel free to ignore"). After the dead-branch cleanup in 07ead3a0 the helper is small enough that the call-overhead vs. register-pressure tradeoff is hard to argue ahead of a real benchmark. Will revisit alongside the benchmark redo (see top-level reply on the helper-trigger-rate thread).

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Applied in 07ead3a0:
```cpp
division_rem = static_cast<uint64_t>(numerator - quotient128 * static_cast<__uint128_t>(pow10_abs_q));
```
Saves the second 128-bit modulo by reusing the quotient. Bit-for-bit equivalent on every helper-triggering input in the unit test.

wjxiz1992:
build

wjxiz1992:
@thirtiseven benchmark redo is now in the PR (b95077e6). The new bench file replaces the old one — three axes with host-verified trigger rates printed before every timed run, so the numbers can't be misread the way the prior round was.

### Setup

- GPU: NVIDIA RTX 5880 Ada Generation (SM 89), CUDA 13.0, 110 SMs, 48 GiB.
- Built with `BUILD_BENCHMARKS=ON BUILD_TESTS=ON`, default
  `CMAKE_CUDA_ARCHITECTURES=RAPIDS`.
- Each axis prints `[helper-trigger]` with the actual share of input
  rows that satisfy the caller's `slow_path_eligible` gate
  (`digits > 2^53 AND |exp_ten| <= 19`), re-evaluated on the host from
  the same input strings the kernel sees.

### Trigger-rate log (verbatim)

```
[helper-trigger] string_to_double_helper_off: 0/1048576 rows triggered
                 (by construction — from_floats caps digits at ~10^10 < 2^53)
[helper-trigger] string_to_double_helper_off: 0/104857600 rows triggered
                 (by construction — from_floats caps digits at ~10^10 < 2^53)
[helper-trigger] string_to_double_helper_on:  1000000/1000000 inspected =
                 100.00% triggered (parse-err: 0.00%)
[helper-trigger] string_to_double_helper_on:  1000000/1000000 inspected =
                 100.00% triggered (parse-err: 0.00%)
```

### Results

| Axis                        | num_rows | GPU Time   | Noise | per-1k-rows |
|---|---:|---:|---:|---:|
| FP32 baseline (unchanged)   |       1M |   1.739 ms | 1.83% |  1.658 us   |
| FP32 baseline (unchanged)   |     100M | 162.177 ms | 0.22% |  1.547 us   |
| FP64 helper **OFF** (legacy) |       1M |   1.743 ms | 4.35% |  1.662 us   |
| FP64 helper **OFF** (legacy) |     100M | 156.919 ms | 0.11% |  1.497 us   |
| FP64 helper **ON**  (100%)   |       1M |   1.265 ms | 1.58% |  1.207 us   |
| FP64 helper **ON**  (100%)   |      10M |  11.619 ms | 0.50% |  1.109 us   |

### Reading

- **No perf regression on the FP32 path.** Wrapped in
  `if constexpr (is_same_v<T, double>)`, codegen unchanged. Confirmed by
  baseline rows.
- **No perf regression on the FP64 legacy path.** Row 4 vs row 2 is
  `156.9 / 162.2 = 0.967` — the FP64-with-helper-not-entered case is
  ~3% faster than FP32 baseline at 100M rows, which is consistent with
  `from_floats` output being slightly shorter when the source is
  upcasted FP32. This is the regression-check row and it passes.
- **Helper path is actually slightly faster per-row than legacy.** The
  caller takes an early return on `slow_path_eligible`, skipping the
  subnormal-shift branch and the `exp10()` + multiply in the legacy
  path. The int128 work in the helper is cheaper. So the new code is
  net-neutral-to-positive on perf and correctness-positive on the
  helper-triggering inputs.

### Caveat on the helper-ON 100M-row axis

`100M handcrafted strings × ~20 chars/row` overflows the strings
column's int32 offset type (`2 * 10^9 > INT32_MAX`). The helper-ON
axis is set to {1M, 10M} as a result; the per-row cost is stable across
both (1.21 us → 1.11 us, in line with cold-start amortization). The
two other axes stay at {1M, 100M} because `from_floats` strings are
~10 chars and don't overflow.

### Verifying yourself

```
cd target/jni/cmake-build && cmake --build . --target STRING_TO_FLOAT_BENCH
LD_LIBRARY_PATH=$PWD:$LD_LIBRARY_PATH benchmarks/STRING_TO_FLOAT_BENCH --md /tmp/out.md
```

The `[helper-trigger]` lines print to stdout before each timed run.

thirtiseven:
@wjxiz1992 Thank you, it looks good!

wjxiz1992:
Follow-up on the benchmark: after applying the code-review fixes (commit `0872512a` — see PR body for the full diff: `cuda::std::bit_cast<double>` swap, `cuda::std::abs(q)`, digit separators, `CUDF_CUDA_TRY` wrapping on the bench `cudaMemcpyAsync`, INT32 offset guard in `pack_host_strings`, etc.), I re-ran the same 3-axis sweep on the rebuilt binary to confirm none of the fixes regressed perf.

Re-run on the same RTX 5880 Ada / CUDA 13.0 box (HEAD = `0872512a`):

| Axis | num_rows | GPU Time | vs prior round (`b95077e6`) |
|---|---:|---:|---:|
| FP32 baseline | 100M | 161.6 ms | −0.4% (within noise) |
| FP64 helper **OFF** (legacy) | 100M | 151.9 ms | −3.2% (within noise + from_floats RNG draw) |
| FP64 helper **ON** (100% trigger) | 10M | 11.64 ms | +0.17% (within noise) |

`[helper-trigger]` log still says `1000000/1000000 = 100.00%` for the helper-ON axis and `0/N` for helper-OFF, so the trigger gate is unchanged.

Headline conclusion unchanged: no perf regression on any axis; helper path is ~25% faster per-row than the legacy path because the caller takes an early return on `slow_path_eligible`, skipping the `exp10()` + subnormal-shift block. PR description has the current table.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -40,6 +43,144 @@ namespace detail {
...
+  //   q == 0: q_lo = digits > 2^53.
+  //   q > 0:  digits * 10^q >= digits > 2^53 (so q_hi or q_lo is non-zero).
+  //   q < 0:  (digits << 64) / 10^|q| > 2^53 (since digits > 2^53 and
+  //           10^|q| <= 10^19 < 2^64, so the quotient exceeds 2^53).
+  // Either q_hi or q_lo may be the most-significant limb in the q < 0 case,
+  // depending on whether digits >= 10^|q|, so both MSB-locator branches
+  // below are reachable.
+  int msb_pos;
+  if (q_hi != 0) {
+    msb_pos = 127 - __clzll(q_hi);
```
revans2: Some comments here about __clzll and what it is doing would be nice. It is not really a self describing function name.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -159,12 +300,32 @@ class string_to_float {
...
+      // preconditions (digits > 2^53 AND |exp_ten| <= 19) are enforced by this
+      // gate; outside them the legacy `digits * exp10(exp_ten)` path is already
+      // correctly rounded (both operands exact as doubles for digits <= 2^53;
+      // the rare |exp_ten| > 19 + large-digits case is the legacy 1-ULP path).
+      if constexpr (cuda::std::is_same_v<T, double>) {
+        bool const slow_path_eligible = (digits > (1ULL << 53)) && (cuda::std::abs(exp_ten) <= 19);
+        if (slow_path_eligible) {
+          _out[_row] = correctly_rounded_uint64_times_pow10(digits, exp_ten, sign);
+          compute_validity(_valid, _except);
+          return;
```
revans2: An early return like this is not very clean code, especially for a function as large as this is. Could we do some refactoring to make the fast path a separate function so that it is a clear and more simple division between the two?

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -159,12 +300,32 @@ class string_to_float {
...
+        }
+      }
+
+      // Legacy path: `static_cast<double>(digits)` is reached only when the
```
revans2: So we have a "legacy" path and a slow path? That is not really consistent.

src/main/cpp/benchmarks/cast_string_to_float.cpp
```diff
@@ -16,14 +16,226 @@
...
+#include <cstdint>
+#include <cstdio>
+#include <cstring>
+#include <limits>
+#include <random>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace {
```
revans2, line 34: nit: There is a lot of new code in this file that was fine for evaluating that the benchmark was touching the problematic values, but it feels like a lot to review and work with just to print out a % of values that hit this condition.

revans2 (commented):
Mostly nits, but the refactoring of the really large function is something I would like to see done. It is not about the current code being right or wrong. It is much more about preventing issues in the future.

While we are at it, it would be nice to break down correctly_rounded_uint64_times_pow10 into smaller functions too, so it is simpler to follow what the algorithm is doing. IT is very complex, but if we can have it look more like 

auto const (quotient128, remainder64) = divide_pows_10(...);
auto const msb_pos = calc_msb_pos(quotient128);
...

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Added an explanatory paragraph at the `__clzll` site in f7af3ace explaining it is the CUDA count-leading-zeros intrinsic on a 64-bit unsigned integer, and how the hi/lo limb split + `63 - __clzll(x)` locates the MSB of the 128-bit value. While at it I also added `(1)..(4)` step-marker comments through the function so the recipe (scale → locate MSB → round 53 bits → assemble IEEE 754) reads top-down.

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Renamed in f7af3ace so the names match across `.cu` and `.cpp`: the int128 path is now uniformly the "high-precision path" (was a mix of `slow_path_eligible` / "correctly rounded"), and the original `digits * exp10(exp_ten)` path is the "default path" (was "legacy path"). Also dropped the `slow_path_eligible` local variable; the predicate now reads inline at the `if` site.

src/main/cpp/benchmarks/cast_string_to_float.cpp
wjxiz1992: Trimmed in f7af3ace. Removed `parsed_decimal`, `parse_for_trigger_gate`, and the `print_trigger_rate` printf (~100 LOC of host code). `make_helper_triggering_strings` already guarantees 100% trigger rate by construction (every row is built with `digits > 2^53` and `|q| <= 19`), so the host re-parse was redundant.

wjxiz1992:
build

wjxiz1992:
build

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Done in 3a387321. Two structural changes:

1. Decomposed `correctly_rounded_uint64_times_pow10` into the four named sub-functions matching the (1)..(4) step block in the original comments — `scale_digits_times_pow10` -> `locate_msb_uint128` -> `round_top_53_bits` -> `assemble_ieee754_double`. All four are `__device__ __inline__` and operate on integers only.
2. Extracted the default `static_cast<double>(digits) * exp10(exp_ten)` path (plus subnormal + overflow handling) into a templated `default_double_path<T>`. The path-selection in `string_to_float::operator()` is now a single bool predicate + if/else with no mid-function early-return:

```cpp
bool const helper_eligible = is_same_v<T, double> &&
                             (digits > (1ULL << 53)) &&
                             (abs(exp_ten) <= 19);
if (helper_eligible) {
  _out[_row] = correctly_rounded_uint64_times_pow10(digits, exp_ten, sign);
} else {
  _out[_row] = default_double_path<T>(digits, exp_ten, sign);
}
```

wjxiz1992:
build

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -215,6 +216,142 @@ void castToFloatsInfTest(){
...
+        "1.0000000000000001",
+        "1.0000000000000002",
+        "3.1415926535897932",
+        "2.7182818284590452",
+        "2.2250738585072014",        // Double.MIN_NORMAL mantissa (no e-308 here)
+        "1.234567890123456789",      // 19-digit mantissa
+        "9.999999999999999999",
+        "-1.7976931348623157",
+        "1.5",
+        "0.1",
```
thirtiseven: really nit: do `"1.5"` and `"0.1"` not actually enter the helper path?

src/main/cpp/benchmarks/cast_string_to_float.cpp
```diff
@@ -16,14 +16,122 @@
...
+    // fail loudly here instead of silently truncating to a negative offset.
+    assert(pos <= static_cast<size_t>(std::numeric_limits<int32_t>::max()));
+    out.offsets.push_back(static_cast<int32_t>(pos));
+  }
+  return out;
+}
+
+// Build a non-nullable cudf STRING column from a host string vector without
+// pulling in cudf::cudftestutil (which isn't linked into benchmarks unless
+// BUILD_CUDF_TESTS=ON).
```
thirtiseven: `ConfigureBench` already links `cudf::cudftestutil`, and other
  benchmarks in this tree use `cudf::test::strings_column_wrapper`. It's a good point, we might need some doc or cmake updates but for now we can just set `BUILD_CUDF_TESTS=ON` so the helper is no longer needed.

src/main/cpp/benchmarks/cast_string_to_float.cpp
```diff
@@ -40,5 +148,60 @@ void string_to_float(nvbench::state& state)
...
+// `convert_floats.cu:134`), so the parsed mantissa is at most ~10^10 — well
+// below 2^53 (~9.007e15). The caller's `use_high_precision_path` is `false` for
+// every row and the default `digits * exp10(exp_ten)` path runs. The trigger
+// rate is asserted statically by `from_floats`'s contract (no host-side
+// re-parse needed). This is the regression-check baseline: cost on the FP64
+// path WITHOUT the new helper.
+void string_to_double_helper_off(nvbench::state& state)
+{
+  cudf::size_type const n_rows{(cudf::size_type)state.get_int64("num_rows")};
+  auto const float_tbl  = create_random_table({cudf::type_id::FLOAT64}, row_count{n_rows});
```
thirtiseven: from AI, really a nit
> Do we want this baseline to be non-nullable? `create_random_table` uses the default profile, which has a null frequency, so some rows will early-return before parsing rather than running the default `digits * exp10(exp_ten)` path. That also makes it a little different from the handcrafted helper-on column, which is non-nullable. Would it be better to pass an explicit no-null profile here, or soften the “every row” wording?

src/main/cpp/benchmarks/cast_string_to_float.cpp
```diff
@@ -40,5 +148,60 @@ void string_to_float(nvbench::state& state)
...
+    auto rows = spark_rapids_jni::string_to_float(
+      cudf::data_type{cudf::type_id::FLOAT64}, string_col->view(), false, stream);
+  });
+}
+
+// Note: helper_on uses smaller row counts than the other axes because the
+// handcrafted inputs average ~20 chars/row; at 100M rows the total chars
+// would exceed `INT32_MAX` and overflow the strings column's int32 offsets.
+// `from_floats`-based inputs (the other two axes) are ~10 chars/row so 100M
+// fits.
```
thirtiseven, line 142: From AI, nit

> Is the intent here to benchmark a general high-precision decimal-string workload, or a 100% helper stress case? The current comments/name say every row enters the helper and this is the worst case, but the generator can produce 20-digit `uint64_t` strings; when those get truncated by `parse_digits()`, `exp_base` increases by 1, so an appended `e19` becomes effective `exp_ten == 20` and misses the helper gate. If this is meant to be a general high-precision workload, should we rename/drop the “every row” and “worst case” claims and maybe just report the expected trigger rate? If it is meant to be a stress case, should the generator cap the mantissa at 19 digits or otherwise mirror the parser’s effective exponent?

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
wjxiz1992: Good catch — they don't. `digits` for those two inputs is 15 and 1 respectively, well below the `digits > 2^53` part of the gate, so they take the default `static_cast<double>(digits) * exp10(exp_ten)` path and just happened to be filed under the wrong array. Dropped both in f77418f8 — short-mantissa happy-path coverage is already provided by the other test arrays.

src/main/cpp/benchmarks/cast_string_to_float.cpp
wjxiz1992: Tried the swap and locally my libcudf install is built with `BUILD_TESTS=OFF`, so `cudf::cudftestutil` doesn't show up as an exported target and `cmake -DBUILD_BENCHMARKS=ON .` fails to configure (`Target "STRING_TO_FLOAT_BENCH" links to: cudf::cudftestutil but the target was not found`) — even though the link line at `benchmarks/CMakeLists.txt:59` already references it for every bench. The link works once libcudf is rebuilt with tests, but that's a substantial detour from this PR's scope (it triggers an Arrow rebuild on this host). I'd rather land the dependency-cleanup as a follow-up once `BUILD_CUDF_TESTS=ON` is the documented default — happy to file that. WDYT?

src/main/cpp/benchmarks/cast_string_to_float.cpp
wjxiz1992: Good point. Swapped to `data_profile const profile = data_profile_builder().no_validity();` and pass it to `create_random_table` in f77418f8 — `helper_off`'s input column is now non-nullable, matching `helper_on` so the comparison stays apples-to-apples and the parser isn't shortcutting on null masks.

src/main/cpp/benchmarks/cast_string_to_float.cpp
wjxiz1992: Right — `[2^53+1, UINT64_MAX]` straddles the 19/20-digit boundary so the 20-digit values get truncated by the parser's `max_safe_digits == 19`, exp_base bumps to 1, exp_ten lands at 20, and the gate fails. Capped the generator to 19 digits (`[2^53+1, 10^19 - 1]` = `9999999999999999999ULL`) in f77418f8 so every row provably enters the helper. Kept the "every row / worst case" framing since with the cap it's now actually true.

src/main/cpp/benchmarks/cast_string_to_float.cpp
wjxiz1992: Correction on my earlier comment: my speculation that "the link works once libcudf is rebuilt with tests" wasn't right *and* hadn't been verified. The actual recipe (thanks to a colleague pointing this out) is `-DBUILD_TESTS=ON` on the **jni cmake** (not libcudf's) — that pulls `cudf::cudftestutil` into the link graph for both tests and benchmarks, no libcudf rebuild needed.

Verified end-to-end this time:
```
cd target/jni/cmake-build && \
  cmake -DBUILD_BENCHMARKS=ON -DBUILD_TESTS=ON -DBUILD_FAULTINJ=OFF -DBUILD_PROFILER=OFF .
LIBRARY_PATH=…/libcudf-install/lib cmake --build . --target STRING_TO_FLOAT_BENCH -j 8
# [3/3] Linking CXX executable benchmarks/STRING_TO_FLOAT_BENCH
```

Swap pushed in 4d48bee9 — removes the custom `pack_host_strings` / `string_column_from_host` helpers (~55 LOC) and uses `cudf::test::strings_column_wrapper` directly. Apologies for the earlier confidence-without-verification.

src/main/cpp/src/cast_string_to_float.cu
```diff
@@ -43,28 +43,27 @@ namespace detail {
...
+__device__ __inline__ scaled_uint128 scale_digits_times_pow10(uint64_t digits, int q)
 {
-  // Caller precondition: digits > 2^53, so digits cannot be 0.
   assert(digits > (1ULL << 53));
-
   int const abs_q = cuda::std::abs(q);
-  // Caller precondition: |q| <= 19, so abs_q is in [0, 19] and indexes k_pow10.
   assert(abs_q <= 19);
 
   // 10^abs_q for abs_q in [0, 19] — exact as uint64. `static constexpr` marks
```
revans2: nit: I think the k_pow10 is fairly self describing and we don't need a lot of explanation around it. We can probably make this a lot smaller.

revans2 (dismissed):
Just a nit that is very small

src/main/cpp/src/cast_string_to_float.cu
wjxiz1992: Fixed in 66bad1fa. Reduced the k_pow10 note to a single line as suggested. Local validation: pre-commit run clang-format --files src/main/cpp/src/cast_string_to_float.cu passed.

src/main/cpp/benchmarks/cast_string_to_float.cpp
```diff
@@ -16,14 +16,55 @@
...
+    }
+    out.push_back(std::move(s));
+  }
+  return out;
+}
+
+}  // namespace
+
+// FP32 cast — the correctly-rounded helper does not run on this code path
+// (the `if constexpr (is_same_v<T, double>)` gate skips it for float). Kept
```
revans2: nit: I think this comment is out of date

src/main/cpp/benchmarks/cast_string_to_float.cpp
```diff
@@ -40,5 +81,65 @@ void string_to_float(nvbench::state& state)
...
 NVBENCH_BENCH(string_to_float)
-  .set_name("Strings to Float Cast")
+  .set_name("Strings to Float Cast (FP32 baseline)")
   .add_int64_axis("num_rows", {1 * 1024 * 1024, 100 * 1024 * 1024});
+
+// FP64 cast where the correctly-rounded helper is NEVER entered. Inputs are
+// produced by `cudf::strings::from_floats`, which caps significant digits at
+// 10 via `nine_digits = 10^9` in `dissect_value` (cudf
+// `convert_floats.cu:134`), so the parsed mantissa is at most ~10^10 — well
+// below 2^53 (~9.007e15). The caller's `use_high_precision_path` is `false` for
```
revans2: nit: Here too I think the comment is out of date with the code.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
```diff
@@ -215,6 +216,140 @@ void castToFloatsInfTest(){
...
+        "1.0000000000000001",
+        "1.0000000000000002",
+        "3.1415926535897932",
+        "2.7182818284590452",
+        "2.2250738585072014",        // Double.MIN_NORMAL mantissa (no e-308 here)
+        "1.234567890123456789",      // 19-digit mantissa
+        "9.999999999999999999",
+        "-1.7976931348623157",
+        // --- helper q > 0 path: 17-digit mantissa with explicit positive
+        // exponent. digits > 2^53 AND |exp_ten| <= 19, so slow_path_eligible
```
revans2: nit: I think slow_path_eligible is removed?

revans2 (dismissed):

src/main/cpp/benchmarks/cast_string_to_float.cpp
wjxiz1992: Updated in 74d37ee9 — replaced the stale `if constexpr (is_same_v<T, double>)` claim with a note that points at the actual gate (the `cuda::std::is_same_v<T, double>` term inside the `helper_eligible` predicate that's now used at the dispatch site).

src/main/cpp/benchmarks/cast_string_to_float.cpp
wjxiz1992: Updated in 74d37ee9 — `use_high_precision_path` was renamed to `helper_eligible` and the FP64 default path fell out into `default_double_path<T>` during the earlier refactor. Both bench comments (helper_off at line 91 and helper_on at line 122) now reference the current names.

src/test/java/com/nvidia/spark/rapids/jni/CastStringsTest.java
wjxiz1992: Updated in 74d37ee9 — replaced `slow_path_eligible is true` with `helper_eligible is true` to match the current predicate name at `cast_string_to_float.cu:399`.

wjxiz1992:
build

thirtiseven (approved):
🎉

PR #4593, opened 2026-05-19
Add a pre-touched pageable memory pool implementation

zpuller:
Adds a pageable host memory pool (`PageableMemoryPool`) as a fallback when the pinned pool is exhausted. The pool pre-touches its backing pages at initialization time, so DtoH copies into pool-allocated buffers avoid first-touch page faults — reaching significantly higher DtoH bandwidth than freshly malloc'd pageable memory.

The implementation is a Java class (`ai.rapids.cudf.PageableMemoryPool`) backed by a C++ `rmm::mr::pool_memory_resource` over a custom `posix_memalign`-based upstream that pre-touches pages in parallel across a configurable number of threads at pool construction time.

The implementation is adapted from the pinned pool equivalent which can be found in cudf and RMM:
* https://github.com/rapidsai/cudf/blob/main/java/src/main/java/ai/rapids/cudf/PinnedMemoryPool.java
* https://github.com/rapidsai/cudf/blob/main/java/src/main/native/src/RmmJni.cpp
* https://github.com/rapidsai/rmm/blob/main/cpp/include/rmm/mr/pinned_host_memory_resource.hpp

Unblocks https://github.com/NVIDIA/spark-rapids/pull/14828


zpuller:
build

zpuller:
build

zpuller:
build

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+ * of pinned DtoH bandwidth than freshly-malloc'd pageable, because the page faults are
+ * paid at pool growth time rather than at DtoH copy time.
+ */
+public final class PageableMemoryPool implements AutoCloseable {
+  private static final Logger log = LoggerFactory.getLogger(PageableMemoryPool.class);
+
+  // These static fields should only ever be accessed when class-synchronized.
+  // Do NOT use singleton_ directly!  Use the getSingleton accessor instead.
+  private static volatile PageableMemoryPool singleton_ = null;
+  private static Future<PageableMemoryPool> initFuture = null;
```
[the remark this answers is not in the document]
zpuller: This shouldn't be an issue because the initialize pattern is that it just happens synchronously in one thread during init time before any threads ever call getSingleton, and all other synchronization happens properly around the volatile singleton_.

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+  private PageableMemoryPool(long poolSize, int pretouchThreads) {
+    this.poolHandle = newPageablePoolMemoryResource(poolSize, poolSize, pretouchThreads);
+    this.poolSize = poolSize;
+  }
+
+  @Override
+  public void close() {
+    releasePageablePoolMemoryResource(this.poolHandle);
+    this.poolHandle = -1;
+  }
```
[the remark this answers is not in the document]
zpuller: Similar to the init case, this isn't really a concern because shutdown will only get called once all tasks complete, if an alloc/free is called concurrently it's a usage error. Also, this code is all based on the pinned pool patterns as mentioned in the description, which are battle tested.

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+    initFuture = initService.submit(() -> new PageableMemoryPool(poolSize, pretouchThreads));
+    initService.shutdown();
+  }
+
+  /**
+   * Check if the pool has been initialized or not.
+   */
+  public static boolean isInitialized() {
+    return getSingleton() != null;
+  }
```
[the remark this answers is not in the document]
zpuller: Valid but again this is fine considering usage pattern and that we're based off PinnedPool design.

src/main/cpp/src/PageablePoolJni.cpp
```diff
@@ -0,0 +1,207 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cudf_jni_apis.hpp"
+
+#include <rmm/aligned.hpp>
+#include <rmm/detail/error.hpp>
+#include <rmm/mr/detail/coalescing_free_list.hpp>
```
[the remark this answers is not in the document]
zpuller: This seems valid but the risk is relatively low, I'll leave it up to other reviewers whether it warrants vendoring a copy.

src/main/cpp/src/PageablePoolJni.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+#include <cstdlib>
+#include <mutex>
+#include <thread>
+#include <vector>
+
+namespace {
+
+static constexpr std::size_t page_size_ = 4096;
```
revans2: I think the right API to use here is `sysconf(_SC_PAGESIZE)` or `_SC_PAGE_SIZE` from `<unistd.h>`, cached once, with error handling. `getpagesize()` also exists but is older/less preferred (At least according to an AI).

src/main/cpp/src/PageablePoolJni.cpp
```diff
@@ -0,0 +1,207 @@
...
+  /**
+   * @brief Try to sub-allocate @p bytes from the pool using a best-fit search.
+   *
+   * The request is rounded up to rmm::CUDA_ALLOCATION_ALIGNMENT. If the chosen block is larger
+   * than needed, the remainder is returned to the free list.
+   *
+   * @param bytes requested allocation size in bytes
+   * @return pointer to the allocated region, or nullptr if no free block is large enough
+   */
+  void* try_allocate(std::size_t bytes)
```
revans2: I think we should look at how 0 bytes is allocated. I think it removes a real free block with a real size, but when we free that 0 length allocation we add a 0 length block back into the free list.

revans2 (commented):
I have a number of concerns with this. We had a pageable pool in the past and stopped using it partly because of memory fragmentation, partly because of the the algorithms used to allocate and free the memory not being great (I wrote them myself after all). And partly because the memory was stuck. we could not free any of it to give it back, let it grow or shrink. 

Could we please get some performance tests on this? How long does it take to touch the pinned memory and how close is it to speed of light? It is a loss if we spend more time touching the memory than we lost trying to allocate it.

Can we run some standard benchmarks and look at memory fragmentation rates? and allocation/deallocation speeds. Ideally we would compare this to other allocators.

zpuller:
> I have a number of concerns with this. We had a pageable pool in the past and stopped using it partly because of memory fragmentation, partly because of the the algorithms used to allocate and free the memory not being great (I wrote them myself after all). And partly because the memory was stuck. we could not free any of it to give it back, let it grow or shrink.
> 
> Could we please get some performance tests on this? How long does it take to touch the pinned memory and how close is it to speed of light? It is a loss if we spend more time touching the memory than we lost trying to allocate it.
> 
> Can we run some standard benchmarks and look at memory fragmentation rates? and allocation/deallocation speeds. Ideally we would compare this to other allocators.

Thanks, these are valid concerns. I ran some focused tests and microbenchmarks for the specific pieces you called out: pre-touch time, DtoH bandwidth into different host-memory types, allocation/deallocation latency, and some deterministic fragmentation probes.

On my workstation (RTX 5880 Ada), the current results look like this:

- Pre-touch cost, one-time at pool initialization:
  - 1 GiB: ~416 ms with 1 thread, ~68 ms with 8 threads, ~46 ms with 16 threads
  - 4 GiB: ~1.60 s with 1 thread, ~252 ms with 8 threads, ~176 ms with 16 threads

- DtoH bandwidth:
  - pinned: ~26.35 GB/s for 1 GiB
  - pre-touched pageable: ~17.29 GB/s for 1 GiB
  - fresh pageable: ~2.12 GB/s for 1 GiB

So direct copies into pre-touched pageable memory are not as fast as pinned memory, but they are much faster than the cold/fresh pageable path. For the 1 GiB case, pre-touching with 16 threads costs ~46 ms once, while avoiding roughly 444 ms of first-touch copy overhead compared with fresh pageable memory. That is the main reason I think the initialization cost is acceptable: it is paid once and then amortized over the lifetime of the application because allocations reuse the same underlying pool pages.

- Allocation latency after pool initialization:
  - pool alloc/free: ~11 us for 64 MiB, 256 MiB, and 1 GiB
  - `cudaMallocHost`: ~34 ms for 64 MiB, ~137 ms for 256 MiB, ~534 ms for 1 GiB
  - `malloc`: ~17 us, but that does not pre-touch pages and does not avoid the cold-copy penalty

You can also see NDS e2e benchmark times with this pool enabled in https://github.com/NVIDIA/spark-rapids/pull/14828

For fragmentation: I agree this is a real concern. I ran some synthetic tests to reproduce the expected failure mode: aggregate free bytes can be high while the largest contiguous free block is too small. For example, in a 1 GiB pool with isolated 16 MiB holes, the benchmark sees 512 MiB free, largest free block 16 MiB, external fragmentation ~0.969, and a 128/256 MiB allocation fails. When adjacent chunks are freed, coalescing recovers the larger block as expected; in the fully coalesced cases the same benchmark reports external fragmentation 0 and the large allocation succeeds.

The synthetic tests intentionally force worst-case allocation patterns and confirm the expected behavior: if free space is split into isolated holes, aggregate free bytes can be high while a large allocation still fails. They also confirm that contiguous frees coalesce and recover the full pool.

I also ran a more realistic Spark workload investigation with pageable fragmentation logging enabled and a deliberately small 64 MiB pinned pool to force traffic through the pageable fallback path. Those results were more encouraging:

- query50: 532 pageable allocs, 531 frees, 0 allocation failures, peak pageable usage ~2.60 GiB, max external fragmentation 8.31%, p99 fragmentation 6.05%.
- full power run: 37,155 pageable allocs, 37,156 frees, 0 allocation failures, peak pageable usage ~5.70 GiB, p99 fragmentation 6.04%, p99.9 13.12%.
- The full run did have a transient max fragmentation spike of 30.14%, but at that point the pool still had a ~9.64 GiB largest contiguous free block, so it was not capacity-threatening for that workload.
- The pool fully coalesced back to one free block by the end of the full run.

So fragmentation is a valid fixed-pool risk, and the synthetic tests demonstrate the failure mode, but the Spark workload data we collected did not show pageable fragmentation becoming a practical problem. Also, this is not a new class of risk relative to the pinned pool path we already use: the pinned pool is also an RMM pool over host memory with the same coalescing free-list behavior. The pageable pool is intended for the same usage pattern as the pinned pool today.

zpuller:
build

zpuller:
build

zpuller:
build

src/main/cpp/src/pageable_pool_resource.hpp
```diff
@@ -0,0 +1,234 @@
...
+  {
+    if (bytes == 0) { return nullptr; }
+    bytes = align_up(bytes);
+    std::lock_guard<std::mutex> lock(mtx_);
+    auto blk = free_list_.get_block(bytes);
+    if (!blk.is_valid()) { throw pageable_pool_exhausted{}; }
+    if (blk.size() > bytes) {
+      free_list_.insert(rmm::mr::detail::block{blk.pointer() + bytes, blk.size() - bytes, false});
+    }
+    return blk.pointer();
```
pmattione-nvidia, line 160: don't you need to align the pointer with the input alignment?  All you've done with align_up is guarantee that you'll be able to align the data within the returned memory block.  I think you may need to embed the offset to the actual first byte of the allocation in the padding bytes right before your returned pointer, so that deallocate() can find where the allocation actually begins.  Which means you need to increase `bytes` by 4 to hold this. Unless rmm has tools already with a different way of handling it.

src/main/cpp/src/pageable_pool_resource.hpp
```diff
@@ -0,0 +1,234 @@
...
+   *
+   * Rounds up to default_cuda_malloc_host_alignment. Throws pageable_pool_exhausted if
+   * no free block is large enough.
+   */
+  [[nodiscard]] void* allocate_sync(
+    std::size_t bytes,
+    [[maybe_unused]] std::size_t alignment = cuda::mr::default_cuda_malloc_host_alignment)
+  {
+    if (bytes == 0) { return nullptr; }
+    bytes = align_up(bytes);
```
pmattione-nvidia, line 153: align_up() doesn't even USE the input alignment.  pass it in

src/main/cpp/src/pageable_pool_resource.hpp
```diff
@@ -0,0 +1,234 @@
...
+        if (t.joinable()) { t.join(); }
+      }
+      throw;
+    }
+  }
+
+  static std::size_t system_page_size()
+  {
+    static std::size_t const page_size = [] {
+      long const size = ::sysconf(_SC_PAGESIZE);
```
pmattione-nvidia, line 217: This won't work if they've enabled large pages

pmattione-nvidia: Or I guess it will, it will just touch more than necessary

src/main/cpp/tests/pageable_pool.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  void* full = pool.allocate_sync(kPoolSize);
+  ASSERT_NE(full, nullptr);
+  pool.deallocate_sync(full, kPoolSize);
+}
+
+// ---------------------------------------------------------------------------
+// Alignment
+// ---------------------------------------------------------------------------
+TEST(PageablePool, PointerIsAligned)
```
pmattione-nvidia, line 94: need some actual real alignment tests.  e.g. a 64-byte-aligned allocation, or a 4kb aligned one

src/main/cpp/src/pageable_pool_resource.hpp
pmattione-nvidia: Also require the alignment to be a power of 2

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+   * for multi-GB pools instead of several seconds serially.
+   *
+   * @param poolSize        size of the pool in bytes
+   * @param pretouchThreads number of worker threads to use for the parallel pre-touch
+   */
+  public static synchronized void initialize(long poolSize, int pretouchThreads) {
+    if (isInitialized()) {
+      throw new IllegalStateException("Can only initialize the pageable pool once.");
+    }
+    ExecutorService initService = Executors.newSingleThreadExecutor(runnable -> {
```
pmattione-nvidia, line 129: what is the point of doing this on a new thread

src/main/cpp/src/pageable_pool_resource.hpp
```diff
@@ -0,0 +1,234 @@
...
+   * @param size           Total pool size in bytes.
+   * @param pretouch_threads Number of threads used to fault in backing pages.
+   * @throws rmm::out_of_memory if the backing allocation fails.
+   */
+  pageable_pool_resource(cuda::mr::any_synchronous_resource<cuda::mr::host_accessible> upstream,
+                         std::size_t size,
+                         int pretouch_threads)
+    : upstream_(std::move(upstream)),
+      pool_size_(size),
+      base_(upstream_.allocate_sync(size, alignof(std::max_align_t)))
```
pmattione-nvidia: you should probably use the page size instead of max_align_t

src/main/cpp/src/pageable_pool_resource.hpp
```diff
@@ -0,0 +1,234 @@
...
+  {
+    std::size_t const page_size = system_page_size();
+    int const n                 = std::max(1, threads);
+    std::vector<std::thread> ts;
+    ts.reserve(n);
+    std::size_t per = (bytes + n - 1) / static_cast<std::size_t>(n);
+    per             = ((per + page_size - 1) / page_size) * page_size;
+    try {
+      for (int i = 0; i < n; ++i) {
+        std::size_t off = static_cast<std::size_t>(i) * per;
```
pmattione-nvidia, line 193: this loop assumes that the byte buffer was allocated at a page boundary, when earlier you only specified alignment of max_align_t.  this means the first and last pages might be missed.  this loop is ok if the alignment of the initial allocation is fixed.

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
pmattione-nvidia: Nevermind, I see why: the getSingleton()

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+      }
+      synchronized (PageableMemoryPool.class) {
+        if (singleton_ == null) {
+          try {
+            singleton_ = initFuture.get();
+          } catch (Exception e) {
+            throw new RuntimeException("Error initializing pageable memory pool", e);
+          }
+          initFuture = null;
+        }
```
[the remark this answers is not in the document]
igorpeshansky: The issue is valid, but the proposed solution, as-is, would introduce a heisenbug, because if `initFuture` is cleared after an exception, any threads waiting on the `synchronized` block after the initial check would find `singleton_ == null` and invoke `initFuture.get()` unconditionally, causing an NPE and resulting in a variant of the "repeated exceptions" behavior this warns about. To mitigate, we'd also need to change the initial condition in the `synchronized` block to `singleton_ == null && initFuture != null`.

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+        origAddress = address;
+        try {
+          PageableMemoryPool.freeInternal(address, origLength);
+        } finally {
+          // Always mark the resource as freed even if an exception is thrown.
+          // We cannot know how far it progressed before the exception, and
+          // therefore it is unsafe to retry.
+          address = -1;
+        }
+        neededCleanup = true;
```
igorpeshansky, line 70: [Optional] If you initialize `origAddress` to -1, you could get rid of `neededCleanup` (because you'd have the `neededCleanup == (origAddress != -1)` invariant).

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+
+  /**
+   * Check if the pool has been initialized or not.
+   */
+  public static boolean isInitialized() {
+    return getSingleton() != null;
+  }
+
+  /**
+   * Shut down the pool, nulling out our reference. Any allocation or free in flight
```
igorpeshansky: Nit: "Any allocation or free in flight" is a bit ambiguous. How about "any in-flight allocation or free"?

```suggestion
   * Shut down the pool, nulling out our reference. Any in-flight allocation or free
```

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+   * @return newly created buffer, or null if the pool is uninitialized or exhausted
+   *         (caller should fall back to a regular malloc'd buffer)
+   */
+  public static HostMemoryBuffer tryAllocate(long bytes) {
+    HostMemoryBuffer result = null;
+    PageableMemoryPool pool = getSingleton();
+    if (pool != null) {
+      result = pool.tryAllocateInternal(bytes);
+    }
+    return result;
```
igorpeshansky: Is there any reason we wouldn't just return the value instead of assigning it to a local (like what you did in `getTotalPoolSizeBytes()`?

```suggestion
    PageableMemoryPool pool = getSingleton();
    if (pool != null) {
      return pool.tryAllocateInternal(bytes);
    }
    return null;
```

[Optional] It might be even cleaner with guard clauses, i.e.:

```java
  public static HostMemoryBuffer tryAllocate(long bytes) {
    PageableMemoryPool pool = getSingleton();
    if (pool == null) return null;
    return pool.tryAllocateInternal(bytes);
  }
…
  public static long getTotalPoolSizeBytes() {
    PageableMemoryPool pool = getSingleton();
    if (pool == null) return 0;
    return pool.poolSize;
  }
```

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+
+  private PageableMemoryPool(long poolSize, int pretouchThreads) {
+    this.poolHandle = newPageablePoolMemoryResource(poolSize, poolSize, pretouchThreads);
+    this.poolSize = poolSize;
+  }
+
+  @Override
+  public void close() {
+    releasePageablePoolMemoryResource(this.poolHandle);
+    this.poolHandle = -1;
```
igorpeshansky: [Optional] It probably doesn't matter, but is it worth clearing out `poolSize` here as well? What are the semantics of calling `getTotalPoolSizeBytes()` on an explicitly `close`d pool?

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  /**
+   * Attempts to allocate from the pageable pool. Returns null rather than throwing if
+   * the pool is exhausted, so callers can fall back gracefully.
+   */
+  private synchronized HostMemoryBuffer tryAllocateInternal(long bytes) {
+    long allocated = allocFromPageablePool(this.poolHandle, bytes);
```
igorpeshansky, line 206: Should `tryAllocateInternal` and `free` throw if the pool has been explicitly `close`d? Today they'd attempt to access `(void*)-1`… Even if this is an error on the part of the caller, an exception is nicer than a segfault, especially in Java…

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+    if (allocated == -1) {
+      return null;
+    } else {
+      return new HostMemoryBuffer(allocated, bytes,
+              new PageableHostBufferCleaner(allocated, bytes));
+    }
```
igorpeshansky: [Optional] I'm a big believer in guard clauses for readability.

```suggestion
    }
    return new HostMemoryBuffer(allocated, bytes,
            new PageableHostBufferCleaner(allocated, bytes));
```

src/main/cpp/tests/pageable_pool.cpp
igorpeshansky, line 1: Do we want Java-level tests as well?

src/main/cpp/tests/pageable_pool.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include <thread>
+#include <vector>
+
+using namespace spark_rapids_jni;
+
+static constexpr std::size_t kPoolSize = 4 * 1024 * 1024;  // 4 MiB
+
+// Explicit any_resource construction is required — the converting constructor
+// of any_resource is not implicit in this CCCL version.
+static pageable_pool_resource make_pool(std::size_t size = kPoolSize, int threads = 1)
```
igorpeshansky, line 31: [Optional] It might be cleaner to group the constant and the function in an unnamed namespace, rather than declaring them static.

src/main/cpp/tests/pageable_pool.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include <new>
+#include <thread>
+#include <vector>
+
+using namespace spark_rapids_jni;
+
+static constexpr std::size_t kPoolSize = 4 * 1024 * 1024;  // 4 MiB
+
+// Explicit any_resource construction is required — the converting constructor
+// of any_resource is not implicit in this CCCL version.
```
igorpeshansky: I initially interpreted this as a function comment, but this is just an implementation detail. I'd put it just before the `cuda::mr::any_synchronous_resource` constructor call…

src/main/cpp/tests/pageable_pool.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+TEST(PageablePool, AllocFillAndFree)
+{
+  auto pool = make_pool();
+  void* p   = pool.allocate_sync(1024);
+  ASSERT_NE(p, nullptr);
+  std::memset(p, 0xAB, 1024);
+  auto* bytes = static_cast<unsigned char*>(p);
+  for (std::size_t i = 0; i < 1024; ++i)
+    ASSERT_EQ(bytes[i], 0xAB);
```
igorpeshansky: The googletest convention is to use `EXPECT_EQ` unless it's pointless to continue.

That said, with the current loop you may end up with very noisy test output on many mismatches, so you might want to add a custom assertion helper[^1], e.g.:

```c++
  auto all_eq = [](unsigned char* b, std::size_t l, unsigned char v) {
    for (std::size_t i = 0; i < l; ++i) {
      if (b[i] != v) return testing::AssertionFailure() << "mismatch at index " << i << std::hex << " expected " << v << " actual " << b[i];
    }
    return testing::AssertionSuccess();
  };
  EXPECT_TRUE(all_eq(bytes, 1024, 0xAB));
```

[^1]: https://google.github.io/googletest/advanced.html#using-a-function-that-returns-an-assertionresult

src/main/cpp/tests/pageable_pool.cpp
```diff
@@ -0,0 +1,195 @@
...
+  pool.deallocate_sync(p, 1024);
+}
+
+TEST(PageablePool, ZeroByteAllocationDoesNotConsumePool)
+{
+  auto pool = make_pool();
+  void* p   = pool.allocate_sync(0);
+  ASSERT_EQ(p, nullptr);
```
igorpeshansky: [Optional] If this assertion fails, `p` would never be deallocated from the pool. That's probably ok, as the entire pool is local and will be deleted when it goes out of scope, but if not, it's worth addressing… In this case, you could use `EXPECT_EQ` with no loss of functionality, but in general, you'd need some sort of RAII.

src/main/cpp/tests/pageable_pool.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+// ---------------------------------------------------------------------------
+// Alignment
+// ---------------------------------------------------------------------------
+TEST(PageablePool, PointerIsAligned)
+{
+  auto pool                   = make_pool();
+  constexpr std::size_t align = cuda::mr::default_cuda_malloc_host_alignment;
+  void* p                     = pool.allocate_sync(1);
+  ASSERT_EQ(reinterpret_cast<std::uintptr_t>(p) % align, 0u);
```
igorpeshansky: `EXPECT_EQ`, since it's not gating the remainder of the test case.

src/main/cpp/tests/pageable_pool.cpp
```diff
@@ -0,0 +1,195 @@
...
+TEST(PageablePool, MultipleAllocsExhaustPool)
+{
+  auto pool                   = make_pool();
+  constexpr std::size_t chunk = 256 * 1024;  // 256 KiB
+  constexpr int n             = static_cast<int>(kPoolSize / chunk);
+
+  std::vector<void*> ptrs(n);
+  for (int i = 0; i < n; ++i) {
+    ptrs[i] = pool.allocate_sync(chunk);
+    ASSERT_NE(ptrs[i], nullptr);
```
igorpeshansky: Might be worth adding the index of the pointer that failed the assertion…

src/main/cpp/tests/pageable_pool.cpp
```diff
@@ -0,0 +1,195 @@
...
+  }
+  EXPECT_THROW({ [[maybe_unused]] void* _ = pool.allocate_sync(chunk); }, pageable_pool_exhausted);
+  for (int i = 0; i < n; ++i)
+    pool.deallocate_sync(ptrs[i], chunk);
```
igorpeshansky: I find that it's always safer to use braces for blocks that don't go on the same line as the loop/conditional header, in case they grow to more than one statement. Also above and below…

```suggestion
  for (int i = 0; i < n; ++i) {
    pool.deallocate_sync(ptrs[i], chunk);
  }
```

res-life:
Have we evaluated existing open-source allocators, such as TLSF (https://github.com/mattconte/tlsf), instead of maintaining a custom allocator?
Also, have we tested this implementation on memory-constrained clusters to evaluate fragmentation, OOM behavior?

src/main/cpp/src/PageablePoolJni.cpp
```diff
@@ -0,0 +1,80 @@
...
+  {                                                                        \
+    return -1; /* Pool exhausted — caller falls back to regular malloc. */ \
+  }                                                                        \
+  CATCH_SPECIAL_EXCEPTION(env, ret_val)                                    \
+  CATCH_STD_EXCEPTION(env, ret_val)
+
+extern "C" {
+
+JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_PageableMemoryPool_newPageablePoolMemoryResource(
+  JNIEnv* env, jclass, jlong pool_size, jlong /* max_size unused */, jint pretouch_threads)
```
ttnghia: Why have unused parameter?

src/main/cpp/src/pageable_pool_resource.hpp
```diff
@@ -0,0 +1,234 @@
...
+   *        adjacent free blocks.
+   */
+  void deallocate_sync(
+    void* ptr,
+    std::size_t bytes,
+    [[maybe_unused]] std::size_t alignment = cuda::mr::default_cuda_malloc_host_alignment) noexcept
+  {
+    if (bytes == 0) { return; }
+    std::lock_guard<std::mutex> lock(mtx_);
+    free_list_.insert(rmm::mr::detail::block{static_cast<char*>(ptr), align_up(bytes), false});
```
ttnghia: `deallocate_sync` is declared `noexcept` (line 163) but calls `free_list_.insert(...)` which can thrown. This will cause potential, silent undefined behavior.

src/main/cpp/src/pageable_pool_resource.hpp
ttnghia: The backing buffer is requested with `alignof(std::max_align_t)` (line 112), so it is only 16-byte aligned, not page-aligned

```suggestion
      base_(upstream_.allocate_sync(size, system_page_size()))
```

src/main/cpp/src/pageable_pool_resource.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * @brief Allocate @p bytes from the pool (best-fit).
+   *
+   * Rounds up to default_cuda_malloc_host_alignment. Throws pageable_pool_exhausted if
+   * no free block is large enough.
+   */
+  [[nodiscard]] void* allocate_sync(
+    std::size_t bytes,
+    [[maybe_unused]] std::size_t alignment = cuda::mr::default_cuda_malloc_host_alignment)
```
ttnghia: I don't see `alignment` is used anywhere. That is not "Rounds up to default_cuda_malloc_host_alignment.".

src/main/cpp/src/pageable_pool_resource.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+ private:
+  static void pretouch_parallel(void* base, std::size_t bytes, int threads)
+  {
+    std::size_t const page_size = system_page_size();
+    int const n                 = std::max(1, threads);
+    std::vector<std::thread> ts;
+    ts.reserve(n);
+    std::size_t per = (bytes + n - 1) / static_cast<std::size_t>(n);
+    per             = ((per + page_size - 1) / page_size) * page_size;
```
ttnghia: Use `rmm::align_up(per, page_size)`.

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
ttnghia: This is NOT declared volatile, yet it is read at :90 OUTSIDE any synchronization, while `singleton_` (:31) IS volatile

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+          initFuture = null;
+        }
+      }
+    }
+    return singleton_;
+  }
+
+  private static void freeInternal(long address, long origLength) {
+    Objects.requireNonNull(getSingleton()).free(address, origLength);
+  }
```
ttnghia, line 114: If `shutdown()` set `singleton_ = null` while live `HostMemoryBuffer` objects exist, the GC cleaner eventually calls `freeInternal`; `getSingleton()` returns null and `requireNonNull` throws NPE inside the cleaner → native memory leaked with no log. The NPE fires before reaching `free()`.

Fix: Replace requireNonNull with an explicit null check + warning log.

Reference: PageableMemoryPool.java:60-86 (cleanImpl).

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+  private PageableMemoryPool(long poolSize, int pretouchThreads) {
+    this.poolHandle = newPageablePoolMemoryResource(poolSize, poolSize, pretouchThreads);
+    this.poolSize = poolSize;
+  }
+
+  @Override
+  public void close() {
+    releasePageablePoolMemoryResource(this.poolHandle);
+    this.poolHandle = -1;
+  }
```
ttnghia: `close()` sets `this.poolHandle = -1`, but `tryAllocateInternal` (195) and `free` (205) call JNI using `this.poolHandle` without checking for -1.

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
```diff
@@ -0,0 +1,208 @@
...
+   */
+  private synchronized HostMemoryBuffer tryAllocateInternal(long bytes) {
+    long allocated = allocFromPageablePool(this.poolHandle, bytes);
+    if (allocated == -1) {
+      return null;
+    } else {
+      return new HostMemoryBuffer(allocated, bytes,
+              new PageableHostBufferCleaner(allocated, bytes));
+    }
+  }
```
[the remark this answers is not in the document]
zpuller: I'm fine with this, I want to match Pinned Pool semantics which allows this

src/main/cpp/src/pageable_pool_resource.hpp
zpuller: Sorry for the confusion. The intention is actually to not support custom alignment whatsoever, to match what the RMM pinned pool does (and to simplify the implementation unless we decide we need to support custom alignment for some reason). I added some comments to clarify that

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
zpuller: It's a nice idea, but I personally prefer to just leave as is because it reads more clearly to me even though it employs an extra bool

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
zpuller: It would already check if the pool is null in which case it would return 0, so setting it to zero here is not necessary

src/main/cpp/src/pageable_pool_resource.hpp
zpuller: Fixed the wording on that

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
zpuller: Well the scenario you describe actually shouldn't happen, so I don't think it's appropriate to just warn log and then continue. But I agree that we should log if there is a potential memory leak. So I tweaked this to both log the address leaked and still throw IllegalStateException afterwards.

zpuller:
build

src/main/cpp/src/pageable_pool_resource.hpp
pmattione-nvidia: Then remove the name alignment from the function signature.  you can still have the default argument.  `std::size_t  = cuda::mr::default_cuda_malloc_host_alignment`

zpuller:
build

src/main/java/ai/rapids/cudf/PageableMemoryPool.java
igorpeshansky: In that case, you can initialize `long origAddress = address;` right away (and even make it `final`), since you'd only use that value when `neededCleanup` is true…

src/main/cpp/tests/pageable_pool.cpp
igorpeshansky: The `static` is now redundant…

zpuller:
> Have we evaluated existing open-source allocators, such as TLSF (https://github.com/mattconte/tlsf), instead of maintaining a custom allocator? 

I don’t think TLSF is a good fit here. The goal is to preserve the behavior of RMM’s existing pool resources, and this implementation reuses RMM’s coalescing best-fit free list, including its splitting, coalescing, and alignment behavior.
TLSF uses a different allocation policy and metadata layout, so its fragmentation and OOM behavior would not match the RMM pool this is intended to mirror. The referenced implementation also defaults to a maximum block size below 4 GiB, making it unsuitable for these pool sizes without customization.
Longer term, this feature may belong in RMM. Staying consistent with RMM avoids introducing a second allocator that we would likely replace later.

> Also, have we tested this implementation on memory-constrained clusters to evaluate fragmentation, OOM behavior?

We have tested forced fragmentation and exhaustion, concurrent allocation/free, and Spark workloads with a constrained pool. The synthetic tests reproduce fragmented OOM; the Spark workloads had no pageable allocation failures and fully coalesced afterward.

igorpeshansky (commented):
Sorry, submitted a review with no top-level comments, so my comments only showed up as responses to earlier unresolved threads. Calling them out here for visibility:
https://github.com/NVIDIA/cudf-spark-jni/pull/4593#discussion_r3573148113
https://github.com/NVIDIA/cudf-spark-jni/pull/4593#discussion_r3573639003
and the previously unresolved thread: https://github.com/NVIDIA/cudf-spark-jni/pull/4593#discussion_r3478747273

nvauto:

nvauto:

PR #4607, opened 2026-05-20
Harden the `noexcept` `deallocate()` path for SparkResourceAdaptorJNI

mythrocks:
This commit hardens `spark_resource_adaptor_impl::deallocate()` against unexpected exceptions in its noexcept path. Instead of allowing an exception to escape and implicitly trigger termination through the RMM/CCCL wrapper stack, the code now catches unexpected failures, attempts best-effort logging, and then explicitly calls `std::terminate()`.

As an aside, this commit also documents why `get_property(..., cuda::mr::device_accessible)` has an empty definition.  (It has to do with tagging the resource as returning device accessible memory.)


mythrocks:
Argh. Closing this in favour of a new PR against 26.06.

PR #4608, opened 2026-05-20
Harden the `noexcept` `deallocate()` path for SparkResourceAdaptorJNI

mythrocks:
This commit hardens `spark_resource_adaptor_impl::deallocate()` against unexpected exceptions in its noexcept path. Instead of allowing an exception to escape and implicitly trigger termination through the RMM/CCCL wrapper stack, the code now catches unexpected failures, attempts best-effort logging, and then explicitly calls `std::terminate()`.

As an aside, this commit also documents why `get_property(..., cuda::mr::device_accessible)` has an empty definition.  (It has to do with tagging the resource as returning device accessible memory.)

### DIFF
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -2865,7 +2865,7 @@ Java_com_nvidia_spark_rapids_jni_SparkResourceAdaptor_shutdownLoggerNative(JNIEn
 }
 
 /**
- * Test utility for RmmSparkDeallocationFailureTest.
+ * Test utility for RmmSparkDeallocationFailureTest.Child.
  *
  * This deliberately drives spark_resource_adaptor_impl::deallocate() into its fatal exception path.
  * It installs a test logger that throws when the normal DEALLOC status is logged, constructs a
@@ -2877,7 +2877,7 @@ Java_com_nvidia_spark_rapids_jni_SparkResourceAdaptor_shutdownLoggerNative(JNIEn
  * terminates.
  */
 JNIEXPORT void JNICALL
-Java_com_nvidia_spark_rapids_jni_RmmSparkDeallocationFailureTest_triggerDeallocationFailureForTesting(
+Java_com_nvidia_spark_rapids_jni_RmmSparkDeallocationFailureTest_00024Child_triggerDeallocationFailureForTesting(
   JNIEnv* env, jclass)
 {
   JNI_TRY
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -22,6 +22,7 @@
 #include <jni_cccl_any_resource.hpp>
 #include <pthread.h>
 #include <spdlog/common.h>
+#include <spdlog/sinks/base_sink.h>
 #include <spdlog/sinks/basic_file_sink.h>
 #include <spdlog/sinks/null_sink.h>
 #include <spdlog/sinks/ostream_sink.h>
@@ -33,10 +34,13 @@
 #include <chrono>
 #include <exception>
 #include <format>
+#include <iostream>
 #include <map>
 #include <optional>
 #include <set>
 #include <sstream>
+#include <stdexcept>
+#include <string>
 #include <unordered_map>
 #include <unordered_set>
 
@@ -2303,6 +2307,62 @@ inline spark_resource_adaptor get_spark_adaptor(jlong handle)
   return it->second;  // Returning by value, while under lock.
 }
 
+// Support for RmmSparkDeallocationFailureTest. The test runs this in a child JVM so the
+// intentional std::terminate() does not kill the JUnit process.
+class test_deallocate_noop_resource {
+ public:
+  void* allocate(cuda::stream_ref, std::size_t, std::size_t = alignof(std::max_align_t))
+  {
+    return nullptr;
+  }
+
+  void deallocate(cuda::stream_ref,
+                  void*,
+                  std::size_t,
+                  std::size_t = alignof(std::max_align_t)) noexcept
+  {
+  }
+
+  void* allocate_sync(std::size_t bytes, std::size_t alignment = alignof(std::max_align_t))
+  {
+    return allocate(cuda::stream_ref{cudaStream_t{nullptr}}, bytes, alignment);
+  }
+
+  void deallocate_sync(void* ptr,
+                       std::size_t bytes,
+                       std::size_t alignment = alignof(std::max_align_t)) noexcept
+  {
+    deallocate(cuda::stream_ref{cudaStream_t{nullptr}}, ptr, bytes, alignment);
+  }
+
+  bool operator==(test_deallocate_noop_resource const&) const noexcept { return true; }
+
+  bool operator!=(test_deallocate_noop_resource const&) const noexcept { return false; }
+
+  friend void get_property(test_deallocate_noop_resource const&,
+                           cuda::mr::device_accessible) noexcept
+  {
+  }
+};
+static_assert(cuda::mr::resource_with<test_deallocate_noop_resource, cuda::mr::device_accessible>);
+
+class throw_on_dealloc_log_sink : public spdlog::sinks::base_sink<std::mutex> {
+ protected:
+  void sink_it_(spdlog::details::log_msg const& msg) override
+  {
+    auto const payload = std::string(msg.payload.data(), msg.payload.size());
+    if (payload.rfind("DEALLOC", 0) == 0) {
+      throw std::runtime_error("injected deallocate failure");
+    }
+
+    spdlog::memory_buf_t formatted;
+    formatter_->format(msg, formatted);
+    std::cerr.write(formatted.data(), static_cast<std::streamsize>(formatted.size()));
+  }
+
+  void flush_() override { std::cerr.flush(); }
+};
+
 }  // namespace
 
 extern "C" {
@@ -2801,6 +2861,39 @@ JNIEXPORT void JNICALL
 Java_com_nvidia_spark_rapids_jni_SparkResourceAdaptor_shutdownLoggerNative(JNIEnv* env, jclass)
 {
   JNI_TRY { shutdown_global_logger(); }
+  JNI_CATCH(env, );
+}
+
+/**
+ * Test utility for RmmSparkDeallocationFailureTest.
+ *
+ * This deliberately drives spark_resource_adaptor_impl::deallocate() into its fatal exception path.
+ * It installs a test logger that throws when the normal DEALLOC status is logged, constructs a
+ * spark_resource_adaptor_impl around a no-op upstream resource, registers the current thread as a
+ * task thread, then calls deallocate(). The injected logging exception is caught by deallocate()'s
+ * noexcept guard, which should log the deallocation failure and call std::terminate().
+ *
+ * This must only be called from a subprocess death-style test, because success means the process
+ * terminates.
+ */
+JNIEXPORT void JNICALL
+Java_com_nvidia_spark_rapids_jni_RmmSparkDeallocationFailureTest_triggerDeallocationFailureForTesting(
+  JNIEnv* env, jclass)
+{
+  JNI_TRY
+  {
+    auto logger = std::make_shared<spdlog::logger>("SPARK_RMM_TEST",
+                                                   std::make_shared<throw_on_dealloc_log_sink>());
+    logger->set_error_handler([](std::string const& msg) { throw std::runtime_error(msg); });
+    set_global_logger(std::make_shared<spark_resource_adaptor_logger>(logger, true));
+
+    auto adaptor = cuda::mr::make_shared_resource<spark_resource_adaptor_impl>(
+      env, cuda::mr::any_resource<cuda::mr::device_accessible>{test_deallocate_noop_resource{}});
+
+    auto const tid = static_cast<long>(pthread_self());
+    adaptor->start_dedicated_task_thread(tid, 1);
+    adaptor->deallocate(cuda::stream_ref{cudaStream_t{nullptr}}, reinterpret_cast<void*>(0x1), 1);
+  }
   JNI_CATCH(env, );
 }
 }
--- a/src/main/cpp/src/SparkResourceAdaptorJni.cpp
+++ b/src/main/cpp/src/SparkResourceAdaptorJni.cpp
@@ -2185,16 +2185,52 @@ class spark_resource_adaptor_impl {
     wake_next_highest_priority_blocked(lock, is_for_cpu);
   }
 
+ private:
+  [[noreturn]] static void log_deallocate_failure_and_terminate(std::exception_ptr failure) noexcept
+  {
+    try {
+      auto const tid = static_cast<long>(pthread_self());
+      try {
+        if (failure) { std::rethrow_exception(failure); }
+        LOG_STATUS("ERROR",
+                   tid,
+                   -2,
+                   thread_state::UNKNOWN,
+                   "deallocate failed with unknown exception; terminating");
+      } catch (std::exception const& e) {
+        LOG_STATUS(
+          "ERROR", tid, -2, thread_state::UNKNOWN, "deallocate failed; terminating: {}", e.what());
+      } catch (...) {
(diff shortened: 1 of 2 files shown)


res-life:
From AI：Missing test coverage for new deallocate error path
  
  Location: src/main/cpp/src/SparkResourceAdaptorJni.cpp:2222-2233 (and helper log_deallocate_failure_and_terminate at lines 2188-2214)

  Issue

  This PR introduces two new behavioral code paths that have no corresponding test coverage:
  1. The new catch(...) error path in deallocate() (lines 2229-2233)
  2. The new log_deallocate_failure_and_terminate() helper (lines 2188-2214)

  The codebase already has injection mechanisms for allocation failures (forceRetryOOM, forceCudfException, etc. in RmmSparkTest.java), but there is no analogous mechanism to force
  resource.deallocate() to throw and exercise the new catch block. The success path of deallocate() is covered implicitly by existing tests, but the new error-hardening path is not verified at all.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -2185,16 +2185,52 @@ class spark_resource_adaptor_impl {
...
+  {
+    try {
+      auto const tid = static_cast<long>(pthread_self());
+      try {
+        if (failure) { std::rethrow_exception(failure); }
+        LOG_STATUS("ERROR",
+                   tid,
+                   -2,
+                   thread_state::UNKNOWN,
+                   "deallocate failed with unknown exception; terminating");
```
igorpeshansky: greptile is correct — this can throw and cause the wrong error to be reported. We should only have the `std::rethrow_exception` in the inner `try`. How about an IIFE (can be a `do`/`while` loop with a `break` instead if you prefer)? As a bonus, we don't repeat the unknown exception logging code:

```c++
    try {
      auto const tid = static_cast<long>(pthread_self());
      [&]() {
        try {
          if (failure) { std::rethrow_exception(failure); }
        } catch (std::exception const& e) {
          LOG_STATUS(
            "ERROR", tid, -2, thread_state::UNKNOWN, "deallocate failed; terminating: {}", e.what());
          return;
        } catch (...) {
        }
        LOG_STATUS("ERROR",
                   tid,
                   -2,
                   thread_state::UNKNOWN,
                   "deallocate failed with unknown exception; terminating");
      }();
    } catch (...) {
      // Logging is best-effort here. This function must not allow any exception to escape.
    }
    std::terminate();
```

src/main/cpp/src/SparkResourceAdaptorJni.cpp
```diff
@@ -2185,16 +2185,52 @@ class spark_resource_adaptor_impl {
...
+      } catch (...) {
+        LOG_STATUS("ERROR",
+                   tid,
+                   -2,
+                   thread_state::UNKNOWN,
+                   "deallocate failed with unknown exception; terminating");
+      }
+    } catch (...) {
+      // Logging is best-effort here. This function must not allow any exception to escape.
+    }
```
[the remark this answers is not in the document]
mythrocks: @greptile-apps:  Please check if the latest commit addresses your concern.

src/main/cpp/src/SparkResourceAdaptorJni.cpp
mythrocks: Agreed. This should be fixed now.  I've also refactored and simplified the checks.

For starters, the `if (failure)` was complicating things.  Having another helper function makes it more readable, I think.  

@igorpeshansky, @pmattione-nvidia:  Does this look better?

igorpeshansky: Yes, looks better. Good observation that `log_deallocate_failure_and_terminate` is only ever called with a non-null `std::exception_ptr` — that simplifies the inner handler considerably. Now that early return is not required, I'm ambivalent about the separate function vs inlining it in the `try` block, though it does save you an indentation level.

igorpeshansky (dismissed):
LGTM :shipit:

mythrocks:
@res-life:  There are a couple of reasons why I'm not inclined to add specific tests for this, unless the reviewers insist on it.

The main reason is that this is a catastrophic (and likely rare) failure case:  In the event of an exception, the program terminates via `std::terminate()`.  It is an improvement on how `spark-rapids-jni` might fail out, but it still terminates the program.

I suppose it might be possible to do the following:
1. Write a test program that launches a separate process, and checks the error code returned.
2. The launched process (from Java) would need to install a custom memory resource that throws. (Alternatively, a custom spdlog logger-sink that throws.). This implies a test-specific native entry point in `SparkResourceAdaptorJni.cpp`.
3. The test should then attempt to allocate memory, thus causing the exception thrown to cause std::terminate() to be called.

I'm not convinced the juice is worth the squeeze.  @pmattione-nvidia, @igorpeshansky:  What do you chaps reckon?

igorpeshansky:
> @res-life: There are a couple of reasons why I'm not inclined to add specific tests for this, unless the reviewers insist on it.
> 
> The main reason is that this is a catastrophic (and likely rare) failure case: In the event of an exception, the program terminates via `std::terminate()`. It is an improvement on how `spark-rapids-jni` might fail out, but it still terminates the program.
> 
> I suppose it might be possible to do the following:
> 
> 1. Write a test program that launches a separate process, and checks the error code returned.
> 2. The launched process (from Java) would need to install a custom memory resource that throws. (Alternatively, a custom spdlog logger-sink that throws.). This implies a test-specific native entry point in `SparkResourceAdaptorJni.cpp`.
> 3. The test should then attempt to allocate memory, thus causing the exception thrown to cause std::terminate() to be called.
> 
> I'm not convinced the juice is worth the squeeze. @pmattione-nvidia, @igorpeshansky: What do you chaps reckon?

googletest does support "death tests"[^1], which do exactly what you've described, and could potentially be reused here (even checking stderr). I don't think Java allows application crash tests anyway, so the tests would have to be initiated from C++ (no new file-specific entrypoint, but definitely a new build target). We could have a googletest test fixture construct an in-process JVM via JNI, which then calls your code — a bit heavyweight, but can be made to work.

[^1]: https://google.github.io/googletest/advanced.html#death-tests

mythrocks:
@igorpeshansky, @res-life: I had an agent have a go at what I was describing.  I think `RmmSparkDeallocationFailureTest` should address the testing for this path.

mythrocks:
Build

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
```diff
@@ -0,0 +1,112 @@
...
+ * after logging the deallocation failure. The parent JUnit test asserts that the child exited with
+ * a nonzero status and that the expected failure log was emitted.
+ */
+public class RmmSparkDeallocationFailureTest {
+  private static final String CHILD_ARG = "--trigger-deallocation-failure";
+  private static final String EXPECTED_LOG =
+      "deallocate failed; terminating: injected deallocate failure";
+
+  static {
+    NativeDepsLoader.loadNativeDeps();
```
igorpeshansky: I realize you're reusing the same class for both the test driver and the code under test, but this call is only needed in the latter, so seems like a bit of a waste…

To solve this yet keep everything in one file, you could isolate the code under test in a nested class, e.g.:

```java
public class RmmSparkDeallocationFailureTest {
  private static final String CHILD_ARG = "--trigger-deallocation-failure";
  private static final String EXPECTED_LOG =
      "deallocate failed; terminating: injected deallocate failure";

  @Test
  public void testDeallocateFailureLogsAndTerminates() throws Exception {
    Process process = new ProcessBuilder(childCommand()).redirectErrorStream(true).start();
    …
  }

  private static List<String> childCommand() {
    List<String> command = new ArrayList<>();
    command.add(javaExecutable());
    …
    command.add(Child.class.getName());
    command.add(CHILD_ARG);
    return command;
  }
  …
  public static class Child {
    static { NativeDepsLoader.loadNativeDeps(); }

    public static void main(String[] args) {
      …
      if (!Arrays.equals(new String[]{CHILD_ARG}, args)) {
        throw new IllegalArgumentException("unexpected child arguments");
      }
      triggerDeallocationFailureForTesting();
    }

    private static native void triggerDeallocationFailureForTesting();
  }
}
```

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    String output = readFully(process.getInputStream());
+    assertTrue(exited, "child JVM did not exit; output:\n" + output);
+    assertNotEquals(0, process.exitValue(), "child JVM unexpectedly succeeded; output:\n" + output);
+    assertTrue(output.contains(EXPECTED_LOG), "expected log line missing; output:\n" + output);
+  }
+
+  public static void main(String[] args) {
+    if (args.length != 1 || !CHILD_ARG.equals(args[0])) {
```
igorpeshansky: Why not use array equality?

```suggestion
    if (!Arrays.equals(new String[]{CHILD_ARG}, args)) {
```

You could, in fact, declare `public static final String[] CHILD_ARGS = {"--trigger-deallocation-failure"};`, and then use `Arrays.equals(CHILD_ARGS, args)`…

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
```diff
@@ -0,0 +1,112 @@
...
+  private static String javaExecutable() {
+    return System.getProperty("java.home") + "/bin/java";
+  }
+
+  private static String testClassPath() {
+    String classPath = System.getProperty("surefire.test.class.path");
+    if (classPath == null || classPath.isEmpty()) {
+      classPath = System.getProperty("java.class.path");
+    }
+    return classPath;
```
igorpeshansky: [Optional] If you invert the test, you can turn it into a guard clause and avoid re-assigning the local:

```suggestion
    String classPath = System.getProperty("surefire.test.class.path");
    if (classPath != null && !classPath.isEmpty()) return classPath;
    return System.getProperty("java.class.path");
```

If an empty value is not a concern, this could even be:

```java
    return System.getProperty("surefire.test.class.path", System.getProperty("java.class.path"));
```

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  private static String readFully(InputStream stream) throws IOException {
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+    byte[] buffer = new byte[8192];
+    int amountRead;
+    while ((amountRead = stream.read(buffer)) != -1) {
+      output.write(buffer, 0, amountRead);
+    }
```
igorpeshansky, line 98: [Optional] If you're on Java 9+:

```suggestion
    stream.transferTo(output);
```

(https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#transferTo(java.io.OutputStream))

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    command.add("-cp");
+    command.add(testClassPath());
+    command.add(RmmSparkDeallocationFailureTest.class.getName());
+    command.add(CHILD_ARG);
+    return command;
+  }
+
+  private static String javaExecutable() {
+    return System.getProperty("java.home") + "/bin/java";
```
igorpeshansky: [Optional] I doubt this would ever run on Windows, but this should technically use `File.separator`.

One possible robust alternative on Java 9+ could be `ProcessHandle.current().info().command().get()` (https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ProcessHandle.Info.html#command()).

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    String libraryPath = System.getProperty("java.library.path");
+    if (libraryPath != null && !libraryPath.isEmpty()) {
+      command.add("-Djava.library.path=" + libraryPath);
+    }
+
+    command.add("-cp");
+    command.add(testClassPath());
+    command.add(RmmSparkDeallocationFailureTest.class.getName());
+    command.add(CHILD_ARG);
```
igorpeshansky: [Optional]

```suggestion
    command.addAll(Arrays.asList("-cp", testClassPath(), RmmSparkDeallocationFailureTest.class.getName(), CHILD_ARG));
```

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
mythrocks: Ah, right.  Thank you.  This should be addressed now.

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
mythrocks: Yep, agreed.  This is cleaner.

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
mythrocks: That would've been shorter.  Sadly, the `<maven.compiler.target>` is `1.8`. :/

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
mythrocks: Again, it's Java 8.

But it turns out that we can use `Paths.get()`.  I'll post that shortly.

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
mythrocks: Yes, sir. Fixed.

mythrocks:
Build

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
igorpeshansky: Not quite what I proposed — with your new code, this would have been:

```java
    command.addAll(Arrays.asList("-cp", testClassPath(), Child.class.getName()));
    command.addAll(Arrays.asList(CHILD_ARGS));
```

But yours is more consistent anyway…

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
```diff
@@ -0,0 +1,116 @@
...
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+    byte[] buffer = new byte[8192];
+    int amountRead;
+    while ((amountRead = stream.read(buffer)) != -1) {
+      output.write(buffer, 0, amountRead);
+    }
+    return output.toString("UTF-8");
+  }
+
+  public static class Child {
```
igorpeshansky, line 102: [Optional] This actually didn't have to be `public`…

igorpeshansky (approved):
What you have here is even more heavyweight than what I proposed in https://github.com/NVIDIA/spark-rapids-jni/pull/4608#issuecomment-4513040600 — it runs two JVMs concurrently instead of one via JNI. But I guess it isn't such a big deal for testing, so your approach is fine too.

src/test/java/com/nvidia/spark/rapids/jni/RmmSparkDeallocationFailureTest.java
mythrocks: I'll check this version of things in, to catch the 26.06 train.  I can circle back to tighten up the access.

mythrocks:
Thank you for the review, @res-life, @igorpeshansky.  I've merged this change.

PR #4628, opened 2026-05-26
Fix row_conversion data corruption, races, and resource bugs

### LINKED ISSUE #4586: [BUG] unsupported types for columnar to row conversion are not rejected
**Describe the bug**
Some data types are not supported by row_conversion.cu including lists, structs, and decimal128 values. These are not rejected when someone tries to use them and can result in data corruption. Even though we do not really support logically sliced columns this too can cause problems if it ever showed up.

```
TEST_F(ColumnToRowTests, Decimal128)
{
  // 3 rows with distinct DECIMAL128 values; scale -2 so the rep values are unambiguous.
  std::vector<__int128_t> vals{12345LL, -67890LL, 999999999999LL};
  cudf::test::fixed_point_column_wrapper<__int128_t> col(
    vals.begin(), vals.end(), numeric::scale_type{-2});
  cudf::table_view in({col});
  std::vector<cudf::data_type> schema{cudf::data_type{cudf::type_id::DECIMAL128, -2}};

  auto rows   = spark_rapids_jni::convert_to_rows(in);
  auto result = spark_rapids_jni::convert_from_rows(cudf::lists_column_view(*rows[0]), schema);

  CUDF_TEST_EXPECT_TABLES_EQUIVALENT(in, *result);
}

// Bug: convert_to_rows does not handle sliced (offset > 0) column views correctly.
// Two problems interact:
// 1. c.data<int8_t>() applies offset_ in bytes (not elements), so for an INT32 column
//    with offset=2 the returned pointer is head+2 rather than head+8. The kernel then
//    reads a 4-byte int32 from a 2-byte-aligned address, causing a GPU misaligned-address
//    fault and a sticky CUDA error on the next API call.
// 2. c.null_mask() is not adjusted for offset_, so validity bits are read from the wrong
//    positions in the bitmask.
// The expected failure is a GPU hardware fault (cudaErrorMisalignedAddress) that poisons
// the CUDA context for all subsequent tests. This test is placed last among the bug repros
// for exactly that reason. See docs/row_conversion_bugs.md Bug 2.
TEST_F(ColumnToRowTests, SlicedColumnValidity)
{
  // 10-row source column; rows 3 and 7 are null.
  std::vector<int32_t> values{10, 11, 12, 13, 14, 15, 16, 17, 18, 19};
  std::vector<bool> validity{true, true, true, false, true, true, true, false, true, true};
  cudf::test::fixed_width_column_wrapper<int32_t> source(
    values.begin(), values.end(), validity.begin());

  // Slice rows [2, 8) → relative rows 0..5; nulls at relative rows 1 and 5.
  auto sliced = cudf::slice(source, {2, 8})[0];
  cudf::table_view in({sliced});
  std::vector<cudf::data_type> schema{cudf::data_type{cudf::type_id::INT32}};

  auto rows   = spark_rapids_jni::convert_to_rows(in);
  auto result = spark_rapids_jni::convert_from_rows(cudf::lists_column_view(*rows[0]), schema);

  // Expected: values [12..17] with nulls at relative positions 1 and 5.
  cudf::test::fixed_width_column_wrapper<int32_t> expected(
    values.begin() + 2, values.begin() + 8, validity.begin() + 2);
  CUDF_TEST_EXPECT_TABLES_EQUIVALENT(cudf::table_view({expected}), *result);
}
```

**Expected behavior**
We get a clear exception instead of data corruption, or it just works.

### LINKED ISSUE #4587: [BUG] row_conversion can hang for very long strings
**Describe the bug**
If we get over 2 GiB of string data in under 32 rows (I know it is a bit extreme) the row conversion code can hang (busy loop) forever.

```
TEST_F(ColumnToRowTests, LargeStringRowBatchHang)
{
  // Two string columns, 35 MiB per string, 33 rows.
  // Each column's total chars ≈ 1155 MiB < INT_MAX; combined per-row size ≈ 70 MiB.
  constexpr std::size_t per_col_bytes = 35ULL * 1024 * 1024;
  constexpr int num_rows              = 33;

  std::string const s(per_col_bytes, 'x');
  std::vector<std::string> data(num_rows, s);

  cudf::test::strings_column_wrapper col_a(data.begin(), data.end());
  cudf::test::strings_column_wrapper col_b(data.begin(), data.end());
  cudf::table_view in({col_a, col_b});
  std::vector<cudf::data_type> schema{cudf::data_type{cudf::type_id::STRING},
                                      cudf::data_type{cudf::type_id::STRING}};

  // Without the fix this hangs; with the guard it throws (which also fails this test).
  // Either way the test fails until the batching logic is corrected.
  auto rows   = spark_rapids_jni::convert_to_rows(in);
  auto result = spark_rapids_jni::convert_from_rows(cudf::lists_column_view(*rows[0]), schema);
  CUDF_TEST_EXPECT_TABLES_EQUIVALENT(in, *result);
}
```

**Expected behavior**
We get back the correct data, as this is valid or at least we throw an exception for this corner case.

```
diff --git a/src/main/cpp/src/row_conversion.cu b/src/main/cpp/src/row_conversion.cu
index 19a299273..083e60c62 100644
--- a/src/main/cpp/src/row_conversion.cu
+++ b/src/main/cpp/src/row_conversion.cu
@@ -1512,6 +1512,12 @@ batch_data build_batches(size_type num_rows,
       thrust::lower_bound(rmm::exec_policy(stream), search_start, search_end, MAX_BATCH_SIZE);
     size_type const batch_size = lb - search_start;
 
+    // If the 2 GiB boundary is hit with fewer than 32 rows the 32-row alignment rounding
+    // produces a zero step and last_row_end never advances, hanging forever. Throw instead.
+    CUDF_EXPECTS(lb == search_end || cudf::util::round_down_safe(batch_size, 32) > 0,
+                 "Row too large: the cumulative encoded size of fewer than 32 rows exceeds "
+                 "MAX_BATCH_SIZE (2 GiB). Reduce per-row data volume.");
+
     size_type const row_end = lb == search_end
                                 ? batch_size + last_row_end
                                 : last_row_end + cudf::util::round_down_safe(batch_size, 32);
```

will turn the hang into an exception, which is better than nothing.

### LINKED ISSUE #4588: [BUG] row_conversion can read uninitialized memory when there are multiple batches
**Describe the bug**
When trying to output multiple batches of string data, the second batch is skipped, which can result in uninitialized memory reads.

```
TEST_F(ColumnToRowTests, StringsMultiBatchSkip)
{
  // 2 string columns × 32.5 MiB per row × 35 rows. Per-column total ≈ 1108 MiB < INT_MAX.
  constexpr std::size_t per_col_bytes = 33ULL * 1024 * 1024;  // 33 MiB keeps totals safe
  constexpr int num_rows              = 35;

  // Use distinct per-row content so wrong output is detectable.
  std::vector<std::string> data_a, data_b;
  data_a.reserve(num_rows);
  data_b.reserve(num_rows);
  for (int i = 0; i < num_rows; ++i) {
    data_a.push_back(std::string(per_col_bytes, static_cast<char>('a' + (i % 26))));
    data_b.push_back(std::string(per_col_bytes, static_cast<char>('A' + (i % 26))));
  }

  cudf::test::strings_column_wrapper col_a(data_a.begin(), data_a.end());
  cudf::test::strings_column_wrapper col_b(data_b.begin(), data_b.end());
  cudf::table_view in({col_a, col_b});
  std::vector<cudf::data_type> schema{cudf::data_type{cudf::type_id::STRING},
                                      cudf::data_type{cudf::type_id::STRING}};

  auto rows = spark_rapids_jni::convert_to_rows(in);

  // Reconstruct each batch and compare — batch 1 (rows 32..34) will have wrong content.
  std::size_t row_start = 0;
  for (auto& batch : rows) {
    auto result  = spark_rapids_jni::convert_from_rows(cudf::lists_column_view(*batch), schema);
    auto in_slice = cudf::slice(in, {static_cast<cudf::size_type>(row_start),
                                     static_cast<cudf::size_type>(row_start + result->num_rows())})[0];
    CUDF_TEST_EXPECT_TABLES_EQUIVALENT(in_slice, *result);
    row_start += result->num_rows();
  }
}
```

results in 
```
[ RUN      ] ColumnToRowTests.StringsMultiBatchSkip
unknown file: Failure
C++ exception with description "std::bad_alloc: out_of_memory: CUDA error (failed to allocate 18446744071629177600 bytes) at: spark-rapids-jni/target/libcudf-install/include/rmm/mr/cuda_memory_resource.hpp:54: cudaErrorMemoryAllocation out of memory" thrown in the test body.
[  FAILED  ] ColumnToRowTests.StringsMultiBatchSkip (6397 ms)
```

  1. The test produces two batches: rows 0–31 in batch 0, rows 32–34 in batch 1.
  2. copy_strings_to_rows is launched for batch 1 with num_rows = 3 (the per-batch count) and batch_row_offset = 32.
  3. Inside the kernel, start_row = batch_row_offset + blockIdx.x * N = 32 + 0 = 32. The loop condition is row < min(num_rows=3, start_row+N), which evaluates to row < 3. Since start_row=32 >= 3, the loop body
  never executes. Nothing is written to the batch-1 output buffer.
  4. rmm::device_buffer is not zero-initialized — it comes from a pool and contains stale bytes from a prior allocation.
  5. When convert_from_rows decodes batch 1, it reads those stale bytes as string offset/length metadata — uint32_t pairs encoding where each string's character data lives. One of those garbage values lands as a
   string length.
  6. RMM faithfully tries to allocate the large memory and gets an OOM

**Expected behavior**
The conversion works, no crashes, multiple batches.

### LINKED ISSUE #4589: [BUG] Agent says there is a race in row_conversion on Ampere+
**Describe the bug**
Please note that the bug is theoretical, but the issue of the locks created and not used is real.

`cuda::barrier` in string kernels is uninitialized and per-thread (P0 on Ampere+)

**File:** `src/main/cpp/src/row_conversion.cu`
**Functions:** `copy_strings_to_rows` (line ~834), `copy_strings_from_rows` (line ~1126)

### What happens

Both string kernels declare the barrier as a plain local (stack) variable, never initialize it, and never wait on it before the kernel exits:

```cpp
cuda::barrier<cuda::thread_scope_block> block_barrier;  // not __shared__, not init'd
...
cuda::memcpy_async(warp, dst, src, string_length, block_barrier);
// kernel exits without block_barrier.arrive_and_wait()
```

Three violations of the `cuda::barrier` contract:
1. Must be `__shared__` (or at minimum visible to all participating threads); a per-thread local gives every thread an independent, incompatible copy.
2. Must be initialized with `init(&barrier, expected_count)` before first use.
3. Must be waited on (`.arrive_and_wait()`) before any thread exits to guarantee that the async copies issued through it have completed.

### Impact

On pre-Ampere hardware (Tesla, Volta, Turing), `cuda::memcpy_async` falls back to synchronous copies, so the barrier is never actually needed and the bug is dormant.

On **Ampere and later** (A100, H100, …), `cuda::memcpy_async` uses dedicated copy hardware (`cp.async`). Without a valid barrier and `arrive_and_wait()`, the hardware copy may still be in-flight when the kernel exits and the completion event is signaled to the host. Subsequent
operations on the same stream may read partially-written string data. The extent of corruption depends on PCIe/NVLink timing and is not deterministic.

The barrier defect is confirmed by code inspection. Reproducing actual data corruption requires an Ampere+ GPU and sufficient memory pressure to force the `cp.async` path to have pending work at kernel exit. A targeted test would need to compare string data byte-by-byte after conversion
on an A100/H100.

**Expected behavior**
No races, no crashes, no data corruption.

### LINKED ISSUE #4590: [BUG] Data race at tile column boundaries in row_conversion
**Describe the bug**
Please note that this was found by an AI agent. I have not reproduced the issue yet as it is a race condition.

**File:** `src/main/cpp/src/row_conversion.cu`
**Functions:** `copy_to_rows` (line ~678), `detail::determine_tiles` (line ~1720)

### What happens

`copy_to_rows` launches one CUDA block per tile. For each row in the tile, the block copies `tile_row_size` bytes from shared memory to the output buffer.

```cpp
auto const tile_row_size = tile.get_shared_row_size(col_offsets, col_sizes);
// tile_row_size = round_up_unsafe(col_offsets[end_col] + col_sizes[end_col]
//                                  - col_offsets[start_col], JCUDF_ROW_ALIGNMENT)
...
cuda::memcpy_async(warp, dst, src, tile_row_size, tile_barrier);
// dst = output_base + row_offset + col_offsets[start_col]
```

`tile_row_size` rounds the actual data size up to 8 bytes. If the last column of tile A ends at a byte position that is not 8-aligned, the round-up makes tile A write past the start of tile B's output region. Both tile A and tile B are independent kernel blocks running on potentially different SMs with no inter-block ordering guarantee. They write to overlapping output bytes concurrently: tile A writes uninitialized shared-memory padding, tile B writes
correct column data. The final value is nondeterministic.

The overlap requires:
- A tile split where tile A's last column ends at a byte offset that is not a multiple of 8.
- Tile B's first column has natural alignment < 8, so no gap is inserted between the tiles.

With all-INT32 or all-INT64 schemas, column offsets are always multiples of 4 or 8 respectively, and the geometry always works out cleanly. The condition can arise with mixed schemas (e.g., INT32 + INT8 + INT32) where the tile boundary lands at a non-8-aligned byte.

Concretely: a schema of [382 × INT32, 1 × INT8, 1 × INT8, …more INT8…, 1 × INT32, …] where the tile split fires after the last INT8 that pushes the padded row size past `shmem_limit / tile_height` (≈ 1536 bytes at 48 KB, tile_height = 32). If at that split point the cumulative data end is, say, 1533 bytes (round-up to 1536), and the next tile starts at byte 1533 (INT8, alignment 1), the first tile's write covers bytes [0, 1536) while the
second tile's write starts at 1533. Bytes [1533, 1536) are written by both blocks concurrently.

The exact schema that triggers the race depends on the interaction between the host-side `row_size` tracking in `determine_tiles` (which uses an approximation after each tile reset and the actual column offsets from `compute_column_information`. For subsequent tiles (after the first reset), the host-side `row_size` can diverge from the true byte offset, making the split fire at a different column than it would for the first tile. This divergence is what can produce a non-8-aligned boundary.

A deterministic repro is difficult because: (a) the triggering schema requires enough columns to exceed the shmem tile budget and a specific alignment pattern at the boundary; (b) even when the race exists, GPU SM scheduling may always order the two blocks' writes consistently, preventing observable corruption in practice. A stress test that runs many iterations with a triggering schema and checks for non-deterministic values in the overlapping bytes is the most reliable approach.

Suggested test schema (requires confirming exact column count via `determine_tiles` debug output): approximately [382 × INT32, 7 × INT8, 1 × INT32, 1 × INT32, …] with enough trailing columns to ensure the split fires at a non-8-aligned boundary.

thirtiseven:
## Summary

Address several latent bugs in `row_conversion.cu` that block re-enabling the `AcceleratedColumnarToRowIterator` fast path in spark-rapids (see NVIDIA/spark-rapids#14651 for context).

- **#4586 — data corruption and missing input guards.**
  - The default branch of `copy_to_rows` (used for any fixed-width column wider than 8 bytes, i.e. DECIMAL128 in practice) dropped the per-byte index in both the shared-memory write and the input read, so it copied one byte 16 times instead of 16 bytes once. Index both pointers by `i`.
  - Add `CUDF_EXPECTS` at the public entry points to reject unsupported column types (LIST / STRUCT / MAP / ...) and sliced columns (`column_view::offset() \!= 0`). The latter is necessary because `column_view::data<int8_t>()` returns `head<int8_t>() + offset`, which is byte-misaligned for any non-1-byte element type. The optimized variants additionally reject STRING up front (fail-fast) so callers do not get a misleading double-message.

- **#4587 — string hang.** `build_batches` looped forever when fewer than 32 rows already exceeded `MAX_BATCH_SIZE`, because `round_down_safe(batch_size, 32)` returned zero and `last_row_end` never advanced. Surface as an exception.

- **#4588 — multi-batch string skip.** `copy_strings_to_rows` was launched with the per-batch row count as `num_rows`, but the kernel computes an absolute `start_row` from `batch_row_offset + ...`. For every batch past the first, `start_row >= num_rows` and the loop body never executed, returning an uninitialized output buffer. Pass `batch_row_offset + batch_num_rows` (the absolute end bound) instead.

- **#4589 — cuda::barrier in string kernels.** Both `copy_strings_to_rows` and `copy_strings_from_rows` declared the barrier as a stack-local, never-initialized variable and never waited on it before kernel exit. Made it `__shared__`, initialized from thread 0 with `my_block.size()`, and added `arrive_and_wait()` before the kernel returns.

- **#4590 — tile-boundary race in copy_to_rows.** `tile_row_size` (`= round_up_8(actual_data_size)`) was used as the `memcpy_async` write length. When the cumulative byte width at a tile boundary was not a multiple of 8, the padding bytes tile A wrote spilled into the output range tile B was writing concurrently, producing non-deterministic corruption that depended on block scheduling. Added `tile_info::get_actual_row_size()`; the padded size stays as the in-shared row pitch (kept 8-aligned for SIMD access), but `memcpy_async` now writes / fetches only the un-padded actual span, so adjacent tiles never touch the same output bytes. Applied symmetrically to the `from-rows` fetch.

## Test plan

- [x] `ColumnToRowTests.Decimal128RoundTrip` — DECIMAL128 round trip that exercises the 16-byte default branch in copy_to_rows.
- [x] `ColumnToRowTests.TileBoundaryWideInt32RoundTrip` — 500-INT32 schema round-tripped 8 times to give the tile-boundary race more chances to surface on the un-fixed code path.
- [x] `ColumnToRowTests.RejectListColumn`, `RejectStructColumn`, `RejectSlicedColumn` — verify the new entry-point guards throw cleanly rather than letting corrupt data propagate.
- [x] `ColumnToRowTests.DISABLED_HugeStringRowThrows`, `DISABLED_MultiBatchStringDoesNotSkip` — the natural reproducers for #4587 and #4588 require ~2.5 GiB of GPU memory to build, so they are disabled by default and runnable manually with `--gtest_also_run_disabled_tests`.
- [x] Existing `ROW_CONVERSION` gtest suite (including `ColumnToRowTests.PivotLikeLayout` for #10062) still passes.

End-to-end integration coverage for the C2R path through `AcceleratedColumnarToRowIterator` lives in the matching spark-rapids PR that lands alongside the spark-rapids-jni version bump.

Fixes #4586, fixes #4587, fixes #4588, fixes #4589, fixes #4590.

### DIFF
--- a/src/main/cpp/src/row_conversion.cu
+++ b/src/main/cpp/src/row_conversion.cu
@@ -156,8 +156,23 @@ struct tile_info {
     // we are not losing data we are just not as efficient as we could be with shared memory. This
     // may be a problem if the tile is computed without regard to variable width offset/length sizes
     // in that we overrun shared memory.
+    //
+    // This rounded-up size is used for the per-row stride inside the shared memory tile only —
+    // global writes use get_actual_row_size() so adjacent tiles do not race on the padding bytes.
     return cudf::util::round_up_unsafe(
       col_offsets[end_col] + col_sizes[end_col] - col_offsets[start_col], JCUDF_ROW_ALIGNMENT);
+  }
+
+  // Real, un-padded data width spanned by this tile. Used as the memcpy_async length when
+  // writing the shared tile out to global memory (and when reading it in for the from-rows
+  // direction). Writing only the actual bytes prevents the JCUDF_ROW_ALIGNMENT rounding in
+  // get_shared_row_size() from spilling into the next tile's output region — a race that can
+  // corrupt data non-deterministically when the cumulative size at the tile boundary is not
+  // an 8-byte multiple (e.g. an odd number of INT32 columns in the first tile).
+  __device__ inline size_type get_actual_row_size(size_type const* const col_offsets,
+                                                  size_type const* const col_sizes) const
+  {
+    return col_offsets[end_col] + col_sizes[end_col] - col_offsets[start_col];
   }
 
   __device__ inline size_type num_cols() const { return end_col - start_col + 1; }
@@ -603,6 +618,7 @@ __launch_bounds__(block_size) CUDF_KERNEL void copy_to_rows(const size_type num_
   auto const num_tile_cols          = tile.num_cols();
   auto const num_tile_rows          = tile.num_rows();
   auto const tile_row_size          = tile.get_shared_row_size(col_offsets, col_sizes);
+  auto const actual_row_size        = tile.get_actual_row_size(col_offsets, col_sizes);
   auto const starting_column_offset = col_offsets[tile.start_col];
 
   // to do the copy we need to do n column copies followed by m element copies OR we have to do m
@@ -658,8 +674,10 @@ __launch_bounds__(block_size) CUDF_KERNEL void copy_to_rows(const size_type num_
         }
         case 1: shared_data[shared_offset] = *input_src; break;
         default: {
+          // Wide fixed-width types (e.g. DECIMAL128). Both destination and source must be
+          // indexed by byte; the original loop wrote to the same byte col_size times.
           for (int i = 0; i < col_size; ++i) {
-            shared_data[shared_offset] = *input_src;
+            shared_data[shared_offset + i] = input_src[i];
           }
           break;
         }
@@ -680,7 +698,9 @@ __launch_bounds__(block_size) CUDF_KERNEL void copy_to_rows(const size_type num_
     auto const src = &shared_data[tile_row_size * copy_row];
     auto const dst = tile_output_buffer + row_offsets(copy_row + tile.start_row, row_batch_start) +
                      starting_column_offset;
-    cuda::memcpy_async(warp, dst, src, tile_row_size, tile_barrier);
+    // Use actual_row_size (un-padded) so the JCUDF_ROW_ALIGNMENT rounding inside the shared
+    // tile never reaches into the next tile's output region — see tile_info::get_actual_row_size.
+    cuda::memcpy_async(warp, dst, src, actual_row_size, tile_barrier);
   }
 
   // wait on the last copies to complete
@@ -802,7 +822,9 @@ __launch_bounds__(block_size) CUDF_KERNEL
  *
  * @tparam block_size number of threads in a block.
  * @tparam RowOffsetFunctor iterator for row offsets into the destination data
- * @param num_rows number of rows in this portion of the table
+ * @param num_rows absolute end-row bound (== batch_row_offset + per-batch row count); the
+ *                 inner loop's `row < num_rows` guard treats this as the exclusive upper
+ *                 bound on absolute row indices for this batch.
  * @param num_variable_columns number of columns of variable-width data
  * @param variable_input_data variable width data column pointers
  * @param variable_col_output_offsets output offset information for variable-width columns
@@ -831,7 +853,14 @@ __launch_bounds__(block_size) CUDF_KERNEL
   // memcpy of the string data.
   auto const my_block = cooperative_groups::this_thread_block();
   auto const warp     = cooperative_groups::tiled_partition<cudf::detail::warp_size>(my_block);
-  cuda::barrier<cuda::thread_scope_block> block_barrier;
+
+  // The barrier must be __shared__ (so all threads in the block share one instance), explicitly
+  // initialized, and waited on before the kernel exits — otherwise cuda::memcpy_async issued
+  // through it on Ampere+ may still be in-flight when the kernel returns, racing with whatever
+  // reads the output next.
+  __shared__ cuda::barrier<cuda::thread_scope_block> block_barrier;
+  if (my_block.thread_rank() == 0) { init(&block_barrier, my_block.size()); }
+  my_block.sync();
 
   auto const start_row =
     blockIdx.x * NUM_STRING_ROWS_PER_BLOCK_TO_ROWS + warp.meta_group_rank() + batch_row_offset;
@@ -858,6 +887,9 @@ __launch_bounds__(block_size) CUDF_KERNEL
       offset += string_length;
     }
   }
+
+  // Ensure all cp.async copies issued through the barrier are complete before the kernel exits.
+  block_barrier.arrive_and_wait();
 }
 /**
  * @brief copy data from row-based format to cudf columns
@@ -910,10 +942,11 @@ __launch_bounds__(block_size) CUDF_KERNEL
   group.sync();
 
   {
-    auto const fetch_tile           = tile_infos[blockIdx.x];
-    auto const fetch_tile_start_row = fetch_tile.start_row;
-    auto const starting_col_offset  = col_offsets[fetch_tile.start_col];
-    auto const fetch_tile_row_size  = fetch_tile.get_shared_row_size(col_offsets, col_sizes);
+    auto const fetch_tile            = tile_infos[blockIdx.x];
+    auto const fetch_tile_start_row  = fetch_tile.start_row;
+    auto const starting_col_offset   = col_offsets[fetch_tile.start_col];
+    auto const fetch_tile_row_size   = fetch_tile.get_shared_row_size(col_offsets, col_sizes);
+    auto const fetch_actual_row_size = fetch_tile.get_actual_row_size(col_offsets, col_sizes);
     auto const row_batch_start =
       fetch_tile.batch_number == 0 ? 0 : batch_row_boundaries[fetch_tile.batch_number];
 
@@ -924,8 +957,9 @@ __launch_bounds__(block_size) CUDF_KERNEL
       auto shared_offset = (absolute_row - fetch_tile_start_row) * fetch_tile_row_size;
       auto dst           = &shared[shared_offset];
       auto src = &input_data[row_offsets(absolute_row, row_batch_start) + starting_col_offset];
-      // copy the data
-      cuda::memcpy_async(warp, dst, src, fetch_tile_row_size, tile_barrier);
+      // Shared stride stays padded (8-aligned for fast access); only fetch the bytes that
+      // actually belong to this tile so we never reach past the tile's column range.
+      cuda::memcpy_async(warp, dst, src, fetch_actual_row_size, tile_barrier);
     }
   }
 
@@ -1123,7 +1157,12 @@ __launch_bounds__(block_size) CUDF_KERNEL
   // Traversing in row-major order to coalesce the offsets and size reads.
   auto my_block = cooperative_groups::this_thread_block();
   auto warp     = cooperative_groups::tiled_partition<cudf::detail::warp_size>(my_block);
-  cuda::barrier<cuda::thread_scope_block> block_barrier;
+
+  // The barrier must be __shared__, initialized, and waited on; see the matching comment in
+  // copy_strings_to_rows.
+  __shared__ cuda::barrier<cuda::thread_scope_block> block_barrier;
+  if (my_block.thread_rank() == 0) { init(&block_barrier, my_block.size()); }
+  my_block.sync();
 
   // workaround for not being able to take a reference to a constexpr host variable
   auto const ROWS_PER_BLOCK = NUM_STRING_ROWS_PER_BLOCK_FROM_ROWS;
@@ -1147,6 +1186,8 @@ __launch_bounds__(block_size) CUDF_KERNEL
       cuda::memcpy_async(warp, dst, src, str_len[row], block_barrier);
     }
   }
+
+  block_barrier.arrive_and_wait();
 }
 
 /**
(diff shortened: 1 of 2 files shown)


thirtiseven:
Thanks @greptile-apps — addressed the misleading-double-message concern in 8439bae5. `is_supported_row_conversion_type` / `check_supported_columns` now take a `fixed_width_only` flag, and the two `_fixed_width_optimized` entry points pass `true`, so STRING is rejected up front with a single clear message rather than triggering both the new guard and the deeper `CUDF_FAIL`.

src/main/cpp/src/row_conversion.cu
```diff
@@ -658,8 +675,10 @@ __launch_bounds__(block_size) CUDF_KERNEL void copy_to_rows(const size_type num_
         }
         case 1: shared_data[shared_offset] = *input_src; break;
         default: {
+          // Wide fixed-width types (e.g. DECIMAL128). Both destination and source must be
+          // indexed by byte; the original loop wrote to the same byte col_size times.
```
revans2: nit: The comments are here to explain what the code does, not what it used to do. If you want to document what it used to do, please use the PR.

revans2 (dismissed):

src/main/cpp/src/row_conversion.cu
thirtiseven: Thanks for the review, removed this comment.

thirtiseven:
build

revans2 (approved):

PR #4635, opened 2026-05-27
[Part 2] [ORC-timezone]: Extract recurring DST rule from JVM timezone APIs

res-life:
## Context

Part 2 of the split of #4432 (full ORC DST GPU support). Part 1 (#4539, runtime build of `OrcTimezoneInfo`) is merged. This PR adds the Java-side machinery to **extract** a recurring DST rule for a zone, but does **not** yet wire it into the GPU dispatch path — production behavior is unchanged.

## What this adds

New file `OrcDstRuleExtractor.java` (package-private `final class`), which hosts:
- `static final class DstRule` — SimpleTimeZone-shaped DST descriptor (raw offset, dst savings, start/end month / day / dayOfWeek / time / mode), 0-based month and Calendar-style 1=Sun..7=Sat to match the GPU consumer.
- `extractDstRule(timezoneId, tz, rules)` — entry point. Returns `null` for non-DST zones (`rules.isFixedOffset()` or `!tz.useDaylightTime()`). Fail-fast sanity check that `tz` and `rules` describe the same zone (guards against the silent-GMT trap where `TimeZone.getTimeZone(id)` returns GMT for offset-style ids).
- **Path B (tried first):** `extractDstRuleByProbing` — probe `tz.getOffset()` hourly across a reference year, find the two transitions, encode them in SimpleTimeZone form. Captures what `java.util.TimeZone.getOffset` actually returns, which is the source of truth the GPU side must match.
- **Path A (fallback):** `extractDstRuleFromZoneRules` — derive from `ZoneRules.getTransitionRules()`. Used for zones whose recurring rule cannot be recovered by probing alone.
- Cross-year verification: the extracted rule is re-evaluated against `tz.getOffset` across anchor years **2060, 2400, 9997** to catch divergence well into the recurring-rule fallback regime.

In `OrcTimezoneInfo.java`:
- Widen `utcMillisForDate` and `binarySearchTransition` from `private` to package-private so `OrcDstRuleExtractor` can reuse the same UTC anchor and bracketed binary search. No behavioral change.

## What this does **not** change

- `GpuTimeZoneDB.convertOrcTimezones` — DST gate still throws `UnsupportedOperationException` for DST zones.
- `buildRuntimeOrcTimezoneInfo` — does not yet populate any `DstRule`.
- `OrcTimezoneInfo`'s public constructor / fields (the two widened methods are package-private, not public).
- JNI, CUDA.

`extractDstRule` is package-private and exercised only by `OrcTimezoneInfoTest`.

## Verification

Extracted rules cross-checked against tzdata; the test suite is green on **both JDK 8 (the build's `maven.compiler.target`) and JDK 17 Zulu**:

| Zone | start | end | dstSavings |
|------|-------|-----|------------|
| America/New_York | 2nd Sun of March, 02:00 standard | 1st Sun of November, 01:00 standard | +1h |
| Europe/London | last Sun of March, 01:00 standard | last Sun of October, 01:00 standard | +1h |
| Australia/Sydney | 1st Sun of October, 02:00 standard | 1st Sun of April, 02:00 standard | +1h |
| Asia/Shanghai, UTC, +05:30 | `null` | `null` | — |

Real-zone tests assert the exact (`month`, `day`, `dayOfWeek`, `time`, `mode`) shape for each zone. Synthetic-zone tests additionally cover both extraction paths (probing success, ZoneRules fallback), every `IllegalStateException` branch (unsupported rule count / shape, zero-/both-positive-/both-negative-delta, mismatched savings, negative day indicator, cross-year verification failure, tz/rules zone mismatch), and edge cases (`isMidnightEndOfDay`, `TimeDefinition.WALL`, February day-of-month clamp, multiple-DST-on probing guard). 26 tests total.

## Diff size

`+1314 / -2` across 3 files:
- `OrcDstRuleExtractor.java` `+529` (new file)
- `OrcTimezoneInfo.java` `+6 / -2` (visibility widening)
- `OrcTimezoneInfoTest.java` `+779`

No changes outside this slice.

## Follow-up parts

- **Part 3**: CUDA civil-calendar utilities + Java↔CUDA `DstRule` serialization scaffolding (still all dead-code).
- **Part 4**: CUDA DST kernel implementation + kernel-level C++ tests.
- **Part 5**: Remove the DST gate in `convertOrcTimezones`, dispatch DST zones through the new kernel, add end-to-end Java tests. This is the only follow-up part that changes runtime behavior.


res-life:
build

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java
```diff
@@ -0,0 +1,529 @@
...
+        // before LocalDate.of so it never throws, and clamp the result to a
+        // valid in-month day so utcMillisForDate cannot receive day <= 0.
+        int anchorDay = Math.min(ruleDay, monthLength);
+        int targetDayOfWeek = toCalendarDayOfWeek(
+            LocalDate.of(year, month + 1, anchorDay).getDayOfWeek().getValue());
+        int diff = targetDayOfWeek - ruleDayOfWeek;
+        if (diff < 0) diff += 7;
+        return Math.max(anchorDay - diff, 1);
+      }
+      case DstRule.MODE_DOM:
```
firestarman, line 524: NIT: Empty `case`, can be removed ?

src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java
```diff
@@ -0,0 +1,529 @@
...
+
+      // All sample points below land within the wall-clock year y for any
+      // real DST rule (boundaries ±12h, plus monthly noon-UTC anchors), so
+      // skip the per-sample year derivation and re-computation of these
+      // bounds inside computeDstOffset -- ~186 computeTransitionUtcMillis
+      // calls per verifyDstRule otherwise. Hand the already-computed bounds
+      // to the variant.
+      long[] boundaries = {dstStart, dstEnd};
+      for (long boundary : boundaries) {
+        long from = boundary - 12 * 3_600_000L;
```
firestarman, line 410: NIT: better name these literal values.

src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java
```diff
@@ -0,0 +1,529 @@
...
+   * for the last Sunday in October.
+   */
+  private static int[] decodeTransition(long utcMs, int rawOffsetMs) {
+    long localMs = utcMs + rawOffsetMs;
+    LocalDateTime ldt = LocalDateTime.ofInstant(Instant.ofEpochMilli(localMs), ZoneOffset.UTC);
+
+    int month = ldt.getMonthValue() - 1; // 0-based for Calendar compat
+    int dayOfMonth = ldt.getDayOfMonth();
+    int dayOfWeek = toCalendarDayOfWeek(ldt.getDayOfWeek().getValue());
+    int timeInDay = ldt.getHour() * 3_600_000
```
firestarman, line 359: NIT: Better name these literal values.

firestarman (approved):
LGTM, some nits, but good to merge it as-is.

src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java
res-life: Keep it → documents intent — it makes explicit that MODE_DOM is a recognized, handled mode (mapped to identity), rather than something silently swept into default.

PR #4636, opened 2026-05-27
Add native find_in_set utility

viadea:
Adds a native Spark-compatible `find_in_set` utility for the scalar-word plus string-column set case.

The new implementation scans each comma-delimited RHS row once on the GPU and returns the 1-based token position, 0 for missing values, and null for null RHS rows. This avoids materializing split list columns and the follow-up list search for the dynamic RHS path in Spark RAPIDS. It also includes a fast path for literal words containing commas, which Spark treats as no match.

Tests:
- Added Java coverage for first-match semantics, missing words, empty tokens, null RHS rows, UTF-8 tokens, and words containing commas.
- `git diff --cached --check`
- Local Maven compile was attempted, but this checkout does not have `thirdparty/cudf` initialized, so the lifecycle stops at the cuDF submodule check and direct Java test compilation cannot resolve `ai.rapids.cudf` classes. This PR is opened as draft for CI/native validation.

### DIFF
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -244,6 +244,7 @@ add_library(
   src/datetime_truncate.cu
   src/decimal_utils.cu
   src/exception_with_row_index_utilities.cu
+  src/find_in_set.cu
   src/format_float.cu
   src/from_json_to_raw_map.cu
   src/from_json_to_structs.cu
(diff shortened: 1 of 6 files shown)


viadea:
pre-commit.ci autofix

nvauto:

nvauto:

PR #4640, opened 2026-05-27
Added comments to empty methods in helper test classes, for clarity. [skip ci]

mythrocks:
This commit adds comments to the `deallocate` and `get_property` methods of the `test_deallocate_noop_resource` test utility, indicating why those method bodies are empty.

The short of it is that these methods are empty by design.  These are test utilities exercised by `RmmSparkDeallocationFailureTest`.  The methods have to have implementations for them to be valid memory resources.

Note:  This change is to address nitpicks in the static analyzer runs.


mythrocks:
We really ought not to waste time and resources on re-running tests on this change.

sameerz (approved):

sameerz:
build

PR #4642, opened 2026-05-28
Floor-divide timestamp to seconds in convert_timestamp

### LINKED ISSUE #14861: [BUG] test_comprehensive_from_utc_timestamp fails on Databricks 14.3 for timezone SystemV/EST5EDT (1-hour GPU/CPU diff) intermittently
**Describe the bug**
Build: rapids_databricks_nightly-dev-github/2247 (Databricks 14.3, CI_PART1)

Databricks 14.3 nightly CI_PART1 integration test failed with a single assertion error in test_comprehensive_from_utc_timestamp for timezone SystemV/EST5EDT. GPU returned -1757278800968948 while CPU returned -1757282400968948 (exactly 3600 seconds / 1 hour difference) when applying from_utc_timestamp under timezone SystemV/EST5EDT. All other 35400 tests in the run passed (1 failed, 35400 passed, 2827 skipped, 817 xfailed, 984 xpassed). The discrepancy looks like a DST/legacy-timezone rule mismatch between the GPU kernel and Spark/JVM for SystemV/EST5EDT.

Error logs:
```
FAILED src/main/python/date_time_test.py::test_comprehensive_from_utc_timestamp[2200-12-30 23:59:59.999999+00:00-SystemV/EST5EDT][DATAGEN_SEED=1779422092, TZ=UTC]
AssertionError: GPU (-1757278800968948) and CPU (-1757282400968948) int values are different at [324, 'from_utc_timestamp(a, SystemV/EST5EDT)']

cpu = -1757282400968948, gpu = -1757278800968948
path = [324, 'from_utc_timestamp(a, SystemV/EST5EDT)']

--- CPU OUTPUT
+++ GPU OUTPUT
@@ -322,7 +322,7 @@
 Row(from_utc_timestamp(a, SystemV/EST5EDT)=None)
 Row(from_utc_timestamp(a, SystemV/EST5EDT)=-29994983373768984)
 Row(from_utc_timestamp(a, SystemV/EST5EDT)=-51642154462707992)
-Row(from_utc_timestamp(a, SystemV/EST5EDT)=-1757282400968948)
+Row(from_utc_timestamp(a, SystemV/EST5EDT)=-1757278800968948)
 Row(from_utc_timestamp(a, SystemV/EST5EDT)=-53083413977565870)

= 1 failed, 35400 passed, 2827 skipped, 817 xfailed, 984 xpassed in 17141.46s =
```

**Environment details**
- Databricks Runtime 14.3 (Spark 3.5.0)
- TEST_MODE=CI_PART1, TZ=UTC
- DATAGEN_SEED=1779422092
- GPU: g5.4xlarge

res-life:
fixes https://github.com/NVIDIA/spark-rapids/issues/14861

## Summary

Fixes a 1-hour off-by-one in `from_utc_timestamp` and the ORC writer-reader rebase path that triggered when a negative microsecond input landed in the last sub-second window before a DST **gap** (spring-forward) transition.

Root cause is in `convert_timestamp` (`src/main/cpp/src/datetime_utils.cuh`):

```cpp
auto const epoch_seconds = static_cast<int64_t>(
  cuda::std::chrono::duration_cast<cudf::duration_s>(timestamp.time_since_epoch()).count());
```

`duration_cast<duration_s>` truncates toward zero. For a negative count with a non-zero sub-second part, this rounds *up* by one second:

- input `-1757264400968948` µs → seconds is mathematically `-1757264400.968948`
- truncation gives `-1757264400`
- correct floor gives `-1757264401`

If that exact second `-1757264400` is the UTC instant of a DST gap transition (e.g. `1914-04-26T07:00:00Z` for `SystemV/EST5EDT`'s last-Sunday-in-April spring-forward), the `upper_bound` call below treats the input as being *at* the transition rather than *before* it and returns the post-transition offset — a 1-hour overshoot.

## Direction matters in `convert_timestamp`

The fix in `datetime_utils.cuh::convert_timestamp` is **directional**:

- **`from_utc` (`to_utc == false`)**: input is a UTC instant. Floor-divide is correct — a UTC instant 1 µs before a gap belongs to the pre-transition offset.
- **`to_utc` (`to_utc == true`)**: input is a local wall clock. Sub-second negative inputs that fall *in* a DST gap (a local time that doesn't exist) must resolve to `offset_after` to match Spark/Java's `LocalDateTime.atZone()` convention. Truncation toward zero happens to land on the post-gap row, so we keep it for this direction.

The existing `TimeZoneTest.convertToUtc{Micro,Nano}SecondsTest` fixtures with local µs `-908838000000456` (in Asia/Shanghai's 1941 spring-forward gap) verify the to_utc semantics — they continue to pass after this PR.

## ORC path

`convert_timestamp_between_timezones` in `timezones.cu` had the same shape of bug: a `duration_cast<duration_ms>` truncation on the `timestamp_us` input. Unlike the bidirectional `convert_timestamp`, **every** lookup in the ORC algorithm is from_utc-style (`tz.getOffset(millis)` interprets its argument as a UTC instant), so floor-divide is the right semantic — no direction split needed.

The Java reference `GpuTimeZoneDBTest.convertOrcTimezonesOnCPU` was using `microseconds / microsPerMillis` (Java integer division, truncate toward zero) for the same µs→ms step, and reconstructed the result with `microseconds[i] % microsPerMillis`. Switched both to `Math.floorDiv` / `Math.floorMod` so:

1. The reference matches the GPU's new floor semantics.
2. The round-trip identity `(millis * 1000) + remainder == microseconds` still holds for negative inputs (which only works when div and mod share the same rounding convention).

## Reproduction

`integration_tests/src/main/python/date_time_test.py::test_comprehensive_from_utc_timestamp` with `DATAGEN_SEED=1779422092` on Databricks 14.3, timezone `SystemV/EST5EDT`. Row 324 hit a UTC instant 0.97 s before the 1914 spring-forward; CPU returned EST offset, GPU returned EDT (NVIDIA/spark-rapids#14861).

Equivalent C++ regression added in `src/main/cpp/tests/timezones.cpp::ConvertFromUTCMicrosecondsSubSecondBeforeGap` using the existing test transition table (gap at `-908870400` UTC seconds, offset 28800 → 32400). Before the fix, input `-908870400000001` µs returned offset 32400 (post-gap); after, it correctly returns 28800.

## Notes

- Positive timestamps are unaffected: truncation and floor agree for non-negative integers.
- Floor-divide goes through `integer_utils::floor_div`, which exists in `integer_utils.cuh` and is already used by other date-math helpers.

## Test plan

- [ ] `ConvertFromUTCMicrosecondsSubSecondBeforeGap` (new C++ regression) passes
- [ ] `convertTo/FromUtc{Micro,Nano}SecondsTest` (existing JUnit) still pass — verifies to_utc gap semantics preserved
- [ ] `testConvertOrcTimezones` (existing JUnit) still passes against the updated CPU reference
- [ ] Rerun `test_comprehensive_from_utc_timestamp` with `DATAGEN_SEED=1779422092` on Databricks 14.3
- [ ] Existing ORC integration tests in spark-rapids (`test_orc_non_utc_timezone`) still pass

### DIFF
--- a/src/test/java/com/nvidia/spark/rapids/jni/TimeZoneTest.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/TimeZoneTest.java
@@ -1,5 +1,5 @@
 /*
-* Copyright (c)  2023-2025, NVIDIA CORPORATION.
+* Copyright (c)  2023-2026, NVIDIA CORPORATION.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -403,7 +403,9 @@ public class TimeZoneTest {
           null);
         ColumnVector expected = ColumnVector.timestampMicroSecondsFromBoxedLongs(
           -1262260800000123L,
-          -908838000000456L,
+          // input -908870400000456 is 456 µs before the 1941 Shanghai spring-forward at
+          // UTC -908870400. Pre-transition offset is +8h (CST), so wall = utc + 28800.
+          -908841600000456L,
           -908837100000789L,
           -888800400000001L,
           -888799500000999L,
@@ -463,7 +465,8 @@ public class TimeZoneTest {
           null);
         ColumnVector expected = ColumnVector.timestampNanoSecondsFromBoxedLongs(
           -1262260800000000123L,
-          -908838000000000456L,
+          // Pre-transition offset (+8h CST) applies; see micros variant above.
+          -908841600000000456L,
           -908837100000000789L,
           -888800400000000001L,
           -888799500000000999L,
--- a/src/test/java/com/nvidia/spark/rapids/jni/TimeZoneTest.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/TimeZoneTest.java
@@ -1,5 +1,5 @@
 /*
-* Copyright (c)  2023-2026, NVIDIA CORPORATION.
+* Copyright (c) 2023-2026, NVIDIA CORPORATION.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
--- a/src/main/cpp/src/datetime_utils.cuh
+++ b/src/main/cpp/src/datetime_utils.cuh
@@ -536,8 +536,14 @@ __device__ static timestamp_type convert_timestamp(
   auto const tz_instants  = fixed_transitions.child().child(1);
   auto const utc_offsets  = fixed_transitions.child().child(2);
 
-  auto const epoch_seconds = static_cast<int64_t>(
-    cuda::std::chrono::duration_cast<cudf::duration_s>(timestamp.time_since_epoch()).count());
+  // Floor-divide the raw count to seconds. `duration_cast` truncates toward zero, which for
+  // negative timestamps with a non-zero sub-second component rounds up by one second. If such a
+  // timestamp lies in the last sub-second window before a DST gap transition, the rounded value
+  // snaps onto the transition instant itself, and the binary search below returns the
+  // post-transition offset (off by one DST step, typically 1 hour).
+  constexpr int64_t sub_seconds_per_second = duration_type::period::den;
+  auto const epoch_seconds =
+    integer_utils::floor_div(timestamp.time_since_epoch().count(), sub_seconds_per_second);
   auto const tz_transitions = cudf::list_device_view{fixed_transitions, tz_index};
   auto const list_size      = tz_transitions.size();
 
--- a/src/main/cpp/src/datetime_utils.cuh
+++ b/src/main/cpp/src/datetime_utils.cuh
@@ -536,14 +536,22 @@ __device__ static timestamp_type convert_timestamp(
   auto const tz_instants  = fixed_transitions.child().child(1);
   auto const utc_offsets  = fixed_transitions.child().child(2);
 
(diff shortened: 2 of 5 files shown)


res-life:
build

res-life:
build

res-life:
build

firestarman (commented):
==(By GPT 5.5)
LGTM overall. 
Nit on test coverage: the C++ side now has a nice targeted regression for the sub-second-before-gap case, but the ORC path still mostly relies on the existing randomized/reference-style coverage. It may be worth adding one focused Java regression for a negative sub-ms timestamp immediately before a gap transition, just to lock down the ORC-specific floorDiv/floorMod behavior as directly as the non-ORC path.

src/test/java/com/nvidia/spark/rapids/jni/TimeZoneTest.java
```diff
@@ -1,5 +1,5 @@
 /*
-* Copyright (c)  2023-2025, NVIDIA CORPORATION.
+* Copyright (c)  2023-2026, NVIDIA CORPORATION.
```
thirtiseven: nit
```suggestion
* Copyright (c) 2023-2026, NVIDIA CORPORATION.
```

thirtiseven (commented):
Code looks good, +1 for test coverage comment from firestarman.

res-life:
build

firestarman (approved):
LGTM

thirtiseven (approved):

PR #4644, opened 2026-05-28
from_protobuf kernel (part3): nested struct decode with scalar children

thirtiseven:
Part 3 of #4107. Implements **JNI-3b.1** and **JNI-3b.2** from the split plan in NVIDIA/spark-rapids#14069.

## Summary

Adds nested message field scan utilities and the first `build_nested_struct_column` path, wiring top-level `STRUCT` fields through the orchestrator for scalar numeric/bool children.

After this PR, the following decode correctly at the top level:
- `STRUCT { int32 / int64 / uint32 / uint64 / float / double / bool ... }`

Nested STRING / LIST<UINT8> / recursive STRUCT children and repeated-in-nested fields still produce typed null columns; those land in 3b.3 – 3b.6.

## Components

**Kernels** (`protobuf_kernels.cu`):
- `scan_nested_message_fields_kernel` — one thread per parent row, records last-one-wins locations for direct singleton children of a nested message. Repeated children are intentionally left to the dedicated count/scan path in 3b.5/3b.6.
- `extract_strided_locations_kernel` — pulls one field's per-row locations out of the 2D nested-locations array, replacing the previous D2H + CPU loop + H2D pattern.

**Builders** (`protobuf_builders.cu`):
- `build_nested_struct_column` — scans child field locations on device, then delegates each numeric/bool child to `extract_typed_column` with a `nested_location_provider`. Child types not yet supported in nested context are filled with schema-aware typed null columns so the output schema stays well-formed.

**Orchestrator** (`protobuf.cu`):
- After repeated-field processing, calls `build_nested_struct_column` for each top-level STRUCT field, using `launch_extract_strided_locations` to derive per-row parent locations from the existing `d_nested_locations` buffer.

## Intentionally out of scope

Deferred to subsequent PRs per the split plan in NVIDIA/spark-rapids#14069:
- **3b.3** — nested string/bytes/enum-as-string, default values inside nested messages, proto2 required-field checks for nested children
- **3b.4** — recursive nested STRUCT support, depth-aware grandchild parent location computation
- **3b.5** — count/scan kernels for repeated fields inside nested messages
- **3b.6** — `build_repeated_child_list_column` for repeated children

## Tests

Seven new tests in `ProtobufTest.java` covering the top-level STRUCT scalar path:
- `testNestedMessageInt32Child`
- `testNestedMessageLastOneWins`
- `testNestedMessageMultipleScalarChildren`
- `testNestedMessageAbsentParentIsNull`
- `testZeroLengthNestedMessageChildNullability`
- `testChildlessNestedMessagePresence`
- `testFailfastNestedRepeatedWrongWireType`

## Size

~740 lines added across 5 source files plus the test file; each source file stays under the ~300 LOC per-file target.


src/main/cpp/src/protobuf/protobuf_builders.cu
igorpeshansky: [Optional] Prior code, but `input_null_count` is unused here.

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1242,6 +1244,44 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     }  // for (ri)
   }
 
+  // Process nested struct fields (3b.2). Only scalar numeric/bool children decode for real;
+  // other child shapes are filled with typed null columns by build_nested_struct_column.
+  if (num_nested > 0) {
```
igorpeshansky: Nit: this is technically redundant — if `num_nested` is 0 or below, the next loop will be a no-op.

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1242,6 +1244,44 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+    for (int ni = 0; ni < num_nested; ni++) {
+      int parent_schema_idx    = nested_field_indices[ni];
+      auto child_field_indices = find_child_field_indices(schema, num_fields, parent_schema_idx);
```
igorpeshansky: [Optional] This does a full pass over all fields for every nested struct, which may become a performance issue for really large schemas with lots of nesting. No action required now, but may be worth at least a comment. If it does become an issue, we could pre-build a tree helper data structure in a single linear pass.

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -447,4 +447,169 @@ std::unique_ptr<cudf::column> build_repeated_string_column(
...
+    cudf::detail::make_pinned_vector_async<field_descriptor>(num_child_fields, stream);
+  for (int i = 0; i < num_child_fields; i++) {
+    int child_idx                             = child_field_indices[i];
+    h_child_field_descs[i].field_number       = schema[child_idx].field_number;
+    h_child_field_descs[i].expected_wire_type = static_cast<int>(schema[child_idx].wire_type);
+    h_child_field_descs[i].is_repeated        = schema[child_idx].is_repeated;
+  }
+
+  auto const scratch_mr          = cudf::get_current_device_resource_ref();
+  auto const child_desc_capacity = num_child_fields > 0 ? num_child_fields : 1;
```
igorpeshansky: Similar to https://github.com/NVIDIA/spark-rapids-jni/pull/4504#discussion_r3163746251, this could be `std::max` and inlined. Also for all other `rmm::device_uvector`s below…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -447,4 +447,169 @@ std::unique_ptr<cudf::column> build_repeated_string_column(
...
+                                    message_data_size,
+                                    list_offsets,
+                                    base_offset,
+                                    d_parent_locs.data(),
+                                    num_rows,
+                                    d_child_field_descs.data(),
+                                    num_child_fields,
+                                    d_child_locations.data(),
+                                    d_error.data(),
+                                    d_row_force_null.size() > 0 ? d_row_force_null.data() : nullptr,
```
igorpeshansky, line 542: [Optional] Technically, `device_uvector::data()` is nullptr when its size is 0, but I don't know if that's documented, so I suppose this is ok.

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -447,4 +447,169 @@ std::unique_ptr<cudf::column> build_repeated_string_column(
...
+                               enum_names,
+                               d_row_force_null,
+                               d_error,
+                               stream,
+                               mr,
+                               top_row_indices,
+                               propagate_invalid_rows));
+        break;
+      }
+      default:
```
igorpeshansky: It might be safer to list out the few remaining types as explicit cases (resulting in a null column), and have the default fail. Otherwise it's too easy to miss a case…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -504,6 +504,168 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
+  int64_t parent_row_start = static_cast<int64_t>(parent_row_offsets[row]) - parent_base_offset;
+  int64_t nested_start_off = parent_row_start + parent_loc.offset;
+  int64_t nested_end_off   = nested_start_off + parent_loc.length;
+  if (nested_start_off < 0 || nested_end_off > message_data_size) {
+    set_error_once(error_flag, ERR_BOUNDS);
+    mark_row_error();
+    return;
+  }
+  uint8_t const* nested_start = message_data + nested_start_off;
+  uint8_t const* nested_end   = nested_start + parent_loc.length;
```
igorpeshansky: Let's assign `message_data + nested_end_off`, to avoid replicating the logic in line 552…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -504,6 +504,168 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
+
+  uint8_t const* cur = nested_start;
+
+  while (cur < nested_end) {
```
igorpeshansky: This can be a `for` loop:


```suggestion
  for (uint8_t const* cur = nested_start; cur < nested_end; ) {
```

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+             new int[]{Protobuf.ENC_DEFAULT, Protobuf.ENC_DEFAULT},
+             new boolean[]{false, false},
+             new boolean[]{false, false},
+             new boolean[]{false, false},
+             new long[]{0, 0},
+             new double[]{0.0, 0.0},
+             new boolean[]{false, false},
+             new byte[][]{null, null},
+             new int[][]{null, null},
+             false)) {
```
igorpeshansky: I think I've mentioned in #4646 that these functions introduce a lot of boilerplate, contrary to the unit testing principles[^1] — it should be obvious what is being tested, by encapsulating the irrelevant bits of test setup. The `decodeRaw` function doesn't make the code easier to read — its javadoc comment says "Avoids verbose ProtobufSchemaDescriptor construction", but you've kept the most verbose part, which is the arguments. Adding a builder pattern would remove most of the default values, and would let you call `Protobuf.decodeToStruct` with the `ProtobufSchemaDescriptor` returned by the builder, which would be a lot shorter.

[^1]: E.g., https://abseil.io/resources/swe-book/html/ch12.html

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+      try (ColumnVector expectedX = ColumnVector.fromBoxedInts(42);
+           ColumnVector expectedInner = ColumnVector.makeStruct(expectedX);
+           ColumnVector expectedOuter = ColumnVector.makeStruct(expectedInner)) {
+        AssertUtils.assertStructColumnsAreEqual(expectedOuter, result);
+      }
+    }
+  }
+
+  @Test
+  void testNestedMessageLastOneWins() {
```
igorpeshansky: `LastOneWins` isn't very clear by itself. How about `testNestedMessageDuplicateFieldTags_LastOneWins` or something?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+  @Test
+  void testNestedMessageMultipleScalarChildren() {
+    // message Inner { int32 a = 1; int64 b = 2; bool c = 3; float d = 4; }
+    // message Outer { Inner inner = 1; }
+    Byte[] inner0 = concat(
+        box(tag(1, WT_VARINT)), box(encodeVarint(7)),
+        box(tag(2, WT_VARINT)), box(encodeVarint(123456789012L)),
+        box(tag(3, WT_VARINT)), box(encodeVarint(1)),
+        box(tag(4, WT_32BIT)), box(encodeFloat(3.5f)));
+    Byte[] row0 = concat(box(tag(1, WT_LEN)), box(encodeVarint(inner0.length)), inner0);
```
igorpeshansky: [Optional] If you add an `encodeMessage` helper, this could save you a local variable and become:

```java
    Byte[] row0 = concat(box(tag(1, WT_LEN)), encodeMessage(concat(
        box(tag(1, WT_VARINT)), box(encodeVarint(7)),
        box(tag(2, WT_VARINT)), box(encodeVarint(123456789012L)),
        box(tag(3, WT_VARINT)), box(encodeVarint(1)),
        box(tag(4, WT_32BIT)), box(encodeFloat(3.5f)))));
```

```java
  /** Encode a submessage. */
  private static Byte[] encodeMessage(Byte[] messageBytes) {
    return concat(box(encodeVarint(messageBytes.length)), messageBytes);
  }
```

The helper would be used for all nested messages.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+
+    Byte[] inner1 = concat(
+        box(tag(1, WT_VARINT)), box(encodeVarint(-1)),
+        box(tag(2, WT_VARINT)), box(encodeVarint(0)),
+        box(tag(3, WT_VARINT)), box(encodeVarint(0)),
+        box(tag(4, WT_32BIT)), box(encodeFloat(-0.25f)));
+    Byte[] row1 = concat(box(tag(1, WT_LEN)), box(encodeVarint(inner1.length)), inner1);
```
igorpeshansky: [Optional] Instead of storing `row0` and `row1` in local variables, only to add them to an array later, could you not do:

```java
    Byte[][] rows = new Byte[][]{concat(…), concat(…)};

    try (Table input = new Table.TestBuilder().column(rows).build(); …) { … }
```

?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+             ColumnVector expD = ColumnVector.fromBoxedFloats(3.5f, -0.25f);
+             ColumnVector expInner = ColumnVector.makeStruct(expA, expB, expC, expD);
+             ColumnVector expOuter = ColumnVector.makeStruct(expInner)) {
+          AssertUtils.assertStructColumnsAreEqual(expOuter, result);
+        }
+    }
+  }
+
+  @Test
+  void testNestedMessageAbsentParentIsNull() {
```
igorpeshansky: `AbsentParent` is ambiguous here, because both the outer and the inner messages are parents. How about `testAbsentNestedMessage_ProducesNull` or something?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+         HostColumnVector hostInner = inner.copyToHost();
+         HostColumnVector hostX = innerX.copyToHost()) {
+      assertNotNull(result);
+      assertTrue(hostInner.isNull(0), "Inner struct should be null when parent field absent");
+      assertTrue(hostX.isNull(0), "Inner x should inherit nullability from absent parent");
+    }
+  }
+
+  @Test
+  void testZeroLengthNestedMessageChildNullability() {
```
igorpeshansky: How about `testZeroLengthNestedMessage_ChildIsNull`?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+         HostColumnVector hostX = innerX.copyToHost()) {
+      assertNotNull(result);
+      assertEquals(DType.STRUCT, result.getType());
+      assertFalse(hostInner.isNull(0), "Inner struct should be present (length=0 nested)");
+      assertTrue(hostX.isNull(0), "Inner x should be null since the field is absent");
+    }
+  }
+
+  @Test
+  void testChildlessNestedMessagePresence() {
```
igorpeshansky: I would suggest using the expected outcome in the test name, i.e., `IsPresent` (or `IsNotNull`) instead of `Presence`…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+             false);
+         ColumnVector inner = result.getChildColumnView(0).copyToColumnVector();
+         HostColumnVector hostInner = inner.copyToHost()) {
+      assertFalse(hostInner.isNull(0), "Present empty message should produce a valid STRUCT<>");
+      assertTrue(hostInner.isNull(1), "Missing empty message should produce a null STRUCT<>");
+    }
+  }
+
+  @Test
+  void testFailfastNestedRepeatedWrongWireType() {
```
igorpeshansky: [Optional] Nit: `Failfast` is an expected outcome, so should follow the setup in the test name (e.g., `testNestedRepeatedWrongWireType_FailsFast`). I realize that this isn't the only test case with that naming pattern, though…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -504,6 +504,168 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
+  auto const& parent_loc = parent_locations[row];
+  if (parent_loc.offset < 0) return;
+
+  // Do the subtraction in int64 to keep the bounds-check honest even if a future caller
+  // ever passes a sliced LIST where parent_base_offset > parent_row_offsets[row].
+  int64_t parent_row_start = static_cast<int64_t>(parent_row_offsets[row]) - parent_base_offset;
+  int64_t nested_start_off = parent_row_start + parent_loc.offset;
+  int64_t nested_end_off   = nested_start_off + parent_loc.length;
+  if (nested_start_off < 0 || nested_end_off > message_data_size) {
+    set_error_once(error_flag, ERR_BOUNDS);
```
igorpeshansky: [Optional] Could we add a 64-bit overload of `check_message_bounds` and use it here?

```suggestion
  if (!check_message_bounds(nested_start_off, nested_end_off, message_data_size, error_flag)) {
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -504,6 +504,168 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
+// ============================================================================
+// Nested message scanning kernels
+// ============================================================================
+
+/**
+ * Scan one nested message per parent row to locate its direct singleton child fields.
+ * Repeated children are intentionally left to a separate count/scan path (3b.5/3b.6);
+ * this kernel only records last-one-wins locations for non-repeated descendants.
+ */
+CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
```
igorpeshansky, line 527: This is very similar to `scan_all_fields_kernel`. How much of this can be shared?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -504,6 +504,168 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
+    proto_tag tag;
+    if (!decode_tag(cur, nested_end, tag, error_flag)) {
+      mark_row_error();
+      return;
+    }
+    int fn = tag.field_number;
+    int wt = tag.wire_type;
+
+    for (int f = 0; f < num_fields; f++) {
+      if (field_descs[f].field_number == fn) {
```
igorpeshansky: [Optional] Instead of the loop with a break and multiple levels of nesting, can we reuse `lookup_field`? Or maybe even build a lookup table, similar to `d_fn_to_nested`…

```c++
    int f = lookup_field(fn, nullptr, 0, num_fields, [&](int i, int fn) {                                                                                                                                                        
      return field_descs[i].field_number == fn;
    });
    if (f >= 0) {
      if (field_descs[f].is_repeated) { … }
      …
    }
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -504,6 +504,168 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
+                                                   bool* row_has_invalid_data,
+                                                   int32_t const* top_row_indices)
+{
+  auto row = static_cast<cudf::size_type>(blockIdx.x * blockDim.x + threadIdx.x);
+  if (row >= num_parent_rows) return;
+
+  auto mark_row_error = [&]() {
+    if (row_has_invalid_data != nullptr) {
+      auto const top_row =
+        top_row_indices != nullptr ? top_row_indices[row] : static_cast<int32_t>(row);
```
igorpeshansky: This can be computed once and captured.

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -447,4 +447,169 @@ std::unique_ptr<cudf::column> build_repeated_string_column(
...
+    child_location_count > 0 ? child_location_count : 1, stream, scratch_mr);
+  // Scan over every child descriptor (including repeated ones) so STRICT-mode wire-type
+  // validation still runs on repeated occurrences. Repeated child slots in d_child_locations
+  // are not consumed; 3b.5 / 3b.6 produce them via dedicated count/scan kernels.
+  launch_scan_nested_message_fields(message_data,
+                                    message_data_size,
+                                    list_offsets,
+                                    base_offset,
+                                    d_parent_locs.data(),
+                                    num_rows,
```
igorpeshansky, line 537: Do we need to check (`CUDF_EXPECTS`) that `d_parent_locs.size() == num_rows`?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3159,6 +3159,268 @@ void testEmptyPackedRepeated() {
...
+           ColumnVector expectedInner = ColumnVector.makeStruct(expectedX);
+           ColumnVector expectedOuter = ColumnVector.makeStruct(expectedInner)) {
+        AssertUtils.assertStructColumnsAreEqual(expectedOuter, result);
+      }
+    }
+  }
+
+  @Test
+  void testNestedMessageMultipleScalarChildren() {
+    // message Inner { int32 a = 1; int64 b = 2; bool c = 3; float d = 4; }
```
igorpeshansky, line 2654: https://github.com/NVIDIA/spark-rapids-jni/pull/4504#discussion_r3171405956: Do we need to verify other types? If we don't, e.g. because we can infer their correctness from just the double tests, can we comment about this explicitly?

igorpeshansky (commented):
None of the comments are technically blocking, so I won't "request changes", but some of the cleanups could be impactful.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Added a `CUDF_EXPECTS`

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: With the `std::max(child_location_count, size_t{1})` change these buffers are always allocated with size >= 1, so the zero-size case no longer arises here.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Dropped the guard

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Ok added a comment for now.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Extracted the common scan into a templated `scan_message_field_locations` helper.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Addressed by the refactor: the helper computes the per-row output base once (`out = output_locations + row * num_fields`) and writes `out[f]`, so the per-field `flat_index` recomputation is gone.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Thanks! It has been addressed by #4663

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Added a comment here

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Added `encodeMessage(Byte[])` and used it for all nested-message construction.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Done, good idea.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Renamed.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Renamed.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Renamed.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Renamed.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
```diff
@@ -234,6 +234,20 @@ __device__ inline bool check_message_bounds(int32_t start,
   return true;
 }
 
+// 64-bit overload: nested message offsets are computed in int64 (parent row offset + relative
+// field offset) before being bounds-checked against the child buffer size.
+__device__ inline bool check_message_bounds(int64_t start,
```
igorpeshansky: [Optional] You could also define this as a generic:

```c++
template <std::integral T = int32_t>
__device__ inline bool check_message_bounds(T start,
                                            T end_pos,
                                            cudf::size_type total_size,
                                            int* error_flag)
{
  if (start < 0 || end_pos < start || end_pos > total_size) {
    set_error_once(error_flag, ERR_BOUNDS);
    return false;
  }
  return true;
}
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -92,78 +174,24 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+    return true;
+  };
+  if (!scan_message_field_locations(
+        msg_start,
+        msg_end,
+        field_descs,
+        num_fields,
+        field_lookup,
+        field_lookup_size,
+        locations + static_cast<size_t>(row) * static_cast<size_t>(num_fields),
```
igorpeshansky: (a) You're basically reimplementing `flat_index` here, and (b) does it make sense to pre-compute this value once (as, e.g., `field_locations`) and use it throughout this function?

```c++
  field_location* field_locations = locations + flat_index(static_cast<size_t>(row), static_cast<size_t>(num_fields), 0);
  for (int f = 0; f < num_fields; f++) {
    field_locations[f] = {-1, 0};
  }
```

[Optional] Also, I did notice that `flat_index` is always called with non-`size_t` values — would it make sense to pull the casts _inside_ the function? So the calls can just become `flat_index(row, num_fields, f)`?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -92,78 +174,24 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
-          mark_row_error();
-          return;
-        }
-        if (len > static_cast<uint64_t>(msg_end - cur - len_bytes) ||
-            len > static_cast<uint64_t>(cuda::std::numeric_limits<int>::max())) {
-          set_error_once(error_flag, ERR_OVERFLOW);
-          mark_row_error();
-          return;
-        }
-        // Record offset pointing to the actual data (after length prefix)
```
igorpeshansky, line 135: [Optional] Were these comments useful? Worth keeping?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -43,6 +43,88 @@ CUDF_KERNEL void set_error_if_unset_kernel(int* error_flag, int error_code)
...
+    proto_tag tag;
+    if (!decode_tag(cur, msg_end, tag, error_flag)) return false;
+    int const wt = tag.wire_type;
+
+    int f = lookup_field(tag.field_number, field_lookup, field_lookup_size, num_fields,
+                         [&](int i, int fn) { return field_descs[i].field_number == fn; });
+    if (f >= 0) {
+      if (field_descs[f].is_repeated) {
+        // Values are produced by the dedicated repeated count/scan path; we still validate the
+        // occurrence here so strict/permissive wire-type errors surface.
```
igorpeshansky: This comment is replicated in lines 590-591 — I'm slightly leaning towards keeping it there, but either location is fine, as long as it's done once.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -550,102 +578,36 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
...
-            static_cast<size_t>(row), static_cast<size_t>(num_fields), static_cast<size_t>(f))] = {
-            data_offset, field_size};
-        }
-        break;
-      }
-    }
+  uint8_t const* nested_end   = message_data + nested_end_off;
+
+  // Repeated children: values come from the dedicated nested repeated count/scan path
+  // (3b.5/3b.6); here we only validate the occurrence so strict/permissive errors surface.
```
igorpeshansky: [Optional] I'd put this comment inside the lambda, since it's effectively referring to the `noop` function…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -92,78 +174,24 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
-
-  while (cur < msg_end) {
-    proto_tag tag;
-    if (!decode_tag(cur, msg_end, tag, error_flag)) {
-      mark_row_error();
-      return;
-    }
-    int fn = tag.field_number;
-    int wt = tag.wire_type;
+  uint8_t const* msg_start = bytes + start;
```
igorpeshansky:
```suggestion
  uint8_t const* const msg_start = bytes + start;
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -550,102 +578,36 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
   int64_t parent_row_start = static_cast<int64_t>(parent_row_offsets[row]) - parent_base_offset;
   int64_t nested_start_off = parent_row_start + parent_loc.offset;
   int64_t nested_end_off   = nested_start_off + parent_loc.length;
-  if (nested_start_off < 0 || nested_end_off > message_data_size) {
-    set_error_once(error_flag, ERR_BOUNDS);
+  if (!check_message_bounds(nested_start_off, nested_end_off, message_data_size, error_flag)) {
     mark_row_error();
     return;
   }
   uint8_t const* nested_start = message_data + nested_start_off;
```
igorpeshansky:
```suggestion
  uint8_t const* const nested_start = message_data + nested_start_off;
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -92,78 +174,24 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
-  while (cur < msg_end) {
-    proto_tag tag;
-    if (!decode_tag(cur, msg_end, tag, error_flag)) {
-      mark_row_error();
-      return;
-    }
-    int fn = tag.field_number;
-    int wt = tag.wire_type;
+  uint8_t const* msg_start = bytes + start;
+  uint8_t const* msg_end   = bytes + end;
```
igorpeshansky: Nit: spacing (the `=` should either be 1-space from `msg_end` or align with the `=` above — both are fine, but pick one).

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -550,102 +578,36 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
...
-            return;
-          }
-          output_locations[flat_index(
-            static_cast<size_t>(row), static_cast<size_t>(num_fields), static_cast<size_t>(f))] = {
-            data_offset, field_size};
-        }
-        break;
-      }
-    }
+  uint8_t const* nested_end   = message_data + nested_end_off;
```
igorpeshansky: Nit: spacing (the `=` should either be 1-space from `nested_end` or align with the `=` above — both are fine, but pick one).

thirtiseven (commented):
oh forgot to submit comments

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -43,6 +43,88 @@ CUDF_KERNEL void set_error_if_unset_kernel(int* error_flag, int error_code)
...
+ * (`scan_nested_message_fields_kernel`) scanners. The caller initializes `out` to {-1, 0} and
+ * owns row-level error marking; this helper only sets `error_flag` and returns false on the first
+ * parse error that leaves the cursor unsafe to advance.
+ *
+ * Matched repeated fields are delegated to `on_repeated(cur, msg_end, msg_start, wt, expected_wt)`
+ * (which returns false on error). Top-level scalars pass a no-op handler since their descriptors
+ * are never repeated; the nested scanner validates repeated occurrences via walk_repeated_element.
+ */
+template <typename RepeatedHandler>
+__device__ bool scan_message_field_locations(uint8_t const* msg_start,
```
igorpeshansky: [Really optional] There's also some overlap with `count_repeated_fields_kernel` and `scan_all_repeated_occurrences_kernel`, but at this point we're approaching the diminishing returns territory, so we can defer that refactoring to a follow-up (though probably the simplest thing we can do is align the various elements — field index lookup via a lambda, `while`→`for` loop — and leave it be, rather than do a full refactor). FWIW, if we did decide to do the refactor to reuse `scan_message_field_locations` in `count_repeated_fields_kernel`, we would probably collapse the two separate field lookup tables into a single one, and use the `.is_repeated` check, as well as expand the `on_repeated` handler…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -43,6 +43,88 @@ CUDF_KERNEL void set_error_if_unset_kernel(int* error_flag, int error_code)
...
+        // Values are produced by the dedicated repeated count/scan path; we still validate the
+        // occurrence here so strict/permissive wire-type errors surface.
+        if (!on_repeated(cur, msg_end, msg_start, wt, field_descs[f].expected_wire_type)) {
+          return false;
+        }
+      } else if (wt != field_descs[f].expected_wire_type) {
+        set_error_once(error_flag, ERR_WIRE_TYPE);
+        return false;
+      } else {
+        int const data_offset = static_cast<int>(cur - msg_start);
```
igorpeshansky: As I mentioned in [my comment above](https://github.com/NVIDIA/spark-rapids-jni/pull/4644/changes/d36c2c7b2efbc978299c9b282a0ec8eac87458a8..eca0de426452f013c2541e3a0bdcf34240bd931f#r3353163458), there is similarity with `count_repeated_fields_kernel`, which leads to a question: is the 64-bit computation (`rel_offset64`) with a range check in `count_repeated_fields_kernel` needed, and if yes, why not here? Seems like the input value to the "end" pointer computation relative to "start" is always 32-bit in both places… We should at least be consistent.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -43,6 +43,90 @@ CUDF_KERNEL void set_error_if_unset_kernel(int* error_flag, int error_code)
...
+template <typename RepeatedHandler>
+__device__ bool scan_message_field_locations(uint8_t const* msg_start,
+                                             uint8_t const* msg_end,
+                                             field_descriptor const* field_descs,
+                                             int num_fields,
+                                             int const* field_lookup,
+                                             int field_lookup_size,
+                                             field_location* out,
+                                             int* error_flag,
+                                             RepeatedHandler&& on_repeated)
```
igorpeshansky: Is this code on C++ 20? If so, you can just make this `auto&&` and remove the template parameter (abbreviated function template[^1]).

[^1]: https://en.cppreference.com/cpp/language/function_template#Abbreviated_function_template

src/main/cpp/src/protobuf/protobuf_builders.cu
igorpeshansky: That invariant only applies to the vectors created in this file. `d_row_force_null` is actually passed in from the outside, so you still need this check (besides, the size of `d_row_force_null` is used to indicate whether to track null rows, and if the size is non-zero, it's expected to be exactly `num_rows`).

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3322,27 +2717,18 @@ void testNestedMessageAbsentParentIsNull() {
   }
 
   @Test
-  void testZeroLengthNestedMessageChildNullability() {
+  void testZeroLengthNestedMessage_ChildIsNull() {
     // Outer carries the nested tag but with length 0 (Inner is empty).
     Byte[] row = concat(box(tag(1, WT_LEN)), box(encodeVarint(0)));
```
igorpeshansky: [Optional] One of these would make it obvious that the inner message is empty:

```java
    Byte[] innerMessage = new Byte[]{};
    Byte[] row = concat(box(tag(1, WT_LEN)), encodeMessage(innerMessage));
```

```java
    Byte[] row = concat(box(tag(1, WT_LEN)), encodeMessage(new Byte[]{}));
```

Also in `testChildlessNestedMessage_IsPresent`…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3356,29 +2742,18 @@ void testZeroLengthNestedMessageChildNullability() {
   }
 
   @Test
-  void testChildlessNestedMessagePresence() {
+  void testChildlessNestedMessage_IsPresent() {
     // message Empty {}
     // message Outer { Empty inner = 1; }
     Byte[] row0 = concat(box(tag(1, WT_LEN)), box(encodeVarint(0)));
     Byte[] row1 = new Byte[]{};
```
igorpeshansky: [Optional] https://github.com/NVIDIA/spark-rapids-jni/pull/4644#discussion_r3328083173 applies here too:

```java
    Byte[][] rows = new Byte[][]{concat(…), new Byte[]{}};

    try (Table input = new Table.TestBuilder().column(rows).build(); …) { … }
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
igorpeshansky: I meant computing `top_row` outside of the lambda, to make the lambda shorter:

```c++
  auto const top_row = top_row_indices != nullptr ? top_row_indices[row] : static_cast<int32_t>(row);
  auto mark_row_error = [&]() {
    if (row_has_invalid_data != nullptr) { row_has_invalid_data[top_row] = true; }
  };
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -92,78 +176,24 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
+  auto unreachable_repeated = [](uint8_t const*, uint8_t const*, uint8_t const*, int, int) {
+    return true;
+  };
+  if (!scan_message_field_locations(
+        msg_start,
+        msg_end,
+        field_descs,
+        num_fields,
+        field_lookup,
+        field_lookup_size,
```
igorpeshansky: These three parameters are only used in the `field_lookup` call. We wouldn't need them if we pass in a `field_lookup_idx`-style lambda instead…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -43,6 +43,90 @@ CUDF_KERNEL void set_error_if_unset_kernel(int* error_flag, int error_code)
...
+  for (uint8_t const* cur = msg_start; cur < msg_end;) {
+    proto_tag tag;
+    if (!decode_tag(cur, msg_end, tag, error_flag)) return false;
+    int const wt = tag.wire_type;
+
+    int f = lookup_field(
+      tag.field_number, field_lookup, field_lookup_size, num_fields, [&](int i, int fn) {
+        return field_descs[i].field_number == fn;
+      });
+    if (f >= 0) {
```
igorpeshansky: [Optional] You could reuse the approach from lines 482-488, by defining `lookup_desc_idx` before the loop. Then this could just become:

```suggestion
    if (int f = lookup_desc_idx(tag.field_number); f >= 0) {
```

---

Your `lookup_desc_idx` would look like this:

```c++
  auto lookup_desc_idx = [&](int fn) -> int {
    return lookup_field(fn, field_lookup, field_lookup_size, num_fields,
                        [&](int f, int fn) { return field_descs[f].field_number == fn; });
  };
```

Per my other comment, you could also pull it out of the function.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Done. TIL

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Thank you. Did _the simplest thing_ part for now.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Removed here.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Good catch. The range check in `count_repeated_fields_kernel` was unnecessary. I dropped it there.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Added them back to `scan_message_field_locations`.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: 👍 done.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Good catches. Done (a), (b) and the optional.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Good catch, add a `CUDF_EXPECTS` above.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
```diff
@@ -206,9 +207,12 @@ __device__ inline bool get_field_data_location(
   return true;
 }
 
-CUDF_HOST_DEVICE inline size_t flat_index(size_t row, size_t width, size_t col)
+// Row-major flat index into a [num_rows x width] array. Takes any integral types and widens to
+// size_t internally so call sites don't need to cast (the multiply happens in size_t).
+template <std::integral Row, std::integral Width, std::integral Col>
+CUDF_HOST_DEVICE inline size_t flat_index(Row row, Width width, Col col)
```
igorpeshansky: [Optional] Since `Row`, `Width`, and `Col` are each used only once, we can use the abbreviated form:

```suggestion
CUDF_HOST_DEVICE inline size_t flat_index(
    std::integral auto row, std::integral auto width, std::integral auto col)
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
igorpeshansky: Great. Leaving unresolved for now to remind us of the larger refactoring opportunities, but won't block the PR on this.

[Really optional] A couple of small things I would point out are the naming mismatches (e.g., `msg_offset` vs `data_location`, `msg_start` vs `msg_base`, `i` vs `f`, etc), and formatting mismatches (e.g., braces around `return`s vs no braces, positions of variable initializers, etc) between these three functions. Might be worth aligning…

---

For when we revisit this (pseudocode, summarized by Claude, but happy to discuss): now that `scan_message_field_locations` takes `lookup_desc_idx` + `on_repeated` lambdas and dispatches on `is_repeated`, most of the enabling work is already done. Folding the other two scanners in would look like:

**Shared-helper change (the only one needed in common code):** give `on_repeated` the matched descriptor index so a handler can route to the right per-field slot:

```cpp
auto&& on_repeated;  // (int f, cur, msg_end, msg_start, wt, expected_wt) -> bool
```

The existing top-level no-op and nested validate-only handlers just ignore `f`.

**`scan_all_repeated_occurrences_kernel` (easy):** single category, single table, no depth filter — its `lookup_desc_idx` is already in the right shape. The per-field state just moves into `on_repeated`:

```cpp
auto on_repeated = [&, row_i32 = static_cast<int32_t>(row)](int f, auto cur, auto end, auto base, int wt, int expected_wt) {
  int& wi = write_idx[f];
  int const we = scan_descs[f].row_offsets[row + 1];
  auto act = [&](int32_t off, int32_t len) {
    if (wi >= we) { set_error_once(error_flag, ERR_REPEATED_COUNT_MISMATCH); return false; }
    scan_descs[f].occurrences[wi++] = {row_i32, off, len};
    return true;
  };
  return walk_repeated_element(cur, end, base, wt, expected_wt, error_flag, act);
};
```

`write_idx[]` init + the post-loop count verification stay in the kernel. Two wrinkles: descriptors here are `repeated_field_scan_desc`, not `field_descriptor`, so we'd either template the helper on the descriptor type or pass a parallel `field_descriptor` view (`is_repeated=true`, `expected_wire_type=wire_type`); and `out` is unused (all repeated), so it'd be `nullptr`/scratch. The explicit `is_packed`/wire-type check drops out since `walk_repeated_element` already does it.

**`count_repeated_fields_kernel` (moderate — the real plumbing):**
- Build a single combined descriptor list in `decode_protobuf_to_struct` (instead of `repeated_field_indices` + `nested_field_indices` and two tables), ordered `[nested | repeated]` so the helper's `out[f]` write lands in `nested_locations` for `f < num_nested`, and `on_repeated` routes repeated via `f - num_nested` into `repeated_info`.
- Preserve the depth filter in `lookup_desc_idx`. `field_descriptor` has no `depth`, so either add one or keep the lookup reading `schema[...].depth` alongside a parallel descriptor array for `is_repeated`/`expected_wire_type`.

```cpp
auto on_repeated = [&](int f, auto cur, auto end, auto base, int wt, int expected_wt) {
  auto& info = repeated_info[flat_index(row, num_repeated_fields, f - num_nested_fields)];
  auto count = [&info](int32_t, int32_t){ info.count++; return true; };
  return walk_repeated_element(cur, end, base, wt, expected_wt, error_flag, count);
};
field_location* out = nested_locations + flat_index(row, num_nested_fields, 0);
```

**Worth measuring** if we do this: register pressure / occupancy of the merged kernels — the `on_repeated` closures capture more state and the helper inlines per instantiation. Likely fine, but an `nvcc -Xptxas -v` / Nsight diff before/after would confirm we don't regress the repeated-heavy path.

Net: a helper-signature tweak + folding each kernel in, with `count`'s host-side descriptor unification as the only non-trivial piece — not a rewrite.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -412,20 +397,17 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
-      auto const rel_offset64 = static_cast<int64_t>(cur - msg_base);
-      if (rel_offset64 < cuda::std::numeric_limits<int32_t>::min() ||
-          rel_offset64 > cuda::std::numeric_limits<int32_t>::max()) {
-        set_error_once(error_flag, ERR_OVERFLOW);
-        return;
-      }
+      // cur - msg_base is bounded by the message length (<= INT32_MAX via check_message_bounds),
+      // so this fits int32; checked_add_int32 still guards the offset + len_bytes addition. Matches
+      // the LEN handling in scan_message_field_locations.
+      int const msg_data_offset = static_cast<int>(cur - msg_base);
```
igorpeshansky: Nit: other functions just call it `data_offset`…

igorpeshansky (dismissed):
Approving, but may be worth addressing the small cleanups and nits before submitting.

src/main/cpp/src/protobuf/protobuf_device_helpers.cuh
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
igorpeshansky: > [Really optional] A couple of small things I would point out are the naming mismatches (e.g., `msg_offset` vs `data_location`, `msg_start` vs `msg_base`, `i` vs `f`, etc), and formatting mismatches (e.g., braces around `return`s vs no braces, positions of variable initializers, etc) between these three functions. Might be worth aligning…

To be clear, I was saying that the above naming mismatches might be worth addressing _in this PR_ without changing the overall structure of the code. That said, I did mark it "really optional", so up to you. I've approved it regardless.

igorpeshansky (dismissed):

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: > To be clear, I was saying that the above naming mismatches might be worth addressing _in this PR_ without changing the overall structure of the code. That said, I did mark it "really optional", so up to you. I've approved it regardless.

Thanks for clarifying. I misread it. Updated now.

thirtiseven:
build

igorpeshansky (approved):

PR #4646, opened 2026-05-28
from_protobuf JNI-3.0: output field flags for pruned schemas

thirtiseven:
Implements **JNI-3.0** from the split plan in NVIDIA/spark-rapids#14069. Independent of #4644 (3b.1+3b.2) — these two PRs can be reviewed in parallel.

depends on https://github.com/NVIDIA/spark-rapids-jni/pull/4663

## Summary

Per-field `isOutput` plumbing through Java, JNI, and C++ validation so a caller can mark fields that should be decoded for validation (required / enum / wire-type / depth checks) but dropped from the returned struct. This is the JNI groundwork the plugin's schema-projection optimization will sit on top of.

## Components

**Java** (`ProtobufSchemaDescriptor.java`, `Protobuf.java`):
- New `isOutput[]` alongside the other parallel arrays.
- Validation rejects parent/child output-flag mismatches — a hidden STRUCT cannot have visible descendants and vice versa.
- The pre-existing 15-arg `ProtobufSchemaDescriptor` constructor is preserved as a back-compat shim that fills `isOutput` with all-true, so existing callers and tests do not need to change.

**JNI bridge** (`ProtobufJni.cpp`):
- Accepts the extra `jbooleanArray`, validates its length alongside the other field-parallel arrays, and forwards it into `protobuf_decode_context.output_fields`.

**C++** (`protobuf.hpp`, `protobuf.cu`):
- New `protobuf_decode_context.output_fields` (empty vector means \"all output\").
- `validate_decode_context` gains size check and parent/child consistency check.
- New `is_output_field` helper.
- `decode_protobuf_to_struct` skips hidden top-level fields when assembling the returned struct, both in the empty-input path and in the final `top_level_children` loop. Decoding itself remains unconditional so validation errors still surface for hidden fields.

## Tests

`ProtobufSchemaDescriptorTest`:
- `testBackCompatConstructorMarksAllFieldsAsOutput`
- `testNestedFieldMustShareOutputFlagWithParent`
- `testHiddenFieldRoundTripsThroughSerialization`

`ProtobufTest`:
- `testHiddenFieldDoesNotAppearInOutput`
- `testHiddenRequiredFieldStillValidates` — proves required-field validation still runs for hidden fields

## Size

~100 LOC implementation + ~150 LOC tests across 7 files; pure plumbing, no kernel changes.


src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3478,4 +3478,110 @@ void testSchemaWithTooManyRepeatedFields() {
...
+        new boolean[]{false, false},
+        new boolean[]{false, false},
+        new boolean[]{false, false},
+        new boolean[]{true, false},  // hide b
+        new long[]{0, 0},
+        new double[]{0.0, 0.0},
+        new boolean[]{false, false},
+        new byte[][]{null, null},
+        new int[][]{null, null},
+        new byte[][][]{null, null});
```
igorpeshansky: This is pretty unreadable — it's very hard to see what this test is doing, and the reader has to rely on the comments, which are notoriously easy to fall out of sync. The test should be obvious from the code, with irrelevant boilerplate hidden behind helpers.

For example, the construction of `ProtobufSchemaDescriptor` is a prime candidate for the builder pattern, as that can easily hide the irrelevancies. With a properly implemented builder the call could look something like this:

```java
    ProtobufSchemaDescriptor schema = new ProtobufSchemaDescriptorBuilder()
        .fieldNumbers(1, 2)
        .outputTypes(DType.INT32, DType.INT32)  // extract type id and use deriveWireType…
        .isOutput(true, false)  // hide b
        .build();
```

or even like this:

```java
    ProtobufSchemaDescriptor schema = new ProtobufSchemaDescriptorBuilder()
        .addField().fieldNumber(1).outputType(DType.INT32)
        .addField().fieldNumber(2).outputType(DType.INT32).isOutput(false)  // hide b
        .build();
```

Other variants are possible too — the point is that it should be obvious at a glance what is being set and what's being left as a default.

Note that such unreadability is pervasive throughout this test file, but I'm only commenting on the lines that changed in this PR.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -338,4 +339,130 @@ void testSerializationRoundTripPreservesContentsAndIsolation() throws Exception
...
   }
+
+  @Test
+  void testBackCompatConstructorMarksAllFieldsAsOutput() {
+    ProtobufSchemaDescriptor schema = new ProtobufSchemaDescriptor(
+        new int[]{1, 2},
+        new int[]{-1, -1},
+        new int[]{0, 0},
+        new int[]{Protobuf.WT_VARINT, Protobuf.WT_LEN},
+        new int[]{ai.rapids.cudf.DType.INT32.getTypeId().getNativeId(),
```
igorpeshansky: `import ai.rapids.cudf.DType`?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
igorpeshansky, line 1: Unrelated, but I didn't see a test that validates that constructor array parameters of different lengths are rejected. Seems like a pretty important property to check. Did I miss one?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -338,4 +339,130 @@ void testSerializationRoundTripPreservesContentsAndIsolation() throws Exception
...
+        new boolean[]{false, false},
+        new long[]{0, 0},
+        new double[]{0.0, 0.0},
+        new boolean[]{false, false},
+        new byte[][]{null, null},
+        new int[][]{null, null},
+        new byte[][][]{null, null});
+    assertEquals(2, schema.isOutput.length);
+    for (boolean flag : schema.isOutput) {
+      assertEquals(true, flag);
```
igorpeshansky: `assertTrue`? Also below.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -338,4 +339,130 @@ void testSerializationRoundTripPreservesContentsAndIsolation() throws Exception
...
+    }
```
igorpeshansky: [Optional] This is the same as the following, right?

```suggestion
    assertArrayEquals(new boolean[]{true, true}, schema.isOutput);
```

Also below.

src/main/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptor.java
```diff
@@ -182,13 +235,22 @@ private static void validate(
...
       validateParentChild(i, parentIndices[i], depthLevels, outputTypeIds);
+      validateOutputFlag(i, parentIndices[i], isOutput);
       validateUniqueFieldKey(i, parentIndices[i], fieldNumbers[i], seenFieldNumbers);
       validateWireTypeAndEncoding(i, wireTypes[i], outputTypeIds[i], encodings[i]);
       validateFieldFlags(i, isRepeated[i], isRequired[i], hasDefaultValue[i], outputTypeIds[i]);
       validateEnumMetadata(i, encodings[i], enumValidValues[i], enumNames[i]);
     }
   }
 
+  private static void validateOutputFlag(int index, int parentIndex, boolean[] isOutput) {
```
igorpeshansky: [Optional] All the previous functions were declared in the order they are called from `validate()`. Any reason not to put this between `validateParentChild` and `validateUniqueFieldKey`?

src/main/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptor.java
```diff
@@ -182,13 +235,22 @@ private static void validate(
...
+    if (parentIndex >= 0 && isOutput[index] != isOutput[parentIndex]) {
```
igorpeshansky: [Optional] Since this is also validating the parent/child relationship, why not just pass `isOutput` into `validateParentChild` and do the validation there?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -352,6 +364,11 @@ void validate_decode_context(protobuf_decode_context const& context)
                std::invalid_argument);
 }
 
+bool is_output_field(protobuf_decode_context const& context, int schema_idx)
+{
+  return context.output_fields.empty() || context.output_fields[schema_idx];
```
igorpeshansky: If this function is ever called with `schema_idx` that is out of bounds, `operator[]` would be undefined behavior. All other accesses do bounds checks. Why not `context.output_fields.at(schema_idx)`?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -402,7 +419,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   if (num_rows == 0) {
     std::vector<std::unique_ptr<cudf::column>> empty_children;
     for (int i = 0; i < num_fields; i++) {
-      if (schema[i].parent_idx == -1) {
+      if (schema[i].parent_idx == -1 && is_output_field(context, i)) {
```
igorpeshansky: [Optional] This would read better as a guard clause:

```c++
    for (int i = 0; i < num_fields; i++) {
      if (schema[i].parent_idx >= 0 || !is_output_field(context, i)) {
        continue;
      }
      auto field_type = …;
      …
    }
```

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1242,10 +1259,11 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     }  // for (ri)
   }
 
-  // Assemble top_level_children in schema order (not processing order)
+  // Assemble top_level_children in schema order (not processing order). Hidden fields are
+  // still decoded above (so validation errors surface), but dropped from the output struct.
   std::vector<std::unique_ptr<cudf::column>> top_level_children;
   for (int i = 0; i < num_fields; i++) {
-    if (schema[i].parent_idx == -1) {
+    if (schema[i].parent_idx == -1 && is_output_field(context, i)) {
```
igorpeshansky: [Optional] This would read better as a guard clause:

```c++
    for (int i = 0; i < num_fields; i++) {
      if (schema[i].parent_idx >= 0 || !is_output_field(context, i)) {
        continue;
      }
      if (column_map[i]) …
      …
    }
```

src/main/cpp/src/protobuf/protobuf.cu
igorpeshansky: [Really optional] I know this is prior code, but this whole conditional could be simplified as:

```c++
      top_level_children.push_back(column_map[i] ?
        std::move(column_map[i]) :
        make_null_column_with_schema(schema, i, num_fields, num_rows, stream, mr));
      }
```

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -338,4 +339,130 @@ void testSerializationRoundTripPreservesContentsAndIsolation() throws Exception
...
+  }
+
+  @Test
+  void testNestedFieldMustShareOutputFlagWithParent() {
+    int structType = ai.rapids.cudf.DType.STRUCT.getTypeId().getNativeId();
+    int intType = ai.rapids.cudf.DType.INT32.getTypeId().getNativeId();
+    // Parent struct hidden, child visible -> illegal.
```
igorpeshansky: Should we test a visible parent with a hidden child as well?

igorpeshansky (changes requested):

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: It's a valuable point! I opened https://github.com/NVIDIA/spark-rapids-jni/pull/4663 to refactor all java tests.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: done.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: Good catch, added `testMismatchedArrayLengthsRejected`

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: Replaced the `assertEquals(true, …)` loops with `assertArrayEquals(new boolean[]{...}, …)`

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: done.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: Added the reverse direction to `testNestedFieldMustShareOutputFlagWithParent`

src/main/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptor.java
thirtiseven: done.

src/main/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptor.java
thirtiseven: Resolved by folding the check into `validateParentChild` (see your next comment), so the standalone function no longer exists.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: 👍 done.

src/main/cpp/src/protobuf/protobuf.cu

src/main/cpp/src/protobuf/protobuf.cu

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Done. Any comments on the prior code are very welcome 😸

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -419,21 +419,19 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+        empty_children.push_back(make_empty_list_column(std::move(empty_struct), stream, mr));
+      } else if (schema[i].is_repeated) {
+        auto empty_child = make_empty_column_safe(field_type, stream, mr);
+        empty_children.push_back(make_empty_list_column(std::move(empty_child), stream, mr));
+      } else if (field_type.id() == cudf::type_id::STRUCT) {
+        empty_children.push_back(
+          make_empty_struct_column_with_schema(schema, i, num_fields, stream, mr));
+      } else {
+        empty_children.push_back(make_empty_column_safe(field_type, stream, mr));
       }
```
igorpeshansky: If we consolidate some common code, this might read better:

```suggestion
      auto empty_child = (field_type.id() == cudf::type_id::STRUCT) ?
          make_empty_struct_column_with_schema(schema, i, num_fields, stream, mr) :
          make_empty_column_safe(field_type, stream, mr);
      if (schema[i].is_repeated) {
        empty_child = make_empty_list_column(std::move(empty_child), stream, mr);
      }
      empty_children.push_back(std::move(empty_child));
```

igorpeshansky (dismissed):
Approving, but strongly recommending the one remaining cleanup before you merge.

thirtiseven:
build

igorpeshansky (approved):

PR #4661, opened 2026-06-01
Match Spark whitespace and separator handling in parseTimestampWithFormat

thirtiseven:
## Summary

Fixes two ways the fused `parseTimestampWithFormat` parser diverged from Spark on space-separated timestamp formats such as `yyyy-MM-dd HH:mm:ss`. Both were silent (GPU result differs from CPU, no error):

1. **Whitespace (LEGACY).** Under `timeParserPolicy=LEGACY` Spark parses with `SimpleDateFormat`, which skips leading `[ \t]` before every numeric field. The kernel only folded whitespace after `-` and `/`, so a multi-space date/time separator (`1999-12-31  11:59:59`), a space/tab run after the separator, or whitespace after `:` (`11: 59: 59`) returned NULL while CPU parses them. A `TOK_SKIP_HT_WS` is now emitted ahead of each non-packed digit field (subsumes the old `REMOVE_WHITESPACE_FROM_MONTH_DAY` rewrite).

2. **`'T'` separator (both policies).** A literal space was compiled to a `'T'`-or-space token, so `1999-12-31T11:59:59` was accepted. Spark matches a literal space with exactly one `' '` and returns NULL for `'T'` under both LEGACY and CORRECTED — consistent with the existing `.{10}T` rejection in the spark-rapids cuDF fallback path. The space now compiles to a plain literal.

The format-less string→timestamp cast uses a different kernel and is unaffected (it still accepts `'T'` per ISO).

## Testing

- `CastStringsTest`: added whitespace cases; `'T'`-separator cases now expect NULL.
- Cross-checked against `java.text.SimpleDateFormat` (lenient) and Spark 3.5.x `to_timestamp` on CPU for both policies.
- End-to-end GPU vs CPU coverage will be added on the spark-rapids side (`date_time_test.py`) in a separate PR.

### DIFF
--- a/src/main/cpp/src/parse_timestamp_with_format.cu
+++ b/src/main/cpp/src/parse_timestamp_with_format.cu
@@ -38,14 +38,20 @@
 #include <string>
 #include <vector>
 
+// This kernel mirrors the formatters Spark uses for to_timestamp / unix_timestamp /
+// ParseToTimestamp: java.text.SimpleDateFormat for LEGACY, and java.time.DateTimeFormatter with a
+// STRICT resolver for CORRECTED. It is NOT a drop-in for the CSV/JSON read path, which uses
+// commons-lang3 FastDateFormat and has different whitespace rules — do not reuse it there.
 namespace spark_rapids_jni {
 
 namespace {
 
 // ---- Low-level string-parsing primitives. ------------------------------------------------------
 
-// Whitespace test consistent with Spark UTF8String.trimAll for char input.
-__device__ bool is_whitespace(unsigned char c) { return c <= 32 || c == 127; }
+// LEGACY parses via java.text.SimpleDateFormat, which skips only ' ' and '\t' before fields — it
+// does not trimAll. So the outer trim must match: a leading control byte like '\r'/'\f'/'\v'
+// rejects on CPU and must not be treated as whitespace here.
+__device__ bool is_whitespace(unsigned char c) { return c == ' ' || c == '\t'; }
 
 // Read between `min_d` and `max_d` digits greedily. Advances pos by the digits read.
 // Returns false (and the partial pos advance is irrelevant since walk_tokens aborts on failure).
@@ -146,7 +152,9 @@ struct format_token {
 //     e.g. yyyyMMdd) get exact width — otherwise the boundary is ambiguous.
 //   - Otherwise CORRECTED uses exact width and LEGACY uses [1, 2].
 //   - CORRECTED `yyyy/MM/dd` keeps the existing spark-rapids compatibility contract and accepts
-//     1-2 digit month/day fields.
+//     1-2 digit month/day fields. This intentionally DEVIATES from Spark CPU, whose STRICT
+//     DateTimeFormatter rejects single-digit fields ("2024/5/6" is null on CPU); the GPU
+//     over-accepts. Pinned by parseTimestampWithFormat_correctedSlashDateDeviation.
 // Literal handling:
 //   - A space matches exactly one ' '. Spark rejects 'T' as the separator for a space pattern
 //     under both policies (unlike the format-less cast, which accepts 'T').
--- a/src/main/cpp/src/parse_timestamp_with_format.cu
+++ b/src/main/cpp/src/parse_timestamp_with_format.cu
@@ -71,16 +71,7 @@ __device__ bool try_parse_char(unsigned char const* p, int& pos, int end, unsign
   return true;
 }
 
-__device__ bool try_parse_t_or_space(unsigned char const* p, int& pos, int end)
-{
-  if (pos >= end) { return false; }
-  unsigned char c = p[pos];
-  if (c != ' ' && c != 'T') { return false; }
-  ++pos;
-  return true;
-}
-
-// Skip [ \t]* — used to fold REMOVE_WHITESPACE_FROM_MONTH_DAY into the legacy state machine.
+// Skip [ \t]* — LEGACY whitespace-before-field fold (see compile_format).
 __device__ void skip_ht_whitespace(unsigned char const* p, int& pos, int end)
 {
   while (pos < end && (p[pos] == ' ' || p[pos] == '\t')) {
@@ -127,7 +118,6 @@ struct parsed_dt {
 enum tok_kind : uint8_t {
   TOK_DIGITS,           // a = field, b = min_digits, c = max_digits
   TOK_LITERAL,          // a = literal char
-  TOK_T_OR_SPACE,       // 'T' or ' ' between date and time
   TOK_SKIP_HT_WS,       // skip [ \t]* (legacy whitespace fold)
   TOK_TRAIL_EOF,        // pos must equal end
   TOK_TRAIL_NON_DIGIT,  // pos == end OR p[pos] is not a digit (legacy tail rule)
@@ -158,9 +148,10 @@ struct format_token {
 //   - CORRECTED `yyyy/MM/dd` keeps the existing spark-rapids compatibility contract and accepts
 //     1-2 digit month/day fields.
 // Literal handling:
-//   - Space ' ' compiles to TOK_T_OR_SPACE (Spark accepts both as the date/time separator).
-//   - In LEGACY, '-' and '/' literals are followed by TOK_SKIP_HT_WS, folding the legacy
-//     `REMOVE_WHITESPACE_FROM_MONTH_DAY` rewrite into the state machine.
+//   - A space matches exactly one ' '. Spark rejects 'T' as the separator for a space pattern
+//     under both policies (unlike the format-less cast, which accepts 'T').
+//   - LEGACY: SimpleDateFormat skips [ \t] before every numeric field, so a TOK_SKIP_HT_WS
+//     precedes each non-packed digit field (subsumes the old REMOVE_WHITESPACE_FROM_MONTH_DAY).
 // The trailing token is TOK_TRAIL_EOF for CORRECTED and TOK_TRAIL_NON_DIGIT for LEGACY.
 
 uint8_t letter_to_field(char c)
@@ -205,21 +196,19 @@ std::vector<format_token> compile_format(std::string const& fmt, bool legacy)
       bool const variable_width = (legacy && !packed) || corrected_variable_width_slash_date;
       uint8_t const min_d       = (c == 'y') ? run : (variable_width ? 1 : run);
       uint8_t const max_d       = run;
+      // Skip [ \t] before each field, except inside a packed run which stays exact-width.
+      bool const abuts_prev_field = (i > 0 && std::isalpha(static_cast<unsigned char>(fmt[i - 1])));
+      if (legacy && !abuts_prev_field) { out.push_back({TOK_SKIP_HT_WS, 0, 0, 0}); }
       out.push_back({TOK_DIGITS, letter_to_field(c), min_d, max_d});
       saw_digit_field = true;
       i               = j;
     } else {
-      if (c == ' ') {
-        out.push_back({TOK_T_OR_SPACE, 0, 0, 0});
-      } else {
-        // Literal char must be ASCII; non-ASCII bytes would alias UTF-8 continuation bytes
-        // when matched against the input.
-        if (static_cast<unsigned char>(c) >= 0x80) {
-          throw std::invalid_argument("non-ASCII literal in pattern is not supported");
-        }
-        out.push_back({TOK_LITERAL, static_cast<uint8_t>(c), 0, 0});
-        if (legacy && (c == '-' || c == '/')) { out.push_back({TOK_SKIP_HT_WS, 0, 0, 0}); }
+      // A space is an ordinary literal (matches one ' '). Reject non-ASCII: it would alias
+      // UTF-8 continuation bytes against the input.
+      if (static_cast<unsigned char>(c) >= 0x80) {
+        throw std::invalid_argument("non-ASCII literal in pattern is not supported");
       }
+      out.push_back({TOK_LITERAL, static_cast<uint8_t>(c), 0, 0});
       ++i;
     }
   }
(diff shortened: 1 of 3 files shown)


src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -71,16 +71,7 @@ __device__ bool try_parse_char(unsigned char const* p, int& pos, int end, unsign
...
-  if (pos >= end) { return false; }
-  unsigned char c = p[pos];
-  if (c != ' ' && c != 'T') { return false; }
-  ++pos;
-  return true;
-}
-
-// Skip [ \t]* — used to fold REMOVE_WHITESPACE_FROM_MONTH_DAY into the legacy state machine.
+// Skip [ \t]* — LEGACY whitespace-before-field fold (see compile_format).
 __device__ void skip_ht_whitespace(unsigned char const* p, int& pos, int end)
```
res-life, line 81: **MUST FIX — CPU/GPU divergence in LEGACY outer trim**

The `is_whitespace` helper at L48 returns `true` for any byte `c <= 32 || c == 127` (matching `UTF8String.isWhitespaceOrISOControl`), and `trim` at L93–L101 applies it to both ends of the input. Spark CPU's `to_timestamp` / `unix_timestamp` / `ParseToTimestamp` do **not** call `trimAll` before handing the string to `LegacySimpleTimestampFormatter` (`java.text.SimpleDateFormat`); `SimpleDateFormat.subParse` only skips `[ \t]+` (space + tab) before each numeric field. So a leading `\r`, `\f`, `\v`, or other control byte is silently consumed on GPU but rejected on CPU.

| Input (pattern `yyyy-MM-dd`, LEGACY) | This PR (GPU) | Spark CPU |
|---|---|---|
| `"\r2024-05-06"` | ✅ accepts | ❌ null |
| `"\f2024-05-06"` | ✅ accepts | ❌ null |
| `"\v2024-05-06"` | ✅ accepts | ❌ null |
| `"\b2024-05-06"` (0x08) | ✅ accepts | ❌ null |
| `"\02024-05-06"` (0x00 prefix) | ✅ accepts | ❌ null |
| `" 2024-05-06"` / `"\t2024-05-06"` | ✅ accepts | ✅ accepts (matches) |
| `"\n2024-05-06"` | ❌ null (special-cased by `has_leading_newline`) | ❌ null (matches) |

The leading-newline path is special-cased so `\n` happens to match, but the broader control-char set still diverges.

Verified the CPU side against Spark v3.3.4, v3.4.4, v3.5.8, v4.0.2, v4.1.2 — all five route `to_timestamp` through `TimestampFormatterHelper` with `SIMPLE_DATE_FORMAT`, i.e. `LegacySimpleTimestampFormatter` (`SimpleDateFormat`), so the deviation reproduces across the entire supported version matrix.

**Suggested fix (smaller change)**: narrow `is_whitespace` at L48 to
```cpp
__device__ bool is_whitespace(unsigned char c) { return c == ' ' || c == '\t'; }
```
This aligns the outer `trim` with `SimpleDateFormat.subParse`'s per-field skip. Also update the L47 comment — "consistent with Spark UTF8String.trimAll for char input" is misleading since the CPU path is **not** a `trimAll` caller.

**Alternative**: drop the outer `trim` entirely and rely on `TOK_SKIP_HT_WS` before each digit field for leading whitespace (you'd need to pick a tail policy explicitly). The narrow-`is_whitespace` option is the smaller change and matches SDF exactly.

**Tests to add** in `CastStringsTest.parseTimestampWithFormat_legacyWhitespaceFold` (or a new method): assert that
- `"\r2024-05-06"`, `"\f2024-05-06"`, `"\v2024-05-06"`, `"\b2024-05-06"` → `null` under LEGACY `yyyy-MM-dd`
- The existing `" 2024-05-06"` / `"\t2024-05-06"` → valid timestamp assertions continue to pass

This pins the alignment so a future refactor can't silently re-widen the trim.

(Cross-version audit confirms no Spark shim under `sql-plugin/src/main/spark330..spark411` overrides this path — the JNI is the single source of truth for the parsed value across all supported Spark versions, so a deviation here surfaces everywhere.)

res-life:
## Cross-version coverage audit — punch list

Audit context: this PR was verified by the author against Spark 3.5.x CPU. I re-checked the plugin (`spark-rapids` `sql-plugin/src/main/scala/org/apache/spark/sql/rapids/datetimeExpressions.scala`) and Apache Spark CPU (`v3.3.4`, `v3.4.4`, `v3.5.8`, `v4.0.2`, `v4.1.2`) to confirm coverage holds for **all** supported Spark version families (3.3.x / 3.4.x / 3.5.x / 4.0.x / 4.1.x).

**Headline findings:**
- The plugin's GPU call site (`datetimeExpressions.scala:709, :758`) is uniform across every version shim (`spark330..spark411`); no shim overrides the timestamp parsing path.
- Spark CPU `to_timestamp` / `unix_timestamp` / `ParseToTimestamp` routes through `TimestampFormatterHelper` with `SIMPLE_DATE_FORMAT` (i.e. `LegacySimpleTimestampFormatter` = `java.text.SimpleDateFormat`) for LEGACY and `Iso8601TimestampFormatter` (`java.time.DateTimeFormatter` with STRICT resolver) for CORRECTED on **every** version inspected.
- Only Spark-side version difference: the default value of `spark.sql.legacy.timeParserPolicy` changed from `EXCEPTION` to `CORRECTED` in Spark 4.0+. This does **not** affect the JNI kernel path — both branches route through the non-legacy code in `doColumnar`.
- `FastDateFormat` (commons-lang3) is only used by Spark's CSV/JSON inferer (`UnivocityParser`, `JacksonParser`), which the plugin handles via a different code path (`GpuTextBasedPartitionReader`) and does **not** call `parseTimestampWithFormat`. So aligning the kernel to `SimpleDateFormat` is the correct reference choice for this PR.

---

### Punch list

| #  | Severity      | Item |
|----|---------------|------|
| R1 | SHOULD        | Four entries of `CORRECTED_COMPATIBLE_FORMATS` lack a direct positive assertion in `CastStringsTest.java`: `yyyy/MM`, `MM/dd`, `MM/yyyy`, `MM-yyyy`. Each can be closed with one `assertParsedTimestamp` line in `parseTimestampWithFormat_correctedDateOnlyFormats` (around L1424) or a new sibling method. |
| R2 | SHOULD        | `parse_timestamp_with_format.cu:175` (`corrected_variable_width_slash_date = !legacy && fmt == "yyyy/MM/dd"`) preserves the pre-existing spark-rapids compatibility contract that accepts 1-digit month/day under CORRECTED `yyyy/MM/dd`. Spark CPU `Iso8601TimestampFormatter` with STRICT resolver **rejects** single-digit fields, so the test at L1404 (`"2024/5/6"` → valid) documents a real CPU deviation. Expand the comment at L175 to call out the deviation explicitly, and add a dedicated test method (`parseTimestampWithFormat_correctedSlashDateDeviation`) that asserts the contract so it cannot be silently broken by future refactoring. |
| R3 | **MUST FIX**  | (See inline comment at L75 → #discussion_r3332630705.) LEGACY outer `trim` uses `is_whitespace = c <= 32 \|\| c == 127` (L48), accepting `\r`, `\f`, `\v`, `\b`, `\0` etc. as leading/trailing whitespace. Spark CPU `SimpleDateFormat.subParse` only skips `[ \t]+`, so inputs like `"\r2024-05-06"` are accepted on GPU but return `null` on CPU. Narrow `is_whitespace` to `c == ' ' \|\| c == '\t'`, fix the L47 comment (`to_timestamp` is **not** a `trimAll` caller), and add tests to pin the alignment. |
| R4 | NICE          | Add a paragraph at the top of `parse_timestamp_with_format.cu` (around L41) noting that the kernel mirrors `SimpleDateFormat` (LEGACY) and `DateTimeFormatter` (CORRECTED, STRICT) as used by `to_timestamp` / `unix_timestamp` / `ParseToTimestamp`. Explicitly say it is **not** a drop-in for the CSV/JSON read path, which uses `FastDateFormat` and has different whitespace rules (only strips before unbounded numeric fields). This prevents future reuse from silently mis-aligning. |
| R5 | NICE          | `parseTimestampWithFormat_correctedSingleSeparatorOnly` (L1614) does not currently assert `"\n1999-12-31 11:59:59"` → `null`. Add this one-line assertion so CORRECTED's rejection of leading newline is pinned. (LEGACY's leading-newline rejection is already pinned at `parseTimestampWithFormat_legacyWhitespaceFold` L1485.) |
| R6 | SHOULD        | The plugin's `getTimeParserPolicy` (`GpuOverrides.scala:4770`) hardcodes its fallback as `"EXCEPTION"`, whereas Spark 4.0+ defaults to `CORRECTED`. They route identically through the non-legacy branch of `doColumnar`, so behavior is equivalent — but worth adding a Python integration test under Spark 4.x that **does not set** `spark.sql.legacy.timeParserPolicy` and verifies GPU/CPU agree on `to_timestamp(s, "yyyy-MM-dd HH:mm:ss")` for inputs `"2024-05-06 11:59:59"`, `"2024-05-06  11:59:59"`, `"2024-05-06T11:59:59"`, `"\r2024-05-06 11:59:59"`. (Lives in `integration_tests`; not in scope for this PR but should follow.) |

---

### Per-version sanity matrix

| Spark tag | LEGACY formatter for `to_timestamp` | CORRECTED formatter | Default `timeParserPolicy` | Notes |
|---|---|---|---|---|
| v3.3.4 | `LegacySimpleTimestampFormatter` (`SimpleDateFormat`) — `sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/TimestampFormatter.scala:400–429` | `Iso8601TimestampFormatter` (`DateTimeFormatter` STRICT) — same file `:151–223` | `EXCEPTION` | Baseline. |
| v3.4.4 | Same class/lines | Same class/lines | `EXCEPTION` | No change. |
| v3.5.8 | Same class, moved to `sql/api/.../TimestampFormatter.scala:457–490` | Same class `:153` | `EXCEPTION` | File moved to `sql/api` module only. |
| v4.0.2 | Same class `sql/api/.../TimestampFormatter.scala:497–530`; CORRECTED enforces strict end-of-input via `parseUnresolved` length-check (`:204`) | Same class | **CORRECTED** | Default policy changed; functionally equivalent for kernel path. |
| v4.1.2 | Same class | Same class | **CORRECTED** | `SQLConf.LEGACY_TIME_PARSER_POLICY.createWithDefault(LegacyBehaviorPolicy.CORRECTED)`. No new pattern letters or whitespace tweaks. |

---

Net of R3 (now an inline MUST FIX): the PR's behavior is correct across the full supported version range. Items R1/R2/R4/R5 are scope-of-this-PR cleanups; R6 is a follow-up integration test.

res-life:
The above comments are from AI. Could you please first add test case and do test to double check what AI said.

res-life (approved):
LGTM

res-life:
### R6 follow-up — Spark 4.x default policy integration test

Confirming R6 was intentionally deferred out of this PR's scope (it lives in `integration_tests` of NVIDIA/spark-rapids, not spark-rapids-jni). Posting the suggested follow-up here so it's tracked.

**Why it matters**

Spark 4.0 changed the default of `spark.sql.legacy.timeParserPolicy` from `EXCEPTION` → `CORRECTED` (`SQLConf.LEGACY_TIME_PARSER_POLICY` in `v4.0.2` / `v4.1.2`). The plugin's `GpuOverrides.getTimeParserPolicy` (`sql-plugin/src/main/scala/com/nvidia/spark/rapids/GpuOverrides.scala:4770`) hardcodes `"EXCEPTION"` as its own fallback. Both branches route through the non-legacy code in `doColumnar`, so behaviour is equivalent today — but nothing pins that equivalence in CI under Spark 4.x's actual default. A regression here would silently drift CPU/GPU.

**Suggested follow-up (open against NVIDIA/spark-rapids)**

Add a Python integration test (e.g. `integration_tests/src/main/python/cast_test.py` or a new `to_timestamp_default_policy_test.py`) parametrized over Spark 4.0+ that:

1. Does **not** set `spark.sql.legacy.timeParserPolicy` (so the 4.x default `CORRECTED` applies).
2. Runs `to_timestamp(s, "yyyy-MM-dd HH:mm:ss")` on inputs that probe the alignment areas this PR touches:
   - `"2024-05-06 11:59:59"` — single-space separator (valid)
   - `"2024-05-06  11:59:59"` — double-space (CORRECTED → null)
   - `"2024-05-06T11:59:59"` — `T` separator (CORRECTED → null)
   - `"\r2024-05-06 11:59:59"` / `"\f..."` / `"..."` — leading control bytes (CORRECTED → null; LEGACY → null after this PR's R3 fix)
   - `" 2024-05-06 11:59:59"` — leading space (CORRECTED → null; LEGACY → valid)
3. Asserts GPU and CPU results match row-by-row via the standard `assert_gpu_and_cpu_are_equal_collect` helper.

A second parametrization with `spark.sql.legacy.timeParserPolicy = LEGACY` covering the same inputs would close the matrix on the LEGACY side at the integration level (the JUnit `CastStringsTest` already covers it at the JNI level, but an integration test catches plugin-layer regressions like the policy lookup or whitelist gating).

If preferred, I can open the spark-rapids follow-up PR — let me know.

(Re-pasting for traceability: this is the only item from the cross-version audit punch list that this PR does **not** address by design. R1, R2, R3, R4, R5 are all closed by `a968a338`.)

res-life:
R6 is a follow-up to add a defensive test case. We can add in this PR or in a follow-up, either is OK.

thirtiseven:
Thanks for the thorough cross-version audit — all items addressed in a968a338, and re-verified against Spark 3.5 CPU where noted:

- **R3 (MUST FIX):** narrowed `is_whitespace` to `' '`/`'\t'` and fixed the stale comment. Confirmed on Spark CPU that leading `\r`/`\f`/`\v`/`\b` return null; added unit tests here plus an end-to-end GPU/CPU regression test in NVIDIA/spark-rapids#14920.
- **R1:** added CORRECTED positive assertions for `yyyy/MM`, `MM/dd`, `MM/yyyy`, `MM-yyyy`.
- **R2:** confirmed on Spark CPU that CORRECTED rejects single-digit `2024/5/6`; documented the deviation and pinned it with `parseTimestampWithFormat_correctedSlashDateDeviation`.
- **R4:** added the file-level note that this mirrors SimpleDateFormat/DateTimeFormatter, not the CSV/JSON FastDateFormat path.
- **R5:** pinned CORRECTED leading-whitespace/newline rejection.
- **R6:** added the Spark 4.0 unset-policy test in NVIDIA/spark-rapids#14920.

Integration tests pass on the release/26.06 base. Thanks for the review\!

thirtiseven:
build

thirtiseven:
> The above comments are from AI. Could you please first add test case and do test to double check what AI said.

Thank you, the R3 is a good catch! Your AI is doing well.

PR #4663, opened 2026-06-01
Add a fluent builder for protobuf schema test descriptors

thirtiseven:
## Summary

Pure test-readability refactor: introduce a fluent `ProtobufSchemaDescriptorBuilder` (test-only) and migrate **all** protobuf tests to it. No production behavior change.

`ProtobufSchemaDescriptor` is a bundle of ~15 parallel arrays. Building one by hand — or via the various `decode*Fields(...)`/`makeScalarSchema(...)`/`decodeRaw(...)` helpers that each wrapped a different slice of those arrays — forced the reader to count positions across a dozen arrays to understand a single field. This was raised in review on #4646 (the unreadability is pervasive across the test file). This PR addresses it across the whole test suite.

Before:
```java
decodeRaw(col,
    new int[]{1, 2, 3, 4}, new int[]{-1, -1, -1, -1}, new int[]{0, 0, 0, 0},
    new int[]{WT_VARINT, WT_VARINT, WT_64BIT, WT_VARINT},
    new int[]{INT32_id, INT32_id, FLOAT64_id, BOOL8_id},
    new int[]{ENC_DEFAULT, ENC_DEFAULT, ENC_DEFAULT, ENC_DEFAULT},
    new boolean[]{false, true, true, true}, /* …8 more arrays… */, false);
```
After:
```java
Protobuf.decodeToStruct(col,
        .addField(1, DType.INT32)               // id
        .addField(2, DType.INT32).repeated()    // int_values (packed)
        .addField(3, DType.FLOAT64).repeated()  // double_values (packed)
        .addField(4, DType.BOOL8).repeated()    // bool_values (packed)
        .build(),
    false);
```

Nested messages are described with `down()`/`up()` instead of hand-counted parent indices:
```java
    .addField(1, DType.STRUCT).down()        // Outer
        .addField(1, DType.INT32)            //   a
        .addField(2, DType.STRUCT).down()    //   b: Mid
            .addField(1, DType.INT32)        //     c
        .up()
    .up()
    .build();
```

## What changed

- **New** `ProtobufSchemaDescriptorBuilder.java`: each attribute setter applies to the most recently added field; only non-default attributes are named. Wire types are derived from output type + encoding, with a `wireType()` override for the negative wire-type-compatibility cases.
  - `down()`/`up()` express message nesting as a structure (parent index and depth derived automatically) — the idiom for nesting.
  - `parent(int)` is retained as a low-level primitive, reserved for loop-built degenerate chains and the malformed parent links that the validation tests deliberately feed to the descriptor (links `down()/up()` can't express). A `build()`-time check rejects an unbalanced `down()/up()`.
- **ProtobufSchemaDescriptorTest** fully migrated (341 → 239 lines).
- **ProtobufTest** fully migrated (3481 → 2845 lines): every `decodeRaw(...)` and the entire helper family — `makeScalarSchema`, both `decodeAllFields` overloads, `decodeAllFieldsWithRequired`, `decodeAllFieldsWithDefaults`, `decodeAllFieldsWithEnums`, `decodeAllFieldsWithEnumStrings`, and the internal `decodeScalarFields` — are removed. Call sites now build a descriptor inline and call `Protobuf.decodeToStruct(...)` directly; the duplicate `deriveWireType` is consolidated into the builder. `parent(int)` no longer appears in `ProtobufTest` at all — all real nesting uses `down()/up()`.
- Removed dead code: the unused `verifyDeepNesting` helper.

## Notes

- The builder targets the current 15-arg `ProtobufSchemaDescriptor` API. The `isOutput` field from #4646 is intentionally not included here; once #4646 lands, the builder gains `isOutput()`.

## Sequencing

Standalone refactor split out from the in-flight protobuf kernel PRs (#4644, #4646) so it can migrate the test file uniformly without a half-migrated state. Conflicts with those PRs are resolved after they merge.

### DIFF
--- a/src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
@@ -18,7 +18,9 @@ package com.nvidia.spark.rapids.jni;
 
 import ai.rapids.cudf.DType;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.List;
 
 /**
@@ -37,6 +39,25 @@ import java.util.List;
  *
  * <p>Each attribute setter applies to the most recently added field. Wire types are derived from
  * the output type and encoding unless overridden via {@link #wireType(int)}.
+ *
+ * <p>For nested messages, {@link #down()} descends into the most recently added field so that
+ * subsequent {@code addField} calls become its children (with parent index and depth derived
+ * automatically); {@link #up()} returns to the enclosing level. This avoids hand-counting flat
+ * parent indices as the nesting deepens:
+ *
+ * <pre>{@code
+ * ProtobufSchemaDescriptor schema = new ProtobufSchemaDescriptorBuilder()
+ *     .addField(1, DType.STRUCT).down()   // message Outer
+ *         .addField(1, DType.INT32)       //   int32 a = 1
+ *         .addField(2, DType.STRUCT).down()  // Inner b = 2
+ *             .addField(1, DType.INT32)   //     int32 x = 1
+ *         .up()
+ *     .up()
+ *     .build();
+ * }</pre>
+ *
+ * <p>{@link #parent(int)} remains available as an explicit escape hatch, e.g. to construct the
+ * malformed schemas that the validation tests deliberately feed to the descriptor.
  */
 public final class ProtobufSchemaDescriptorBuilder {
   private static final class Field {
@@ -58,19 +79,54 @@ public final class ProtobufSchemaDescriptorBuilder {
   }
 
   private final List<Field> fields = new ArrayList<>();
+  // Indices of the fields we have descended into via down(); the top is the current parent.
+  private final Deque<Integer> parentStack = new ArrayDeque<>();
 
-  /** Start a new top-level field with the given field number and output type. */
+  /**
+   * Add a field. By default it is top-level; inside a {@link #down()} scope it becomes a child of
+   * the enclosing field, with parent index and depth derived automatically.
+   */
   public ProtobufSchemaDescriptorBuilder addField(int fieldNumber, DType outputType) {
     Field f = new Field();
     f.fieldNumber = fieldNumber;
     f.outputTypeId = outputType.getTypeId().getNativeId();
+    if (!parentStack.isEmpty()) {
+      int parentIndex = parentStack.peek();
+      f.parentIndex = parentIndex;
+      f.depth = fields.get(parentIndex).depth + 1;
+    }
     fields.add(f);
     return this;
   }
 
   /**
-   * Mark the current field as a child of {@code parentIndex}, setting its depth to
-   * parent depth + 1.
+   * Descend into the most recently added field so that subsequent {@code addField} calls become
+   * its children (until the matching {@link #up()}). This is the idiom for expressing message
+   * nesting; prefer it over {@link #parent(int)}, which is a low-level primitive.
+   */
+  public ProtobufSchemaDescriptorBuilder down() {
+    if (fields.isEmpty()) {
+      throw new IllegalStateException("down() requires a field to descend into");
+    }
+    parentStack.push(fields.size() - 1);
+    return this;
+  }
+
+  /** Return to the enclosing nesting level opened by {@link #down()}. */
+  public ProtobufSchemaDescriptorBuilder up() {
+    if (parentStack.isEmpty()) {
+      throw new IllegalStateException("up() called without a matching down()");
+    }
+    parentStack.pop();
+    return this;
+  }
+
+  /**
+   * Low-level: assign a raw parent index to the current field (depth = parent depth + 1),
+   * overriding any parent set by {@link #down()}. Prefer {@link #down()}/{@link #up()} for ordinary
+   * nesting. Reserve this for (a) loop-built degenerate chains where {@code parent(i - 1)} reads
+   * cleaner, and (b) malformed parent links that {@code down()}/{@code up()} cannot express (e.g. a
+   * non-ancestor index) in validation tests.
    */
   public ProtobufSchemaDescriptorBuilder parent(int parentIndex) {
     Field f = current();
@@ -169,6 +225,9 @@ public final class ProtobufSchemaDescriptorBuilder {
   }
 
   public ProtobufSchemaDescriptor build() {
+    if (!parentStack.isEmpty()) {
+      throw new IllegalStateException("build() called inside a down() scope; missing up()");
+    }
     int n = fields.size();
     int[] fieldNumbers = new int[n];
     int[] parentIndices = new int[n];
--- a/src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
@@ -94,6 +94,12 @@ public final class ProtobufSchemaDescriptorBuilder {
     return this;
   }
 
+  /** Conditional form so callers can pass a flag inline; {@code repeated(false)} is a no-op. */
+  public ProtobufSchemaDescriptorBuilder repeated(boolean value) {
+    current().isRepeated = value;
+    return this;
+  }
+
   public ProtobufSchemaDescriptorBuilder required() {
     current().isRequired = true;
     return this;
@@ -133,6 +139,12 @@ public final class ProtobufSchemaDescriptorBuilder {
    */
   public ProtobufSchemaDescriptorBuilder hasDefault() {
     current().hasDefaultValue = true;
+    return this;
+  }
+
+  /** Conditional form so callers can pass a flag inline; {@code hasDefault(false)} is a no-op. */
+  public ProtobufSchemaDescriptorBuilder hasDefault(boolean value) {
+    current().hasDefaultValue = value;
     return this;
   }
 
--- a/src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
@@ -1,0 +1,222 @@
+/*
+ * Copyright (c) 2026, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nvidia.spark.rapids.jni;
+
+import ai.rapids.cudf.DType;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Test-only fluent builder for {@link ProtobufSchemaDescriptor}.
+ *
+ * <p>The descriptor is a set of parallel arrays, which is error-prone to assemble by hand: the
+ * reader has to count positions across a dozen arrays to see what a single field looks like. This
+ * builder lets each field be described in one place, with only the non-default attributes named:
+ *
+ * <pre>{@code
+ * ProtobufSchemaDescriptor schema = new ProtobufSchemaDescriptorBuilder()
+ *     .addField(1, DType.INT32)
+ *     .addField(2, DType.STRING).required()
+ *     .build();
+ * }</pre>
+ *
+ * <p>Each attribute setter applies to the most recently added field. Wire types are derived from
+ * the output type and encoding unless overridden via {@link #wireType(int)}.
+ */
+public final class ProtobufSchemaDescriptorBuilder {
+  private static final class Field {
+    int fieldNumber;
+    int outputTypeId;
+    int parentIndex = -1;
+    int depth = 0;
+    int encoding = Protobuf.ENC_DEFAULT;
+    Integer wireType = null;  // null => derive from output type + encoding
+    boolean isRepeated = false;
+    boolean isRequired = false;
+    boolean hasDefaultValue = false;
+    long defaultInt = 0L;
+    double defaultFloat = 0.0;
+    boolean defaultBool = false;
+    byte[] defaultString = null;
+    int[] enumValidValues = null;
+    byte[][] enumNames = null;
+  }
+
+  private final List<Field> fields = new ArrayList<>();
+
+  /** Start a new top-level field with the given field number and output type. */
+  public ProtobufSchemaDescriptorBuilder addField(int fieldNumber, DType outputType) {
+    Field f = new Field();
+    f.fieldNumber = fieldNumber;
+    f.outputTypeId = outputType.getTypeId().getNativeId();
+    fields.add(f);
+    return this;
+  }
+
+  /**
+   * Mark the current field as a child of {@code parentIndex}, setting its depth to
+   * parent depth + 1.
+   */
+  public ProtobufSchemaDescriptorBuilder parent(int parentIndex) {
+    Field f = current();
+    f.parentIndex = parentIndex;
+    f.depth = fields.get(parentIndex).depth + 1;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder encoding(int encoding) {
+    current().encoding = encoding;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder wireType(int wireType) {
+    current().wireType = wireType;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder repeated() {
+    current().isRepeated = true;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder required() {
+    current().isRequired = true;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder defaultInt(long value) {
+    Field f = current();
+    f.hasDefaultValue = true;
+    f.defaultInt = value;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder defaultFloat(double value) {
+    Field f = current();
+    f.hasDefaultValue = true;
+    f.defaultFloat = value;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder defaultBool(boolean value) {
+    Field f = current();
+    f.hasDefaultValue = true;
+    f.defaultBool = value;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder defaultString(byte[] value) {
+    Field f = current();
+    f.hasDefaultValue = true;
+    f.defaultString = value;
+    return this;
+  }
+
+  /**
+   * Set the has-default flag without a value. Used by negative tests that pair a default with an
+   * incompatible field (e.g. a repeated or STRUCT field) to exercise validation.
+   */
+  public ProtobufSchemaDescriptorBuilder hasDefault() {
+    current().hasDefaultValue = true;
+    return this;
+  }
+
+  /** Provide enum-as-string metadata for the current field (also sets ENC_ENUM_STRING). */
+  public ProtobufSchemaDescriptorBuilder enumMetadata(int[] validValues, byte[][] names) {
+    Field f = current();
+    f.encoding = Protobuf.ENC_ENUM_STRING;
+    f.enumValidValues = validValues;
+    f.enumNames = names;
+    return this;
+  }
+
+  /** Attach raw enum metadata without forcing the encoding (for validation-only tests). */
+  public ProtobufSchemaDescriptorBuilder enumValidValues(int[] validValues) {
+    current().enumValidValues = validValues;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptorBuilder enumNames(byte[][] names) {
+    current().enumNames = names;
+    return this;
+  }
+
+  public ProtobufSchemaDescriptor build() {
+    int n = fields.size();
+    int[] fieldNumbers = new int[n];
+    int[] parentIndices = new int[n];
+    int[] depthLevels = new int[n];
+    int[] wireTypes = new int[n];
+    int[] outputTypeIds = new int[n];
+    int[] encodings = new int[n];
+    boolean[] isRepeated = new boolean[n];
+    boolean[] isRequired = new boolean[n];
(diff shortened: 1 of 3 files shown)


src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -36,25 +37,14 @@ private ProtobufSchemaDescriptor makeDescriptor(
...
-        new byte[][]{null},
-        new int[][]{enumValidValues},
-        new byte[][][]{enumNames});
+    DType outputType = (encoding == Protobuf.ENC_ENUM_STRING) ? DType.STRING : DType.INT32;
+    ProtobufSchemaDescriptorBuilder b =
+        new ProtobufSchemaDescriptorBuilder().addField(1, outputType).encoding(encoding);
+    if (isRepeated) b.repeated();
+    if (hasDefaultValue) b.hasDefault();
+    if (enumValidValues != null) b.enumValidValues(enumValidValues);
+    if (enumNames != null) b.enumNames(enumNames);
```
igorpeshansky: These two methods already support assigning `null` (which would be a no-op), so you can just call these unconditionally. Note that this suggestion is already accounted for in [my comment on lines 43-44](https://github.com/NVIDIA/spark-rapids-jni/pull/4663/changes#r3338391477).

```suggestion
    b.enumValidValues(enumValidValues).enumNames(enumNames);
```

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -36,25 +37,14 @@ private ProtobufSchemaDescriptor makeDescriptor(
...
-        new double[]{0.0},
-        new boolean[]{false},
-        new byte[][]{null},
-        new int[][]{enumValidValues},
-        new byte[][][]{enumNames});
+    DType outputType = (encoding == Protobuf.ENC_ENUM_STRING) ? DType.STRING : DType.INT32;
+    ProtobufSchemaDescriptorBuilder b =
+        new ProtobufSchemaDescriptorBuilder().addField(1, outputType).encoding(encoding);
+    if (isRepeated) b.repeated();
+    if (hasDefaultValue) b.hasDefault();
```
igorpeshansky: [Optional] If you add overloads that accept a `boolean` value, this whole builder could be a single expression, e.g.:

```java
    DType outputType = (encoding == Protobuf.ENC_ENUM_STRING) ? DType.STRING : DType.INT32;
    return new ProtobufSchemaDescriptorBuilder()
        .addField(1, outputType).encoding(encoding)
            .repeated(isRepeated).hasDefault(hasDefaultValue)
            .enumValidValues(enumValidValues).enumNames(enumNames)
        .build();
```

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -146,173 +97,70 @@ void testEnumStringRejectsEmptyEnumArrays() {
...
-            new boolean[]{false},
-            new long[]{0},
-            new double[]{0.0},
-            new boolean[]{false},
-            new byte[][]{null},
-            new int[][]{{0, 1}},
-            new byte[][][]{new byte[][]{"A".getBytes(), "B".getBytes()}}));
+        new ProtobufSchemaDescriptorBuilder()
+            .addField(1, DType.STRING).wireType(Protobuf.WT_LEN)
+            .enumMetadata(new int[]{0, 1}, new byte[][]{"A".getBytes(), "B".getBytes()})
```
igorpeshansky: I would indent this 4 spaces, since it technically pertains to the above field, not the builder (you already do this in lines 160-162, so I assume there's no auto-formatter to overrule it).

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
igorpeshansky: I'd revisit the "out of scope" decision. I read your "their call sites are already concise and converting int ids back to DType would add noise" as "diff noise" (i.e., a larger PR diff), which is true — the diff will be larger. But the purpose of this PR is descriptor readability, and these helpers suffer from the same parallel array issues that this PR was intended to address. A later reader would see the builder everywhere else and has to wonder why these spots don't use it — especially since the builder is actually more concise here, because `addField` would eliminate the `.getTypeId().getNativeId()` and array creation boilerplate. You already call `ProtobufSchemaDescriptorBuilder.deriveWireType`, so it's half-migrated. I'd lean toward finishing it in this PR rather than deferring.

igorpeshansky: Same argument as for `makeScalarSchema` — all of these `decode*Fields*` functions are basically wrappers around the different ways to construct a `ProtobufSchemaDescriptor`, which are made obsolete by the new builder, i.e., they now all reduce to one `private static ColumnVector decodeFields(ColumnView binaryInput, ProtobufSchemaDescriptor descriptor, boolean failOnErrors)` function, or even direct calls to `Protobuf.decodeToStruct`.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -1757,25 +1707,13 @@ void testUnpackedRepeatedInt32() {
         box(tag(1, WT_VARINT)), box(encodeVarint(3)));
 
     try (Table input = new Table.TestBuilder().column(new Byte[][]{row}).build()) {
-      // Use the new nested API for repeated fields
-      // Field: ids (field_number=1, parent=-1, depth=0, wire_type=VARINT, type=INT32, repeated=true)
-      try (ColumnVector result = decodeRaw(
+      // Field: ids (field_number=1, repeated int32)
```
igorpeshansky: [Optional] This could become a line comment on the `addField` below (just the name, the rest of the info is obvious from the `addField`), i.e.:

```java
              .addField(1, DType.INT32).repeated()  // ids
              .build(),
```

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -1842,27 +1780,14 @@ void testPackedRepeatedDoubleWithMultipleFields() {
...
-          new boolean[]{false, true, true, true},
-          new boolean[]{false, false, false, false},
-          new boolean[]{false, false, false, false},
-          new long[]{0, 0, 0, 0},
-          new double[]{0.0, 0.0, 0.0, 0.0},
-          new boolean[]{false, false, false, false},
-          new byte[][]{null, null, null, null},
-          new int[][]{null, null, null, null},
+          new ProtobufSchemaDescriptorBuilder()
+              .addField(1, DType.INT32)            // id
```
igorpeshansky: Nit: should the comment indentation align for all fields?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -1921,22 +1846,13 @@ void testNestedMessage() {
...
-          new boolean[]{false, false},     // isRequired
-          new boolean[]{false, false},     // hasDefaultValue
-          new long[]{0, 0},
-          new double[]{0.0, 0.0},
-          new boolean[]{false, false},
-          new byte[][]{null, null},
-          new int[][]{null, null},
+          new ProtobufSchemaDescriptorBuilder()
+              .addField(1, DType.STRUCT)
+              .addField(1, DType.INT32).parent(0)
```
igorpeshansky: [Really optional] It's still a bit error-prone that one has to count the added fields to figure out which is the parent, but it looks a lot better than before. It may be possible to create an `addChildField` method that automatically assigns the parent index, but I can't think of a good way to do this without adding some complex machinery to keep track of the current field depth to know which of the recently-added fields is the parent, or exposing the `Field` class in some manner, so this'll have to do…
If you did want to make it more robust, a couple of ways that come to mind are:

1. Field stack:

   ```java
                 .addField(1, DType.STRUCT).down()
                     .addField(1, DType.INT32)
                 .up()
                 .build()
   ```

   This would coexist with explicit `.parent()` setters, which can be used to construct invalid nesting examples.

2. Nested builder (similar to what protobuf itself generates):

   ```java
             new ProtobufSchemaDescriptorBuilder().fields(
                 new FieldBuilder(1, DType.STRUCT).children(
                     new FieldBuilder(1, DType.INT32).build())
                     .build())
                 .build()
   ```

   Caveat: because the parent index computation is encapsulated, it would be structurally impossible to construct invalid nestings.

3. Computed field indexes (avoids explicit computation and counting):

   ```java
         ProtobufSchemaDescriptorBuilder b = new ProtobufSchemaDescriptorBuilder();
         int structFieldIdx = b.addField(1, DType.STRUCT).fieldIndex();
         b.addField(1, DType.INT32).parent(structFieldIdx);
             …
             …b.build()…
   ```

   Caveat: breaks up the fluent builder pattern; can no longer chain `addField` with nested structure; the computed indexes cannot be used in an expression unless you use something like an inline lambda (IILE) pattern.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2953,50 +2745,21 @@ private void verifyDeepNesting(int numLevels) {
...
-    boolean[] hasDefaultValue = new boolean[numFields];
-    long[] defaultInts = new long[numFields];
-    double[] defaultFloats = new double[numFields];
-    boolean[] defaultBools = new boolean[numFields];
-    byte[][] defaultStrings = new byte[numFields][];
-    int[][] enumValidValues = new int[numFields][];
-
+    // Each level contributes an INT32 leaf (field 1) and, except the last, a STRUCT (field 2)
+    // that parents the next level. Flat indices: int at 2*level, struct at 2*level+1.
+    ProtobufSchemaDescriptorBuilder builder = new ProtobufSchemaDescriptorBuilder();
```
igorpeshansky: Before we look deeper into it (pun intended), is `verifyDeepNesting` dead code? I couldn't find any callers…

If you decide to keep it, you could use the same peeled loop trick as you did in `ProtobufSchemaDescriptorTest.testDepthAboveSupportedLimitRejected`, i.e.:

```java
    // Each level contributes an INT32 leaf (field 1) and, except the last, a STRUCT (field 2)
    // that parents the next level. Flat indices: int at 2*level, struct at 2*level+1.
    ProtobufSchemaDescriptorBuilder builder = new ProtobufSchemaDescriptorBuilder();
    if (numLevels > 1) {
      builder.addField(1, DType.INT32);
      builder.addField(2, DType.STRUCT);
    }
    for (int level = 1; level < numLevels - 1; level++) {
      int parentIdx = 2 * (level - 1) + 1;
      builder.addField(1, DType.INT32).parent(parentIdx);
      builder.addField(2, DType.STRUCT).parent(parentIdx);
    }
    builder.addField(1, DType.INT32);
    if (numLevels > 1) builder.parent(2 * (numLevels - 2) + 1);
```

Since `2 * (0 - 1) + 1 == -1`, if you let `builder.parent(-1)` be a graceful no-op, you could eliminate the conditionals and start the loop at 0:

```java
    // Each level contributes an INT32 leaf (field 1) and, except the last, a STRUCT (field 2)
    // that parents the next level. Flat indices: int at 2*level, struct at 2*level+1.
    ProtobufSchemaDescriptorBuilder builder = new ProtobufSchemaDescriptorBuilder();
    for (int level = 0; level < numLevels - 1; level++) {
      int parentIdx = 2 * (level - 1) + 1;  // == -1 when level is 0
      builder.addField(1, DType.INT32).parent(parentIdx);  // requires parent(-1) to be a no-op
      builder.addField(2, DType.STRUCT).parent(parentIdx);  // requires parent(-1) to be a no-op
    }
    builder.addField(1, DType.INT32).parent(2 * (numLevels - 2) + 1);
```

or even:

```java
    // Each level contributes an INT32 leaf (field 1) and, except the last, a STRUCT (field 2)
    // that parents the next level. Flat indices: int at 2*level, struct at 2*level+1.
    ProtobufSchemaDescriptorBuilder builder = new ProtobufSchemaDescriptorBuilder();
    int parentIdx = -1;
    for (int level = 0; level < numLevels - 1; level++) {
      builder.addField(1, DType.INT32).parent(parentIdx);  // requires parent(-1) to be a no-op
      builder.addField(2, DType.STRUCT).parent(parentIdx);  // requires parent(-1) to be a no-op
      parentIdx += 2;
    }
    builder.addField(1, DType.INT32).parent(parentIdx);
```

You might not need the comment either, if the above code is obvious enough.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: Done, combining both suggestions. Added `repeated(boolean)` / `hasDefault(boolean)` overloads to the builder, and since `enumValidValues/enumNames` already no-op on null, the whole helper is now a single expression.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: ditto

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: done

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Totally agreed. I cleaned up all those helpers, please take another look.

thirtiseven: ditto

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: done

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: It's dead code, I removed it for now. Thank you for the ideas!

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Thank you, it's a good idea! Implemented option 1.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
```diff
@@ -58,19 +79,54 @@ private static final class Field {
...
+    parentStack.push(fields.size() - 1);
+    return this;
+  }
+
+  /** Return to the enclosing nesting level opened by {@link #down()}. */
+  public ProtobufSchemaDescriptorBuilder up() {
+    if (parentStack.isEmpty()) {
+      throw new IllegalStateException("up() called without a matching down()");
+    }
+    parentStack.pop();
```
igorpeshansky, line 120: [Really optional] FWIW, my initial thought was that you don't actually need a stack, because fields form a tree, so if you keep track of the "current parent" index, on `up()` you can make _its_ parent the current parent. But then I realized that this would only work if you assume no invalid parents. Since we're explicitly allowing to override parent indices, the "current parent" field may not have a valid parent to pop to. Might be worth a comment — your choice.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
```diff
@@ -58,19 +79,54 @@ private static final class Field {
...
   public ProtobufSchemaDescriptorBuilder addField(int fieldNumber, DType outputType) {
     Field f = new Field();
     f.fieldNumber = fieldNumber;
     f.outputTypeId = outputType.getTypeId().getNativeId();
+    if (!parentStack.isEmpty()) {
+      int parentIndex = parentStack.peek();
+      f.parentIndex = parentIndex;
+      f.depth = fields.get(parentIndex).depth + 1;
+    }
     fields.add(f);
```
igorpeshansky, line 98: [Optional] If you move this `.add` above the conditional, you could reuse `.parent` and turn the `if` into a guard clause:

```java
    fields.add(f);
    if (parentStack.isEmpty()) return this;
    return parent(parentStack.peek());
```

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
```diff
@@ -58,19 +79,54 @@ private static final class Field {
...
+      throw new IllegalStateException("up() called without a matching down()");
+    }
+    parentStack.pop();
+    return this;
+  }
+
+  /**
+   * Low-level: assign a raw parent index to the current field (depth = parent depth + 1),
+   * overriding any parent set by {@link #down()}. Prefer {@link #down()}/{@link #up()} for ordinary
+   * nesting. Reserve this for (a) loop-built degenerate chains where {@code parent(i - 1)} reads
```
igorpeshansky, line 127: [Really optional] If you reconsider allowing `.build()` from an unbalanced nested struct (as [suggested in lines 228-230](https://github.com/NVIDIA/spark-rapids-jni/pull/4663/changes/73ccafa9f0ba1c0ad9a747855a4a83daf3bfddd9..103cbe2bac92f61d7de3e0d77970c7ff0a987122#r3343248446)), you could also use `.down()` in a loop…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorBuilder.java
```diff
@@ -157,6 +225,9 @@ public ProtobufSchemaDescriptorBuilder enumNames(byte[][] names) {
   }
 
   public ProtobufSchemaDescriptor build() {
+    if (!parentStack.isEmpty()) {
+      throw new IllegalStateException("build() called inside a down() scope; missing up()");
+    }
```
igorpeshansky, line 230: [Really optional] I know this is a sanity check, but if you are building a descending chain, you may not want to bother calling `.up()` repeatedly to get back to the top. Just a thought…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -153,6 +151,57 @@ void testDepthAboveSupportedLimitRejected() {
...
+        .addField(2, DType.STRUCT).parent(2)
+        .addField(1, DType.INT32).parent(4)
+        .addField(3, DType.INT32).parent(0)
+        .build();
+
+    assertArrayEquals(viaExplicitParent.parentIndices, viaDownUp.parentIndices);
+    assertArrayEquals(viaExplicitParent.depthLevels, viaDownUp.depthLevels);
+    assertArrayEquals(viaExplicitParent.fieldNumbers, viaDownUp.fieldNumbers);
+    assertArrayEquals(viaExplicitParent.outputTypeIds, viaDownUp.outputTypeIds);
+    // Spot-check the derived flat indices directly.
```
igorpeshansky, line 187: Nit: "spot-check" usually means not checking all of them, while here you do. So I'd just say "check" instead.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -405,11 +295,11 @@ void testSchemaProjectionDecodeNone() {
...
          // With no fields in the schema, the GPU returns an empty struct
-         ColumnVector actualStruct = decodeAllFields(
+         ColumnVector actualStruct = Protobuf.decodeToStruct(
              input.getColumn(0),
-             new int[]{},  // no field numbers
-             new int[]{},  // no types
-             new int[]{})) {  // no encodings
+             new ProtobufSchemaDescriptorBuilder()
+                 .build(),
+             true)) {
```
igorpeshansky, line 302: Want to use the same formatting as in lines 344-345, i.e.:

```java
         ColumnVector actualStruct = Protobuf.decodeToStruct(input.getColumn(0),
             new ProtobufSchemaDescriptorBuilder().build(), true)) {
```

? Or be consistent about putting the builder and the arguments on separate lines…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -1383,17 +1267,12 @@ void testDefaultValueFieldPresent_OverridesDefault() {
...
-             new boolean[]{false, false},  // not required
-             new boolean[]{true, true},    // has default value
-             new long[]{42, 100},          // default values - NOT used since field is present
-             new double[]{0.0, 0.0},
-             new boolean[]{false, false},
-             new byte[][]{null, null},
+         ColumnVector actualStruct = Protobuf.decodeToStruct(
+             input.getColumn(0),
+             new ProtobufSchemaDescriptorBuilder()
+                 .addField(1, DType.INT32).defaultInt(42)
```
igorpeshansky, line 1273: [Optional] Keep the comment (`// NOT used since field is present`)?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -1356,17 +1245,12 @@ void testDefaultValueForMissingFields() {
...
-             new boolean[]{false, false},  // not required
-             new boolean[]{true, true},    // has default value
-             new long[]{42, 100},          // default int values (42, 100)
-             new double[]{0.0, 0.0},       // default float values (unused for int fields)
-             new boolean[]{false, false},  // default bool values (unused)
-             new byte[][]{null, null},     // default string values (unused)
+         ColumnVector actualStruct = Protobuf.decodeToStruct(
+             input.getColumn(0),
+             new ProtobufSchemaDescriptorBuilder()
+                 .addField(1, DType.INT32).defaultInt(42)
```
igorpeshansky, line 1251: [Really optional] Since the value types are fairly unique and visually distinguishable at call sites, these could all be overloads named `defaultValue` (i.e., `defaultValue(int)`, `defaultValue(double)`, `defaultValue(boolean)`, `defaultValue(byte[])`)… But it's also fine to keep the types in the names if you want to be explicit.

igorpeshansky (approved):
Only optional comments and formatting nits. I'll approve to unblock you, but if you want me to take another look after addressing them, I'd be happy to.

thirtiseven:
build

thirtiseven:
> Only optional comments and formatting nits. I'll approve to unblock you, but if you want me to take another look after addressing them, I'd be happy to.只有些可选的评论和格式上的小问题而已。我会同意解封你，不过如果你希望我在处理这些问题后再检查一遍，我也很乐意这么做。

Thanks! I'm going to merge it, as #4644 and #4646 depend on it. I will address those comments in a follow-up.

PR #4690, opened 2026-06-09
Honor configured POM file during deploy [skip ci]

NvTimLiu:
Use the POM_FILE environment variable for the final deploy-file invocation instead of hardcoding pom.xml.

This matches the configurable deployment flow from NVIDIA/spark-rapids#15002, where we can stage rewritten

Maven POMs for CUDF artifacts and pass their paths through POM_FILE. Without this change, ci/deploy.sh

evaluates the rewritten POM to derive the CUDF artifactId but then deploys with the original pom.xml,

causing coordinate mismatches for cudf-spark-jni artifacts.

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -94,7 +94,7 @@ echo "Deploy CMD: $DEPLOY_CMD"
 
 ###### Deploy spark-rapids-jni jar with all its additions ######
 $DEPLOY_CMD -Dfile=$FPATH.jar \
-            -DpomFile=pom.xml \
+            -DpomFile=$POM_FILE \
             -Dsources=$FPATH-sources.jar \
             -Djavadoc=$FPATH-javadoc.jar \
             -Dfiles=$CLASS_FILES \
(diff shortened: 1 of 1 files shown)


pxLi (approved):

NvTimLiu:
Skip CI because this file(ci/deploy.sh) cannot be run or covered by the pre-merge CI job

ci/deploy.sh
```diff
@@ -94,7 +94,7 @@ echo "Deploy CMD: $DEPLOY_CMD"
 
 ###### Deploy spark-rapids-jni jar with all its additions ######
 $DEPLOY_CMD -Dfile=$FPATH.jar \
-            -DpomFile=pom.xml \
+            -DpomFile=$POM_FILE \
```
pxLi, line 97: Q: Do we need a default here?

NvTimLiu:
build

ci/deploy.sh
NvTimLiu: > Q: Do we need a default here?

Do you mean `POM_FILE=${POM_FILE:-"pom.xml"}` ? 

Yes it is set ahead: https://github.com/NvTimLiu/spark-rapids-jni/blob/e61036bd0e6cb7cec37c96a01aa13eedcf9d5f62/ci/deploy.sh#L42

PR #4692, opened 2026-06-09
from_protobuf part3 followup: schema_context_view + kernel refactor

thirtiseven:
## Summary

Followup cleanup after PR #4644 (and #4663), preparing the codebase for upcoming part3 sub-tasks (3b.3-3b.6).

- **`schema_context_view` parameter bundle** - new `schema_context_view` in `protobuf_host_helpers.hpp` bundles the six per-decode metadata vectors into a single non-owning view. This reduces parameter pressure on recursive nested builders while leaving the enum lookup helper local until a follow-up PR needs a shared cache.
- **`build_nested_struct_column` API update** - declaration, definition, and call site updated to accept `schema_context_view` by value.
- **`on_repeated` callback signature** - `scan_message_field_locations` now passes the matched descriptor index `f` as the first argument to `on_repeated`. All three call sites are updated (`unreachable_repeated`, `validate_repeated`, `on_repeated_scan`).
- **`scan_all_repeated_occurrences_kernel` refactor** - delegates message walking to `scan_message_field_locations`, removing duplicated parse logic that diverged from the shared helper. The shared scanner now takes caller-supplied descriptor accessors so the repeated-occurrence hot path reads `repeated_field_scan_desc` directly instead of materializing a per-thread `field_descriptor[]`.
- **`ProtobufSchemaDescriptorBuilder` API cleanup** - `defaultInt/Float/Bool/String` collapsed into overloaded `defaultValue`; `addField` parent-tracking fixed to call `parent()` instead of mutating `parentIndex`/`depth` directly (a `Deque` is required because `parent()` allows non-ancestor jumps).
- **Test updates** - all existing tests migrated to the new `defaultValue` API.

No behavior change for any currently-supported decode path.

## Test plan

- [x] `git diff --check -- src/main/cpp/src/protobuf/protobuf_kernels.cu`
- [ ] Existing `ProtobufTest` and `ProtobufSchemaDescriptorTest` pass unchanged
- [ ] No new tests added (this is a pure refactor)

Part of NVIDIA/spark-rapids#14069


thirtiseven:
@greptile review

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -468,12 +462,7 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
   std::vector<int> const& child_field_indices,
   std::vector<nested_field_descriptor> const& schema,
   int num_fields,
-  std::vector<int64_t> const& default_ints,
-  std::vector<double> const& default_floats,
-  std::vector<bool> const& default_bools,
-  std::vector<cudf::detail::host_vector<uint8_t>> const& default_strings,
-  std::vector<cudf::detail::host_vector<int32_t>> const& enum_valid_values,
-  std::vector<std::vector<cudf::detail::host_vector<uint8_t>>> const& enum_names,
+  schema_context_view const& ctx,
```
igorpeshansky: [Really optional] This is just a bunch of pointers, so it may be worth passing it by value and letting the compiler optimize it.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -460,52 +460,41 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
-  // Descriptor-index lookup. No depth filter here (scan kernel runs against a flat list
-  // of top-level repeated descriptors), so the predicate is just the field-number match.
-  auto lookup_desc_idx = [&](int fn) -> int {
-    return lookup_field(fn, fn_to_desc_idx, fn_to_desc_size, num_scan_fields, [&](int f, int fn) {
+  // Build field_descriptor[] from scan_descs; all entries are repeated (on_repeated handles them).
+  field_descriptor fd[MAX_REPEATED_FIELDS_PER_KERNEL];
+  field_location dummy_out[MAX_REPEATED_FIELDS_PER_KERNEL];
+  for (int f = 0; f < num_scan_fields; f++) {
+    fd[f]        = {scan_descs[f].field_number, scan_descs[f].wire_type, true};
+    dummy_out[f] = {-1, 0};
```
igorpeshansky: `out` is a write-only parameter — why do you need to initialize it?

[Optional] In fact, we could even support passing in `nullptr` as `out`, and either conditionally storing the locations, or, better yet, just allocating the dummy on the stack within `scan_message_field_locations` if `nullptr` is passed.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -460,52 +460,41 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
+  }
+
+  auto lookup_by_fn = [&](int fn) -> int {
```
igorpeshansky: [Optional] The return type would be inferred here anyway.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -460,52 +460,41 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
 
-  // Descriptor-index lookup. No depth filter here (scan kernel runs against a flat list
-  // of top-level repeated descriptors), so the predicate is just the field-number match.
-  auto lookup_desc_idx = [&](int fn) -> int {
-    return lookup_field(fn, fn_to_desc_idx, fn_to_desc_size, num_scan_fields, [&](int f, int fn) {
+  // Build field_descriptor[] from scan_descs; all entries are repeated (on_repeated handles them).
+  field_descriptor fd[MAX_REPEATED_FIELDS_PER_KERNEL];
+  field_location dummy_out[MAX_REPEATED_FIELDS_PER_KERNEL];
+  for (int f = 0; f < num_scan_fields; f++) {
+    fd[f]        = {scan_descs[f].field_number, scan_descs[f].wire_type, true};
```
igorpeshansky: [Optional] `field_descs[f].field_number` is never used within `scan_message_field_locations`… And you could use designated initializers for the other fields, making this a bit more readable:

```suggestion
    fd[f] = {.expected_wire_type = scan_descs[f].wire_type, .is_repeated = true};
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -460,52 +460,41 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
-        occs[wi] = {row_i32, off, len};
-        wi++;
-        return true;
-      };
-      if (!walk_repeated_element(cur, msg_end, msg_base, wt, target_wt, error_flag, scan_action)) {
-        return;
+  auto row_i32 = static_cast<int32_t>(row);
+  auto on_repeated_scan =
+    [&](int f, uint8_t const* cur, uint8_t const* me, uint8_t const* mb, int wt, int expected_wt)
+    -> bool {
```
igorpeshansky: [Optional] The return type would be inferred here anyway. Also in `scan_action`…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -572,7 +561,8 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
         return field_descs[f].field_number == n;
       });
   };
-  auto validate_repeated = [&](uint8_t const* cur,
+  auto validate_repeated = [&](int /*f*/,
```
igorpeshansky: [Optional] Let's match `noop` below:

```suggestion
  auto validate_repeated = [&]([[maybe_unused]] int f,
```

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
```diff
@@ -211,8 +211,8 @@ void testSerializationRoundTripPreservesContentsAndIsolation() throws Exception
     ProtobufSchemaDescriptor original = new ProtobufSchemaDescriptorBuilder()
         .addField(1, DType.STRING)
             .enumMetadata(new int[]{0, 1}, new byte[][]{"A".getBytes(), "B".getBytes()})
-            .defaultString("def".getBytes())
-            .defaultInt(7)  // non-zero numeric default to exercise scalar round-trip
+            .defaultValue("def".getBytes())
+            .defaultValue(7L)  // non-zero numeric default to exercise scalar round-trip
```
igorpeshansky: [Optional] In `ProtobufTest`, you didn't have to add the `L` suffix to explicitly make this a `long`, and the tests still passed (and thus resolved to the `(long)` overload instead of the `(double)`)…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -1275,8 +1275,8 @@ void testDefaultValueFieldPresent_OverridesDefault() {
          ColumnVector actualStruct = Protobuf.decodeToStruct(
              input.getColumn(0),
              new ProtobufSchemaDescriptorBuilder()
-                 .addField(1, DType.INT32).defaultInt(42)
-                 .addField(2, DType.INT64).defaultInt(100)
+                 .addField(1, DType.INT32).defaultValue(42)
```
igorpeshansky, line 1279: [Optional] There was a comment here that it might make sense to keep (https://github.com/NVIDIA/spark-rapids-jni/pull/4663#discussion_r3343307036).

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -396,7 +397,19 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
+  enum_string_lookup_cache enum_lookup_cache;
+  // Bundle defaults + enum metadata into a single view so the recursive nested/repeated
+  // builders take one parameter instead of six identical references.
+  schema_context_view const schema_ctx{default_ints,
+                                       default_floats,
+                                       default_bools,
+                                       default_strings,
+                                       enum_valid_values,
+                                       enum_names,
+                                       &enum_lookup_cache};
```
igorpeshansky: No point in naming this — you can use the braced list initialization directly at the call site…

src/main/cpp/src/protobuf/protobuf_kernels.cu
igorpeshansky: Nit: `scan_message_field_locations` does not depend on `out` having any value at all — it's an output-only field, so the only reason to mention initialization here is to warn the callers that the field is written sparsely. So maybe let's just say that: ``The caller is responsible for initializing `out` if they care about the result.``

FWIW, it's also not clear how "row-level error marking" is related to `out`, so I would make that a separate sentence, e.g., ``The caller also owns row-level error marking; this helper only sets `error_flag` and…``.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -331,19 +364,14 @@ std::unique_ptr<cudf::column> build_repeated_struct_column(
   cudf::size_type message_data_size,
   cudf::size_type const* list_offsets,
   cudf::size_type base_offset,
-  rmm::device_uvector<int32_t> const& d_field_counts,
+  rmm::device_uvector<int32_t> d_field_offsets,
```
igorpeshansky: Is this change in scope for this PR? If so, please mention it in the PR description.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -30,10 +30,53 @@
...
+ * across recursive call sites instead of re-running `make_enum_string_lookup_tables`.
+ */
+struct schema_context_view {
+  std::vector<int64_t> const& default_ints;
+  std::vector<double> const& default_floats;
+  std::vector<bool> const& default_bools;
+  std::vector<cudf::detail::host_vector<uint8_t>> const& default_strings;
+  std::vector<cudf::detail::host_vector<int32_t>> const& enum_valid_values;
+  std::vector<std::vector<cudf::detail::host_vector<uint8_t>>> const& enum_names;
+  enum_string_lookup_cache* enum_lookup_cache = nullptr;
```
igorpeshansky: `enum_lookup_cache` seems to be unused — does it make sense to defer adding it (and the related infrastructure) until the follow-up PR that actually uses it?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -460,52 +460,41 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
 
-    uint8_t const* next;
-    if (!skip_field(cur, msg_end, wt, next)) {
-      set_error_once(error_flag, ERR_SKIP);
-      return;
-    }
-    cur = next;
-  }
+  scan_message_field_locations(
+    msg_base, msg_end, fd, dummy_out, error_flag, lookup_by_fn, on_repeated_scan);
```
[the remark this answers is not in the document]
igorpeshansky: +1, this should check the return value and `return` if `false`.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf.cu

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Agreed. Deferred the whole cache infrastructure to follow-up PRs.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Reverted here.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Reworded: `out` is output-only (only matched non-repeated fields are written), so the caller initializes it if it cares about the result; split the row-level-error-marking note into its own sentence; and added the third caller (`scan_all_repeated_occurrences_kernel`) to the list.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: TIL about designated initializers

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Done. `scan_message_field_locations` now accepts `out == nullptr` for validation-only scans — both `out[f]` writes are guarded. `scan_all_repeated_occurrences_kernel` (all-repeated schema) passes `nullptr` and drops the unused dummy array entirely. I went with conditional stores rather than a stack dummy because `out[f]` is indexed by descriptor index, so a single reused slot wouldn't be in-bounds.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: updated.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: updated.

src/main/cpp/src/protobuf/protobuf_kernels.cu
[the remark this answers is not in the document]
thirtiseven: Good catch — fixed. `scan_all_repeated_occurrences_kernel` now checks the return value and `return`s on `false`, matching the other two call sites and the old early-return behavior. The error flag was already correct (first-writer-wins), but bailing keeps the count check from running on a half-scanned message.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/test/java/com/nvidia/spark/rapids/jni/ProtobufSchemaDescriptorTest.java
thirtiseven: removed the 'L'.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Added it back

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -47,24 +47,29 @@ CUDF_KERNEL void set_error_if_unset_kernel(int* error_flag, int error_code)
...
  * pass a no-op handler since their descriptors are never repeated.
  */
 __device__ bool scan_message_field_locations(uint8_t const* msg_base,
                                              uint8_t const* msg_end,
-                                             field_descriptor const* field_descs,
                                              field_location* out,
                                              int* error_flag,
                                              auto&& lookup_desc_idx,
+                                             auto&& is_repeated_field,
+                                             auto&& get_expected_wire_type,
```
igorpeshansky, line 73: FWIW, I don't have a great intuition on how this would affect performance, so I defer to your judgment on making these abstractions. We probably want to do some side-by-side comparisons with prior code to avoid performance regressions.

src/main/cpp/src/protobuf/protobuf_kernels.cu
igorpeshansky: > I went with conditional stores rather than a stack dummy because `out[f]` is indexed by descriptor index, so a single reused slot wouldn't be in-bounds.

FWIW, I meant allocating the same `field_location dummy_out[MAX_REPEATED_FIELDS_PER_KERNEL]`, within `scan_message_field_locations`, rather than doing an `out != nullptr` check on each iteration. But it looks like doing the conditional write may actually be cheaper, so what you have is great.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -460,41 +467,42 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
-    int const we     = scan_descs[f].row_offsets[row + 1];
-    auto scan_action = [&](int32_t off, int32_t len) -> bool {
-      if (wi >= we) {
-        set_error_once(error_flag, ERR_REPEATED_COUNT_MISMATCH);
-        return false;
-      }
-      occs[wi] = {row_i32, off, len};
-      wi++;
-      return true;
+    [&](int f, uint8_t const* cur, uint8_t const* me, uint8_t const* mb, int wt, int expected_wt) {
```
igorpeshansky: [Optional] Now that we're passing `f` in, you can actually compute `expected_wt` within `on_repeated` instead of passing it in, and save a parameter (albeit at the cost of replicating the lookup)…

```suggestion
    [&](int f, uint8_t const* cur, uint8_t const* me, uint8_t const* mb, int wt) {
      int expected_wt  = scan_descs[f].wire_type;
```

Also in `scan_nested_message_fields_kernel`.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -349,7 +325,12 @@ std::unique_ptr<cudf::column> build_repeated_child_list_column(
   int child_schema_idx,
   std::vector<nested_field_descriptor> const& schema,
   int num_fields,
-  schema_context_view const& ctx,
+  std::vector<int64_t> const& default_ints,
+  std::vector<double> const& default_floats,
+  std::vector<bool> const& default_bools,
+  std::vector<cudf::detail::host_vector<uint8_t>> const& default_strings,
+  std::vector<cudf::detail::host_vector<int32_t>> const& enum_valid_values,
+  std::vector<std::vector<cudf::detail::host_vector<uint8_t>>> const& enum_names,
```
igorpeshansky: This one was in scope and probably worth keeping, since you're introducing `schema_context_view` across the board… I was only flagging that the `d_field_counts`→`d_field_offsets` replacement may have been out of scope, not the entire change to the `build_repeated_child_list_column` and `build_repeated_struct_column` signatures…

thirtiseven: You're right, I over-reverted. Restored `schema_context_view` (by value) on both `build_repeated_child_list_column` and `build_repeated_struct_column`.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Agreed. I will revisit it in the benchmark pr, let's keep it open.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Good idea, done.

igorpeshansky (approved):
LGTM :shipit:

thirtiseven:
build

PR #4717, opened 2026-06-15
Support non-repeated nested protobuf messages

thirtiseven:
## Description

This PR extends the protobuf JNI decoder for non-repeated nested messages.

It adds support for:

- nested string, bytes, and enum-as-string child fields
- nested string defaults and proto2 required-field checks
- recursive nested `STRUCT` children by computing grandchild parent locations on device
- depth-3 nested message decoding tests

Repeated children inside nested messages are intentionally still emitted as typed null `LIST`
columns. Repeated `MessageType` support remains out of scope for a follow-up PR.

## Validation

- `git diff --check`
- Verified touched protobuf C++ files do not contain raw `cudaMemcpyAsync`,
  `rmm::exec_policy(...)`, or single-argument `rmm::exec_policy_nosync(stream)` calls.

### DIFF
--- a/src/main/cpp/src/protobuf/protobuf.cu
+++ b/src/main/cpp/src/protobuf/protobuf.cu
@@ -488,31 +488,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   int const num_nested   = static_cast<int>(nested_field_indices.size());
   int const num_scalar   = static_cast<int>(scalar_field_indices.size());
 
-  // Error flag
-  rmm::device_uvector<int> d_error(1, stream, cudf::get_current_device_resource_ref());
-  CUDF_CUDA_TRY(cudaMemsetAsync(d_error.data(), 0, sizeof(int), stream.value()));
-  auto error_message = [](protobuf_error error) -> std::string {
-    switch (error) {
-      case protobuf_error::BOUNDS: return "Protobuf decode error: message data out of bounds";
-      case protobuf_error::VARINT: return "Protobuf decode error: invalid or truncated varint";
-      case protobuf_error::FIELD_NUMBER: return "Protobuf decode error: invalid field number";
-      case protobuf_error::WIRE_TYPE: return "Protobuf decode error: unexpected wire type";
-      case protobuf_error::OVERFLOW:
-        return "Protobuf decode error: length-delimited field overflows message";
-      case protobuf_error::FIELD_SIZE: return "Protobuf decode error: invalid field size";
-      case protobuf_error::SKIP: return "Protobuf decode error: unable to skip unknown field";
-      case protobuf_error::FIXED_LEN:
-        return "Protobuf decode error: invalid fixed-width or packed field length";
-      case protobuf_error::REQUIRED: return "Protobuf decode error: missing required field";
-      case protobuf_error::SCHEMA_TOO_LARGE:
-        return "Protobuf decode error: schema exceeds maximum supported repeated fields per "
-               "kernel (" +
-               std::to_string(MAX_REPEATED_FIELDS_PER_KERNEL) + ")";
-      case protobuf_error::REPEATED_COUNT_MISMATCH:
-        return "Protobuf decode error: repeated-field count/scan mismatch";
-      default: return "Protobuf decode error: unknown error";
-    }
-  };
+  auto d_error = cudf::detail::make_zeroed_device_uvector_async<protobuf_error>(
+    1, stream, cudf::get_current_device_resource_ref());
   // PERMISSIVE-mode row nulling support. Unknown enum values and malformed rows should both
   // surface as null structs instead of partially decoded data.
   bool const track_permissive_null_rows = !fail_on_errors;
@@ -604,10 +581,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
     rmm::device_uvector<field_descriptor> d_field_descs(
       num_scalar, stream, cudf::get_current_device_resource_ref());
-    CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_field_descs.data(),
-                                             h_field_descs.data(),
-                                             h_field_descs.size() * sizeof(field_descriptor),
-                                             stream));
+    CUDF_CUDA_TRY(cudf::detail::memcpy_async(
+      d_field_descs.data(), h_field_descs.data(), num_scalar * sizeof(field_descriptor), stream));
 
     rmm::device_uvector<field_location> d_locations(
       static_cast<size_t>(num_rows) * num_scalar, stream, cudf::get_current_device_resource_ref());
@@ -981,6 +956,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
     // Phase B: allocate occurrence buffers and launch the combined scan kernel.
     auto h_scan_descs = cudf::detail::make_pinned_vector_async<repeated_field_scan_desc>(0, stream);
+    h_scan_descs.reserve(num_repeated);
 
     for (auto& w : rep_work) {
       if (w.total_count > 0) {
@@ -1273,17 +1249,17 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
                     : make_null_column_with_schema(schema, i, num_fields, num_rows, stream, mr));
   }
 
-  CUDF_CUDA_TRY(cudaPeekAtLastError());
-  int h_error = 0;
-  CUDF_CUDA_TRY(cudf::detail::memcpy_async(&h_error, d_error.data(), sizeof(int), stream));
-  stream.synchronize();
-  auto const error = static_cast<protobuf_error>(h_error);
-  if (error == protobuf_error::SCHEMA_TOO_LARGE ||
-      error == protobuf_error::REPEATED_COUNT_MISMATCH) {
-    throw cudf::logic_error(error_message(error));
-  }
-  if (fail_on_errors && error != protobuf_error::NONE) {
-    throw cudf::logic_error(error_message(error));
+  {
+    using enum protobuf_error;
+    CUDF_CUDA_TRY(cudaPeekAtLastError());
+    protobuf_error h_error = NONE;
+    CUDF_CUDA_TRY(
+      cudf::detail::memcpy_async(&h_error, d_error.data(), sizeof(protobuf_error), stream));
+    stream.synchronize();
+    if (h_error == SCHEMA_TOO_LARGE || h_error == REPEATED_COUNT_MISMATCH) {
+      throw cudf::logic_error(error_message(h_error));
+    }
+    if (fail_on_errors && h_error != NONE) { throw cudf::logic_error(error_message(h_error)); }
   }
 
   // Build final struct null mask by combining input nulls with PERMISSIVE-mode row invalidation.
--- a/src/main/cpp/src/protobuf/protobuf.cu
+++ b/src/main/cpp/src/protobuf/protobuf.cu
@@ -162,6 +162,16 @@ struct repeated_field_work {
   }
 };
 
+template <typename T>
+cudf::detail::host_vector<T> make_pinned_staging_copy(std::vector<T> const& source,
+                                                      rmm::cuda_stream_view stream)
+{
+  // Stream-ordered pinned deallocation keeps this staging safe without a local sync.
+  auto staging = cudf::detail::make_pinned_vector_async<T>(source.size(), stream);
+  std::copy(source.begin(), source.end(), staging.begin());
+  return staging;
+}
+
 }  // namespace
 
 std::unique_ptr<cudf::column> make_null_column_with_schema(
@@ -453,7 +463,6 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   // Scratch allocations consumed inside this function go through the current device resource;
   // only buffers that flow into the returned column should use the caller-supplied `mr`.
   auto const scratch_mr = cudf::get_current_device_resource_ref();
-  pinned_staging_buffer_store pinned_staging_buffers;
 
   // Copy schema to device
   std::vector<device_nested_field_descriptor> h_device_schema(num_fields);
@@ -562,17 +571,15 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       build_index_lookup_table(schema.data(), nested_field_indices.data(), num_nested);
 
     if (!h_fn_to_rep.empty()) {
-      auto& h_pinned_fn_to_rep =
-        make_pinned_staging_copy(h_fn_to_rep, stream, pinned_staging_buffers);
-      d_fn_to_rep = rmm::device_uvector<int>(h_fn_to_rep.size(), stream, scratch_mr);
+      auto h_pinned_fn_to_rep = make_pinned_staging_copy(h_fn_to_rep, stream);
+      d_fn_to_rep             = rmm::device_uvector<int>(h_fn_to_rep.size(), stream, scratch_mr);
       CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_fn_to_rep.data(),
                                                h_pinned_fn_to_rep.data(),
                                                h_pinned_fn_to_rep.size() * sizeof(int),
                                                stream));
     }
     if (!h_fn_to_nested.empty()) {
-      auto& h_pinned_fn_to_nested =
-        make_pinned_staging_copy(h_fn_to_nested, stream, pinned_staging_buffers);
+      auto h_pinned_fn_to_nested = make_pinned_staging_copy(h_fn_to_nested, stream);
       d_fn_to_nested = rmm::device_uvector<int>(h_fn_to_nested.size(), stream, scratch_mr);
       CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_fn_to_nested.data(),
                                                h_pinned_fn_to_nested.data(),
@@ -614,8 +621,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
     rmm::device_uvector<field_descriptor> d_field_descs(
       num_scalar, stream, cudf::get_current_device_resource_ref());
-    auto& h_pinned_field_descs =
-      make_pinned_staging_copy(h_field_descs, stream, pinned_staging_buffers);
+    auto h_pinned_field_descs = make_pinned_staging_copy(h_field_descs, stream);
     CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_field_descs.data(),
                                              h_pinned_field_descs.data(),
                                              h_pinned_field_descs.size() * sizeof(field_descriptor),
@@ -628,8 +634,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     rmm::device_uvector<int> d_field_lookup(
       h_field_lookup.size(), stream, cudf::get_current_device_resource_ref());
     if (!h_field_lookup.empty()) {
-      auto& h_pinned_field_lookup =
-        make_pinned_staging_copy(h_field_lookup, stream, pinned_staging_buffers);
+      auto h_pinned_field_lookup = make_pinned_staging_copy(h_field_lookup, stream);
       CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_field_lookup.data(),
                                                h_pinned_field_lookup.data(),
                                                h_pinned_field_lookup.size() * sizeof(int),
@@ -658,7 +663,6 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
                                 track_permissive_null_rows ? d_row_force_null.data() : nullptr,
                                 nullptr,
                                 d_error.data(),
-                                pinned_staging_buffers,
                                 stream);
 
     // Batched scalar extraction: group non-special fixed-width fields by extraction
@@ -736,9 +740,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
         std::vector<std::unique_ptr<scalar_buf_pair>> bufs;
         bufs.reserve(nf);
-        auto& h_descs = keep_pinned_staging(
-          cudf::detail::make_pinned_vector_async<batched_scalar_desc>(nf, stream),
-          pinned_staging_buffers);
+        auto h_descs = cudf::detail::make_pinned_vector_async<batched_scalar_desc>(nf, stream);
 
         for (int j = 0; j < nf; j++) {
           int li   = idxs[j];
@@ -897,15 +899,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
             CUDF_EXPECTS(!valid_enums.empty() && valid_enums.size() == enum_name_bytes.size(),
                          "Protobuf decode error: missing or mismatched enum metadata for "
                          "enum-as-string field");
-            column_map[schema_idx] = build_enum_string_column(out,
-                                                              valid,
-                                                              valid_enums,
-                                                              enum_name_bytes,
-                                                              pinned_staging_buffers,
-                                                              d_row_force_null,
-                                                              num_rows,
-                                                              stream,
-                                                              mr);
+            column_map[schema_idx] = build_enum_string_column(
+              out, valid, valid_enums, enum_name_bytes, d_row_force_null, num_rows, stream, mr);
           } else {
             // Regular protobuf STRING (length-delimited)
             bool has_def_str    = has_def;
@@ -1023,8 +1018,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     if (!h_scan_descs.empty()) {
       rmm::device_uvector<repeated_field_scan_desc> d_scan_descs(
         h_scan_descs.size(), stream, scratch_mr);
-      auto& h_pinned_scan_descs =
-        make_pinned_staging_copy(h_scan_descs, stream, pinned_staging_buffers);
+      auto h_pinned_scan_descs = make_pinned_staging_copy(h_scan_descs, stream);
       CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_scan_descs.data(),
                                                h_pinned_scan_descs.data(),
                                                h_pinned_scan_descs.size() * sizeof(h_scan_descs[0]),
@@ -1038,8 +1032,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       rmm::device_uvector<int> d_fn_to_scan(0, stream, scratch_mr);
       int fn_to_scan_size = 0;
       if (!h_fn_to_scan.empty()) {
-        auto& h_pinned_fn_to_scan =
-          make_pinned_staging_copy(h_fn_to_scan, stream, pinned_staging_buffers);
+        auto h_pinned_fn_to_scan = make_pinned_staging_copy(h_fn_to_scan, stream);
         d_fn_to_scan = rmm::device_uvector<int>(h_fn_to_scan.size(), stream, scratch_mr);
         CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_fn_to_scan.data(),
                                                  h_pinned_fn_to_scan.data(),
@@ -1215,7 +1208,6 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
                                                                        num_rows,
                                                                        field_meta.enum_valid_values,
                                                                        field_meta.enum_names,
-                                                                       pinned_staging_buffers,
                                                                        d_row_force_null,
                                                                        d_error,
                                                                        stream,
@@ -1285,7 +1277,6 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       schema,
       num_fields,
       {default_ints, default_floats, default_bools, default_strings, enum_valid_values, enum_names},
-      pinned_staging_buffers,
       d_row_force_null,
       d_error,
       num_rows,
--- a/src/main/cpp/src/protobuf/protobuf.cu
+++ b/src/main/cpp/src/protobuf/protobuf.cu
@@ -544,10 +544,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
                                              stream));
   }
   if (num_nested > 0) {
-    CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_nested_indices.data(),
-                                             nested_field_indices.data(),
-                                             num_nested * sizeof(int),
-                                             stream));
+    CUDF_CUDA_TRY(cudf::detail::memcpy_async(
+      d_nested_indices.data(), nested_field_indices.data(), num_nested * sizeof(int), stream));
   }
 
   // Count repeated fields at depth 0 (with O(1) field_number lookup tables)
@@ -562,17 +560,13 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
     if (!h_fn_to_rep.empty()) {
       d_fn_to_rep = rmm::device_uvector<int>(h_fn_to_rep.size(), stream, scratch_mr);
-      CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_fn_to_rep.data(),
-                                               h_fn_to_rep.data(),
-                                               h_fn_to_rep.size() * sizeof(int),
-                                               stream));
+      CUDF_CUDA_TRY(cudf::detail::memcpy_async(
+        d_fn_to_rep.data(), h_fn_to_rep.data(), h_fn_to_rep.size() * sizeof(int), stream));
     }
     if (!h_fn_to_nested.empty()) {
       d_fn_to_nested = rmm::device_uvector<int>(h_fn_to_nested.size(), stream, scratch_mr);
-      CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_fn_to_nested.data(),
-                                               h_fn_to_nested.data(),
-                                               h_fn_to_nested.size() * sizeof(int),
-                                               stream));
+      CUDF_CUDA_TRY(cudf::detail::memcpy_async(
+        d_fn_to_nested.data(), h_fn_to_nested.data(), h_fn_to_nested.size() * sizeof(int), stream));
     }
 
     launch_count_repeated_fields(*d_in,
@@ -609,10 +603,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
     rmm::device_uvector<field_descriptor> d_field_descs(
       num_scalar, stream, cudf::get_current_device_resource_ref());
-    CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_field_descs.data(),
-                                             h_field_descs.data(),
-                                             num_scalar * sizeof(field_descriptor),
-                                             stream));
+    CUDF_CUDA_TRY(cudf::detail::memcpy_async(
+      d_field_descs.data(), h_field_descs.data(), num_scalar * sizeof(field_descriptor), stream));
 
     rmm::device_uvector<field_location> d_locations(
       static_cast<size_t>(num_rows) * num_scalar, stream, cudf::get_current_device_resource_ref());
@@ -621,10 +613,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     rmm::device_uvector<int> d_field_lookup(
       h_field_lookup.size(), stream, cudf::get_current_device_resource_ref());
     if (!h_field_lookup.empty()) {
-      CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_field_lookup.data(),
-                                               h_field_lookup.data(),
-                                               h_field_lookup.size() * sizeof(int),
-                                               stream));
+      CUDF_CUDA_TRY(cudf::detail::memcpy_async(
+        d_field_lookup.data(), h_field_lookup.data(), h_field_lookup.size() * sizeof(int), stream));
     }
 
     launch_scan_all_fields(*d_in,
@@ -968,11 +958,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       // overflow before truncating into the int32 LIST offsets buffer (cudf list offsets are
       // int32). The same total is reused as the `offsets[num_rows]` sentinel below, so this
       // single reduce serves both the overflow guard and the offsets construction.
-      int64_t total_64 =
-        thrust::reduce(rmm::exec_policy_nosync(stream, scratch_mr),
-                       counts_begin,
-                       counts_end,
-                       int64_t{0});
+      int64_t total_64 = thrust::reduce(
+        rmm::exec_policy_nosync(stream, scratch_mr), counts_begin, counts_end, int64_t{0});
       CUDF_EXPECTS(total_64 <= std::numeric_limits<int32_t>::max(),
                    "Top-level repeated field total element count exceeds 2^31-1");
       w.total_count = static_cast<int32_t>(total_64);
@@ -985,10 +972,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
                                w.offsets.begin(),
                                int32_t{0});
       }
-      thrust::fill_n(rmm::exec_policy_nosync(stream, scratch_mr),
-                     w.offsets.data() + num_rows,
-                     1,
-                     w.total_count);
+      thrust::fill_n(
+        rmm::exec_policy_nosync(stream, scratch_mr), w.offsets.data() + num_rows, 1, w.total_count);
     }
 
     // Phase B: allocate occurrence buffers and launch the combined scan kernel.
@@ -1023,10 +1008,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       int fn_to_scan_size = 0;
       if (!h_fn_to_scan.empty()) {
         d_fn_to_scan = rmm::device_uvector<int>(h_fn_to_scan.size(), stream, scratch_mr);
-        CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_fn_to_scan.data(),
-                                                 h_fn_to_scan.data(),
-                                                 h_fn_to_scan.size() * sizeof(int),
-                                                 stream));
+        CUDF_CUDA_TRY(cudf::detail::memcpy_async(
+          d_fn_to_scan.data(), h_fn_to_scan.data(), h_fn_to_scan.size() * sizeof(int), stream));
         fn_to_scan_size = static_cast<int>(h_fn_to_scan.size());
       }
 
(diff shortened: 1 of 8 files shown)


thirtiseven:
@greptile full review

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -567,20 +570,20 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-      CUDF_CUDA_TRY(cudaMemcpyAsync(d_fn_to_nested.data(),
-                                    h_fn_to_nested.data(),
-                                    h_fn_to_nested.size() * sizeof(int),
-                                    cudaMemcpyHostToDevice,
-                                    stream.value()));
+      CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_fn_to_nested.data(),
+                                               h_pinned_fn_to_nested.data(),
+                                               h_pinned_fn_to_nested.size() * sizeof(int),
+                                               stream));
     }
```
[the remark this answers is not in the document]
thirtiseven: Thanks for flagging this. I revisited the cuDF/RMM lifetime rules here.

These staging buffers are created with cudf::detail::make_pinned_vector_async(..., stream), which uses cuDF's stream-ordered pinned memory resource. When the host_vector is destroyed, deallocation is ordered on the same stream, so the buffer cannot be reused before the earlier memcpy_async work on that stream has consumed it.
If these were pageable std::vector buffers or host buffers from a non-stream-ordered allocator, I agree we would need either a longer-lived owner or an explicit sync. For the pinned staging used here, the local lifetime is safe and matches existing cuDF patterns.

I reverted the keepalive-store change and added short comments at the staging sites to make this explicit.

thirtiseven:
build

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -567,20 +571,20 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       build_index_lookup_table(schema.data(), nested_field_indices.data(), num_nested);
 
     if (!h_fn_to_rep.empty()) {
-      d_fn_to_rep = rmm::device_uvector<int>(h_fn_to_rep.size(), stream, scratch_mr);
-      CUDF_CUDA_TRY(cudaMemcpyAsync(d_fn_to_rep.data(),
-                                    h_fn_to_rep.data(),
-                                    h_fn_to_rep.size() * sizeof(int),
-                                    cudaMemcpyHostToDevice,
-                                    stream.value()));
+      auto h_pinned_fn_to_rep = make_pinned_staging_copy(h_fn_to_rep, stream);
```
igorpeshansky: I'm wondering, why do we need a pinned _copy_? Can we not have these be pinned vectors from the outset (like you do below with `h_descs`)?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1280,7 +1284,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
       mr,
       nullptr,
       0,
-      track_permissive_null_rows);
+      false);
```
igorpeshansky, line 1237: Other calls maintain the invariant that `d_row_force_null.size() > 0` == `propagate_invalid_rows`. Does this break that invariant? Should `d_row_force_null` change here as well?

In fact, why did we need two parameters in the first place, given the invariant?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+      case cudf::type_id::STRING: {
+        nested_location_provider loc_provider{list_offsets,
+                                              base_offset,
+                                              d_parent_locs.data(),
+                                              d_child_locations.data(),
+                                              ci,
+                                              num_child_fields};
+        if (enc == encoding_value(proto_encoding::ENUM_STRING)) {
+          rmm::device_uvector<int32_t> out(num_rows, stream, mr);
+          rmm::device_uvector<bool> valid((num_rows > 0 ? num_rows : 1), stream, mr);
```
igorpeshansky: Lines 492-495 already guarantee that `num_rows` is at least 1.

```suggestion
          rmm::device_uvector<bool> valid(num_rows, stream, mr);
```

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+          extract_varint_kernel<int32_t, false, nested_location_provider>
+            <<<blocks, threads, 0, stream.value()>>>(
+              message_data,
+              loc_provider,
+              num_rows,
+              out.data(),
+              valid.data(),
+              d_error.data(),
+              has_def,
+              has_def ? ctx.default_ints[child_schema_idx] : 0);
```
igorpeshansky: [Optional] Aren't we passing appropriately-sized arrays of default values anyway, even if `has_def` is false? Can we just unconditionally access `ctx.default_ints[child_schema_idx]`? FWIW, this applies throughout…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+          if (child_schema_idx < static_cast<int>(ctx.enum_valid_values.size()) &&
+              child_schema_idx < static_cast<int>(ctx.enum_names.size()) &&
```
igorpeshansky: Given the size parity, the second check is redundant. In fact, the entire bounds check is redundant, because `validate_decode_context` checks that `ctx.enum_valid_values.size() == num_fields`, and `child_schema_idx` is in `[0, num_fields)` by construction. So the `else` block in lines 643-646 is effectively unreachable. And since that seems to be the only place that sets `ERR_MISSING_ENUM_META`, that error may be dead as well…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+                           flocs = d_child_locations.data(),
+                           ci,
+                           num_child_fields,
+                           has_def] __device__(cudf::size_type row) {
+            return (plocs[row].offset >= 0 &&
+                    flocs[flat_index(static_cast<size_t>(row),
+                                     static_cast<size_t>(num_child_fields),
+                                     static_cast<size_t>(ci))]
+                        .offset >= 0) ||
+                   has_def;
```
igorpeshansky: [Optional] If you check `has_def` first, it'll short-circuit the index computation and the array indexing…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+                                                     has_def,
+                                                     ctx.default_strings[child_schema_idx],
+                                                     d_error,
+                                                     stream,
+                                                     mr));
+        }
+        break;
+      }
+      case cudf::type_id::LIST: {  // bytes represented as LIST<UINT8>
+        nested_location_provider loc_provider{list_offsets,
```
igorpeshansky: This entire case is identical to the non-enum string case. Can we combine them, e.g.:

```c++
      case cudf::type_id::STRING:
      case cudf::type_id::LIST: {  // bytes represented as LIST<UINT8>
        nested_location_provider loc_provider{…};
        if (dt.id() == cudf::type_id::STRING && enc == encoding_value(proto_encoding::ENUM_STRING)) {
          …
        } else {
          auto valid_fn = …;
          …
        }
        break;
      }
```

?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+                            [plocs = d_parent_locs.data(),
+                             flocs = d_child_locations.data(),
+                             ci,
+                             num_child_fields] __device__(cudf::size_type row) {
+                              return plocs[row].offset >= 0 &&
+                                     flocs[flat_index(static_cast<size_t>(row),
+                                                      static_cast<size_t>(num_child_fields),
+                                                      static_cast<size_t>(ci))]
+                                         .offset >= 0;
+                            });
```
igorpeshansky: This would benefit from having a named local (e.g., `valid_fn`)…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+                                                   has_def,
+                                                   ctx.default_strings[child_schema_idx],
+                                                   d_error,
+                                                   stream,
+                                                   mr));
         break;
+      }
+      case cudf::type_id::STRUCT: {
+        auto gc_indices = find_child_field_indices(schema, num_fields, child_schema_idx);
+        if (gc_indices.empty()) {
```
igorpeshansky: Is this condition even necessary? It looks like calling `build_nested_struct_column` with an empty `child_field_indices` would produce exactly the same result (well, almost — but if we pass in an empty `child_locations` at the top level, we can then propagate `d_child_locations` to the nested calls, and it _would_ be the same)…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
-      case cudf::type_id::STRUCT:
-        // Nested string/bytes/enum-as-string (3b.3) and recursive struct (3b.4) children are
-        // not decoded yet; emit a typed null column so the output schema still matches.
+      case cudf::type_id::STRING: {
+        nested_location_provider loc_provider{list_offsets,
+                                              base_offset,
+                                              d_parent_locs.data(),
+                                              d_child_locations.data(),
+                                              ci,
+                                              num_child_fields};
```
igorpeshansky: This is the same for most cases — can we pull it out of the `switch`?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+        } else {
+          auto valid_fn = [plocs = d_parent_locs.data(),
+                           flocs = d_child_locations.data(),
+                           ci,
+                           num_child_fields,
+                           has_def] __device__(cudf::size_type row) {
+            return (plocs[row].offset >= 0 &&
+                    flocs[flat_index(static_cast<size_t>(row),
+                                     static_cast<size_t>(num_child_fields),
+                                     static_cast<size_t>(ci))]
```
igorpeshansky: This part of the function is just `loc_provider.get(row, offset); offset >= 0`. Also below; we can express all of these via a nested location provider. In fact, we could even add a `__device__ inline bool valid(int thread_idx) const` function to `nested_location_provider` to encapsulate this and avoid the dummy locals and throwaway computations…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -593,6 +593,34 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
...
+  }
+
+  auto const abs_offset = static_cast<int64_t>(parent_loc.offset) + child_loc.offset;
+  if (abs_offset > cuda::std::numeric_limits<int32_t>::max()) {
+    gc_parent_locs[row] = {-1, 0};
+    set_error_once(error_flag, ERR_OVERFLOW);
+    return;
+  }
+
+  gc_parent_locs[row] = {static_cast<int32_t>(abs_offset), child_loc.length};
```
igorpeshansky: This kernel seems to be doing almost the same exact computations as `nested_location_provider::get`. I wonder if we could just pass the location provider around and add a function to do this (or modify `get` appropriately)?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2543,26 +2543,6 @@ void testNestedMessageOutputShape() {
     }
   }
 
-  @Test
-  void testZeroLengthNestedMessage() {
```
igorpeshansky, line 2547: Is this test no longer relevant? Or is it now covered by some other test?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2693,6 +2673,227 @@ void testNestedMessageMultipleScalarChildren() {
     }
   }
 
+  @Test
+  void testNestedMessageVarlenChildren() {
+    // message Inner { string name = 1; bytes payload = 2; }
+    // message Outer { Inner inner = 1; }
+    byte[] payload = new byte[]{1, 2, 3};
+    Byte[] innerMessage = concat(
+        box(tag(1, WT_LEN)), box(encodeVarint(5)), box("alice".getBytes()),
```
igorpeshansky: The second part of it is the same as `encodeMessage(box("alice".getBytes()))`. Now, I know it's not a message, but we could extract `encodeBytes`, and even add an overload with `byte[]`, e.g.:

```java
  /** Encode a length-delimited byte sequence: varint length prefix followed by the bytes. */
  private static Byte[] encodeBytes(Byte[] bytes) {
    return concat(box(encodeVarint(bytes.length)), bytes);
  }
  private static Byte[] encodeBytes(byte[] bytes) {
    return encodeBytes(box(bytes));
  }

  /** Encode a length-delimited submessage: varint length prefix followed by the message bytes. */
  private static Byte[] encodeMessage(Byte[] messageBytes) {
    return encodeBytes(messageBytes);
  }
```

Then these become:

```java
    Byte[] innerMessage = concat(
        box(tag(1, WT_LEN)), encodeBytes("alice".getBytes()),
        box(tag(2, WT_LEN)), encodeBytes(payload));
```

You could even add an `encodeString` wrapper…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2693,6 +2673,227 @@ void testNestedMessageMultipleScalarChildren() {
...
+    // message Inner  { int32 a = 1; string b = 2; bool c = 3; }
+    // message Middle { Inner inner = 1; int64 m = 2; }
+    // message Outer  { Middle middle = 1; float score = 2; }
+    Byte[] inner = concat(
+        box(tag(1, WT_VARINT)), box(encodeVarint(7)),
+        box(tag(2, WT_LEN)), box(encodeVarint(3)), box("abc".getBytes()),
+        box(tag(3, WT_VARINT)), box(encodeVarint(1)));
+    Byte[] middle = concat(
+        box(tag(1, WT_LEN)), encodeMessage(inner),
+        box(tag(2, WT_VARINT)), box(encodeVarint(123L)));
```
igorpeshansky: Do we need the `L` here (and below)?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2693,6 +2673,227 @@ void testNestedMessageMultipleScalarChildren() {
...
+    // message Outer { int32 id = 1; Inner inner = 2; string name = 3; }
+    // message Inner { enum Status { UNKNOWN=0; OK=1; BAD=2; } Status status = 1;
+    //                 int32 count = 2; }
+    Byte[] innerValid = concat(
+        box(tag(1, WT_VARINT)), box(encodeVarint(1)),
+        box(tag(2, WT_VARINT)), box(encodeVarint(10)));
+    Byte[] innerInvalid = concat(
+        box(tag(1, WT_VARINT)), box(encodeVarint(999)),
+        box(tag(2, WT_VARINT)), box(encodeVarint(20)));
+    Byte[] row0 = concat(
```
igorpeshansky: Can this be a `Byte[][] rows` instead?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+              out.data(),
+              valid.data(),
+              d_error.data(),
+              has_def,
+              has_def ? ctx.default_ints[child_schema_idx] : 0);
+          if (child_schema_idx < static_cast<int>(ctx.enum_valid_values.size()) &&
+              child_schema_idx < static_cast<int>(ctx.enum_names.size()) &&
+              !ctx.enum_valid_values[child_schema_idx].empty() &&
+              ctx.enum_valid_values[child_schema_idx].size() ==
+                ctx.enum_names[child_schema_idx].size()) {
```
igorpeshansky: These are already checked by `validate_decode_context`.

src/main/cpp/src/protobuf/protobuf_builders.cu
igorpeshansky: Missed this in #4644: let's include the offending output type in the error?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -592,14 +604,158 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+                                                   valid_fn,
+                                                   has_def,
+                                                   ctx.default_strings[child_schema_idx],
+                                                   d_error,
+                                                   stream,
+                                                   mr));
         break;
+      }
+      case cudf::type_id::STRUCT: {
+        auto gc_indices = find_child_field_indices(schema, num_fields, child_schema_idx);
```
igorpeshansky, line 656: [Optional] Should we add the same kind of note about the linear complexity of this as in [`protobuf.cu:1259-1261`](https://github.com/thirtiseven/spark-rapids-jni/blob/4f97965e31c97ea26a9bc83382ca5757360f2326/src/main/cpp/src/protobuf/protobuf.cu#L1259-L1261)? Given that this is now called recursively, the complexity may compound…

igorpeshansky (commented):
None of the comments are technically blocking, so I won't 'request changes', but some of the cleanups could be impactful.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Good point! I changed these to be pinned vectors from the outset and removed the pinned staging copy helper.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Good catch. The old invariant was true when `d_row_force_null` was only used for enum invalid-row propagation. In the nested path, the buffer is also needed for row-level failures such as missing required fields, while invalid nested enum values should only null the enum child field. I renamed the flag to `propagate_invalid_enum_rows` and added a comment to make that narrower meaning explicit.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Good catch, removed.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Added a comment.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: good catch, removed.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Done. I added `get_nested_parent_location()` to `nested_location_provider` and now reuse it from the grandchild parent-location kernel.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: This case is still covered by `testZeroLengthNestedMessage_ChildIsNull`. It uses a present nested tag with a length-0 body, an INT32 child with no default, and verifies that the struct row is present while the child field is null.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Done. I added `encodeBytes` overloads and an `encodeString` helper, and updated the new length-delimited test payloads to use them.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Done. I added `nested_location_provider::valid()` and use it for the nested string/bytes validity check.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -61,34 +62,44 @@ struct schema_context_view {
...
+  rmm::cuda_stream_view stream)
 {
-  return build_lookup_table([&](int i) { return schema[field_indices[i]].field_number; },
-                            num_indices);
+  return build_pinned_lookup_table(
+    [&](int i) { return schema[field_indices[i]].field_number; }, num_indices, stream);
 }
 
-inline std::vector<int> build_field_lookup_table(field_descriptor const* descs, int num_fields)
+inline cudf::detail::host_vector<int> build_pinned_field_lookup_table(field_descriptor const* descs,
```
igorpeshansky: [Optional] If you make this generic in the field descriptor type, it could apply as-is to `repeated_field_scan_desc` as well…

```c++
template <typename FD>
inline cudf::detail::host_vector<int> build_pinned_field_lookup_table(FD const* descs,
                                                                      int num_fields,
                                                                      rmm::cuda_stream_view stream)
```

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -61,34 +62,44 @@ struct schema_context_view {
  * Build a host-side direct-mapped lookup table: field_number -> index.
  * @param get_field_number Callable: (int i) -> field_number for the i-th entry.
  * @param num_entries Number of entries.
- * @return Empty vector if the max field number exceeds the threshold.
+ * @return Empty pinned vector if the max field number exceeds the threshold.
  */
 template <typename FieldNumberFn>
-inline std::vector<int> build_lookup_table(FieldNumberFn get_field_number, int num_entries)
+inline cudf::detail::host_vector<int> build_pinned_lookup_table(FieldNumberFn get_field_number,
```
igorpeshansky: IMO, this rename is superfluous, since all of these functions now unconditionally return a pinned vector. I'd just keep them as they were (`build_lookup_table`, `build_index_lookup_table`, `build_field_lookup_table`) and just change the return type…

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1001,49 +979,46 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
 
     for (auto& w : rep_work) {
       if (w.total_count > 0) {
         w.occurrences = std::make_unique<rmm::device_uvector<repeated_occurrence>>(
           w.total_count, stream, scratch_mr);
-        h_scan_descs.push_back({schema[w.schema_idx].field_number,
-                                static_cast<int>(schema[w.schema_idx].wire_type),
-                                w.offsets.data(),
-                                w.occurrences->data()});
+        h_scan_descs[num_scan_descs++] = {schema[w.schema_idx].field_number,
```
igorpeshansky: This seems like a superfluous change — `cudf::detail::host_vector` inherits `push_back` from `thrust::host_vector`, so the previous code would have worked just fine. Ditto for `empty()` and `size()` below…

src/main/cpp/src/protobuf/protobuf_types.cuh
```diff
@@ -37,8 +37,7 @@ constexpr int ERR_SKIP                    = 7;
 constexpr int ERR_FIXED_LEN               = 8;
 constexpr int ERR_REQUIRED                = 9;
 constexpr int ERR_SCHEMA_TOO_LARGE        = 10;
-constexpr int ERR_MISSING_ENUM_META       = 11;
-constexpr int ERR_REPEATED_COUNT_MISMATCH = 12;
+constexpr int ERR_REPEATED_COUNT_MISMATCH = 11;
```
igorpeshansky: Just checking, is it safe to change the enum value here? Any non-test code that relies on this?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -546,14 +560,23 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
       continue;
     }
 
+    nested_location_provider loc_provider{list_offsets,
+                                          base_offset,
+                                          d_parent_locs.data(),
+                                          d_child_locations.data(),
+                                          ci,
+                                          num_child_fields};
+    auto valid_fn = [loc_provider, has_def] __device__(cudf::size_type row) {
```
igorpeshansky: [Really optional] This is only used in the `STRING`/`LIST` `else` block, but it doesn't hurt to define it here, so feel free to keep…

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2679,8 +2693,8 @@ void testNestedMessageVarlenChildren() {
     // message Outer { Inner inner = 1; }
     byte[] payload = new byte[]{1, 2, 3};
     Byte[] innerMessage = concat(
-        box(tag(1, WT_LEN)), box(encodeVarint(5)), box("alice".getBytes()),
-        box(tag(2, WT_LEN)), box(encodeVarint(payload.length)), box(payload));
+        box(tag(1, WT_LEN)), encodeString("alice"),
+        box(tag(2, WT_LEN)), encodeBytes(payload));
```
igorpeshansky: [Optional] Now that it's only referenced once, you could inline `payload` here:

```java
        box(tag(2, WT_LEN)), encodeBytes(new byte[]{1, 2, 3}));
```

src/main/cpp/src/protobuf/protobuf_types.cuh
igorpeshansky: [Really optional] Definitely out of scope for this PR, but just wondering why this isn't an `enum class`…

src/main/cpp/src/protobuf/protobuf_kernels.cuh
```diff
@@ -93,16 +94,32 @@ struct nested_location_provider {
   int field_idx;
   int num_fields;
 
-  __device__ inline field_location get(int thread_idx, int32_t& data_offset) const
+  __device__ inline field_location get_nested_parent_location(int thread_idx, int* error_flag) const
```
igorpeshansky: Just a note that this has a couple of descriptors piled up, and doesn't make it clear which descriptors refer to which nouns — i.e., is it "get location within a nested parent", or "get location of the nested parent", or "get parent location for a nested field"? A comment would probably help, or we could brainstorm a better name…

src/main/cpp/src/protobuf/protobuf_builders.cu
igorpeshansky: Here's where the _almost_ comes in: `build_nested_struct_column` eventually calls `launch_scan_nested_message_fields`, which the block you removed does not. So a malformed encoding of a childless struct child would now error out, while the previous code would silently accept it. Arguably, the previous state was buggy and the current state is a bug fix, but wanted you to be aware of it.

Note that we don't have any tests that would have caught the difference — should we?

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Thanks! I changed them back

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Agreed, done.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Moved to the only branch that uses it.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Done. I made `build_field_lookup_table` generic in the descriptor type and reused it for `repeated_field_scan_desc`.

src/main/cpp/src/protobuf/protobuf_kernels.cuh
thirtiseven: Done. I renamed it to `get_rebased_child_location` and added a short comment explaining that it rebases child offsets from the parent message to the row for recursive STRUCT
decode.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf_types.cuh
thirtiseven: Good catch, updated to an enum

src/main/cpp/src/protobuf/protobuf_types.cuh
thirtiseven: It's safe, and we don't need the `ERR_MISSING_ENUM_META` any more

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Thanks for the explanation! Added a new case `testMalformedChildlessNestedMessage_FailsFast`.

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -491,23 +491,24 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-      case ERR_SCHEMA_TOO_LARGE:
+      case protobuf_error::REQUIRED: return "Protobuf decode error: missing required field";
+      case protobuf_error::SCHEMA_TOO_LARGE:
         return "Protobuf decode error: schema exceeds maximum supported repeated fields per "
                "kernel (" +
                std::to_string(MAX_REPEATED_FIELDS_PER_KERNEL) + ")";
-      case ERR_REPEATED_COUNT_MISMATCH:
+      case protobuf_error::REPEATED_COUNT_MISMATCH:
         return "Protobuf decode error: repeated-field count/scan mismatch";
       default: return "Protobuf decode error: unknown error";
```
igorpeshansky: If you want to take advantage of exhaustive switch checks, put this `return` _after_ the `switch` statement. You'd need to add a case for `protobuf_error::NONE`, though…

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -491,23 +491,24 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-      case ERR_VARINT: return "Protobuf decode error: invalid or truncated varint";
-      case ERR_FIELD_NUMBER: return "Protobuf decode error: invalid field number";
-      case ERR_WIRE_TYPE: return "Protobuf decode error: unexpected wire type";
-      case ERR_OVERFLOW: return "Protobuf decode error: length-delimited field overflows message";
-      case ERR_FIELD_SIZE: return "Protobuf decode error: invalid field size";
-      case ERR_SKIP: return "Protobuf decode error: unable to skip unknown field";
-      case ERR_FIXED_LEN:
+  auto error_message = [](protobuf_error error) -> std::string {
+    switch (error) {
+      case protobuf_error::BOUNDS: return "Protobuf decode error: message data out of bounds";
```
igorpeshansky: [Optional] If you add `using enum protobuf_error;`[^1], this becomes much shorter:

```c++
  auto error_message = [](protobuf_error error) -> std::string {
    switch (error) {
      using enum protobuf_error;
      case BOUNDS: return "Protobuf decode error: message data out of bounds";
      case VARINT: return "Protobuf decode error: invalid or truncated varint";
      case FIELD_NUMBER: return "Protobuf decode error: invalid field number";
      case WIRE_TYPE: return "Protobuf decode error: unexpected wire type";
      case OVERFLOW: return "Protobuf decode error: length-delimited field overflows message";
      case FIELD_SIZE: return "Protobuf decode error: invalid field size";
      case SKIP: return "Protobuf decode error: unable to skip unknown field";
      case FIXED_LEN: return "Protobuf decode error: invalid fixed-width or packed field length";
      case REQUIRED: return "Protobuf decode error: missing required field";
      case SCHEMA_TOO_LARGE:
        return "Protobuf decode error: schema exceeds maximum supported repeated fields per "
               "kernel (" +
               std::to_string(MAX_REPEATED_FIELDS_PER_KERNEL) + ")";
      case REPEATED_COUNT_MISMATCH:
        return "Protobuf decode error: repeated-field count/scan mismatch";
      default: return "Protobuf decode error: unknown error";
    }
  };
```

[^1]: https://en.cppreference.com/cpp/language/enum#using_enum_declaration

src/main/cpp/src/protobuf/protobuf_types.cuh
```diff
@@ -26,18 +26,22 @@ constexpr int MAX_VARINT_BYTES = 10;
...
-constexpr int ERR_OVERFLOW                = 5;
-constexpr int ERR_FIELD_SIZE              = 6;
-constexpr int ERR_SKIP                    = 7;
-constexpr int ERR_FIXED_LEN               = 8;
-constexpr int ERR_REQUIRED                = 9;
-constexpr int ERR_SCHEMA_TOO_LARGE        = 10;
-constexpr int ERR_REPEATED_COUNT_MISMATCH = 11;
+enum class protobuf_error : int {
+  NONE                    = 0,
+  BOUNDS                  = 1,
```
igorpeshansky: [Really optional] You could leave just the initial assignment and rely on auto-increment…

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -491,23 +491,24 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-    switch (code) {
-      case ERR_BOUNDS: return "Protobuf decode error: message data out of bounds";
-      case ERR_VARINT: return "Protobuf decode error: invalid or truncated varint";
-      case ERR_FIELD_NUMBER: return "Protobuf decode error: invalid field number";
-      case ERR_WIRE_TYPE: return "Protobuf decode error: unexpected wire type";
-      case ERR_OVERFLOW: return "Protobuf decode error: length-delimited field overflows message";
-      case ERR_FIELD_SIZE: return "Protobuf decode error: invalid field size";
-      case ERR_SKIP: return "Protobuf decode error: unable to skip unknown field";
-      case ERR_FIXED_LEN:
+  auto error_message = [](protobuf_error error) -> std::string {
```
igorpeshansky: [Really optional] You could also put this as an inline function in `protobuf_types.cuh`, next to the `protobuf_error` declaration…

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1278,10 +1277,14 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-  if (h_error == ERR_SCHEMA_TOO_LARGE || h_error == ERR_REPEATED_COUNT_MISMATCH) {
-    throw cudf::logic_error(error_message(h_error));
+  auto const error = static_cast<protobuf_error>(h_error);
+  if (error == protobuf_error::SCHEMA_TOO_LARGE ||
+      error == protobuf_error::REPEATED_COUNT_MISMATCH) {
+    throw cudf::logic_error(error_message(error));
+  }
+  if (fail_on_errors && error != protobuf_error::NONE) {
+    throw cudf::logic_error(error_message(error));
   }
```
igorpeshansky, line 1263: [Optional] Same here, you could shorten it with `using enum protobuf_error;`:

```c++
  auto const error = static_cast<protobuf_error>(h_error);
  using enum protobuf_error;
  if (error == SCHEMA_TOO_LARGE || error == REPEATED_COUNT_MISMATCH) {
    throw cudf::logic_error(error_message(error));
  }
  if (fail_on_errors && error != NONE) throw cudf::logic_error(error_message(error));
```

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1278,10 +1277,14 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   int h_error = 0;
```
igorpeshansky: [Really optional] This is a much more pervasive change, and thus probably best left to a follow-up PR, but you should be able to declare both `d_error` and `h_error` directly in terms of the enum class, i.e.:

```c++
  rmm::device_uvector<protobuf_error> d_error = cudf::detail::make_zeroed_device_uvector_async<protobuf_error>(
      1, stream, cudf::get_current_device_resource_ref());
  …
  protobuf_error h_error = protobuf_error::NONE;
  CUDF_CUDA_TRY(cudf::detail::memcpy_async(&h_error, d_error.data(), sizeof(protobuf_error), stream));
  …
  using enum protobuf_error;
  if (h_error == SCHEMA_TOO_LARGE || h_error == REPEATED_COUNT_MISMATCH) { … }
```

This would require changing `error_flag` to `protobuf_error*` everywhere, and using `atomic_ref<protobuf_error>` for underlying storage…

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1001,8 +980,7 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     }
 
     // Phase B: allocate occurrence buffers and launch the combined scan kernel.
-    std::vector<repeated_field_scan_desc> h_scan_descs;
-    h_scan_descs.reserve(num_repeated);
+    auto h_scan_descs = cudf::detail::make_pinned_vector_async<repeated_field_scan_desc>(0, stream);
```
igorpeshansky, line 958: Just curious, do we no longer need a `reserve` here?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -603,13 +604,15 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
     rmm::device_uvector<field_descriptor> d_field_descs(
       num_scalar, stream, cudf::get_current_device_resource_ref());
-    CUDF_CUDA_TRY(cudf::detail::memcpy_async(
-      d_field_descs.data(), h_field_descs.data(), num_scalar * sizeof(field_descriptor), stream));
+    CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_field_descs.data(),
+                                             h_field_descs.data(),
+                                             h_field_descs.size() * sizeof(field_descriptor),
```
igorpeshansky: [Optional] We could still use `num_scalar` here, right?

src/main/cpp/src/protobuf/protobuf_builders.cu
igorpeshansky: Do you want to also add one for the permissive (non-strict) case, and check it now does not throw and returns null? Also one for a field that's unknown but well-formed, which should just pass through with no error?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -489,26 +491,24 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   // Error flag
   rmm::device_uvector<int> d_error(1, stream, cudf::get_current_device_resource_ref());
   CUDF_CUDA_TRY(cudaMemsetAsync(d_error.data(), 0, sizeof(int), stream.value()));
```
igorpeshansky: [Really optional] …or use `make_zeroed_device_uvector_async`…

src/main/cpp/src/protobuf/protobuf_types.cuh
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Done. It's in `cudf::detail` but still looks better than a `cudaMemsetAsync`.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: TIL, done.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Yes, updated.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Updated here. I will clean up in the dev branch.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Ah, good catch, we still need the `reserve`.

src/main/cpp/src/protobuf/protobuf_types.cuh
```diff
@@ -54,6 +54,30 @@ constexpr int FIELD_LOOKUP_TABLE_MAX = 4096;
 // error surface depends on the schema, not on which fields happen to have data in a given batch.
 constexpr int MAX_REPEATED_FIELDS_PER_KERNEL = 32;
 
+inline std::string error_message(protobuf_error error)
+{
+  switch (error) {
+    using enum protobuf_error;
+    case NONE: return "Protobuf decode error: none";
```
igorpeshansky, line 61: [Optional] Technically, `"Protobuf decode error:"` is not part of the error name, so you could prepend it within `decode_protobuf_to_struct` when you create a `cudf::logic_error`…

src/main/cpp/src/protobuf/protobuf_types.cuh
```diff
@@ -27,22 +29,20 @@ constexpr int MAX_VARINT_BYTES = 10;
 constexpr int THREADS_PER_BLOCK = 256;
 
 enum class protobuf_error : int {
```
igorpeshansky, line 31: [Optional] I would keep this and `error_message` together, so it may make sense to move this after all of the constants, just before `error_message` is defined.

src/main/cpp/src/protobuf/protobuf.cu
igorpeshansky: Ok, leaving this thread for you to resolve once that's done.

igorpeshansky (approved):
A couple of optional suggestions, otherwise LGTM :shipit:

Approving to unblock.

thirtiseven:
build

thirtiseven:
build

thirtiseven:
@igorpeshansky  Thanks for your review! I will address the new comments in dev branch together with the enum cleanup.

PR #4723, opened 2026-06-16
Reduce ORC DST rule probing complexity

res-life:
## Summary

Refactor `extractDstRuleByProbing` to reduce cognitive complexity while preserving the original probing behavior.

The transition scan now delegates transition collection to `findDstTransitionsByProbing`, keeps DST-on/off transition state in a small helper object, and builds the final `DstRule` after both transitions are found.

## Details

- Preserve the original one-hour probing loop and binary search for exact transition instants.
- Preserve duplicate DST-on / DST-off detection and fallback behavior by returning `null`.
- Preserve early break once both transitions are found and the running offset returns to the initial offset.
- Reduce the maximum method cognitive complexity below the threshold of 15.

## Validation

Ran:

```bash
./build/run-in-docker mvn -Dsubmodule.check.skip=true -Dtest=OrcTimezoneInfoTest test
```

Result:

```text
OrcTimezoneInfoTest: Tests run: 26, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java
@@ -725,6 +725,61 @@ public class OrcTimezoneInfoTest {
   }
 
   @Test
+  void testExtractDstRuleByProbingReturnsNullOnMultipleDstOffTransitions() {
+    // extractDstRuleByProbing returns null when probing finds more than one
+    // offset-decreasing transition in a year. This is the symmetric counterpart
+    // of testExtractDstRuleByProbingReturnsNullOnMultipleDstOnTransitions: it
+    // exercises the second-DST-off early-return in DstTransitions
+    // .recordDstOffTransition, which no other test reaches.
+    //
+    // This synthetic TimeZone springs forward once (March 2nd Sunday) and falls
+    // back twice (July 1st Sunday and November 1st Sunday). The first fall-back
+    // records dstOffTransition; the second trips the "more than one DST-off"
+    // guard, so recordTransition returns false and probing returns null. The
+    // ZoneRules supplies no recurring transition rules, so Path A also returns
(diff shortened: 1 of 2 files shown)


res-life:
build

src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java
```diff
@@ -724,6 +724,61 @@ void testExtractDstRuleByProbingReturnsNullOnMultipleDstOnTransitions() {
...
+    // "Failed to extract" throw as the expected outcome.
+    ZoneOffsetTransition historicalTransition = ZoneOffsetTransition.of(
+        LocalDateTime.of(1900, 1, 1, 0, 0), ZoneOffset.ofHours(-1), base);
+    ZoneRules rules = ZoneRules.of(base, base,
+        Collections.emptyList(),
+        Collections.singletonList(historicalTransition),
+        Collections.emptyList());
+    IllegalStateException ex = assertThrows(IllegalStateException.class,
+        () -> OrcDstRuleExtractor.extractDstRule("Synthetic/MultipleDstOff", tz, rules));
+    assertTrue(ex.getMessage().contains("Failed to extract"),
```
firestarman, line 778: Nit: this does exercise the current multiple-DST-off path, but the final `"Failed to extract"` assertion doesn't uniquely prove that probing returned `null` because `recordDstOffTransition` rejected the second DST-off transition. If that guard were weakened, probing could still build a rule and then fail later during cross-year verification for this synthetic three-offset shape.

Could we assert the probing result more directly, e.g. by reflectively invoking `extractDstRuleByProbing(TimeZone, int)` and checking it returns `null`? That would make this test specifically protect the new duplicate-DST-off guard.

firestarman (approved):
LGTM, thanks for the cleanup. I left one nit on making the new test assert the intended guard more directly.

PR #4731, opened 2026-06-18
Add stricter compiler warning flags

### LINKED ISSUE #4497: [FEA] Add more stricter compiler warnings `-Wshadow`, `-Wnon-virtual-dtor`, `-Woverloaded-virtual`
There are ongoing work to adopt stricter compiler warnings `-Wshadow`, `-Wnon-virtual-dtor`, `-Woverloaded-virtual` in RAPIDS. For example, https://github.com/rapidsai/rmm/pull/2327 is adding them to `rmm` and that set of flags could be adopted more in the future.

We are using the same set of compiler flags as cudf and would have these flags in our build config when cudf adopts such flags. However, instead of waiting for cudf to adopt which could take a long time, we can jump in faster to add these flags into our CMake build config without waiting for cudf. The flags could bring some benefits to the codebase:
 * `-Wshadow`: flags when a name in an inner scope (parameter, local, lambda capture) hides a name from an outer scope, since the inner one  silently takes over and any reader who thought the outer name was being used has just misread the code.
 * `-Wnon-virtual-dtor`: flags a class that has virtual functions but a non-virtual public destructor, because delete `base_ptr` on a derived object then runs only the base destructor and skips the derived one — undefined behavior, typically manifesting as a memory/resource leak.
 * `-Woverloaded-virtual`: flags a derived-class function that has the same name as a base-class virtual but a different signature, because C++ name lookup hides all base overloads of that name in the derived scope, so calls you expected to dispatch to the base silently fail to compile or pick the wrong overload.

As such, having them in our codebase will help us to identify and fix potential bugs before they could even show up.

In case we don't want to add them explicitly but wait for cudf to adopt and we will inherit, we also need to test these flags ASAP to see if they break our build (due to potential bugs) and fix our code ASAP. That would make our codebase cleaner and prepare for adopting the flags from cudf in a smooth way.

WilliamK112:
## Summary

- add project-level strict warning flags for `-Wshadow`, `-Wnon-virtual-dtor`, and `-Woverloaded-virtual`
- pass the same warnings to CUDA host compilation via `-Xcompiler`
- apply the warning helper across the JNI library, test, benchmark, profiler, fault-injection, and auxiliary JNI targets

## Related Issue

Closes #4497

## Testing

- [x] `git diff --check`
- [ ] Full native/CUDA build not run locally: this Windows shell does not have `cmake`/CUDA configured, and the cuDF submodule is not initialized here.

### DIFF
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -94,6 +94,23 @@ rapids_find_package(CUDAToolkit REQUIRED)
 # Set a default build type if none was specified
 rapids_cmake_build_type("Release")
 
+set(SPARK_RAPIDS_JNI_STRICT_CXX_WARNING_FLAGS
+    -Wshadow
+    -Wnon-virtual-dtor
+    -Woverloaded-virtual
+)
+set(SPARK_RAPIDS_JNI_STRICT_CUDA_WARNING_FLAGS
+    -Xcompiler=-Wshadow,-Wnon-virtual-dtor,-Woverloaded-virtual
+)
+
+function(spark_rapids_jni_add_strict_warning_flags target_name)
+  target_compile_options(
+    ${target_name}
+    PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${SPARK_RAPIDS_JNI_STRICT_CXX_WARNING_FLAGS}>"
+            "$<$<COMPILE_LANGUAGE:CUDA>:${SPARK_RAPIDS_JNI_STRICT_CUDA_WARNING_FLAGS}>"
+  )
+endfunction()
+
 # ##################################################################################################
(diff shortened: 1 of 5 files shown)


sameerz:
Hi @WilliamK112 , thank you for the contribution.  Have you tried compiling with the stricter warning flags?  Did any new warnings arise?

WilliamK112:
Moving this back to draft for now. The current feedback needs a compile-capable CUDA/CMake validation pass: Greptile has two unresolved CMake comments, and @sameerz asked whether the stricter warning flags were compiled and whether they introduce new warnings. I cannot honestly provide that evidence from my current Windows environment, so this should not stay ready for review until I can validate it in a suitable native/CUDA build setup and address the comments.

nvauto:

nvauto:

PR #4734, opened 2026-06-23
Quote non-finite float values for JSON casts

thirtiseven:
## Summary
- Add a JSON mode to the native float-to-string conversion path.
- Quote `NaN`, `Infinity`, and `-Infinity` when Spark JSON serialization casts float/double values.
- Keep the existing float-to-string cast behavior on the default path.

## Problem
Spark JSON serialization does not use the same textual representation as a regular float-to-string cast for non-finite floating point values.

For `to_json`, Spark emits non-finite float and double values as JSON strings, for example `"NaN"`, `"Infinity"`, and `"-Infinity"`. The existing JNI float-to-string path emits the regular cast strings without JSON quoting. When cudf-spark routes `to_json` through that path, GPU output can produce bare `NaN`/`Infinity` tokens instead of matching Spark's JSON output.

## Fix
This PR adds an explicit JSON mode to the native float-to-string conversion API. The default path remains unchanged for normal casts, while the JSON path quotes non-finite values so cudf-spark can use it from `to_json` without adding a per-row Spark-side workaround.

Contributes to NVIDIA/cudf-spark#15118.

### DIFF
--- a/src/main/cpp/src/cast_string.hpp
+++ b/src/main/cpp/src/cast_string.hpp
@@ -79,7 +79,7 @@ std::unique_ptr<cudf::column> string_to_integer(
   bool ansi_mode,
   bool strip,
   rmm::cuda_stream_view stream,
-  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
+  rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
 
 /**
  * @brief Convert a string column into an decimal column.
@@ -101,7 +101,7 @@ std::unique_ptr<cudf::column> string_to_decimal(
   bool ansi_mode,
   bool strip,
   rmm::cuda_stream_view stream,
-  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
+  rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
 
 /**
  * @brief Convert a string column into an float column.
@@ -119,18 +119,18 @@ std::unique_ptr<cudf::column> string_to_float(
   cudf::strings_column_view const& string_col,
   bool ansi_mode,
   rmm::cuda_stream_view stream,
-  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
+  rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
 
 std::unique_ptr<cudf::column> format_float(
   cudf::column_view const& input,
   int const digits,
   rmm::cuda_stream_view stream,
-  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
+  rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
 
 std::unique_ptr<cudf::column> float_to_string(
   cudf::column_view const& input,
   rmm::cuda_stream_view stream,
-  rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
+  rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
 
 [[nodiscard]] std::unique_ptr<cudf::column> float_to_string(
   cudf::column_view const& input,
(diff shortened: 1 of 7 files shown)


src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1200,25 +1211,53 @@ __device__ inline int compute_d2s_size_json(double value)
...
 //===== APIs =====
 
-__device__ inline int compute_ftos_size(double value, bool is_float)
+template <typename FloatType, bool json_string = false>
+__device__ inline int compute_ftos_size(FloatType value)
 {
-  if (is_float) {
-    return compute_f2s_size(value);
+  static_assert(cuda::std::is_same_v<FloatType, float> || cuda::std::is_same_v<FloatType, double>);
+  if constexpr (cuda::std::is_same_v<FloatType, float>) {
```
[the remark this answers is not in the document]
thirtiseven: There is no FLOAT16 in cudf so I think we don't need to do this.

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1200,25 +1211,53 @@ __device__ inline int compute_d2s_size_json(double value)
...
 }
 
-__device__ inline int float_to_string(double value, bool is_float, char* output)
+template <typename FloatType, bool json_string = false>
+__device__ inline int float_to_string(FloatType value, char* output)
 {
-  if (is_float) {
-    return f2s_buffered_n(value, output);
+  static_assert(cuda::std::is_same_v<FloatType, float> || cuda::std::is_same_v<FloatType, double>);
+  if constexpr (cuda::std::is_same_v<FloatType, float>) {
```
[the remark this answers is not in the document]
thirtiseven: ditto

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1156,7 +1156,10 @@ __device__ inline int copy_special_str_json(char* const result,
                                             bool const exponent,
                                             bool const mantissa)
 {
-  // no NaN in json
+  if (mantissa) {
```
igorpeshansky: [Optional] The only difference between JSON and non-JSON are the quotes around the `NaN`/`±Infinity` strings. So could we make that apparent, and remove some code duplication in the process (especially given that the JSON vs non-JSON code is already using slightly different implementations of the same logic)? E.g.:

```c++
__device__ inline int copy_special_str_json(char* const result,
                                            bool const sign,
                                            bool const exponent,
                                            bool const mantissa)
{
  int offset = 0;
  int quotes = 0;
  if (mantissa || exponent) {
    result[0] = '"';
    offset = 1;
    quotes = 2;
  }
  int length = copy_special_str(result + offset, sign, exponent, mantissa);
  if (mantissa || exponent) {
    result[offset + length] = '"';
  }
  return length + quotes;
}

__device__ inline int special_str_size_json(bool const sign,
                                            bool const exponent,
                                            bool const mantissa)
{
  int const quotes = (mantissa || exponent) ? 2 : 0;
  return special_str_size(sign, exponent, mantissa) + quotes;
}
```

src/main/cpp/src/cast_string.hpp
```diff
@@ -132,6 +132,12 @@ std::unique_ptr<cudf::column> float_to_string(
   rmm::cuda_stream_view stream,
   rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource_ref());
 
+[[nodiscard]] std::unique_ptr<cudf::column> float_to_string(
+  cudf::column_view const& input,
+  bool json_string,
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
```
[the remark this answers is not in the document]
thirtiseven: we should use `cudf::get_current_device_resource_ref()` for all. Updated.

src/main/cpp/src/ftos_converter.cuh
thirtiseven: Good idea, done.

src/main/cpp/src/ftos_converter.cuh
```diff
@@ -1156,12 +1156,15 @@ __device__ inline int copy_special_str_json(char* const result,
...
-  int const length = copy_special_str(result + offset, sign, exponent, mantissa);
-  if (quote) { result[offset + length] = '"'; }
+  int offset = 0;
+  int quotes = 0;
+  if (mantissa || exponent) {
+    result[0] = '"';
+    offset    = 1;
+    quotes    = 2;
+  }
+  int length = copy_special_str(result + offset, sign, exponent, mantissa);
```
igorpeshansky: FWIW, your previous attempt was shorter and a bit cleaner (and possibly faster) — any reason for reverting to my verbatim proposal here rather than keeping the `const` version?

If you choose to stay with this version, I would at least keep the `const` in this line.

```suggestion
  int const length = copy_special_str(result + offset, sign, exponent, mantissa);
```

igorpeshansky (dismissed):
One minor readability question/nit, but will approve to unblock. Up to you how to address it.

src/main/cpp/src/ftos_converter.cuh
thirtiseven: Thanks, reverted back to the const version now.

thirtiseven:
build

igorpeshansky (approved):
LGTM :shipit:

PR #4748, opened 2026-06-25
Support nested repeated protobuf non-message fields

thirtiseven:
## Description

This adds the JNI-3b.5/JNI-3b.6 slice for `from_protobuf`: repeated non-message fields inside nested protobuf messages.

The change:
- adds count/scan kernels for repeated fields inside nested parent messages
- builds LIST children for nested repeated scalar, string, bytes, and enum-as-string fields
- keeps nested repeated `MessageType` unsupported for a later repeated-message PR
- adds Java coverage for nested packed repeated ints, enum-as-string, strings, and bytes

This is part of NVIDIA/cudf-spark#14069.

## Testing

- `git diff --check`
- JNI build/local protobuf tests passed locally


src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -296,6 +296,13 @@ __device__ bool walk_repeated_element(uint8_t const* cur,
   return true;
 }
 
+__device__ bool repeated_wire_type_matches(int wire_type, int expected_wire_type)
+{
+  return wire_type == expected_wire_type ||
+         (wire_type == wire_type_value(proto_wire_type::LEN) &&
+          expected_wire_type != wire_type_value(proto_wire_type::LEN));
+}
```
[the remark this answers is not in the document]
thirtiseven: Thanks for checking. This is intentional: protobuf-java treats a nested occurrence with the same field number but an incompatible wire type as an unknown field and skips it. Spark's explicit unknown-field collision check is applied at the top-level DynamicMessage, not recursively for nested messages. The JNI behavior and tests were updated to match that reference behavior: strict mode still validates malformed compatible repeated encodings, while wrong-wire nested occurrences are ignored.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2782,6 +2793,251 @@ void testDeepNestedMessageDepth3() {
     }
   }
 
+  @Test
+  void testPackedRepeatedInsideNestedMessage() {
+    // message Inner { repeated int32 ids = 1 [packed=true]; }
+    // message Outer { Inner inner = 1; }
+    byte[] packedIds = concatBytes(encodeVarint(10), encodeVarint(20), encodeVarint(30));
+    Byte[] inner = concat(
+        box(tag(1, WT_LEN)), box(encodeVarint(packedIds.length)), box(packedIds));
```
igorpeshansky: This is just `encodeBytes`, right? Also for all the other packed-payload sites in the new tests…

```suggestion
        box(tag(1, WT_LEN)), encodeBytes(packedIds));
```

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2782,6 +2793,251 @@ void testDeepNestedMessageDepth3() {
...
+    }
+  }
+
+  @Test
+  void testNestedRepeatedScalarEmptyAndAbsentParent() {
+    // message Inner { repeated int32 ids = 1 [packed=true]; }
+    // message Outer { Inner inner = 1; }
+    Byte[][] rows = new Byte[][] {
+        concat(box(tag(1, WT_LEN)), encodeMessage(new Byte[]{})),
+        new Byte[] {}
```
igorpeshansky: [Optional] Should we be consistent about the spaces in these? We have a bunch with no space between `]` and `{`, and a bunch with. Let's pick one?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -2782,6 +2793,251 @@ void testDeepNestedMessageDepth3() {
...
+         HostColumnVector hostInner = inner.copyToHost();
+         HostColumnVector hostPriorityValues = priorityValues.copyToHost()) {
+      assertEquals(0, actual.getNullCount(),
+          "Invalid nested repeated enum should not null outer rows");
+      assertFalse(hostOuter.isNull(0));
+      assertFalse(hostOuter.isNull(1));
+      assertEquals(0, inner.getNullCount(), "Nested struct rows should stay present");
+      assertFalse(hostInner.isNull(0));
+      assertFalse(hostInner.isNull(1));
+      assertListOffsets(priorities, 0, 2, 4);
```
igorpeshansky: [Really optional] I wonder if we should be using a higher-level comparison, something like `assertStructColumnsAreEqual `, or even `assertColumnsAreEqual`…

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -54,6 +55,30 @@ struct schema_context_view {
...
 };
 
+// ============================================================================
+// Nested decode view bundles
+// ============================================================================
+
+struct protobuf_input_view {
+  uint8_t const* message_data;
+  cudf::size_type message_data_size;
+  cudf::size_type const* row_offsets;
```
igorpeshansky, line 74: Simplifying multi-argument functions by grouping arguments into structs is a good call, but `row_offsets` requires `num_rows`, so I would move the latter to `protobuf_input_view` to guarantee they are provided together. `nested_parent_view` would be impossible to use without `protobuf_input_view`, but it's never used without `protobuf_input_view` anyway.

Also, `protobuf_input_view` and `protobuf_decode_state` can now be reused in many other functions — are you leaving that cleanup for a follow-up?

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -54,6 +55,30 @@ struct schema_context_view {
...
+  field_location const* locations;
+  std::size_t location_count;
+  int num_rows;
+  int32_t const* top_row_indices;
+};
+
+struct protobuf_decode_state {
+  rmm::device_uvector<bool>* row_force_null;
+  rmm::device_uvector<protobuf_error>* error;
+  bool propagate_invalid_enum_rows;
```
igorpeshansky: Since the associated parameter defaulted to `true`, you could default this one to `true` as well — then you'd only need to initialize it when overriding to `false`…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -458,45 +558,37 @@ std::unique_ptr<cudf::column> build_repeated_string_column(
...
-  bool propagate_invalid_enum_rows)
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr)
 {
   CUDF_EXPECTS(depth < MAX_NESTING_DEPTH,
                "Nested protobuf struct depth exceeds supported decode recursion limit");
-  CUDF_EXPECTS(d_parent_locs.size() == static_cast<size_t>(num_rows),
-               "build_nested_struct_column: parent locations size must match row count");
-  // d_row_force_null comes from the caller; empty means "don't track", otherwise it must be
+  validate_nested_parent_view(parent, "build_nested_struct_column");
```
igorpeshansky: [Optional] Can this be `__func__`? Also throughout…

[Really optional] The standard trick is to use an argument that always defaults to `std::source_location::current()`[^1] and extract the function from there. Is that available to us?

[^1]: https://en.cppreference.com/cpp/utility/source_location/current

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -690,17 +779,211 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+                                  state.error->data(),
+                                  stream);
+
+  rmm::device_uvector<int32_t> d_rep_counts(parent.num_rows, stream, scratch_mr);
+  thrust::transform(rmm::exec_policy_nosync(stream, scratch_mr),
+                    d_rep_info.begin(),
+                    d_rep_info.end(),
+                    d_rep_counts.begin(),
+                    [] __device__(repeated_field_info const& info) { return info.count; });
+  auto const total_count_64 = thrust::reduce(rmm::exec_policy_nosync(stream, scratch_mr),
```
igorpeshansky: This code already exists in `decode_protobuf_to_struct` repeated fields loop (around line 943 in `protobuf.cu`). Can that be refactored to use this builder? If not, can the common bits be extracted into helpers?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -458,45 +558,37 @@ std::unique_ptr<cudf::column> build_repeated_string_column(
...
-               "build_nested_struct_column: parent locations size must match row count");
-  // d_row_force_null comes from the caller; empty means "don't track", otherwise it must be
+  validate_nested_parent_view(parent, "build_nested_struct_column");
+  validate_protobuf_decode_state(state, "build_nested_struct_column");
+  // row_force_null comes from the caller; empty means "don't track", otherwise it must be
   // row-aligned since the scan/extract kernels index it by row.
-  CUDF_EXPECTS(
-    d_row_force_null.is_empty() || d_row_force_null.size() == static_cast<size_t>(num_rows),
-    "build_nested_struct_column: row-force-null buffer must be empty or row-sized");
+  CUDF_EXPECTS(state.row_force_null->is_empty() ||
```
igorpeshansky: If you pass `parent.num_rows` into `validate_protobuf_decode_state`, you could fold this check into it as well…

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -54,6 +55,30 @@ struct schema_context_view {
...
+};
+
+struct nested_parent_view {
+  field_location const* locations;
+  std::size_t location_count;
+  int num_rows;
+  int32_t const* top_row_indices;
+};
+
+struct protobuf_decode_state {
```
igorpeshansky: Minor nit: this isn't just state — it has one input parameter, one in/out scratch, and one pointer to error storage. `_context`, maybe?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -593,6 +602,174 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
...
+                                    error_flag,
+                                    lookup_by_fn,
+                                    is_repeated_field,
+                                    get_expected_wire_type,
+                                    on_repeated_count)) {
+    return;
+  }
+}
+
+CUDF_KERNEL void scan_repeated_in_nested_kernel(
```
igorpeshansky: This seems to overlap a lot with `scan_all_repeated_occurrences_kernel` — can they be consolidated? Especially if we change `walk_repeated_element` to add permissive mode…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -690,17 +779,211 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+    make_null_mask_from_parent_locations(parent.locations, parent.num_rows, stream, mr);
+  return cudf::make_structs_column(parent.num_rows,
+                                   std::move(struct_children),
+                                   struct_null_count,
+                                   std::move(struct_mask),
+                                   stream,
+                                   mr);
+}
+
+std::unique_ptr<cudf::column> build_repeated_child_list_column(
```
igorpeshansky, line 861: This seems to have a lot of overlap with `build_nested_struct_column`. Can we consolidate?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -574,11 +581,13 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
   auto get_expected_wire_type = [&](int f) { return field_descs[f].expected_wire_type; };
   auto validate_repeated =
     [&](int f, uint8_t const* cur, uint8_t const* msg_end, uint8_t const* msg_base, int wt) {
-      // Values come from the dedicated nested repeated path; here we only validate the occurrence
-      // so strict/permissive errors surface.
+      auto const expected_wire_type = get_expected_wire_type(f);
+      // Values come from the dedicated nested repeated path; here we only validate compatible
+      // occurrences so malformed payloads surface while wrong-wire occurrences remain unknown.
+      if (!repeated_wire_type_matches(wt, expected_wire_type)) { return true; }
```
igorpeshansky: [Optional] This is the same check as at the start of `walk_repeated_element`, so this is just to prevent that from reporting `protobuf_error::WIRE_TYPE`, right? Is it possible to change `walk_repeated_element` to have a mode where it simply returns `true` instead of reporting an error? Especially since this appears more than once…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -593,6 +602,174 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
...
+}
+
+/**
+ * Count repeated field occurrences within nested messages.
+ *
+ * Unlike the top-level count kernel, this kernel does not check schema depth. Callers pass a
+ * schema slice whose repeated fields are already known to be direct children of the parent
+ * message location.
+ */
+CUDF_KERNEL void count_repeated_in_nested_kernel(
```
igorpeshansky: How much overlap is there with `count_repeated_fields_kernel`?

This is the same overlap I flagged earlier in [#4644](https://github.com/NVIDIA/cudf-spark-jni/pull/4644#discussion_r3358116516); we can keep deferring, but the eventual refactor surface grows as a result…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -593,6 +602,174 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
...
+    return walk_repeated_element(current,
+                                 message_end,
+                                 message_base,
+                                 wire_type,
+                                 repeated_fd.expected_wire_type,
+                                 error_flag,
+                                 count_action);
+  };
+
+  if (!scan_message_field_locations(msg_start,
```
igorpeshansky: [Optional] This effectively ignores the return value, so should we just assign it to a dummy variable?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -1217,24 +1221,18 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-      d_row_force_null,
-      d_error,
-      num_rows,
-      stream,
-      mr,
-      nullptr,
-      0,
-      false);
+    auto const parent =
+      nested_parent_view{d_parent_locs.data(), d_parent_locs.size(), num_rows, nullptr};
```
igorpeshansky: This local doesn't help you — you can inline it in the call (and lose the type)…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -690,17 +779,211 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+                                  input.row_offsets,
+                                  input.base_offset,
+                                  parent.locations,
+                                  parent.num_rows,
+                                  d_repeated_field.data(),
+                                  d_rep_info.data(),
+                                  state.error->data(),
+                                  stream);
+
+  rmm::device_uvector<int32_t> d_rep_counts(parent.num_rows, stream, scratch_mr);
```
igorpeshansky: Can we use `make_transform_iterator` instead (https://github.com/NVIDIA/cudf-spark-jni/pull/4504#discussion_r3171865047)?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Done.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Done. Cleaned up all to use no space.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Good idea, updated to `assertStructColumnsAreEqual` in this pr. I will updated all cases in the dev branch.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Done. `num_rows` is now part of `protobuf_input_view`, and the input view is reused by the repeated builders as well.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Renamed to `protobuf_decode_runtime_context`.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Updated to `std::source_location::current()`.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Done. Singleton and repeated children now share `build_protobuf_field_values_column` for value construction.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: done.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Done. The common reduce, overflow check and list-offset construction are now shared in `make_list_offsets_from_counts`.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Done. `walk_repeated_element` now has a mismatch policy, and nested decoding uses the `skip` mode.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Done. Updated to `count_repeated_fields_kernel`.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Done. Top-level and nested occurrence scans now share `scan_all_repeated_occurrences_in_message`.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
```diff
@@ -339,18 +394,18 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
   std::vector<nested_field_descriptor> const& schema,
   int num_fields,
   schema_context_view ctx,
-  protobuf_decode_state state,
+  protobuf_decode_runtime_context context,
```
igorpeshansky: Careful — `context` and `ctx` mean pretty much the same thing (the latter is the conventional abbreviation of the former). Want to distinguish them (e.g., `schema_ctx` and `decode_ctx`)?

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -612,51 +682,116 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+  auto h_scan_descs = cudf::detail::make_pinned_vector_async<repeated_field_scan_desc>(0, stream);
+  h_scan_descs.reserve(repeated_child_positions.size());
+  for (auto const ci : repeated_child_positions) {
+    auto const child_schema_idx = child_field_indices[ci];
+    auto counts_begin           = thrust::make_transform_iterator(
+      thrust::make_counting_iterator<int>(0),
+      extract_strided_count{d_repeated_info.data(), ci, num_child_fields});
+    repeated_work[ci].emplace(
+      child_schema_idx,
+      make_list_offsets_from_counts(counts_begin,
```
igorpeshansky: There's a subtle invariant at play here: `make_list_offsets_from_counts` expects exactly `input.num_rows` entries, `d_repeated_info` is sized at `input.num_rows * num_child_fields`, and we're passing in a strided iterator which will produce one entry every `num_child_fields`, yielding exactly `input.num_rows` results. Additionally, there is a branch where `d_repeated_info` has size 0 (and thus, potentially, where its `.data()` returns nullptr), but since that branch is predicated on `repeated_child_positions.empty()`, and this call is within a loop over `repeated_child_positions`, that branch cannot affect this call. Both might be worth mentioning in a comment (or possibly even adding an invariant check)…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -612,51 +682,116 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+                                    input.num_rows,
+                                    "Repeated nested-field total element count exceeds 2^31-1",
```
igorpeshansky: Nit: the "total element count exceeds 2^31-1" part would apply to any caller. Can we just pass in `"Repeated nested-field"` and construct the error within `make_list_offsets_from_counts`?

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
igorpeshansky: What about the other structs? I saw at least three more functions that could reuse `protobuf_decode_runtime_context`: `extract_typed_column`, `build_enum_string_column`, `build_repeated_enum_string_column`.

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -20,11 +20,188 @@
...
+namespace {
+
+inline std::pair<rmm::device_buffer, cudf::size_type> make_null_mask_from_parent_locations(
+  field_location const* parent_locs,
+  int num_rows,
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr)
+{
+  CUDF_EXPECTS(num_rows >= 0,
+               "make_null_mask_from_parent_locations: row count must be non-negative");
```
igorpeshansky: [Optional] Can this be `__func__`[^1]? Also throughout…

[^1]: https://en.cppreference.com/cpp/language/function#func

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -123,6 +99,107 @@ inline std::unique_ptr<cudf::column> make_list_column_with_parent_nulls(
...
+          stream,
+          mr,
+          context.propagate_invalid_enum_rows ? get_top_row_indices() : nullptr,
+          context.propagate_invalid_enum_rows);
+      }
+      return extract_and_build_string_or_bytes_column(value_type.id() == cudf::type_id::LIST,
+                                                      input.message_data,
+                                                      num_values,
+                                                      loc_provider,
+                                                      loc_provider,
```
igorpeshansky: [Optional] Probably for a follow-up, but these two seem to be identical in all calls to this function. Do we need two separate ones?

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -438,37 +436,21 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
-  int32_t end       = in.offset_at(row + 1) - base;
-  if (!check_message_bounds(start, end, child.size(), error_flag)) return;
-
-  uint8_t const* const msg_base = bytes + start;
-  uint8_t const* const msg_end  = bytes + end;
-
-  // Defense-in-depth: host-side validate_decode_context enforces this cap, so the check is
-  // unreachable on a correct config. Using set_error_once instead of `assert` because the
-  // failure mode (overrunning `write_idx` below) is silent UB — `assert` is a no-op under
-  // NDEBUG and would leave the OOB write live in release.
```
igorpeshansky, line 460: This seems like it would've been a useful comment to keep.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -509,25 +491,61 @@ CUDF_KERNEL void scan_all_repeated_occurrences_kernel(cudf::column_device_view c
...
+
+  auto const base   = in.offset_at(0);
+  auto const child  = in.get_sliced_child();
+  auto const* bytes = reinterpret_cast<uint8_t const*>(child.data<int8_t>());
+  int32_t start     = in.offset_at(row) - base;
+  int32_t end       = in.offset_at(row + 1) - base;
+  if (!check_message_bounds(start, end, child.size(), error_flag)) return;
+
+  uint8_t const* const msg_base = bytes + start;
+  uint8_t const* const msg_end  = bytes + end;
```
igorpeshansky: These locals can now be inlined into the call arguments…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -716,58 +660,17 @@ CUDF_KERNEL void scan_repeated_in_nested_kernel(
...
-                                    get_expected_wire_type,
-                                    on_repeated_scan)) {
-    return;
-  }
-
-  if (write_idx != write_end) {
-    set_error_once(error_flag, protobuf_error::REPEATED_COUNT_MISMATCH);
-  }
+  auto const* msg_start = message_data + msg_start_off;
+  auto const* msg_end   = message_data + msg_end_off;
```
igorpeshansky: These locals can be inlined into the call arguments…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -210,18 +212,21 @@ CUDF_KERNEL void scan_all_fields_kernel(
...
-                                      F&& f)
+__device__ bool walk_repeated_element(
+  uint8_t const* cur,
+  uint8_t const* msg_end,
+  uint8_t const* msg_base,
+  int wt,
+  int expected_wt,
+  protobuf_error* error_flag,
+  F&& f,
+  wire_type_mismatch_policy mismatch_policy = wire_type_mismatch_policy::report_error)
```
igorpeshansky: [Optional] I wonder why you chose to make this a function argument rather than a template parameter (NTTP)? For the latter you could use `if constexpr` below…

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -602,111 +628,29 @@ CUDF_KERNEL void scan_nested_message_fields_kernel(uint8_t const* message_data,
...
+  auto row = static_cast<cudf::size_type>(blockIdx.x * blockDim.x + threadIdx.x);
   if (row >= num_rows) return;
 
-  int write_idx          = occ_prefix_sums[row];
-  int const write_end    = occ_prefix_sums[row + 1];
   auto const& parent_loc = parent_locs[row];
   if (parent_loc.offset < 0) {
-    if (write_idx != write_end) {
-      set_error_once(error_flag, protobuf_error::REPEATED_COUNT_MISMATCH);
+    for (int f = 0; f < num_scan_fields; f++) {
```
igorpeshansky: The top-level path (`scan_all_repeated_occurrences_kernel`) simply skips null rows, so omits this check. Is it redundant? I would remove it here as well.

[Optional] If you do want to keep this check, you could set `msg_start`==`msg_end`==`message_data` for an empty parent, and simply launch `scan_all_repeated_occurrences_in_message`, which will skip over the (empty) message and do this exact loop in lines 497-502.

```c++
  auto const* msg_start = message_data;
  auto const* msg_end = message_data;
  if (parent_loc.offset >= 0) {
    int64_t const row_off       = static_cast<int64_t>(row_offsets[row]) - base_offset;
    int64_t const msg_start_off = row_off + parent_loc.offset;
    int64_t const msg_end_off   = msg_start_off + parent_loc.length;
    if (!check_message_bounds(msg_start_off, msg_end_off, message_data_size, error_flag)) { return; }
    msg_start = message_data + msg_start_off;
    msg_end   = message_data + msg_end_off;
  }
  …
  … = scan_all_repeated_occurrences_in_message<wire_type_mismatch_policy::skip>(msg_start, msg_end, scan_descs, …);
```

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -20,11 +20,188 @@
...
+{
+  auto [list_mask, list_null_count] =
+    make_null_mask_from_parent_locations(parent_locs, num_rows, stream, mr);
+  return cudf::make_lists_column(
+    num_rows, std::move(offsets_col), std::move(child_col), list_null_count, std::move(list_mask));
+}
+
+template <typename LocationProvider, typename ValidityFn, typename TopRowIndexProvider>
+std::unique_ptr<cudf::column> build_protobuf_field_values_column(
+  protobuf_input_view input,
```
igorpeshansky: Note that there's a bit of redundancy here, because the location provider encapsulates the same row offsets and base offset as `input`. The only field used within the helper is `input.message_data`, so we might as well pass that in directly.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Done. I updated them all. There may be other places with too many arguments that could be wrapped in structs like this, or other functions that could use the current structs in the protobuf codebase. I plan to do a full clean-up, but maybe in a follow-up.

src/main/cpp/src/protobuf/protobuf_host_helpers.hpp
thirtiseven: Good point, renamed to `schema_ctx` and `decode_ctx`.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Thanks, I forgot to check it when creating this pr.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Done. `build_protobuf_field_values_column` now takes only `message_data`, row offsets and the base offset remain encapsulated by the location provider.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Good catch, merged them.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Done. Added an invariant check and a comment for it.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Thanks, NTTP is better.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Restored

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Yes, removed.

thirtiseven:
pre-commit.ci autofix

thirtiseven:
build

igorpeshansky (approved):
LGTM :shipit:

PR #4765, opened 2026-06-29
Use larger copy buffer for vectored read fallback

binmahone:
## Description

This updates the default `RapidsInputFile.readVectored` fallback to copy through an explicit byte-buffer loop and write into the destination `HostMemoryBuffer` with `setBytes`.

The default fallback does not cache a scratch buffer in the interface implementation. It allocates a bounded buffer for each `readVectored` call, using the smaller of the default 8 MiB limit and the largest requested range. Callers with a well-defined lifecycle can use `RapidsInputFile.readVectoredUsingCopyBuffer(..., byte[])` to provide a caller-managed scratch buffer.

This avoids routing default fallback vectored reads through `HostMemoryBuffer.copyFromStream`, whose internal copy chunk is 128 KiB. Internal validation and stream-copy helpers live in a package-private utility class so they do not expand the interface API surface beyond the caller-supplied helper.

Related issue: NVIDIA/cudf-spark#15163.

The dependent NVIDIA/cudf-spark#15164 uses the helper in `HadoopInputFile.readVectored`. It keeps the `GpuParquetScan` `readVectored` abstraction introduced by NVIDIA/cudf-spark#14674 while allowing Hadoop-backed fallback reads to honor `parquet.read.allocation.size`.

## Testing

The current PR head (`901ca4f570cb843c814121a7698ba8b3d5896fbf`) was tested with:

```text
mvn -B -Dmaven.antrun.skip=true -Dtest=RapidsInputFileTest \
  test-compile surefire:test@default-test
```

Result:

```text
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The JNI artifact was then installed locally and the dependent cudf-spark head (`ba4487c750ba73cf2f2281288ca319ad7f3752be`) was packaged on an x86 host with:

```text
mvn -B -pl sql-plugin -am -DskipTests \
  -Dspark-rapids-jni.version=26.06.1-SNAPSHOT package
```

Result: `BUILD SUCCESS`.

## Performance Validation

End-to-end performance validation was performed with the dependent NVIDIA/cudf-spark#15164. Five interleaved NDS runs did not reproduce a stable regression after the fix. See NVIDIA/cudf-spark#15164 for the complete results.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
@@ -39,19 +39,42 @@ final class RapidsInputFileUtils {
 
     byte[] copyBuffer = new byte[getCopyBufferAllocationSize(copyRanges, copyBufferSize)];
     try (SeekableInputStream input = inputFile.open()) {
-      for (RapidsInputFile.CopyRange copyRange : copyRanges) {
-        if (input.getPos() != copyRange.getInputOffset()) {
-          input.seek(copyRange.getInputOffset());
-        }
-        long outputOffset = copyRange.getOutputOffset();
-        long bytesLeft = copyRange.getLength();
-        while (bytesLeft > 0) {
-          int readLength = (int) Math.min(bytesLeft, copyBuffer.length);
-          readFully(input, copyBuffer, readLength);
-          output.setBytes(outputOffset, copyBuffer, 0, readLength);
-          outputOffset += readLength;
-          bytesLeft -= readLength;
-        }
+      readVectored(input, output, copyRanges, copyBuffer);
+    }
+  }
+
+  static void readVectoredUsingCopyBuffer(
+      RapidsInputFile inputFile,
+      HostMemoryBuffer output,
+      List<RapidsInputFile.CopyRange> copyRanges,
+      byte[] copyBuffer) throws IOException {
+    validateReadVectoredArgs(inputFile, output, copyRanges, copyBuffer);
+    if (copyRanges.isEmpty()) {
+      return;
+    }
+
+    try (SeekableInputStream input = inputFile.open()) {
+      readVectored(input, output, copyRanges, copyBuffer);
+    }
+  }
+
+  private static void readVectored(
+      SeekableInputStream input,
+      HostMemoryBuffer output,
+      List<RapidsInputFile.CopyRange> copyRanges,
+      byte[] copyBuffer) throws IOException {
+    for (RapidsInputFile.CopyRange copyRange : copyRanges) {
+      if (input.getPos() != copyRange.getInputOffset()) {
+        input.seek(copyRange.getInputOffset());
+      }
+      long outputOffset = copyRange.getOutputOffset();
+      long bytesLeft = copyRange.getLength();
+      while (bytesLeft > 0) {
+        int readLength = (int) Math.min(bytesLeft, copyBuffer.length);
+        readFully(input, copyBuffer, readLength);
+        output.setBytes(outputOffset, copyBuffer, 0, readLength);
+        outputOffset += readLength;
+        bytesLeft -= readLength;
       }
     }
   }
@@ -62,11 +85,32 @@ final class RapidsInputFileUtils {
       List<RapidsInputFile.CopyRange> copyRanges,
       int copyBufferSize) {
     Objects.requireNonNull(inputFile, "inputFile can't be null");
-    Objects.requireNonNull(output, "output can't be null");
-    Objects.requireNonNull(copyRanges, "copyRanges can't be null");
     if (copyBufferSize <= 0) {
       throw new IllegalArgumentException("copyBufferSize must be positive");
     }
+    validateReadVectoredArgs(inputFile, output, copyRanges);
+  }
+
+  private static void validateReadVectoredArgs(
+      RapidsInputFile inputFile,
+      HostMemoryBuffer output,
+      List<RapidsInputFile.CopyRange> copyRanges,
+      byte[] copyBuffer) {
+    Objects.requireNonNull(inputFile, "inputFile can't be null");
+    Objects.requireNonNull(copyBuffer, "copyBuffer can't be null");
+    if (copyBuffer.length == 0) {
+      throw new IllegalArgumentException("copyBuffer must not be empty");
+    }
+    validateReadVectoredArgs(inputFile, output, copyRanges);
+  }
+
+  private static void validateReadVectoredArgs(
+      RapidsInputFile inputFile,
+      HostMemoryBuffer output,
+      List<RapidsInputFile.CopyRange> copyRanges) {
+    Objects.requireNonNull(inputFile, "inputFile can't be null");
+    Objects.requireNonNull(output, "output can't be null");
+    Objects.requireNonNull(copyRanges, "copyRanges can't be null");
 
     long outputLength = output.getLength();
     for (RapidsInputFile.CopyRange copyRange : copyRanges) {
--- a/src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright (c) 2026, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nvidia.spark.rapids.jni.fileio;
+
+import ai.rapids.cudf.HostMemoryBuffer;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.util.List;
+import java.util.Objects;
+
+final class RapidsInputFileUtils {
+  private RapidsInputFileUtils() {
+  }
+
+  static void readVectoredUsingCopyBuffer(
+      RapidsInputFile inputFile,
+      HostMemoryBuffer output,
+      List<RapidsInputFile.CopyRange> copyRanges,
+      int copyBufferSize) throws IOException {
+    validateReadVectoredArgs(inputFile, output, copyRanges, copyBufferSize);
+    if (copyRanges.isEmpty()) {
+      return;
+    }
+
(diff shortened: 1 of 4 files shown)


binmahone:
build

res-life (dismissed):
LGTM

src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
```diff
@@ -0,0 +1,108 @@
...
+      throws IOException {
+    int totalRead = 0;
+    while (totalRead < readLength) {
+      int amountRead = input.read(copyBuffer, totalRead, readLength - totalRead);
+      if (amountRead < 0) {
+        throw new EOFException(
+            "Unexpected end of stream, expected " + (readLength - totalRead) + " more bytes");
+      }
+      if (amountRead == 0) {
+        throw new IOException(
```
gerashegalov, line 149: **[PLAUSIBLE]** Throwing `IOException` on `read()` returning 0 is stricter than the `InputStream` contract requires. The spec says `read(byte[], int, int)` returns "the number of bytes actually read" — it does not guarantee ≥ 1 when `len > 0`, so a non-blocking or peek-buffered `SeekableInputStream` could legitimately return 0 and be killed here with a confusing error instead of retrying.

For this PR's target use case (S3A blocking streams) the behaviour is intentional and tested. But since `SeekableInputStream` is an abstract class in this package, a future implementation backed by a non-blocking NIO channel would silently break. Consider documenting this expectation in the `SeekableInputStream` contract so implementors know 0 is not a valid return when `len > 0`.

src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
```diff
@@ -0,0 +1,108 @@
...
+
+  private static int getCopyBufferAllocationSize(
+      List<RapidsInputFile.CopyRange> copyRanges,
+      int copyBufferSize) {
+    long maxRangeLength = 0;
+    for (RapidsInputFile.CopyRange copyRange : copyRanges) {
+      maxRangeLength = Math.max(maxRangeLength, copyRange.getLength());
+    }
+    return (int) Math.min(copyBufferSize, maxRangeLength);
+  }
```
gerashegalov, line 137: **[Cleanup]** The scan over all ranges is unnecessary when `copyBufferSize` is already the smaller value — the common case with the 8 MiB default and typical parquet column-chunk sizes. An early-exit makes the hot path O(1):

```java
for (RapidsInputFile.CopyRange copyRange : copyRanges) {
    maxRangeLength = Math.max(maxRangeLength, copyRange.getLength());
    if (maxRangeLength >= copyBufferSize) return copyBufferSize;
}
return (int) maxRangeLength;
```

src/test/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileTest.java
```diff
@@ -53,6 +65,17 @@ public void readVectoredThrowsOnShortRead() throws IOException {
     }
   }
 
+  @Test
+  public void readVectoredThrowsIOExceptionOnZeroProgressRead() throws IOException {
+    RapidsInputFile inputFile = new ZeroProgressInputFile(FILE_DATA.length);
+    try (HostMemoryBuffer output = HostMemoryBuffer.allocate(3)) {
+      IOException e = assertThrows(IOException.class,
+          () -> inputFile.readVectored(output,
+              Collections.singletonList(new RapidsInputFile.CopyRange(0, 3, 0))));
```
gerashegalov, line 75: **[Nit]** `assertThrows(IOException.class, ...)` already guarantees the thrown exception is exactly `IOException`. The `assertEquals(IOException.class, e.getClass())` is redundant and can be removed.

src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
binmahone: Good point. I documented the `SeekableInputStream` contract in b8720e0b: implementations of `read(byte[], int, int)` must not return 0 when `len > 0`; they should block until progress, return -1 at EOF, or throw `IOException`.

src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
binmahone: Done in b8720e0b. `getCopyBufferAllocationSize` now returns early once the max range reaches the configured copy buffer size.

src/test/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileTest.java
binmahone: Done in b8720e0b. Removed the redundant `assertEquals` and kept the `assertThrows(IOException.class, ...)` assertion.

binmahone:
build

gerashegalov:
**Follow-up: remaining `copyFromStream` callers still use the 128 KiB chunk**

This PR fixes the `readVectored` fallback path, but two other call sites remain on the slow path and would benefit from the same treatment:

- `RapidsInputFile.java:127` — `readTail` calls `output.copyFromStream(0, input, length)` with the default 128 KiB internal buffer
- `KudoTable.java:70` — `buffer.copyFromStream(0, din, header.getTotalDataLen())` likewise

The root cause is `HostMemoryBuffer.copyFromStream`'s hard-coded 128 KiB chunk size. A follow-up that adds a `copyFromStream(long destOffset, InputStream in, long length, int chunkSize)` overload (or changes the default) would let all three call sites benefit without needing per-site copy-buffer wrappers.

src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
```diff
@@ -0,0 +1,111 @@
...
+      RapidsInputFile inputFile,
+      HostMemoryBuffer output,
+      List<RapidsInputFile.CopyRange> copyRanges,
+      int copyBufferSize) throws IOException {
+    validateReadVectoredArgs(inputFile, output, copyRanges, copyBufferSize);
+    if (copyRanges.isEmpty()) {
+      return;
+    }
+
+    byte[] copyBuffer = new byte[getCopyBufferAllocationSize(copyRanges, copyBufferSize)];
```
gerashegalov, line 40: The buffer is described as "reusable" in the PR but it is only reused across ranges *within a single call* — a fresh  is allocated on every invocation. The old  path allocated 128 KiB per call; this is a 64x increase in per-call heap pressure for high-frequency callers (e.g. one  per Parquet row-group read).

Consider adding an overload that accepts a caller-supplied buffer, letting the caller amortize the allocation:



The existing  overload can then allocate once and delegate to it. Callers in hot paths (e.g. cudf-spark’s Hadoop adapter) can hold a single buffer and reuse it across reads.

gerashegalov, line 40: The buffer is described as "reusable" in the PR but it is only reused across ranges *within a single call* — a fresh `new byte[up to 8 MiB]` is allocated on every invocation. The old `copyFromStream` path allocated 128 KiB per call; this is a 64x increase in per-call heap pressure for high-frequency callers (e.g. one `readVectored` per Parquet row-group read).

Consider adding an overload that accepts a caller-supplied buffer, letting the caller amortize the allocation:

```java
static void readVectoredUsingCopyBuffer(
    RapidsInputFile inputFile,
    HostMemoryBuffer output,
    List<CopyRange> copyRanges,
    byte[] copyBuffer) throws IOException
```

The existing `int copyBufferSize` overload can then allocate once and delegate to it. Callers in hot paths (e.g. cudf-spark's Hadoop adapter) can hold a single buffer and reuse it across reads.

gerashegalov: https://github.com/NVIDIA/cudf-spark-jni/pull/4765#discussion_r3490230267

binmahone: Agreed that the previous wording was imprecise: the allocated buffer was reusable only within one `readVectored` invocation. The 64x allocation increase is the worst case because the allocated buffer is capped by the largest range, but the cross-call allocation concern is valid. I addressed this in f1269668 by adding a caller-supplied `byte[]` overload and updating the docs so repeated-read callers can manage buffer reuse explicitly. I did not add a `ThreadLocal` policy here; the helper now exposes the mechanism and leaves buffer lifetime to the caller.

binmahone: Handled by f1269668 as above: added a caller-supplied `byte[]` overload and clarified the allocation-vs-reuse wording.

binmahone: Follow-up cleanup pushed in 17938f2a: the public interface now exposes only the caller-supplied `byte[]` helper. The `int copyBufferSize` allocation helper remains package-private in `RapidsInputFileUtils` for the default fallback implementation, so hot callers have an explicit reuse API without adding another public size-based helper.

binmahone:
Good point. I agree those remaining `copyFromStream` callers still inherit the 128 KiB chunk behavior. I would keep that as a follow-up rather than fold it into this PR: this PR is scoped to the `readVectored` fallback regression that cudf-spark now hits from parquet scan, while `readTail` and `KudoTable` have different call patterns and should probably be handled by a general `HostMemoryBuffer.copyFromStream(..., chunkSize)` API or default-size change. The dependent cudf-spark PR only needs the caller-supplied `readVectored` helper from this PR.

binmahone:
build

sameerz:
@binmahone please retarget to `release/26.06` as this is a regression and we will need to release a hotfix.

src/main/java/com/nvidia/spark/rapids/jni/fileio/RapidsInputFileUtils.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+
+  private static void validateReadVectoredArgs(
+      RapidsInputFile inputFile,
+      HostMemoryBuffer output,
+      List<RapidsInputFile.CopyRange> copyRanges,
+      int copyBufferSize) {
+    Objects.requireNonNull(inputFile, "inputFile can't be null");
+    if (copyBufferSize <= 0) {
```
gerashegalov, line 88: **[Nit]** `inputFile` is null-checked here but the 3-arg delegate at line 112 checks it again. The same redundancy is at line 100 in the `byte[]` overload. The 4-arg overloads can drop their own `requireNonNull(inputFile)` and let the delegate handle it consistently.

gerashegalov (dismissed):
LGTM

binmahone:
Done. This PR targets `release/26.06`. I also updated the description with the final dependent-build verification and the five-pair interleaved EMR on EKS results. The JNI code itself was left unchanged after review.

binmahone:
build

gerashegalov (approved):
LGTM

PR #4782, opened 2026-07-02
Fix misuse of pointer type in `KudoGpuSerializer.java#assembleFromDeviceRaw`

ttnghia:
`KudoGpuSerializer.java#AssembleResult` only stores the buffer handle (i.e., pointer of type `rmm::device_buffer*`). When calling `DeviceMemoryBuffer.fromRmm(gpuAddress, size, handle)`, it passes handle `rmm::device_buffer*` twice for both `gpuAddress` and `handle`. This is incorrect, and such call will result in crash or data corruption.

This fix properly stores both the GPU address of the buffer as well as buffer handle and pass to `fromRmm` by the proper way.

### DIFF
--- a/src/main/cpp/src/KudoGpuSerializerJni.cpp
+++ b/src/main/cpp/src/KudoGpuSerializerJni.cpp
@@ -112,10 +112,15 @@ Java_com_nvidia_spark_rapids_jni_kudo_KudoGpuSerializer_assembleFromDeviceRawNat
                                             cudf::get_default_stream(),
                                             cudf::get_current_device_resource_ref());
 
-    // Create buffer metadata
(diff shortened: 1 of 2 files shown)


ttnghia:
build

ttnghia:
build

ttnghia:
build

zpuller (approved):

zpuller:
should we add a test that executes the spill path, which would presumably fail before this change and now pass?

ttnghia:
> should we add a test that executes the spill path, which would presumably fail before this change and now pass?

Unfortunately testing it is awkward in JNI because the problematic buffer is never exposed, it is used internally in a fuction, which currently does not trigger memory access to the wrongly provided GPU address. This bug has never triggered, only discovered by accident.

PR #4783, opened 2026-07-02
Synchronize stream to drain H2D uploads before host staging vectors are destroyed

ttnghia:
This fixes the bug that the host staging vectors are destroyed before their data was completely uploaded to device memory through async memcpy. 

From CUDA 13, `memcpy_batch_async` routes a non-default-stream copy through `cudaMemcpyBatchAsync` with `cudaMemcpySrcAccessOrderStream`, so the host source is read when the stream executes the copy, not at enqueue. As such, the bug already stay hidden for a long time but didn't show up until being tested with CUDA 13 build.

### DIFF
--- a/src/main/cpp/src/row_conversion.cu
+++ b/src/main/cpp/src/row_conversion.cu
@@ -1998,6 +1998,10 @@ std::vector<std::unique_ptr<column>> convert_to_rows(
           batch_row_offset,
           reinterpret_cast<int8_t*>(output_data[i]));
     }
+
+    // Drain the async H2D uploads above before variable_width_input_data goes out of scope:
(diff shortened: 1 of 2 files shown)


ttnghia:
Tried to add tests but non of them could reproduce the issue in native test. The bugs only show up in Spark query with multiple concurrent GPU tasks.

ttnghia:
build

ttnghia:
build

abellina (approved):
This looks great to me and also appears to be the root cause of something I am seeing when enabling GPU Kudo serialization and deserialization. I can't repro my illegal access with this patch.

PR #4794, opened 2026-07-06
Prevent NUL from being selected as the JSON row delimiter

nartal1:
## Summary

This PR contributes to https://github.com/NVIDIA/cudf-spark/issues/15212

`concat_json` can select NUL (`\0`) as the delimiter used to concatenate JSON rows.
When an input string contains embedded NUL characters, the cuDF JSON reader can interpret
them as additional row boundaries and return more rows than were present in the input.

This can cause downstream expressions such as `CASE WHEN` to fail with:

```
Boolean mask column must be the same size as lhs column
```

This change excludes NUL from the possible JSON delimiters.

## Test plan

- Added FromJsonToStructsTest.testEmbeddedNulIsNotUsedAsRowDelimiter.
- Added `FromJsonToRawMapTest.testExtractRawMapArrayEmbeddedNul` to cover the
  `MAP<STRING, ARRAY<STRING>>` raw-extraction path.
- Verified that the output struct and its children retain the input row count.
- Ran the focused JNI test successfully.

A separate cudf-spark PR will add Spark integration coverage for the
from_json and CASE WHEN reproduction.

### DIFF
--- a/src/test/java/com/nvidia/spark/rapids/jni/FromJsonToRawMapTest.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/FromJsonToRawMapTest.java
@@ -26,6 +26,7 @@ import ai.rapids.cudf.JSONOptions;
 import org.junit.jupiter.api.Test;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import static ai.rapids.cudf.AssertUtils.assertColumnsAreEqual;
@@ -203,6 +204,25 @@ public class FromJsonToRawMapTest {
          ColumnVector result = JSONUtils.extractRawMapFromJsonString(input, getOptions(),
              JSONUtils.MapValueType.ARRAY_OF_STRING);
          ColumnVector expected = ColumnVector.fromLists(LIST_TYPE, row0)) {
+      assertColumnsAreEqual(expected, result);
+      assertKeyAndStructNonNullable(result);
+    }
+  }
+
+  // Empty keys, lists, and strings are valid values and must remain distinct from JSON null.
+  @Test
+  void testExtractRawMapArrayEmptyBoundaries() {
+    List<HostColumnVector.StructData> row0 =
+        Arrays.asList(pair("", Collections.emptyList()));
+    List<HostColumnVector.StructData> row1 =
+        Arrays.asList(pair("k", Collections.emptyList()));
+    List<HostColumnVector.StructData> row2 =
+        Arrays.asList(pair("k", Arrays.asList("", "null", null)));
+    try (ColumnVector input = ColumnVector.fromStrings(
+             "{\"\":[]}", "{\"k\":[]}", "{\"k\":[\"\",\"null\",null]}");
+         ColumnVector result = JSONUtils.extractRawMapFromJsonString(input, getOptions(),
+             JSONUtils.MapValueType.ARRAY_OF_STRING);
+         ColumnVector expected = ColumnVector.fromLists(LIST_TYPE, row0, row1, row2)) {
       assertColumnsAreEqual(expected, result);
       assertKeyAndStructNonNullable(result);
     }
--- a/src/test/java/com/nvidia/spark/rapids/jni/FromJsonToRawMapTest.java
+++ b/src/test/java/com/nvidia/spark/rapids/jni/FromJsonToRawMapTest.java
@@ -192,6 +192,22 @@ public class FromJsonToRawMapTest {
     }
   }
 
+  // The delimiter used to concatenate rows must not split an embedded NUL in an array element.
+  @Test
+  void testExtractRawMapArrayEmbeddedNul() {
+    String valueWithNul = "a\u0000b";
+    List<HostColumnVector.StructData> row0 =
+        Arrays.asList(pair("k", Arrays.asList(valueWithNul)));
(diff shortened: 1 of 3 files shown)


nartal1:
build

src/test/java/com/nvidia/spark/rapids/jni/FromJsonToStructsTest.java
```diff
@@ -0,0 +1,61 @@
...
+    Schema schema = root.build();
+
+    try (ColumnVector input = ColumnVector.fromStrings(json);
+         ColumnVector output = JSONUtils.fromJSONToStructs(
+             input, schema, getOptions(), true);
+         ColumnView outputAniviaData = output.getChildColumnView(0);
+         ColumnView outputAsset = outputAniviaData.getChildColumnView(0)) {
+      assertEquals(input.getRowCount(), output.getRowCount());
+      assertEquals(input.getRowCount(), outputAniviaData.getRowCount());
+      assertEquals(input.getRowCount(), outputAsset.getRowCount());
```
amahussein, line 64: `MAP<STRING,ARRAY<STRING>>` raw extraction also reaches the fixed `concat_json` function through `unify_json_strings`/`tokenize_and_classify`. Could we add a sibling embedded-NUL regression through `extractRawMapFromJsonString(..., MapValueType.ARRAY_OF_STRING)`, or explicitly track it in the planned cudf-spark integration follow-up? The current test proves the struct caller but not the newly exposed raw-map-array caller.

CC: @ttnghia for visibility related to https://github.com/NVIDIA/cudf-spark/pull/15134

amahussein (changes requested):

nartal1:
build

src/test/java/com/nvidia/spark/rapids/jni/FromJsonToStructsTest.java
nartal1: Thanks for the suggestion @amahussein ! I added a regression test here using an embedded NUL inside an array element.  PTAL

amahussein, line 64: Could we also assert the expected value or validity of `assetId` here? These assertions pin the
cardinality failure, which is the primary regression, but a one-row null or otherwise corrupted
parse would still pass. Since the input deliberately places a valid `"assetId":"va"` before the
malformed banner, checking its expected recovered result would also catch silent data loss. If
cuDF intentionally nullifies that field for this malformed record, explicitly asserting that null
state would still strengthen the contract.

src/test/java/com/nvidia/spark/rapids/jni/FromJsonToRawMapTest.java
```diff
@@ -192,6 +192,22 @@ void testExtractRawMapArrayFromJsonString() {
...
+    String valueWithNul = "a\u0000b";
+    List<HostColumnVector.StructData> row0 =
+        Arrays.asList(pair("k", Arrays.asList(valueWithNul)));
+    try (ColumnVector input =
+             ColumnVector.fromStrings("{\"k\":[\"" + valueWithNul + "\"]}");
+         ColumnVector result = JSONUtils.extractRawMapFromJsonString(input, getOptions(),
+             JSONUtils.MapValueType.ARRAY_OF_STRING);
+         ColumnVector expected = ColumnVector.fromLists(LIST_TYPE, row0)) {
+      assertColumnsAreEqual(expected, result);
+      assertKeyAndStructNonNullable(result);
```
amahussein, line 208: Could we add a compact sibling case for the zero-length/null boundaries of the
`ARRAY_OF_STRING` path—for example `{"":[]}`, `{"k":[]}`, and
`{"k":["", "null", null]}`? The implementation distinguishes valid empty keys, lists, and
strings from nulls, but the direct JNI tests currently pin only the null list and literal-null
element cases below. cudf-spark #15134 covers most of this matrix at integration level, so I see
this as non-blocking; keeping a small case next to the CUDA/JNI implementation would make future
regressions easier to isolate. A null key itself is not valid JSON and does not need a value case.

amahussein (dismissed):
The centralized delimiter fix looks correct, and the newly added raw-map regression closes the caller-coverage gap: the struct path pins row cardinality, while the `ARRAY_OF_STRING` path verifies exact preservation of a NUL-containing element. I found no blocking implementation issue. Null map values/lists and
literal-null elements are directly covered; empty keys, empty lists, and empty-string elements are implemented but currently rely mainly on cudf-spark #15134's integration matrix. 
I left those as optional test-hardening suggestions. Final merge should still wait for the blossom/GPU result, and #15134's final validation should consume a JNI artifact containing this fix, as already noted in the existing review.

nartal1:
build

src/test/java/com/nvidia/spark/rapids/jni/FromJsonToStructsTest.java
nartal1: Done. PTAL

src/test/java/com/nvidia/spark/rapids/jni/FromJsonToRawMapTest.java
nartal1: Done

amahussein (approved):
LGTM
Thanks @nartal1

PR #4797, opened 2026-07-07
Fix protobuf decoding for sliced LIST inputs

thirtiseven:
## Summary

- Use the sliced LIST child and sliced offset range when decoding protobuf input.
- Add regression coverage for a sliced nullable input view containing malformed data.

## Root cause

The decoder read the unsliced LIST child and offsets directly. For a sliced LIST view, this could map rows to bytes outside the slice and produce incorrect row boundaries or validity.


src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -421,9 +421,10 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-  auto const* message_data = reinterpret_cast<uint8_t const*>(in_list_view.child().data<int8_t>());
-  auto const message_data_size = in_list_view.child().size();
-  auto const* list_offsets     = in_list_view.offsets().data<cudf::size_type>();
+  auto const message_bytes     = in_list_view.get_sliced_child(stream);
+  auto const* message_data     = reinterpret_cast<uint8_t const*>(message_bytes.data<int8_t>());
+  auto const message_data_size = message_bytes.size();
+  auto const* list_offsets     = in_list_view.offsets_begin();
 
   // Stage list_offsets[0] through pinned host memory so the D2H stays truly async.
   auto h_base_offset = cudf::detail::make_pinned_vector_async<cudf::size_type>(1, stream);
```
igorpeshansky: [Optional] Doesn't `get_sliced_child` already do exactly this? Is this now redundant?

We should be able to just compute `base_offset` on the host:

```c++
  cudf::size_type const base_offset = static_cast<cudf::size_type>(message_data - reinterpret_cast<uint8_t const*>(in_list_view.child().data<int8_t>()));
```

and skip the `memcpy_async`/`synchronize`.

This becomes even cleaner with https://github.com/NVIDIA/cudf-spark-jni/pull/4797#discussion_r3575194012 …

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3710,6 +3710,32 @@ void testHiddenRepeatedMessageAbsentDoesNotFail() {
...
 
+  @Test
+  void testSlicedNullableInputMapsMalformedRootScalarRow_Permissive() {
+    Byte[] sentinel = concat(box(tag(99, WT_VARINT)), box(encodeVarint(7)));
+    Byte[] malformed = concat(box(tag(1, WT_VARINT)), new Byte[]{(byte) 0x80});
+    Byte[] valid = concat(box(tag(1, WT_VARINT)), box(encodeVarint(42)));
+    ProtobufSchemaDescriptor schema = new ProtobufSchemaDescriptorBuilder()
+        .addField(1, DType.INT32)
+        .build();
+    StructType outputType = new StructType(true, new BasicType(true, DType.INT32));
```
igorpeshansky, line 3722: [Optional] Do we need to test other data types?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3710,6 +3710,32 @@ void testHiddenRepeatedMessageAbsentDoesNotFail() {
...
+
+    try (Table input = new Table.TestBuilder()
+             .column(new Byte[][]{sentinel, null, malformed, valid, sentinel})
+             .build();
+         ColumnVector expected = ColumnVector.fromStructs(
+             outputType, (StructData) null, (StructData) null, struct(42))) {
+      ColumnView[] views = input.getColumn(0).splitAsViews(1, 4);
```
igorpeshansky: Should we also add a test for a truncated slice with offset 0?

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
igorpeshansky, line 1: [Optional] Is it possible to run this test before the merge?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -421,9 +421,10 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
 
   // Extract shared input data pointers (used by scalar, repeated, and nested sections)
   cudf::lists_column_view const in_list_view(binary_input);
-  auto const* message_data = reinterpret_cast<uint8_t const*>(in_list_view.child().data<int8_t>());
-  auto const message_data_size = in_list_view.child().size();
-  auto const* list_offsets     = in_list_view.offsets().data<cudf::size_type>();
+  auto const message_bytes     = in_list_view.get_sliced_child(stream);
+  auto const* message_data     = reinterpret_cast<uint8_t const*>(message_bytes.data<int8_t>());
```
igorpeshansky: Out of curiosity, why are we not just using `message_bytes.data<uint8_t>()` here?

thirtiseven: Good catch, we can just use it. Updated.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Thanks, you are right. Updated.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Yes, I am running all protobuf tests locally before each commit. They will also be run in pre-merge. Updated the pr descirption to remove the `tests not run` statement.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: I added a non-zero sliced-input test for a nested STRING field. This exercises both nested scanning and variable-width string extraction in addition to the existing INT32 scalar path.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Yes, added one.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
```diff
@@ -3710,6 +3710,89 @@ void testHiddenRepeatedMessageAbsentDoesNotFail() {
...
+      try (ColumnVector actual = Protobuf.decodeToStruct(views[1], schema, false)) {
+        AssertUtils.assertStructColumnsAreEqual(expected, actual);
+      } finally {
+        for (ColumnView view : views) {
```
igorpeshansky: We could avoid these explicit loops by using the `ai.rapids.cudf.CloseableArray` helper, e.g.:

```java
    try (Table input = new Table.TestBuilder()
             .column(new Byte[][]{sentinel, null, malformed, valid, sentinel})
             .build();
         ColumnVector expected = ColumnVector.fromStructs(
             outputType, (StructData) null, (StructData) null, struct(42));
         CloseableArray<ColumnView> views =
             CloseableArray.wrap(input.getColumn(0).splitAsViews(1, 4));
         ColumnVector actual = Protobuf.decodeToStruct(views.get(1), schema, false)) {
      AssertUtils.assertStructColumnsAreEqual(expected, actual);
    }
```

Now that we have 3 of them in this file, it might be worth the investment… The helper is already used in, e.g., `KudoSerializerTest`. As a follow-up, we can also change `MapUtilsTest.slicedInputThrows`…

igorpeshansky (dismissed):
Approving to unblock, but there's an optional cleanup proposal.

src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java
thirtiseven: Good suggestion. I changed the three new `ProtobufTest` cases and the existing `MapUtilsTest.slicedInputThrows` case to use `CloseableArray`, so all splitAsViews call sites in our test sources now use the helper.

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
```diff
@@ -265,17 +266,11 @@ void slicedInputThrows() {
     List<HostColumnVector.StructData> row0 = Arrays.asList(entry(1, 10));
     List<HostColumnVector.StructData> row1 = Arrays.asList(entry(2, 20));
     List<HostColumnVector.StructData> row2 = Arrays.asList(entry(3, 30));
-    try (ColumnVector full = ColumnVector.fromLists(LIST_TYPE, row0, row1, row2)) {
-      ColumnView[] views = full.splitAsViews(1);  // [0..1), [1..3) — second has offset != 0
```
igorpeshansky, line 269: Would it be helpful to keep this comment?

igorpeshansky (approved):
One minor nit in the new file.

thirtiseven:
build

src/test/java/com/nvidia/spark/rapids/jni/MapUtilsTest.java
thirtiseven: Yes, done in #4799

PR #4798, opened 2026-07-07
Fix protobuf malformed-row handling before field decoding

thirtiseven:
## Summary

Fix malformed-input handling in protobuf scan and sizing paths that run before value extraction:

- Propagate errors from repeated-field counting to the per-row invalid mask, so PERMISSIVE mode nulls the entire malformed struct instead of returning partial or empty data.
- After a safely skippable nested-field wire-type mismatch, mark the row invalid but continue scanning so repeated-field counting remains consistent with occurrence scanning.
- Validate protobuf wire data even when the output schema has no fields.
- Add regression coverage for the affected PERMISSIVE and FAILFAST paths.

## Background

Repeated fields are decoded in multiple passes:

1. Count repeated occurrences and record nested-message locations.
2. Calculate LIST offsets and allocate occurrence storage.
3. Scan the input again and record each occurrence.
4. Extract values and build the output columns.

The count and occurrence scans must observe the same repeated fields. Decode errors are tracked in two ways:

- A batch-wide error flag is used by FAILFAST mode.
- A per-row invalid flag is used by PERMISSIVE mode to null malformed struct rows.

## Root cause

The repeated-field count path previously recorded only the batch-wide error. In PERMISSIVE mode, malformed input detected during counting could therefore produce a partial or empty result while leaving the struct row non-null.

The count path also stopped immediately when a nested field used the wrong wire type. The occurrence scan could safely skip that field and continue to later repeated fields, causing the two passes to disagree about the required occurrence storage.

Separately, the empty-schema path returned an empty struct without parsing the payload. Spark still parses and skips unknown fields before producing an empty struct, so malformed payloads must still be handled according to the configured mode.

## Behavior

After this change:

- PERMISSIVE mode returns a null struct row for malformed input detected by these paths.
- FAILFAST mode throws for malformed input, including when the output schema is empty.
- Valid input behavior is unchanged.
- A malformed row does not affect subsequent valid rows in the same batch.

## Testing

Added regression coverage for:

- Empty output schemas in PERMISSIVE and FAILFAST modes.
- Repeated fields with incorrect wire types.
- Misaligned packed fixed32 and fixed64 payloads.
- A nested-field wire mismatch followed by a repeated field.
- A malformed repeated field followed by another field and subsequent valid rows.

Ran:

- `pre-commit run --files src/main/cpp/src/protobuf/protobuf.cu src/main/cpp/src/protobuf/protobuf_kernels.cu src/main/cpp/src/protobuf/protobuf_kernels.cuh src/test/java/com/nvidia/spark/rapids/jni/ProtobufTest.java`


src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -447,10 +437,12 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
 
   rmm::device_uvector<device_nested_field_descriptor> d_schema(num_fields, stream, scratch_mr);
-  CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_schema.data(),
-                                           h_device_schema.data(),
-                                           num_fields * sizeof(device_nested_field_descriptor),
-                                           stream));
+  if (num_fields > 0) {
+    CUDF_CUDA_TRY(cudf::detail::memcpy_async(d_schema.data(),
+                                             h_device_schema.data(),
+                                             num_fields * sizeof(device_nested_field_descriptor),
```
igorpeshansky: Isn't `cudf::detail::memcpy_async` a no-op when given 0 bytes to copy? What would be the drawback of making this call unconditionally?

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -487,6 +479,23 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   }
   auto const decode_ctx = protobuf_decode_runtime_context{&d_row_force_null, &d_error};
 
+  // Spark still parses unknown fields before producing an empty struct.
+  if (num_fields == 0) {
+    // The zero-width kernel still forms a row pointer; one element keeps it valid, and no
+    // locations are written.
+    rmm::device_uvector<field_location> d_unused_location(1, stream, scratch_mr);
+    launch_scan_all_fields(*d_in,
```
igorpeshansky: The invariant you want to maintain is that at least one of `launch_scan_all_fields` or `launch_count_repeated_fields` runs on the entire message to ensure validation.

FWIW, it seems that a lot of the code below can be made 0-friendly (use `std::max(1, num_scalar)` for vector sizes, etc), and then this `launch_scan_all_fields` call would naturally happen as part of the normal flow in line 595, without special-casing.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -397,36 +408,42 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
+        }
+        // cur - msg_base is bounded by the message length (<= INT32_MAX via
+        // check_message_bounds), so this fits int32; checked_add_int32 still guards the offset +
+        // len_bytes addition. Matches the LEN handling in scan_message_field_locations.
+        int const data_offset = static_cast<int>(cur - msg_base);
+        int32_t data_location;
+        if (!checked_add_int32(data_offset, len_bytes, data_location)) {
+          set_error_once(error_flag, protobuf_error::OVERFLOW);
+          mark_row_error();
+          return;
```
igorpeshansky: If we need to keep scanning in row 412 above, why is it ok to `return` early here (and possibly also in lines 419, 425, and 447)? Here specifically, shouldn't it match `scan_all_repeated_occurrences_in_message` and keep going?
Otherwise would it not stop counting, causing a `REPEATED_COUNT_MISMATCH`, which will fail even in `PERMISSIVE` mode?

[Optional] FWIW, the real fix for this whole "keep scanning vs return" mismatch issue would be to have a shared loop driver kernel for all such loops (the refactoring suggested earlier in https://github.com/NVIDIA/cudf-spark-jni/pull/4644#discussion_r3358116516 that we keep deferring) — then all loops stay in lock-step regarding how they process the fields (the driver would return on unrecoverable errors, but be configurable to either continue or exit on recoverable errors, similar to `wire_type_mismatch_policy`).

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -397,36 +408,42 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
-      // so this fits int32; checked_add_int32 still guards the offset + len_bytes addition. Matches
-      // the LEN handling in scan_message_field_locations.
-      int const data_offset = static_cast<int>(cur - msg_base);
-      int32_t data_location;
-      if (!checked_add_int32(data_offset, len_bytes, data_location)) {
-        set_error_once(error_flag, protobuf_error::OVERFLOW);
-        return;
+        mark_row_error();
+        // Keep scanning so later repeated-field counts stay aligned with the occurrence scan.
+      } else {
```
igorpeshansky, line 413: [Optional] Instead of an `else`, it would be great if we could just use a `continue` and let the block in lines 442-449 happen naturally as part of the `for` loop. In fact, we can, with something like this:

```c++
  int wt = -1;  // dummy value for capture
  auto advance = [&](uint8_t const* cur) {
    uint8_t const* next;
    if (!skip_field(cur, msg_end, wt, next)) {
      set_error_once(error_flag, protobuf_error::SKIP);
      mark_row_error();
      return msg_end;
    }
    return next;
  };
  for (uint8_t const* cur = msg_base; cur < msg_end; cur = advance(cur)) {
    …
    wt = tag.wire_type;
    …
    // Check nested message fields at this depth
    if (int f = lookup_field_idx(fn, fn_to_nested_idx, …); f >= 0) {
      …
      if (wt != wire_type_value(proto_wire_type::LEN)) {
        set_error_once(error_flag, protobuf_error::WIRE_TYPE);
        mark_row_error();
        // Keep scanning so later repeated-field counts stay aligned with the occurrence scan.
        continue;
      }
      uint64_t len;
      …
    }
  }
```

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Yes. `cudf::detail::memcpy_async` returns `cudaSuccess` immediately, so the guard is unnecessary. Removed.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Thanks! In dev branch I'm using a 
```
scan_message_field_locations<report_error_and_skip>(...)
```
that can address this. I will put a separate pr for the refactor.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Great idea. I removed the special-case launch and made the normal field-scan path accept zero descriptors and locations.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Thanks! I have implemented https://github.com/NVIDIA/cudf-spark-jni/pull/4644#discussion_r3358116516 in dev branch and will submit separately.

These returns are for unrecoverable errors where the cursor can no longer be advanced safely. Continuing could reinterpret bytes inside the malformed field as subsequent tags. The wire-type mismatch case is different because `skip_field` can still advance from a valid cursor, so counting continues there.

I added `testMalformedNestedLengthBeforeRepeatedField_Permissive` to cover this distinction. The malformed row is nulled in PERMISSIVE mode without producing `REPEATED_COUNT_MISMATCH`, while the valid row still decodes.

src/main/cpp/src/protobuf/protobuf.cu
igorpeshansky, line 564: [Optional] Just pointing out that in the past we've used `std::max(…, 1)` in such cases, to ensure that `.data()` does not return a `nullptr`. We really only need this in cases where we use `.data()` for pointer arithmetic, which can also be guarded via a ternary expression or a conditional. So maybe we should do a pass to make this consistent, possibly in a follow-up if it feels out of scope here.

src/main/cpp/src/protobuf/protobuf_kernels.cu
igorpeshansky: `scan_all_repeated_occurrences_in_message` will skip nested fields, regardless of the overflow status, so it effectively treats it as a recoverable error. I'm fine with either, but shouldn't the two be consistent?

src/main/cpp/src/protobuf/protobuf_kernels.cu
igorpeshansky: FWIW, this also applies to `scan_message_field_locations`, and thus is orthogonal to the refactor — it just makes the loop cleaner. Feel free to defer to the follow-up.

thirtiseven: Thanks for clarifying. You're right that this loop-shape cleanup is orthogonal to the shared-driver refactor. Will do it in a follow-up.

src/main/cpp/src/protobuf/protobuf_kernels.cu

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Agreed. Will do in follow-up

igorpeshansky (approved):
LGTM :shipit:
Leaving follow-up comments open for easier tracking.

thirtiseven:
build

PR #4799, opened 2026-07-07
Refactor protobuf scan and decode context interfaces

thirtiseven:
## Summary

- Replace correlated protobuf scan/count kernel arguments with typed views.
- Generalize occurrence-scan names and ownership beyond repeated fields.
- Add a host `protobuf_schema` facade for validated metadata access, output selection, and ordered child lookup.
- Thread schema and runtime state through recursive builders as a single decode context.
- Preserve existing protobuf decoding behavior and feature support.

## Motivation

The protobuf decoder passed related pointers, sizes, lookup tables, and output buffers as independent kernel arguments. The host path also passed the schema and six parallel metadata vectors separately and repeatedly scanned the schema to find child fields.

This made ownership and lifetime requirements difficult to see and left similar top-level and nested paths with slightly different interfaces. This PR makes those contracts explicit without changing decoding semantics.

### Device scan interfaces

- Add `device_schema_view`, `repeated_field_count_view`, `nested_field_location_view`, `field_scan_view`, and `field_occurrence_scan_view`.
- Use the same protobuf input and nested-parent views across nested scan launchers.
- Centralize occurrence descriptors and field-number lookup buffers in an owning host bundle.
- Keep these device views as the kernel-facing API; the host schema refactor does not replace them.

### Host schema and decode context

- Replace `schema_context_view` and scattered metadata parameters with `protobuf_schema`.
- Validate the original `protobuf_decode_context` before building an ordered parent-to-children cache.
- Keep `protobuf_schema` non-owning with respect to the decode context and prevent construction from temporary contexts.
- Use `protobuf_field_meta_view` for per-field defaults and enum metadata.
- Pass `protobuf_schema` and runtime error/null state together through nested and repeated-child builders.
- Keep pinned field-descriptor staging and its device copy under one explicit owner.
- Remove the unused `build_repeated_struct_column` declaration.


src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -631,79 +641,59 @@ std::unique_ptr<cudf::column> build_nested_struct_column(
...
+    {child_field_descs.device.data(),
+     num_child_fields,
+     nullptr,
+     0,
+     d_child_locations.data(),
+     d_repeated_info.data()},
     decode_ctx.error->data(),
     !decode_ctx.row_force_null->is_empty() ? decode_ctx.row_force_null->data() : nullptr,
-    parent.top_row_indices,
     stream);
```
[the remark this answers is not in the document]
thirtiseven: Thanks. The null lookup is intentional and preserves the existing nested-scan path. A direct table would add pinned/device allocations and an H2D copy sized by the maximum field number for every nested builder invocation, so it is not unconditionally faster for the common narrow-schema case. We prefer to benchmark that tradeoff separately rather than include it in this interface-only refactor.

thirtiseven:
Thanks for the review. On the two observations outside the diff:

- The `host_lookup` lifetime is intentional. `make_pinned_vector_async(..., stream)` binds the host allocator to that stream, and `host_vector` destruction deallocates through the same stream, so pinned-pool reuse is ordered after the preceding H2D copy. The device MR passed to this helper does not replace the host allocator. Keeping the host vector in the returned bundle would still rely on the same stream-ordered deallocation contract. We previously resolved the same concern in [#4717](https://github.com/NVIDIA/cudf-spark-jni/pull/4717#discussion_r3418999783), after which Greptile confirmed that the pinned staging lifetimes were correctly managed.
- Repeated fields cannot carry defaults in an accepted decode context. `protobuf_schema` construction calls `validate_decode_context()`, which rejects `is_repeated && has_default_value`, and this builder also requires the target field to be repeated. The schema-derived value is therefore guaranteed to be `false`, matching the previous explicit argument. Occurrences are emitted only after `walk_repeated_element` successfully parses an element, so malformed elements do not produce negative-offset occurrences.

No code changes are needed for these two points.

thirtiseven:
@greptile review

src/main/cpp/src/protobuf/protobuf_types.cuh
```diff
@@ -156,4 +163,34 @@ struct device_nested_field_descriptor {
...
+  int lookup_size;
+};
+
+struct field_scan_view {
+  field_descriptor const* descriptors;
+  int size;
+  int const* field_number_lookup;
+  int lookup_size;
+  field_location* locations;
+  field_occurrence_count* repeated_info;
```
igorpeshansky, line 186: Can we put these two fields first, to match the other structs? As an added benefit, you'd be able to value-initialize the (new) last two fields (i.e., `field_number_lookup` and `lookup_size`) by omitting them, instead of having to explicitly specify `nullptr` and `0`…

src/main/cpp/src/protobuf/protobuf_types.cuh
```diff
@@ -116,11 +116,18 @@ struct repeated_occurrence {
...
+  int32_t const* row_offsets;     // Pre-computed prefix-sum offsets [num_rows + 1]
+  field_occurrence* occurrences;  // Output buffer [total_count]
+};
+
+struct field_occurrence_scan_view {
+  field_occurrence_scan_desc const* descriptors;
+  int size;
+  int const* field_number_lookup;
+  int lookup_size;
 };
```
igorpeshansky: Can we define this as a generic struct? E.g.:

```c++
template <typename T>
struct lookup_view {
  T const* data;      // [size]
  int size;
  int const* direct;  // direct-mapped lookup table (nullable)
  int direct_size;    // size of lookup table (0 if null)
};
using field_occurrence_scan_view = lookup_view<field_occurrence_scan_desc>;
…
struct repeated_field_count_view {
  field_occurrence_count* info;
  lookup_view<int> schema_lookup;
};
struct nested_field_location_view {
  field_location* locations;
  lookup_view<int> schema_lookup;
};

struct field_scan_view {
  field_location* locations;
  field_occurrence_count* repeated_info;
  lookup_view<field_descriptor> field_lookup;
};
```

Then we can even do:

```c++
template <typename T, typename Match> requires std::is_invocable_r_v<bool, Match, int, int>
__device__ __forceinline__ int lookup_field(int field_number, lookup_view<T> table, Match&& match) { … }
template <typename T>
__device__ __forceinline__ int lookup_field(int field_number, lookup_view<T> table) {
  return lookup_field(field_number, table, [&table](int f, int n) { return table.data[f].field_number == n; });
}
…
  auto lookup_desc_idx = [&](int fn) { return lookup_field(fn, fields.field_lookup); };
…
  auto lookup_field_idx = [&](int fn, lookup_view<int> table) {
    return lookup_field(fn, table, [&](int local_i, int fn) {
      auto const& field_schema = schema.fields[table.data[local_i]];
      return field_schema.field_number == fn && field_schema.depth == schema.depth;
    });
  };
```

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -434,45 +427,43 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
+    write_idx[f] = fields.descriptors[f].row_offsets[row];
   }
 
   auto lookup_by_fn = [&](int fn) {
-    return lookup_field(fn, fn_to_desc_idx, fn_to_desc_size, num_scan_fields, [&](int f, int) {
-      return scan_descs[f].field_number == fn;
-    });
+    return lookup_field(
+      fn, fields.field_number_lookup, fields.lookup_size, fields.size, [&](int f, int) {
+        return fields.descriptors[f].field_number == fn;
```
igorpeshansky: Nit: we should be using the value that's passed in, not a captured one:

```suggestion
      fn, fields.field_number_lookup, fields.lookup_size, fields.size, [&](int f, int n) {
        return fields.descriptors[f].field_number == n;
```

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -349,39 +345,49 @@ void validate_decode_context(protobuf_decode_context const& context)
...
+  children_by_parent_.resize(context.schema.size() + 1);
+  for (int schema_idx = 0; schema_idx < static_cast<int>(context.schema.size()); ++schema_idx) {
+    children_by_parent_[context.schema[schema_idx].parent_idx + 1].push_back(schema_idx);
+  }
+}
+
+protobuf_field_meta_view protobuf_schema::field(int schema_idx) const
+{
+  auto const idx = static_cast<size_t>(schema_idx);
+  return protobuf_field_meta_view{context_.schema.at(idx),
```
igorpeshansky: The type is redundant here.

src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -349,39 +345,49 @@ void validate_decode_context(protobuf_decode_context const& context)
...
 }
 
-bool is_output_field(protobuf_decode_context const& context, int schema_idx)
+protobuf_schema::protobuf_schema(protobuf_decode_context const& context) : context_(context)
 {
-  return context.output_fields.empty() || context.output_fields.at(schema_idx);
+  validate_decode_context(context);
+  children_by_parent_.resize(context.schema.size() + 1);
+  for (int schema_idx = 0; schema_idx < static_cast<int>(context.schema.size()); ++schema_idx) {
+    children_by_parent_[context.schema[schema_idx].parent_idx + 1].push_back(schema_idx);
```
igorpeshansky, line 360: Do we want to (and can we) pre-`reserve` the inner vectors?

thirtiseven: Yes, done.

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Removed.

src/main/cpp/src/protobuf/protobuf_types.cuh
thirtiseven: Thanks. Looks great. Done.

src/main/cpp/src/protobuf/protobuf_types.cuh
thirtiseven: Done. I moved `locations` and `repeated_info` before the lookup state. With the trailing `lookup_view<field_descriptor>`, callers without a direct lookup table now initialize only `data` and `size`; `direct` and `direct_size` are value-initialized.

src/main/cpp/src/protobuf/protobuf_kernels.cu
thirtiseven: Fixed. The matcher now compares against the passed-in `n`. After the `lookup_view` refactor, this path uses the shared default matcher through `lookup_field(fn, fields)`.

src/main/cpp/src/protobuf/protobuf_types.cuh
```diff
@@ -170,27 +173,18 @@ struct device_schema_view {
...
 
 struct field_scan_view {
-  field_descriptor const* descriptors;
-  int size;
-  int const* field_number_lookup;
-  int lookup_size;
   field_location* locations;
   field_occurrence_count* repeated_info;
+  lookup_view<field_descriptor> field_lookup;
```
igorpeshansky: [Really optional] Given the struct name and the field type, it would be unambiguous to just name it `lookup`, saving some characters. But up to you.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -360,14 +353,9 @@ CUDF_KERNEL void count_repeated_fields_kernel(cudf::column_device_view const d_i
...
-    if (int f = lookup_field_idx(fn,
-                                 repeated.field_number_lookup,
-                                 repeated.lookup_size,
-                                 repeated.schema_indices,
-                                 repeated.size);
-        f >= 0) {
-      int schema_idx    = repeated.schema_indices[f];
-      auto& info        = repeated.info[flat_index(row, repeated.size, f)];
+    if (int f = lookup_field_idx(fn, repeated.schema_lookup); f >= 0) {
+      int schema_idx    = repeated.schema_lookup.data[f];
```
igorpeshansky: [Really optional] `schema_idx` is only used once, to look up the field's wire type. If we hoist that up to here, we might get nicer formatting for the `walk_repeated_element` call:

```c++
      int expected_wt   = schema.fields[repeated.schema_lookup.data[f]].wire_type;
      …
      if (!walk_repeated_element<wire_type_mismatch_policy::report_error>(
            cur, msg_end, msg_base, wt, expected_wt, error_flag, count_action)) { … }
```

or even (for symmetry with other kernels):

```c++
    auto get_expected_wire_type = [&](int f) {
      return schema.fields[repeated.schema_lookup.data[f]].wire_type;
    };
    if (int f = lookup_field_idx(fn, repeated.schema_lookup); f >= 0) {
      …
      if (!walk_repeated_element<wire_type_mismatch_policy::report_error>(
            cur, msg_end, msg_base, wt, get_expected_wire_type(f), error_flag, count_action)) { … }
    }
```

igorpeshansky (dismissed):
Approving with two optional tweaks.

src/main/cpp/src/protobuf/protobuf_types.cuh
thirtiseven: Renamed to `lookup`

src/main/cpp/src/protobuf/protobuf_kernels.cu

igorpeshansky (approved):
LGTM :shipit:

igorpeshansky:
@greptileai please re-review.

thirtiseven:
build

PR #4807, opened 2026-07-08
Fix JSON row delimiter selection for control-byte inputs

nartal1:
This PR addresses NVIDIA/cudf-spark#15231.

#4794 prevented NUL from being selected as the delimiter used to join JSON rows. That change exposed another problem in the existing delimiter-selection code that is - the histogram used to find an unused character could fail, and the
failure was not checked.

As a result, the code could select a control character that was already present in the JSON input. cuDF would then treat that character as a row separator which could turn one input row into multiple rows and cause downstream row-count mismatch errors.

This PR fixes the histogram calculation, checks for CUDA errors and makes sure the selected delimiter is not present in the input. 

It also adds tests for embedded NUL characters, control characters, printable characters, DEL and the case where no delimiter is available.

## Testing

- JNI build passed
- All 23 native C++ tests passed
- 11 focused JNI JSON tests passed
- The three failing tests from NVIDIA/cudf-spark#15231 passed
- The full JSON matrix passed with 946 tests and no failures

### DIFF
--- a/src/main/cpp/src/json_utils.cu
+++ b/src/main/cpp/src/json_utils.cu
@@ -36,6 +36,8 @@
 #include <thrust/transform.h>
 #include <thrust/uninitialized_fill.h>
 
+#include <cstdint>
+
 namespace spark_rapids_jni {
 
 namespace detail {
@@ -47,9 +49,9 @@ __host__ __device__ constexpr bool not_whitespace(cudf::char_utf8 ch)
   return ch != ' ' && ch != '\r' && ch != '\n' && ch != '\t';
 }
 
-__host__ __device__ constexpr bool can_be_delimiter(char c)
+__host__ __device__ constexpr bool can_be_delimiter(std::uint8_t c)
 {
-  // The character list below is from `json_reader_options.set_delimiter`.
+  // The character list below is from `json_reader_options::set_delimiter`.
   switch (c) {
     case '{':
     case '[':
@@ -66,6 +68,43 @@ __host__ __device__ constexpr bool can_be_delimiter(char c)
     case '\r': return false;
     default: return true;
   }
+}
+
+// Delimiter preference, in the order `delimiter_candidate` enumerates it:
+//   1. '\n'                         - the natural JSON-lines delimiter
+//   2. printable ASCII 0x21..0x7e   - safe, never confused with unquoted control chars
+//   3. DEL 0x7f
+//   4. C0 controls 0x01..0x1f       - last-resort fallback (see note in `delimiter_candidate`)
+// NUL (0x00) is never a candidate; the other excluded bytes are filtered by `can_be_delimiter`.
+constexpr std::uint8_t first_printable = 0x21;  // '!'
+constexpr std::uint8_t last_printable  = 0x7e;  // '~'
+constexpr std::uint8_t del_byte        = 0x7f;
+constexpr std::uint8_t last_c0         = 0x1f;  // highest C0 control byte used in the fallback
+
+constexpr int num_printable         = last_printable - first_printable + 1;
+constexpr int first_printable_index = 1;  // index 0 is '\n'
+constexpr int del_index             = first_printable_index + num_printable;
+constexpr int first_c0_index        = del_index + 1;
+
+// One index per distinct preference slot: '\n' + printables + DEL + C0 fallback (0x01..0x1f).
+constexpr int num_delimiter_candidates = first_c0_index + last_c0;
+
+__host__ __device__ constexpr std::uint8_t delimiter_candidate(int candidate_index)
+{
+  if (candidate_index == 0) { return '\n'; }
+  if (candidate_index < del_index) {
+    return static_cast<std::uint8_t>(first_printable + candidate_index - first_printable_index);
+  }
+  if (candidate_index == del_index) { return del_byte; }
+
+  // C0 bytes (0x01..0x1f) are a last-resort fallback, reached only if '\n', every eligible
+  // printable, and DEL are all already present in the input. A C0 delimiter can still be
+  // mishandled by the reader when `allow_unquoted_control` is set (the very interaction this
+  // ordering avoids), so this branch is intentionally the least preferred. NUL (0x00) is not
+  // reachable here; tab (0x09) and CR (0x0d) fall in this range but are rejected by
+  // `can_be_delimiter`, while LF (0x0a) is already covered at index 0 (its count is nonzero
+  // whenever this branch is reached, so the duplicate is harmless).
+  return static_cast<std::uint8_t>(candidate_index - del_index);
 }
 
 }  // namespace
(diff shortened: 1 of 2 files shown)


nartal1:
build

nartal1:
@ttnghia - It would be good if you could review this since you are working on json related issues.

src/main/cpp/tests/from_json.cpp
```diff
@@ -104,6 +104,14 @@ std::unique_ptr<cudf::column> make_expected_raw_map(std::vector<std::vector<kv>>
 // Convenience: every row valid.
 std::vector<bool> all_valid(std::size_t n) { return std::vector<bool>(n, true); }
 
+// Returns the delimiter `concat_json` picks for `input`; used to pin delimiter selection.
+char selected_delimiter(std::string const& input)
+{
+  auto const input_col = cudf::test::strings_column_wrapper{input};
+  auto const result    = spark_rapids_jni::concat_json(cudf::strings_column_view{input_col});
+  return std::get<1>(result);
```
firestarman, line 114: [SHOULD FIX] Can you add a delimiter-selection test where every earlier candidate is present and only C0 byte `0x1f` remains available? The implementation defines `last_c0 = 0x1f` and derives the candidate count from it, but current tests cover the first C0 fallback and the all-exhausted path, not the final fallback byte. A boundary regression that omits the last C0 slot would still pass.

Suggested test: add a `ConcatJsonDelimiter_UsesLastC0Fallback` case that populates LF, printable bytes, DEL, and C0 bytes `0x01..0x1e`, then asserts `selected_delimiter(input) == '\x1f'`.

src/main/cpp/src/json_utils.cu
```diff
@@ -68,6 +70,43 @@ __host__ __device__ constexpr bool can_be_delimiter(char c)
...
+constexpr std::uint8_t del_byte        = 0x7f;
+constexpr std::uint8_t last_c0         = 0x1f;  // highest C0 control byte used in the fallback
+
+constexpr int num_printable         = last_printable - first_printable + 1;
+constexpr int first_printable_index = 1;  // index 0 is '\n'
+constexpr int del_index             = first_printable_index + num_printable;
+constexpr int first_c0_index        = del_index + 1;
+
+// One index per distinct preference slot: '\n' + printables + DEL + C0 fallback (0x01..0x1f).
+constexpr int num_delimiter_candidates = first_c0_index + last_c0;
```
firestarman: NIT:
The C0 fallback math would be clearer if it used explicit `first_c0` and `num_c0` constants instead of relying on `last_c0` as both the highest byte value and the candidate count. That would make the `0x01..0x1f` range explicit in both `num_delimiter_candidates` and the candidate mapping.

src/main/cpp/src/json_utils.cu
```diff
@@ -138,59 +177,57 @@ std::tuple<std::unique_ptr<rmm::device_buffer>, char, std::unique_ptr<cudf::colu
...
-                                      num_chars,
-                                      stream.value());
-
-  auto const it             = thrust::make_counting_iterator(0);
-  auto const zero_level_idx = -lower_level;  // the bin storing count for character `\0`
-  auto const zero_level_it  = it + zero_level_idx;
-  auto const end            = it + num_levels;
-
-  auto const first_zero_count_pos =
+  CUDF_CUDA_TRY(cub::DeviceHistogram::HistogramEven(d_temp.data(),
```
firestarman: [SHOULD FIX] Can you scope the CUB histogram temporary buffer to just the execution call? `d_temp` is only needed for the second `cub::DeviceHistogram::HistogramEven`, but it remains alive through delimiter search, mask construction, and `join_strings`. Wrapping the allocation and execution call in a local block releases that scratch allocation before later device allocations and lowers peak memory.

src/main/cpp/src/json_utils.cu
```diff
@@ -138,59 +177,57 @@ std::tuple<std::unique_ptr<rmm::device_buffer>, char, std::unique_ptr<cudf::colu
...
-  // This should never happen since the input should never cover the entire char range.
-  if (first_zero_count_pos == end) {
+  if (first_available == candidates_end) {
     throw std::logic_error(
-      "Cannot find any character suitable as delimiter during joining json strings.");
+      "Cannot find an unused cuDF-supported ASCII delimiter while joining JSON strings.");
   }
   auto const delimiter =
-    static_cast<char>(cuda::std::distance(zero_level_it, first_zero_count_pos));
+    static_cast<char>(delimiter_candidate(cuda::std::distance(candidates_begin, first_available)));
```
firestarman, line 266: [SHOULD FIX] Please add the direct libcu++ iterator header for the new `cuda::std::distance` use. This file currently includes `<cuda/std/functional>` and `<cuda/std/tuple>` but not `<cuda/std/iterator>`, so the new symbol depends on a transitive include.

src/main/cpp/tests/from_json.cpp
firestarman, line 114: [SHOULD FIX] Please add direct standard headers for the new symbols used by these tests: `std::get` needs `<tuple>`, and `std::logic_error` needs `<stdexcept>`. The current include block uses both symbols via transitive includes.

firestarman (commented):
Posted inline findings from nt-pr-review --quick.

src/main/cpp/src/json_utils.cu
```diff
@@ -138,59 +177,57 @@ std::tuple<std::unique_ptr<rmm::device_buffer>, char, std::unique_ptr<cudf::colu
...
+  auto constexpr lower_level = 0;
+  auto constexpr upper_level = 128;
   auto const num_chars       = input.chars_size(stream);
 
-  rmm::device_uvector<uint32_t> histogram(num_levels, stream, default_mr);
+  rmm::device_uvector<uint32_t> histogram(num_bins, stream, default_mr);
   thrust::uninitialized_fill(
     rmm::exec_policy_nosync(stream), histogram.begin(), histogram.end(), 0);
 
+  auto const byte_samples   = reinterpret_cast<std::uint8_t const*>(input.chars_begin(stream));
```
res-life, line 198: [MUST FIX] Can you exclude rows already marked invalid from the delimiter histogram? When `nullify_invalid_rows` is true, a non-object row is marked invalid above and is later replaced with `"{}"` by `join_strings`, but this histogram still scans that row’s bytes in the original chars buffer. An invalid row containing bytes `0x00..0x7f` can therefore make every delimiter candidate appear occupied and throw, even though none of those bytes reach the JSON parser. Please build the histogram from rows selected by `is_valid_input` (or retry with a row-aware histogram when the raw-buffer histogram is exhausted), and add a regression test with a valid object plus an invalid row containing all ASCII bytes.

nartal1:
@firestarman @res-life  Addressed review comments. PTAL.

nartal1:
build

firestarman (approved):
Verified latest b80ed87 addresses my review comments. Code review LGTM. Note: blossom-ci is still reporting failure and needs to be rerun/pass before merge.

firestarman:
LGTM, premerge has failed.

nartal1:
build

PR #4809, opened 2026-07-09
Fix ORC base-offset borrow

res-life:
Contributes to https://github.com/NVIDIA/cudf-spark/issues/13437
Related to rapidsai/cudf#21993.

### Problem to fix
CPU and GPU have 1 second diff when reading timestamps near epoch(1970-01-01 00:00:00)  when writer is non-UTC timezone(e.g.: Shanghai). 

### Root cause
In order to follow Java behaviors, cudf-spark-jni has own impementations do the timezone conversions,  jni is using  `ignoreTimezoneInStripeFooter=true` to read ORC, all the timestamps are recognized as in UTC timezone, although they are in Shanghai timezone. So JNI must adjust for some extra logic in ORC: borrow logic.

#### ORC borrow logic, aka **non-negative** nanos
ORC stores timestamps in seconds plus a **non-negative** nanos field.  So for a negative nano value, MU
ST borrow a second and make nano positive. E.g.:  value -1ns, the second part is 0, the nano part is -1. After apply non-negative nano, the second part is -1, and the nano part is 999,999,999ns.

When writer timezone is Shanghai,  for the timestamps near epoch, using UTC timezone get positive value, while using Shanghai get negative value, becuase of time diff between UTC and Shanghai is 8 hours.

[borrow logic link](https://github.com/apache/orc/blob/rel/release-1.9.1/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L1284-L1286)

JNI only implements `SerializationUtils.convertBetweenTimezones`, lack the borrow logic.

Please refer to the following detail example
<details>
<summary>a detail example</summary>

  Assume the ORC timestamp stores:

  ORC stored seconds = -1'420'049'313
  ORC stored nanos   =    883'873'000 ns
                      =    883'873 us

  ORC timestamp semantics are:

  timestamp = ORC base timestamp + stored seconds + stored nanos

  The key point is: the ORC base timestamp is 2015-01-01 00:00:00 in the writer timezone.

  Here the writer timezone is Asia/Shanghai.

  ### 1. Two ORC base timestamps

  If cuDF ignores the stripe footer timezone, it decodes using the UTC ORC base:

  UTC ORC base:
  2015-01-01 00:00:00 UTC
  = 1'420'070'400'000'000 us

  But Apache ORC / Spark semantics use the writer timezone ORC base:

  Asia/Shanghai ORC base:
  2015-01-01 00:00:00 Asia/Shanghai
  = 2014-12-31 16:00:00 UTC
  = 1'420'041'600'000'000 us

  The difference is:

  base_offset_us
  = UTC ORC base - Shanghai ORC base
  = 28'800'000'000 us
  = 8 hours

  ### 2. cuDF decodes with the UTC ORC base

  decoded_us
  = UTC_ORC_BASE_US + stored_seconds * 1'000'000 + stored_nanos_us

  = 1'420'070'400'000'000
    + (-1'420'049'313 * 1'000'000)
    + 883'873

  = 21'087'883'873 us

  This is positive, so cuDF does not apply the negative timestamp borrow.

  ### 3. Correct semantics switch back to the writer timezone ORC base

  Equivalent correction:

  plain_adjusted_us = decoded_us - base_offset_us

  = 21'087'883'873 - 28'800'000'000
  = -7'712'116'127 us

  This value is now negative.

  ### 4. Recompute the Apache ORC borrow

  The fractional part is:

  floor_mod(-7'712'116'127, 1'000'000)
  = 883'873 us

  Because:

  883'873 us >= 1'000 us

  Apache ORC applies the 1-second negative timestamp borrow:

  result_us
  = -7'712'116'127 - 1'000'000
  = -7'713'116'127 us

  ### 5. Wrong vs correct result

  plain offset result        = -7'712'116'127
  Apache-compatible result   = -7'713'116'127
  difference                 = -1'000'000 us
                             = -1 second

  The bug is that cuDF made the borrow decision in the UTC ORC base frame. After applying the writer timezone base offset, the value is in the writer ORC base frame and the borrow decision changes.

</details>

### Fix
Implement [borrow logic link](https://github.com/apache/orc/blob/rel/release-1.9.1/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L1284-L1286)

### DIFF
--- a/src/main/cpp/src/timezones.hpp
+++ b/src/main/cpp/src/timezones.hpp
@@ -151,9 +151,11 @@ struct orc_tz_side {
  * recurring DST rules derived from JVM timezone APIs for dates beyond the table.
  *
  * @param input The input timestamp column in microseconds.
- * @param base_offset_us Fixed microsecond offset to apply before timezone conversion.
- *        Fuses ORC's base-timestamp adjustment (writer TZ offset at 2015-01-01) into
- *        the kernel, eliminating a separate pass. Pass 0 for no adjustment.
+ * @param writer_2015_year_base_offset_us Writer timezone offset at ORC's 2015-01-01 base instant.
+ *        ORC applies this base-timestamp adjustment, including any DST savings in effect at that
+ *        instant, before running SerializationUtils.convertBetweenTimezones. The native path
+ *        applies the same adjustment first so the negative-timestamp nanos borrow is decided in
+ *        the same frame as Apache ORC. Pass 0 for no adjustment.
  * @param writer writer timezone transition data, offsets, and DST rule.
  * @param reader reader timezone transition data, offsets, and DST rule.
  * @param stream CUDA stream.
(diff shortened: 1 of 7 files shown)


res-life:
build

res-life:
build

res-life:
build

src/main/cpp/src/timezones.hpp
```diff
@@ -151,9 +151,11 @@ struct orc_tz_side {
  * recurring DST rules derived from JVM timezone APIs for dates beyond the table.
  *
  * @param input The input timestamp column in microseconds.
- * @param base_offset_us Fixed microsecond offset to apply before timezone conversion.
- *        Fuses ORC's base-timestamp adjustment (writer TZ offset at 2015-01-01) into
- *        the kernel, eliminating a separate pass. Pass 0 for no adjustment.
+ * @param writer_2015_year_base_offset_us Writer timezone offset at ORC's 2015-01-01 base instant.
```
thirtiseven, line 154: Could we rename this consistently to `writer_tz_offset_at_orc_epoch_us` (`writerTzOffsetAtOrcEpochUs` in Java)? The value is the UTC offset of the writer timezone at the ORC base timestamp; `writer_2015_year_base_offset_us` reads as though it describes the year itself. It would also help to document that positive values are east of UTC and are subtracted from the cuDF-decoded timestamp.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -591,11 +607,17 @@ public static ColumnVector convertOrcTimezones(
...
+      // including any DST offset in effect at that instant, then applies the negative nanos borrow.
+      // The native path applies this adjustment before matching ORC's writer/reader timezone
+      // conversion.
+      long writer2015YearBaseOffsetUs = TimeUnit.MILLISECONDS.toMicros(
+          getOrc2015YearBaseOffsetMillis(writerTimezone, writerTzInfo));
       return new ColumnVector(convertOrcTimezones(
           input.getNativeView(),
           writerTzInfoTable != null ? writerTzInfoTable.getNativeView() : 0L,
           writerTzInfo.rawOffset,
+          writer2015YearBaseOffsetUs,
```
thirtiseven, line 620: Could we clarify how this reaches the production ORC scan path and ensure the same-timezone case does not bypass it? In current cudf-spark main, the scan still uses `GpuOrcTimezoneUtils.rebaseTimeZone`, and companion PR NVIDIA/cudf-spark#14544 keeps a same-rules fast path that only subtracts the base offset. That path would skip this borrow recomputation for the exact Shanghai writer=reader reproducer. Please link or update the consumer and cover this case with an end-to-end ORC-file regression.

res-life: Good catch, Updated #14544.

src/main/cpp/src/timezones.hpp
res-life: “Epoch” can be confused with the Unix epoch at 1970-01-01, while this value specifically represents the  offset from 2015-01-01 base timestamp in the write timezone. 2015 makes it more clear.

res-life:
build

thirtiseven (dismissed):
LGTM

thirtiseven (dismissed):

res-life:
build

thirtiseven (approved):

PR #4811, opened 2026-07-09
Fix includes of cudf-spark generate_input.hpp

davidwendt:
Fixes the include statements for the 'generate_input.hpp'.
There is a duplicate name in the libcudf library and it was being picked up by these includes and then failing to compile.
The fix here ensures the correct is included.

Example failure: https://github.com/rapidsai/cudf/actions/runs/29039260419/job/86192283985?pr=23199

Looks like cudf-spark-jni has an older copy of the generate_input.h/.cu files so this PR copies the latest from that repo.

### DIFF
--- a/src/main/cpp/benchmarks/common/generate_input.cu
+++ b/src/main/cpp/benchmarks/common/generate_input.cu
@@ -1117,7 +1117,7 @@ std::pair<rmm::device_buffer, cudf::size_type> create_random_null_mask(
 
 std::unique_ptr<cudf::column> create_ascii_string_column(data_profile const& profile,
                                                          cudf::size_type num_rows,
-                                                         unsigned seed = 1)
+                                                         unsigned seed)
 {
   auto engine = deterministic_engine(seed);
   return create_random_utf8_string_column<string_encoding::ASCII>(profile, engine, num_rows);
--- a/src/main/cpp/benchmarks/common/generate_input.cu
+++ b/src/main/cpp/benchmarks/common/generate_input.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2026, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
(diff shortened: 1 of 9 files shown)


davidwendt:
Greptile's flowchart here is something else.

mythrocks:
Build

mythrocks:
Build

mythrocks:
This looks good to me.  Awaiting CI pass.

src/main/cpp/benchmarks/common/generate_input.cu
```diff
@@ -850,14 +1107,22 @@ std::pair<rmm::device_buffer, cudf::size_type> create_random_null_mask(
...
                                   bool_generator{seed, 1.0 - *null_probability},
                                   cudf::get_default_stream(),
-                                  rmm::mr::get_current_device_resource_ref());
+                                  cudf::get_current_device_resource_ref());
   }
 }
 
+std::unique_ptr<cudf::column> create_ascii_string_column(data_profile const& profile,
+                                                         cudf::size_type num_rows,
+                                                         unsigned seed = 1)
```
mythrocks: Greptile found one that slipped past me:

This function is defined in the header as:
```c++
std::unique_ptr<cudf::column> create_ascii_string_column(cudf::size_type num_rows,
                                                         data_profile const& profile,
                                                         unsigned seed = 1);
```

Notice how the parameter order is not in sync.

mythrocks (changes requested):
Minor nitpick, but looks good to go otherwise.

src/main/cpp/benchmarks/common/generate_input.cu
davidwendt: Ok, good catch. I'll put up a fix by Monday.

davidwendt: @mythrocks Should be fixed now if you can kick this off again.

mythrocks:
Build

mythrocks (approved):

mythrocks:
Thank you for the fix, @davidwendt.  I've merged this change now.

PR #4812, opened 2026-07-10
[Part 4] [ORC-timezone]: Wire recurring DST metadata through JNI

res-life:
## Context

Part 4 of the split of #4432.

Merged prerequisites:
- Part 1: #4539
- Part 2: #4635
- Part 3: #4733

Additional prerequisite:
- ORC 2015 base-offset borrow correction: #4809

The end-to-end cudf-spark consumer remains tracked by NVIDIA/cudf-spark#14544.

## Changes

- Extend runtime `OrcTimezoneInfo` metadata with the historical initial offset and the recurring
  `OrcDstRuleExtractor.DstRule` produced by Part 2.
- Add a reusable `GpuTimeZoneDB.OrcTimezoneContext` that owns writer/reader transition tables on
  the GPU and stores the writer timezone offset at ORC's 2015 base timestamp.
- Serialize the 13 recurring-rule fields in a stable Java/JNI order.
- Add `convertOrcTimezonesWithRules` JNI plumbing that invokes the DST-capable native API added in
  Part 3.
- Add focused metadata tests for fixed, historical non-DST, and recurring-DST zones.

## Behavior

Production behavior is unchanged in this part. The existing public conversion path retains its DST
gate and continues to use the existing non-DST JNI entry point. The new full-path context API is
staged for Part 5.

## Verification

- Repository-pinned clang-format 20.1.8
- `git diff --check`
- Pre-commit hooks
- Build and tests were not run locally; CI will run them.

Signed-off-by: Chong Gao <chongg@nvidia.com>

PR #4733, opened 2026-06-23
[Part 3] [ORC-timezone]: GPU kernel for DST-aware timezone conversion

res-life:
## Context

Part 3 of the split of #4432 (full ORC DST GPU support). Part 1 (#4539, runtime build of `OrcTimezoneInfo`) and Part 2 (#4635, Java-side DST rule extraction) are merged. This PR adds the **CUDA implementation** that converts ORC writer/reader timestamps with Daylight Saving Time support. It does **not** wire DST into the production dispatch path — production behavior is unchanged.

## What this adds

In `timezones.hpp`:
- `struct dst_rule` — a SimpleTimeZone-shaped DST descriptor (raw/dst offsets, start/end month/day/dayOfWeek/time/mode) consumed by the GPU kernel. Field layout mirrors the Java `OrcDstRuleExtractor.DstRule` (0-based month; dayOfWeek 1=Sun..7=Sat; mode 0=DOM/1=DOW_IN_MONTH/2=DOW_GE_DOM/3=DOW_LE_DOM; timeMode 0=WALL/1=STANDARD/2=UTC).
- A DST-capable `convert_orc_writer_reader_timezones` overload (adds `base_offset_us`, per-side `initial_offset` and `dst_rule`).

In `timezones.cu`:
- Device civil-calendar / DST-offset helpers: `is_leap_year`, `days_in_month`, `days_before_month`, `days_from_epoch_to_year`, `compute_rule_day`, `compute_transition_utc_ms`, `millis_to_year`, `compute_dst_offset`, `get_transition_index`, `is_fixed_offset_tz`. These implement `SimpleTimeZone.getOffset()` semantics for dates beyond the historical transition table.
- A shared-memory conversion kernel and its host launcher.
- A **backward-compatible 5-arg overload** of `convert_orc_writer_reader_timezones` that forwards to the new path with `has_dst=false` and `initial_offset=raw_offset`. This reproduces the legacy non-DST behavior exactly (the only old/new divergence is the before-first-transition branch, which the mapping pins), so existing callers — the JNI binding and C++ tests — are unaffected.
- NVTX instrumentation and CUDA launch error checking on the shared launcher path.

In `tests/timezones.cpp`:
- `ConvertOrcTimezonesReaderDstBeyondTable` — fixed-UTC writer, US-style DST reader with no transition table; a winter instant is unchanged, a summer instant shifts back one hour.
[...]

PR #4813, opened 2026-07-10
[Part 5] [ORC-timezone]: Enable DST-aware ORC timezone conversion

res-life:
## Context

Part 5 of the split of #4432 and the only split part that enables recurring-DST conversion in the
existing public dispatch path.

Prerequisites:
- Part 1: #4539
- Part 2: #4635
- Part 3: #4733
- ORC 2015 base-offset borrow correction: #4809
- Part 4: #4812

The end-to-end cudf-spark consumer remains tracked by NVIDIA/cudf-spark#14544.

## Changes

- Remove the Java DST gate from `GpuTimeZoneDB.convertOrcTimezones`.
- Route fixed, historical, and recurring-DST zones through the reusable full-path timezone context.
- Remove the superseded raw JNI entry point.
- Enable DST writer/reader pairs in the deterministic Java conversion matrix.
- Add a focused year-9999 test for northern- and southern-hemisphere recurring-rule fallback.
- Keep the ORC 2015 base-offset borrow regression from #4809 active for same-timezone conversion.

## Behavior

This part changes runtime behavior: ORC writer/reader timezone conversion now supports recurring DST
rules instead of rejecting DST zones.

## Verification

- Repository-pinned clang-format 20.1.8
- `git diff --check`
- Pre-commit hooks
- Build and tests were not run locally; CI will run them.

### DIFF
--- a/src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
@@ -693,14 +693,6 @@ public class GpuTimeZoneDB {
   }
 
   /**
-   * Does the given timezone have Daylight Saving Time(DST) rules.
-   */
-  private static boolean hasDaylightSavingTime(String timezoneId) {
-    ZoneId zoneId = ZoneId.of(timezoneId, ZoneId.SHORT_IDS);
-    return !zoneId.getRules().getTransitionRules().isEmpty();
-  }
-
-  /**
    * Convert timestamps between writer/reader timezones for ORC reading.
    * Similar to Apache ORC, this first reconstructs the timestamp from ORC's
    * writer-timezone 2015 base instant and applies the negative nanos borrow,
@@ -721,33 +713,9 @@ public class GpuTimeZoneDB {
       ColumnVector input,
       String writerTimezone,
       String readerTimezone) {
-    // Does not support DST timezone now, just throw exception.
-    if (hasDaylightSavingTime(writerTimezone) ||
-        hasDaylightSavingTime(readerTimezone)) {
-      throw new UnsupportedOperationException("Daylight Saving Time is not supported now.");
-    }
-
-    // get timezone info from `java.util.TimeZone`
-    OrcTimezoneInfo writerTzInfo = OrcTimezoneInfo.get(writerTimezone);
-    OrcTimezoneInfo readerTzInfo = OrcTimezoneInfo.get(readerTimezone);
-    try (Table writerTzInfoTable = getTableForUtilTZ(writerTzInfo);
-        Table readerTzInfoTable = getTableForUtilTZ(readerTzInfo)) {
-
-      // ORC first reconstructs the timestamp using the writer timezone's 2015 base instant,
-      // including any DST offset in effect at that instant, then applies the negative nanos borrow.
-      // The native path applies this adjustment before matching ORC's writer/reader timezone
-      // conversion.
-      long writer2015YearBaseOffsetUs = TimeUnit.MILLISECONDS.toMicros(
-          getOrc2015YearBaseOffsetMillis(writerTimezone, writerTzInfo));
-      return new ColumnVector(convertOrcTimezones(
-          input.getNativeView(),
-          writerTzInfoTable != null ? writerTzInfoTable.getNativeView() : 0L,
-          writerTzInfo.rawOffset,
-          writer2015YearBaseOffsetUs,
-          readerTzInfoTable != null ? readerTzInfoTable.getNativeView() : 0L,
-          readerTzInfo.rawOffset));
-    } catch (Exception e) {
-      throw new IllegalStateException("convert between timezones failed!", e);
+    try (OrcTimezoneContext context =
+        buildOrcTimezoneContext(writerTimezone, readerTimezone)) {
+      return convertOrcTimezones(input, context);
     }
   }
 
@@ -758,14 +726,6 @@ public class GpuTimeZoneDB {
   private static native long convertTimestampColumnToUTCWithTzCv(
       long input_seconds, long input_microseconds, long invalid, long tzType,
       long tzOffset, long timezoneInfo, long tzIndex);
-
-  private static native long convertOrcTimezones(
-      long input,
-      long writerTzInfoTable,
-      int writerTzRawOffset,
-      long writer2015YearBaseOffsetUs,
-      long readerTzInfoTable,
-      int readerTzRawOffset);
 
   private static native long convertOrcTimezonesWithRules(
       long input,
--- a/src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
@@ -212,6 +212,7 @@ class OrcTimezoneInfo {
     List<Integer> offsets = new ArrayList<>();
     long scanCursor = MIN_SUPPORTED_ORC_UTC_MILLIS;
     int currentOffset = getInitialOffset(tz);
+    boolean hasPreviousCandidate = false;
 
     for (ZoneOffsetTransition transition : transitionList) {
       long transitionMs = transition.getInstant().toEpochMilli();
@@ -221,18 +222,19 @@ class OrcTimezoneInfo {
 
       long beforeTransitionMs = transitionMs - 1;
       int offsetBeforeTransition = tz.getOffset(beforeTransitionMs);
-      // Invariant: between two consecutive entries returned by
-      // ZoneRules.getTransitions(), the wall offset is constant — no hidden
-      // paired round-trips (e.g. A->B->A) net to zero between entries. If
-      // that ever breaks (DST zones, future tzdata revisions), the guard
-      // below will not fire and both transitions in the pair will be
-      // silently dropped. ZoneRules enumerates every transition for this zone,
-      // while the TimeZone gap scan below reconciles any offset differences
-      // between the two APIs using a six-hour base probe that is shorter than
-      // the minimum spacing between real IANA transitions.
-      if (beforeTransitionMs >= scanCursor && offsetBeforeTransition != currentOffset) {
-        currentOffset = collectTimeZoneTransitionsByScanning(
-            tz, scanCursor, beforeTransitionMs, currentOffset, transitions, offsets);
+      if (beforeTransitionMs >= scanCursor) {
+        if (hasPreviousCandidate) {
+          // Reconcile every interval between ZoneRules candidates. Endpoints with the same
+          // offset do not prove that the interval is transition-free: TimeZone may contain an
+          // A -> B -> A round trip that ZoneRules does not expose.
+          currentOffset = collectTimeZoneTransitionsByScanning(
+              tz, scanCursor, beforeTransitionMs, currentOffset, transitions, offsets);
+        } else if (offsetBeforeTransition != currentOffset) {
+          // Keep the year-0001-to-first-candidate path logarithmic. ZoneRules has no candidate
+          // in this interval, which is typically about 1,900 years long.
+          currentOffset = collectInitialTimeZoneTransitions(
+              tz, scanCursor, beforeTransitionMs, currentOffset, transitions, offsets);
+        }
       }
 
       int offsetAtTransition = tz.getOffset(transitionMs);
@@ -241,7 +243,19 @@ class OrcTimezoneInfo {
         offsets.add(offsetAtTransition);
         currentOffset = offsetAtTransition;
       }
-      scanCursor = transitionMs;
+
+      // Some JDK tzdata versions expose a ZoneRules candidate that TimeZone observes for only
+      // one millisecond. Sampling T+1 catches that immediate return and anchors the following
+      // bounded scan with the correct running offset.
+      long afterTransitionMs = transitionMs + 1;
+      int offsetAfterTransition = tz.getOffset(afterTransitionMs);
+      if (offsetAfterTransition != currentOffset) {
+        transitions.add(afterTransitionMs);
+        offsets.add(offsetAfterTransition);
+        currentOffset = offsetAfterTransition;
+      }
+      scanCursor = afterTransitionMs;
+      hasPreviousCandidate = true;
     }
 
     if (transitions.isEmpty()) {
@@ -250,7 +264,7 @@ class OrcTimezoneInfo {
     return new HistoricalTransitions(toLongArray(transitions), toIntArray(offsets));
   }
 
-  private static int collectTimeZoneTransitionsByScanning(
+  static int collectTimeZoneTransitionsByScanning(
       TimeZone tz,
       long scanStartMs,
       long scanEndMs,
@@ -260,15 +274,34 @@ class OrcTimezoneInfo {
     long cursor = scanStartMs;
     int currentOffset = startOffset;
     while (cursor < scanEndMs) {
-      // Exponentially expand the probe step while the offset stays equal to
-      // currentOffset. This collapses long no-transition stretches (e.g. the
-      // year-0001-to-first-historical-transition gap, ~1880 years for typical
-      // IANA zones) from O(N) day probes to O(log N). Once the probe lands on
-      // a different offset, the [lo, hi] bracket contains a transition and we
-      // hand it to binarySearchTransition. The bracket may be wider than the
-      // base 6h step, so this assumes at most one offset transition lives in
-      // the expanded window — which holds for real IANA data; A->B->A pairs
-      // narrower than the base step are addressed separately by the step size.
+      long lo = cursor;
+      long hi = Math.min(lo + HISTORICAL_TRANSITION_SCAN_STEP_MILLIS, scanEndMs);
+      int hiOffset = tz.getOffset(hi);
+      if (hiOffset == currentOffset) {
+        cursor = hi;
+        continue;
+      }
+
+      long exactTransition = binarySearchTransition(tz, lo, hi);
+      int offsetAfterTransition = tz.getOffset(exactTransition);
+      transitions.add(exactTransition);
+      offsets.add(offsetAfterTransition);
+      currentOffset = offsetAfterTransition;
+      cursor = exactTransition;
+    }
+    return currentOffset;
+  }
+
+  private static int collectInitialTimeZoneTransitions(
+      TimeZone tz,
+      long scanStartMs,
+      long scanEndMs,
+      int startOffset,
+      List<Long> transitions,
+      List<Integer> offsets) {
+    long cursor = scanStartMs;
+    int currentOffset = startOffset;
+    while (cursor < scanEndMs) {
       long lo = cursor;
       long step = HISTORICAL_TRANSITION_SCAN_STEP_MILLIS;
       long hi = Math.min(lo + step, scanEndMs);
@@ -280,7 +313,6 @@ class OrcTimezoneInfo {
         hiOffset = tz.getOffset(hi);
       }
       if (hiOffset == currentOffset) {
-        // Reached scanEndMs without seeing any transition.
         cursor = hi;
         continue;
       }
--- a/src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
+++ b/src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
@@ -226,9 +226,10 @@ class OrcTimezoneInfo {
       // paired round-trips (e.g. A->B->A) net to zero between entries. If
       // that ever breaks (DST zones, future tzdata revisions), the guard
       // below will not fire and both transitions in the pair will be
-      // silently dropped. The DST guard in
-      // GpuTimeZoneDB.convertOrcTimezones currently keeps this dormant;
-      // any follow-up that relaxes it must revisit this code.
+      // silently dropped. ZoneRules enumerates every transition for this zone,
+      // while the TimeZone gap scan below reconciles any offset differences
+      // between the two APIs using a six-hour base probe that is shorter than
+      // the minimum spacing between real IANA transitions.
       if (beforeTransitionMs >= scanCursor && offsetBeforeTransition != currentOffset) {
         currentOffset = collectTimeZoneTransitionsByScanning(
             tz, scanCursor, beforeTransitionMs, currentOffset, transitions, offsets);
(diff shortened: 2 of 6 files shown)


res-life:
Closing this draft because a fork-local stacked PR is required to use the Part 4 head branch as its actual GitHub base. Replaced by https://github.com/res-life/spark-rapids-jni/pull/12.

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
```diff
@@ -226,9 +226,10 @@ private static HistoricalTransitions buildHistoricalTransitions(
       // paired round-trips (e.g. A->B->A) net to zero between entries. If
       // that ever breaks (DST zones, future tzdata revisions), the guard
       // below will not fire and both transitions in the pair will be
-      // silently dropped. The DST guard in
-      // GpuTimeZoneDB.convertOrcTimezones currently keeps this dormant;
-      // any follow-up that relaxes it must revisit this code.
+      // silently dropped. ZoneRules enumerates every transition for this zone,
```
firestarman: **SHOULD FIX — add a deterministic scanner regression test**

The six-hour, endpoint-based scan can miss a complete `A -> B -> A` round trip when both probe endpoints have the same offset. No test directly exercises `collectTimeZoneTransitionsByScanning`, and the broad conversion matrix only uses JVM-provided zones, so it cannot force this reconciliation edge case.

Please add a synthetic `TimeZone` regression test whose offset changes `A -> B -> A` between equal-offset endpoints and assert that both transition instants and offsets are retained. Making the collector package-private for this test would keep the behavior deterministic.

src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
```diff
@@ -226,9 +226,10 @@ private static HistoricalTransitions buildHistoricalTransitions(
...
+      // while the TimeZone gap scan below reconciles any offset differences
```
firestarman: **SHOULD FIX — this comment overstates what the scan reconciles**

The text says the scan reconciles "any offset differences", but the guard below only invokes it when `offsetBeforeTransition != currentOffset`. A paired `A -> B -> A` change returns to `currentOffset`, so the endpoints compare equal and the scan does not run.

Either fix the guard so every interval is scanned, or document this limitation explicitly; the current wording contradicts the adjacent control flow.

src/test/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDBTest.java
```diff
@@ -208,4 +195,33 @@ void testConvertOrcTimezones() {
...
+    GpuTimeZoneDB.cacheDatabase();
+    GpuTimeZoneDB.verifyDatabaseCached();
+
+    long[] microseconds = {
+        LocalDateTime.of(9999, 1, 15, 12, 0, 0)
+            .toEpochSecond(ZoneOffset.UTC) * TimeUnit.SECONDS.toMicros(1),
+        LocalDateTime.of(9999, 7, 15, 12, 0, 0)
+            .toEpochSecond(ZoneOffset.UTC) * TimeUnit.SECONDS.toMicros(1)
+    };
+    String[][] cases = {
```
firestarman, line 270: **SHOULD FIX — cover the exact recurring-DST boundaries**

The newly enabled matrix uses random instants, while this focused fallback test samples only January 15 and July 15. Neither checks immediately before, exactly at, and immediately after the gap/overlap transitions. The native rule path uses inclusive/exclusive boundary comparisons, so an off-by-one mutation can survive these tests.

Please derive the year-9999 transitions with `ZoneOffsetTransitionRule.createTransition(9999)` and test transition `-1 ms`, the transition instant, and `+1 ms` for Los Angeles and Sydney in both writer and reader positions.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -721,33 +713,9 @@ public static ColumnVector convertOrcTimezones(
...
-          writerTzInfoTable != null ? writerTzInfoTable.getNativeView() : 0L,
-          writerTzInfo.rawOffset,
-          writer2015YearBaseOffsetUs,
-          readerTzInfoTable != null ? readerTzInfoTable.getNativeView() : 0L,
-          readerTzInfo.rawOffset));
-    } catch (Exception e) {
-      throw new IllegalStateException("convert between timezones failed!", e);
+    try (OrcTimezoneContext context =
+        buildOrcTimezoneContext(writerTimezone, readerTimezone)) {
+      return convertOrcTimezones(input, context);
```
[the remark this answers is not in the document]
firestarman: **MUST FIX — confirmed with a concrete `Asia/Gaza` reproduction**

On the project's Java 8 target, `ZoneRules` has a transition at `2037-10-09T23:00:00Z`, while `TimeZone` reports UTC+03 one millisecond before, UTC+02 at the transition, and UTC+03 again one millisecond later. The builder records UTC+02, but the endpoint-equality guard skips the following gap scan and never records the return to UTC+03. As a result, converting writer `UTC` to reader `Asia/Gaza` at `2_139_177_600_000_000` microseconds returns a value one hour too high.

Scanning every interval between consecutive `ZoneRules` entries, even when endpoint offsets match, removed all 168 mismatches in an all-zone diagnostic. Please update the guard and add `Asia/Gaza` around 2037-10-15 as a regression case in both writer and reader positions.

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -721,33 +713,9 @@ public static ColumnVector convertOrcTimezones(
...
-      return new ColumnVector(convertOrcTimezones(
-          input.getNativeView(),
-          writerTzInfoTable != null ? writerTzInfoTable.getNativeView() : 0L,
-          writerTzInfo.rawOffset,
-          writer2015YearBaseOffsetUs,
-          readerTzInfoTable != null ? readerTzInfoTable.getNativeView() : 0L,
-          readerTzInfo.rawOffset));
-    } catch (Exception e) {
-      throw new IllegalStateException("convert between timezones failed!", e);
+    try (OrcTimezoneContext context =
```
firestarman, line 716: #### 1. [🔴 MUST FIX] Can the newly enabled all-zone dispatch at src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java:716 avoid rejecting a valid recurring-DST zone whose standard offset changed after the extractor's fixed reference date? The dispatch builds metadata for every writer/reader ID through src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java:629, which calls the rule extractor at src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:159

**Line:** 716 ([diff](https://github.com/NVIDIA/cudf-spark-jni/pull/4813/files#diff-08f7ad4a7c0bb6b12259f2f005bc3489e6785d29acc1ff8f2289c3557d6480e6R716))
**Issue:** Can the newly enabled all-zone dispatch at src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java:716 avoid rejecting a valid recurring-DST zone whose standard offset changed after the extractor's fixed reference date? The dispatch builds metadata for every writer/reader ID through src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java:629, which calls the rule extractor at src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:159. For `` `America/Scoresbysund` ``, src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java:126 compares the current raw offset (-7,200,000 ms) with the historical standard offset on 2024-01-15 (-3,600,000 ms), decides that the two APIs describe different zones, and throws. Both APIs report the same actual offset (-3,600,000 ms) at that instant, and the ZoneRules fallback at src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java:138 successfully produces the recurring rule when invoked directly. Consequently, this valid ID cannot use the DST support enabled by this change.
**Confidence:** 🟣❗ CERTAIN
**Code:**
```` ```java ````
```java
// src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java:716
try (OrcTimezoneContext context =
    buildOrcTimezoneContext(writerTimezone, readerTimezone)) {
  return convertOrcTimezones(input, context);
}
```
```` ``` ````
**Fix:** Compare the actual TimeZone and ZoneRules offsets at the same reference instant in src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java:126. That retains the mismatched-zone defense while allowing zones with a historical standard-offset change to reach the verified ZoneRules fallback at src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java:138.
**Diff:**
```` ```diff ````
```diff
--- src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java
+++ src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java
@@
     Instant ref = Instant.parse("2024-01-15T00:00:00Z");
-    int rulesStandardOffsetMillis = rules.getStandardOffset(ref).getTotalSeconds() * 1000;
-    if (tz.getRawOffset() != rulesStandardOffsetMillis) {
+    int rulesOffsetMillis = rules.getOffset(ref).getTotalSeconds() * 1000;
+    int timeZoneOffsetMillis = tz.getOffset(ref.toEpochMilli());
+    if (timeZoneOffsetMillis != rulesOffsetMillis) {
       throw new IllegalStateException(
           "TimeZone and ZoneRules describe different zones for timezone: " + timezoneId
-              + " (tz.rawOffset=" + tz.getRawOffset()
-              + ", rules.standardOffset=" + rulesStandardOffsetMillis + ")");
+              + " (tz.offset=" + timeZoneOffsetMillis
+              + ", rules.offset=" + rulesOffsetMillis + ")");
     }
```
```` ``` ````
**Reference:** Temurin 8u492 compiled the production sources src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:45 and src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java:45 with the focused diagnostic /tmp/tzdiag/com/nvidia/spark/rapids/jni/TimezoneDiag.java:24. Running `` `/tmp/temurin8/bin/java -cp /tmp/tzdiag-classes com.nvidia.spark.rapids.jni.TimezoneDiag` `` printed `` `Scoresbysund raw=-7200000 rulesStandard=-3600000 tzActual=-3600000 rulesActual=-3600000` ``, `` `Scoresbysund ZoneRules rule=true` ``, and `` `BUILD_FAILURE America/Scoresbysund ... TimeZone and ZoneRules describe different zones` ``. The concrete call chain is src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java:716 -> src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java:629 -> src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:159 -> src/main/java/com/nvidia/spark/rapids/jni/OrcDstRuleExtractor.java:126.
*(Reviewer: correctness)*

**Review by firestarman — COMMENTED** [review]
Inline review findings.

File: src/test/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDBTest.java [hunk 6]
```diff
@@ -168,8 +201,8 @@ void testConvertOrcTimezones() {
     long max = LocalDateTime.of(9999, 12, 31, 23, 59, 59)
         .toEpochSecond(ZoneOffset.UTC) * TimeUnit.SECONDS.toMicros(1);
 
-    // use today as the random seed so we get different values each day
-    Random rng = new Random(LocalDate.now().toEpochDay());
+    // Keep the DST matrix deterministic so failures are reproducible.
```
firestarman, line 204: #### 2. [🟢 NIT] Could you add a deterministic integration-level regression for the candidate-interval routing at src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:225? The new test at src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:178 calls the six-hour scanner directly, so it still passes if the `` `hasPreviousCandidate` `` branch at src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:226 is deleted or changed back to an endpoint-offset guard

**Line:** 204 ([diff](https://github.com/NVIDIA/cudf-spark-jni/pull/4813/files#diff-a914dc98eef534e8662cd079a5c80ae0c4deecbad76cecf6ca4e742201941663R204))
**Issue:** Could you add a deterministic integration-level regression for the candidate-interval routing at src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:225? The new test at src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:178 calls the six-hour scanner directly, so it still passes if the `` `hasPreviousCandidate` `` branch at src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:226 is deleted or changed back to an endpoint-offset guard. The real-zone regression at src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:227 instead locks the separate one-millisecond return path at src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:247. An integration fixture with two candidate transitions and a hidden A -> B -> A interval would ensure the performance-sensitive split between the logarithmic initial scan and bounded inter-candidate scan remains wired correctly.
**Confidence:** 🟣❗ CERTAIN
**Code:**
```` ```java ````
```java
int finalOffset = OrcTimezoneInfo.collectTimeZoneTransitionsByScanning(pairedTransitions, 0L, 24L * 3_600_000L, 0, transitions, offsets);
```
```` ``` ````
**Fix:** Expose the package-private historical builder/result to the same-package test, then pass a synthetic `` `TimeZone` `` containing a hidden round trip plus two surrounding `` `ZoneOffsetTransition` `` candidates and assert that the builder retains both hidden transitions.
**Diff:**
```` ```diff ````
```diff
|
--- src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
+++ src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java
@@
-  private static HistoricalTransitions buildHistoricalTransitions(
+  static HistoricalTransitions buildHistoricalTransitions(
       TimeZone tz,
       List<ZoneOffsetTransition> transitionList) {
@@
-  private static final class HistoricalTransitions {
+  static final class HistoricalTransitions {
--- src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java
+++ src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java
@@
+  @Test
+  void testBuilderScansPairedTransitionsBetweenEqualCandidateEndpoints() {
+    long firstHiddenTransition = 8L * 3_600_000L;
+    long secondHiddenTransition = 16L * 3_600_000L;
+    TimeZone pairedTransitions = new TimeZone() {
+      private int rawOffset;
+
+      @Override
+      public int getOffset(long date) {
+        return date >= firstHiddenTransition && date < secondHiddenTransition
+            ? 3_600_000 : 0;
+      }
+
+      @Override
+      public int getOffset(int era, int year, int month, int day,
+          int dayOfWeek, int milliseconds) {
+        return rawOffset;
+      }
+
+      @Override
+      public void setRawOffset(int offsetMillis) {
+        rawOffset = offsetMillis;
+      }
+
+      @Override
+      public int getRawOffset() {
+        return rawOffset;
+      }
+
+      @Override
+      public boolean useDaylightTime() {
+        return true;
+      }
+
+      @Override
+      public boolean inDaylightTime(Date date) {
+        return getOffset(date.getTime()) != rawOffset;
+      }
+    };
+    List<ZoneOffsetTransition> candidates = Arrays.asList(
+        ZoneOffsetTransition.of(LocalDateTime.of(1970, 1, 1, 4, 0),
+            ZoneOffset.UTC, ZoneOffset.ofHours(1)),
+        ZoneOffsetTransition.of(LocalDateTime.of(1970, 1, 1, 20, 0),
+            ZoneOffset.UTC, ZoneOffset.ofHours(1)));
+
+    OrcTimezoneInfo.HistoricalTransitions actual =
+        OrcTimezoneInfo.buildHistoricalTransitions(pairedTransitions, candidates);
+
+    assertEquals(2, actual.transitions.length);
+    assertEquals(firstHiddenTransition, actual.transitions[0]);
+    assertEquals(3_600_000, actual.offsets[0]);
+    assertEquals(secondHiddenTransition, actual.transitions[1]);
+    assertEquals(0, actual.offsets[1]);
+  }
```
```` ``` ````
**Reference:** src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:227 demonstrates the same-package integration pattern by building full metadata through src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:118 and asserting exact transition positions and offsets.
*(Reviewer: test-coverage)*

File: src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java [hunk 7]
```diff
@@ -174,6 +175,70 @@ void testGetDstZoneIncludesKernelMetadata() {
...
+    OrcTimezoneInfo info = OrcTimezoneInfo.get("Asia/Gaza");
+    long transition = ZoneOffsetTransition.of(
+        LocalDateTime.of(2037, 10, 10, 2, 0),
+        ZoneOffset.ofHours(3),
+        ZoneOffset.ofHours(2)).getInstant().toEpochMilli();
+    int index = Arrays.binarySearch(info.transitions, transition);
+
+    assertTrue(index >= 0, "expected the ZoneRules candidate in the transition table");
+    assertEquals(7_200_000, info.offsets[index]);
+    assertEquals(transition + 1, info.transitions[index + 1]);
```
firestarman: #### 3. [🔴 MUST FIX] Can this regression test derive the expected one-millisecond return from the runtime TimeZone data instead of requiring it on every JDK? src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:38 explicitly states that transition tables depend on the JVM's bundled tzdata, but src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:238 unconditionally requires an `` `Asia/Gaza` `` transition at `` `T+1` ``

**Line:** 238 ([diff](https://github.com/NVIDIA/cudf-spark-jni/pull/4813/files#diff-27cdeecbb2e48e1cd43f9ad66855e1c0f0d96f5d9d2a198277dc2875def6f623R238))
**Issue:** Can this regression test derive the expected one-millisecond return from the runtime TimeZone data instead of requiring it on every JDK? src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:38 explicitly states that transition tables depend on the JVM's bundled tzdata, but src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:238 unconditionally requires an `` `Asia/Gaza` `` transition at `` `T+1` ``. Temurin 17.0.19 keeps UTC+02 at `` `T+1` `` and places the next transition at 2,153,260,800,000 ms, so the assertion compares 2,138,742,000,001 with 2,153,260,800,000 and fails even though the generated table exactly matches that runtime's TimeZone. The same file explicitly exercises JDK 17 behavior at src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:58, so this environment-specific expected value makes the suite internally inconsistent.
**Confidence:** 🟣❗ CERTAIN
**Code:**
```` ```java ````
```java
// src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:236
assertTrue(index >= 0, "expected the ZoneRules candidate in the transition table");
assertEquals(7_200_000, info.offsets[index]);
assertEquals(transition + 1, info.transitions[index + 1]);
assertEquals(10_800_000, info.offsets[index + 1]);
```
```` ``` ````
**Fix:** Derive both offsets from the canonical TimeZone in src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:228, and assert the extra `` `T+1` `` row only when that runtime actually exposes an immediate return. Temurin 8u492 still exercises all three immediate-return assertions, while Temurin 17.0.19 validates the candidate without indexing a nonexistent `` `T+1` `` entry.
**Diff:**
```` ```diff ````
```diff
--- src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java
+++ src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java
@@
     int index = Arrays.binarySearch(info.transitions, transition);
+    TimeZone tz = TimeZone.getTimeZone(GpuTimeZoneDB.getZoneId("Asia/Gaza").getId());
+    int offsetAtTransition = tz.getOffset(transition);
+    int offsetAfterTransition = tz.getOffset(transition + 1);

     assertTrue(index >= 0, "expected the ZoneRules candidate in the transition table");
-    assertEquals(7_200_000, info.offsets[index]);
-    assertEquals(transition + 1, info.transitions[index + 1]);
-    assertEquals(10_800_000, info.offsets[index + 1]);
+    assertEquals(offsetAtTransition, info.offsets[index]);
+    if (offsetAfterTransition != offsetAtTransition) {
+      assertEquals(transition + 1, info.transitions[index + 1]);
+      assertEquals(offsetAfterTransition, info.offsets[index + 1]);
+    }
   }
```
```` ``` ````
**Reference:** Running the production table builder src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:204 through /tmp/tzdiag/com/nvidia/spark/rapids/jni/TimezoneDiag.java:24 on Temurin 17.0.19 printed `` `old table[149]=2138742000000 -> 7200000` ``, `` `old table[150]=2153260800000 -> 10800000` ``, and `` `Gaza old delta=1 TimeZone=7200000 table=7200000` ``. The same command on Temurin 8u492 printed `` `old table[150]=2138742000001 -> 10800000` ``, confirming that the changed assertion at src/test/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfoTest.java:238 is tied to one runtime's tzdata rather than the runtime-derived contract documented at src/main/java/com/nvidia/spark/rapids/jni/OrcTimezoneInfo.java:38.
*(Reviewer: correctness)*

res-life:
build

firestarman (approved):
LGTM

PR #4827, opened 2026-07-15
Fix kudo shuffle uninitialized offset for deeply nested lists

pmattione-nvidia:
Kudo GPU shuffle reassembly didn't zero-initialized list offsets, so a zero-row list/string column's lone terminal offset was left uninitialized, causing sporadic invalid global reads when the Parquet writer later dereferenced it. This fix zeroes those zero-row offset buffers alongside the validity buffers in the existing batched memset. This restores cudf's list invariant that every advertised offset entry is initialized. 

Note that the EmptyOffsets test actually tests this case already, but it is unlikely to ever catch it.  This is because the bug is an uninitialized read, which is usually zero, the correct value.  It only triggers if the memory is reused from something else previously, or if all allocated memory is initialized to some other default value besides zero (e.g. 0xab).

### DIFF
--- a/src/main/cpp/src/shuffle_assemble.cu
+++ b/src/main/cpp/src/shuffle_assemble.cu
@@ -922,8 +922,17 @@ std::pair<shuffle_assemble_result, rmm::device_uvector<assemble_batch>> assemble
   // for better memory management and reduced allocation overhead.
   std::vector<buffer_slice> buffer_slices(num_dst_buffers);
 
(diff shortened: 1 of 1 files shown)


pmattione-nvidia:
build

ttnghia (approved):

jihoonson (approved):
Tested the fix manually. The test failing previously now runs successfully for 2 hours with no error.

hyperbolic2346 (approved):

PR #4880, opened 2026-07-23
Reduce packaged native library size

gerashegalov:
## Summary
- link `libcudf.so` against the `libnvcomp.so` already packaged by the Java distribution instead of embedding a second static nvCOMP copy
- localize symbols from statically linked CUDA compiler implementation archives with `--exclude-libs`
- retain an opt-out via `HIDE_PRIVATE_CUDA_SYMBOLS=OFF`

No GPU architectures are removed.

## Dependency
Draft until rapidsai/cudf#23415 merges and the submodule revision is available from upstream cuDF.

Related to NVIDIA/cudf-spark#15145.

## Size results
CUDA 12.9.1 Release builds for `sm_75`, changing only the two settings measured below:

| Variant | `libcudf.so` | ZIP level 6 |
| --- | ---: | ---: |
| static nvCOMP, CUDA symbols exported | 711,163,904 B | 359,193,871 B |
| shared nvCOMP, CUDA symbols exported | 691,845,960 B | 344,989,786 B |
| shared nvCOMP, CUDA symbols localized | 668,512,072 B | 336,370,271 B |

Combined reduction: 40.68 MiB uncompressed and 21.77 MiB after DEFLATE. The Maven-built JAR stores `libnvcomp.so`, `libcudf.so`, and `libcudfjni.so` as DEFLATED entries.

## Symbol compatibility
The final library removes 195,864 private CUDA toolkit symbols from `.dynsym`, with no added symbols and no changes to JNI or cuDF mangled exports. The regular ELF `.symtab` is retained.

Focused crash tests retained private function names in HotSpot fatal-error logs and names/source lines in GDB. As expected for localized symbols, glibc `backtrace_symbols` reports offsets rather than private CUDA implementation names; production sources do not use that API.

## Validation
- full native Release build on CUDA 12.9.1 (`sm_75`)
- loaded `libnvcomp.so`, `libcudf.so`, and `libcudfjni.so` in Java dependency order
- `mvn package -DskipTests` in the project CUDA container
- `NativeDepsLoaderTest`: 4 passed
- `git diff --check`


gerashegalov:
Superseded by #4881, which adds the required DCO sign-off on a new branch without rewriting published history.

PR #4885, opened 2026-07-24
Support using prebuilt cudf native binary

ttnghia:
When building this repository, we also have to build `libcudf` and `libcudfjni` (native library compiled from cudf-java JNI) from `thirdparty/cudf`. This is the slowest part of every build as cudf is very heavy and building it is very time-consuming.

This PR implements the ability that allows to skip the `libcudf` and `libcudfjni` build phases. Instead, the linking process will use their prebuilt binary directly, allowing to reduce the build time significantly during development especially in the situations when we need to test many worktrees using just the same cudf commit. This feature is turned on by a new Maven option `cudf.reuse.prebuilt`:
```
mvn package -Dcudf.reuse.prebuilt=true \
    -Dcudf.path=/path/to/cudf-source \
    -Dlibcudf.install.path=/path/to/prebuilt/libcudf-install \
    -Dlibcudfjni.build.path=/path/to/prebuilt/libcudfjni
```

The default behavior is unchanged — `cudf.reuse.prebuilt` default to `false` which means always compile cudf.

To protect against dependency incompatibility issue, the prebuilt binary are carefully verified to make sure their dependency pins or ABI-affecting build options match exactly with the current build. In the situations when the users want to intentionally reuse the binary with different dependencies or build options, there is also a new option `cudf.reuse.force` added for that purpose (users proceed with their own risk).


gerashegalov (changes requested):
I think the core use case—avoiding recompilation across worktrees—is already achievable with the existing path and skip properties. For example:

```bash
mvn package \
  -Dcudf.path=/shared/cudf \
  -Dlibcudf.build.path=/seed/target/libcudf/cmake-build \
  -Dlibcudf.install.path=/seed/target/libcudf-install \
  -Dlibcudfjni.build.path=/seed/target/libcudfjni \
  -Dlibcudf.clean.skip=true \
  -Dsubmodule.check.skip=true \
  -Dsubmodule.patch.skip=true
```

With the shared CMake build trees already up to date, the libcudf/libcudfjni build commands should be no-ops; only the per-worktree spark JNI target needs rebuilding. The incremental benefit here appears to be skipping the lightweight CMake invocations and setting the existing skip flags automatically.

Could we first validate the timing difference versus this existing workflow? If the invocation is simply too cumbersome, a small shortcut script such as `build/build-with-prebuilt-cudf <seed-worktree> [maven args...]` could derive these paths and invoke Maven consistently. That would make the workflow handy while avoiding the new manifest format, nonce pairing, force mode, and roughly 300 lines of validation/cache-management logic.

ttnghia:
> I think the core use case—avoiding recompilation across worktrees—is already achievable with the existing path and skip properties. 

Unfortunately not, as the current code is incomplete. Although we can specify the path of prebuilt cudf elsewhere, the libcudfjni cmake configure is unconditional, thus every build re-runs cmake there (find_package + CPM/setup.cmake for arrow/parquet), plus the libcudf install re-scan and ninja's up-to-date walk over all cudf objects. The PR skips all of these.

I'll run a test to compare runtime difference.

build/apply-patches
```diff
@@ -18,6 +18,9 @@
 
 set -e
 
+# Prebuilt-cudf reuse never builds from the submodule; skip patching an empty/stale checkout.
+[[ "${CUDF_REUSE_PREBUILT:-false}" == "true" ]] && exit 0
```
gerashegalov, line 22: Reuse still compiles the cuDF Java sources from `${cudf.path}`, so skipping patches is unsafe now that `main` carries `0001-use-static-nvcomp-from-libcudf.patch`. With the documented workflow and a clean cuDF checkout at the correct HEAD, fingerprint validation passes, but `NativeDepsLoader` still requests `nvcomp` while the current POM no longer packages `libnvcomp.so`; native initialization can then fail at runtime. Please apply the project patches to the actual `${cudf.path}` in reuse mode, or fingerprint and require the exact patched source content.

build/buildcpp.sh
```diff
@@ -82,88 +84,252 @@ WARNING: CMAKE_CUDA_ARCHITECTURES is overridden by GPU_ARCHS.
...
+      -C="$CUDF_PIN_PATH/setup.cmake"
+  fi
+  # submodule-sync.sh phase 1 calls this script with LIBCUDF_CONFIGURE_ONLY=ON
+  if [[ $LIBCUDF_CONFIGURE_ONLY == ON ]]; then
+    echo "Skip build..."
+    exit 0
+  fi
+  echo "Building cudf native libs"
+  cmake --build "$LIBCUDF_BUILD_PATH" --target install "-j$CPP_PARALLEL_LEVEL"
+  cudf_fingerprint_stamp "$LIBCUDF_INSTALL_PATH/cudf-prebuilt-fingerprint.txt"
```
gerashegalov: This stamps the requested environment, not necessarily the configuration that produced `libcudf.a`. Above, libcudf configuration is skipped whenever an existing `CMakeCache.txt` is present. For example, a tree configured for architecture 80 can be invoked with architecture 90 without `libcudf.build.configure=true`; the old build remains in effect, but this manifest records 90 and a later reuse build for 90 passes validation. Please stamp effective build/cache metadata, reject requested-vs-effective discrepancies, or force configuration before creating the reusable manifest.

build/buildcpp.sh
```diff
@@ -82,88 +84,252 @@ WARNING: CMAKE_CUDA_ARCHITECTURES is overridden by GPU_ARCHS.
...
+  LIBDIR="$(dirname "$LIBCUDF_A")"
+  [[ -f "$LIBCUDF_A" ]] \
+    || { echo "ERROR: no prebuilt lib{64,}/libcudf.a under $LIBCUDF_INSTALL_PATH" >&2; exit 1; }
+  [[ -f "$LIBCUDFJNI_BUILD_PATH/libcudfjni.a" ]] \
+    || { echo "ERROR: no prebuilt libcudfjni.a at $LIBCUDFJNI_BUILD_PATH" >&2; exit 1; }
+  [[ -f "$LIBCUDFJNI_BUILD_PATH/_deps/arrow-build/release/libarrow.a" ]] \
+    || { echo "ERROR: incomplete libcudfjni _deps (arrow)" >&2; exit 1; }
+  # jar packaging copies libnvcomp*.so from here (pom.xml copy-native-libs) but does not fail if
+  # absent, so a trimmed prebuilt would silently ship a jar without nvcomp.
+  [[ -f "$LIBCUDFJNI_BUILD_PATH/libnvcomp.so" ]] \
```
gerashegalov: This requirement and its packaging rationale are stale after merging current `main`: the POM no longer copies any nvcomp shared libraries, and the final JNI library uses statically linked nvcomp. Requiring this unused byproduct can reject an otherwise complete trimmed prebuilt. Please remove the check unless there is another current consumer that requires this file.

ttnghia:
Some measurement:
```
 Timing difference — PR reuse (A) vs. build using existing flags (B)

┌────────────────────────────────────────────────────────────────────────────────────────────────┬────────┬────────┬───────────────┬────────────────┐
│                                            Workflow                                            │ Run 1  │ Run 2  │      Avg      │ vs. full build │
├────────────────────────────────────────────────────────────────────────────────────────────────┼────────┼────────┼───────────────┼────────────────┤
│ B — existing flags (build.path+install.path+libcudfjni.build.path+clean.skip+submodule.*.skip) │ 19.75s │ 18.00s │ 18.9s         │ —              │
├────────────────────────────────────────────────────────────────────────────────────────────────┼────────┼────────┼───────────────┼────────────────┤
│ A — PR cudf.reuse.prebuilt                                                                     │ 17.35s │ 17.08s │ 17.2s         │ —              │
├────────────────────────────────────────────────────────────────────────────────────────────────┼────────┼────────┼───────────────┼────────────────┤
│ Seed / full build (reference)                                                                  │ —      │ —      │ 1076s (17:55) │ 1×             │
└────────────────────────────────────────────────────────────────────────────────────────────────┴────────┴────────┴───────────────┴────────────────┘
```

So if using existing flags can save time similar to this PR, turning out that the advantage of this one is just to safeguard the rebuild process against skew dependencies.

build/apply-patches
ttnghia: Fixed this. Thanks.

build/buildcpp.sh
ttnghia: Fixed this in latest commit.

build/buildcpp.sh
ttnghia: Remove this in latest commit. Thanks.

gerashegalov (changes requested):
Re-review of `c0c4b6d0dc5fe32bc507e8ecc842cf155e7111a3` against `781d7d4a0f6b5fcc1ea68f2d2111be34f6cedc9a`.

The three findings from my prior pass were addressed correctly: patch presence is now checked, stale nvcomp validation was removed, and skipped libcudf configuration no longer restamps requested settings. I still see the following blockers:

1. **The guard permits source/prebuilt ABI mismatch by default.** Patch presence is checked, but a different cuDF commit or additional working-tree modifications only produce a warning. Phase 3 then compiles headers/JNI sources from that tree while linking older archives. Compatible symbol names do not guarantee compatible layouts, inline code, or semantics, so this can link successfully and fail or corrupt data at runtime. Since skew protection is now the principal value of this feature, exact source identity and expected patch state should be required by default; `cudf.reuse.force` is the appropriate escape hatch.

2. **A successful retry can produce an unstamped, unusable seed.** The libcudf manifest is written only when configuration occurred in the same invocation. If configure succeeds but compilation is interrupted, the retry sees `CMakeCache.txt`, skips configure, successfully builds/installs, and never writes a manifest. The explicit configure-only path has the same outcome. Reuse then rejects the valid artifact as unstamped. Persist configuration metadata and finalize it after a successful build, or reconstruct the fingerprint from effective cache metadata when a successful install lacks a manifest.

3. **Identical CUDA toolkits can compare unequal.** `CUDA_VERSION` is recorded verbatim (the repository commonly uses values such as `12.9.1`), while the `nvcc` fallback records `12.9`. Seed and consumer builds using the same toolkit through different invocation paths can therefore be falsely rejected. Normalize both inputs to one representation.

**Worthwhileness:** the posted measurements show 1076s for a full build, 18.9s using existing properties, and 17.2s with this PR. Existing functionality already removes about 98.2% of the build time; this PR saves another 1.7s while adding 374 lines, a persistent manifest/nonce protocol on normal builds, new failure modes, and no retained reuse test suite. A small `build/build-with-prebuilt-cudf <seed-worktree> [maven args...]` wrapper around the existing properties would deliver nearly identical performance and much better ergonomics with substantially less maintenance. If automated safety is desired, a focused standalone verifier could require exact revision, patch state, and effective CMake-cache values.

nvauto:

PR #4896, opened 2026-07-27
Fix profile converter crash on markers with missing strings

### LINKED ISSUE #4895: [BUG] Profile converter crashes on markers with missing name or domain
## Description

`spark_rapids_profile_converter` dereferences optional FlatBuffers strings without checking whether they are present. A RAPIDS 26.08 self-profiler capture containing an NVTX marker without a domain crashes the current converter with `SIGSEGV` when using the documented direct NVTXW conversion path.

The profile schema does not require `MarkerActivity.name` or `MarkerActivity.domain`, and `profiler_serializer.cpp` intentionally omits either field when the corresponding CUPTI pointer is null.

## Reproduction

Converter built from current `main`:

```text
spark_rapids_profile_converter 26.10.00 f96622fd097a9b9d7cf9822cbac4f613a89999a6
```

Command:

```bash
spark_rapids_profile_converter \
  -w \
  --nvtxw-backend=/opt/nvidia/nsight-systems/2025.2.1/host-linux-x64/libNvtxwBackend.so \
  -o output.nsys-rep \
  rapids-profile.bin
```

Result:

```text
Backend implementation loaded!  Applying config string...
Loader config key/value pairs not provided
Creating report: "output.nsys-rep"
Segmentation fault (exit code 139)
```

A RelWithDebInfo build under GDB identifies the dereference in `convert_to_nvtxw`:

```text
Program received signal SIGSEGV
spark_rapids_profile_converter.cpp:974
    m->domain()->str()
```

The same unchecked pattern exists for `m->name()` in both the NVTXT and NVTXW paths.

## Expected behavior

Missing marker names and domains should be accepted and represented as empty strings. Conversion should not crash on a profile emitted by `profiler_serializer`.

## Proposed fix

Use an empty string when either optional field is absent:

```cpp
m->name() ? m->name()->str() : ""
m->domain() ? m->domain()->str() : ""
```

nvliyuan:
## Summary

- treat absent NVTX marker names as empty strings during NVTXT conversion
- treat absent marker names and domains as empty strings during direct NVTXW conversion
- prevent valid profiler output from crashing the offline converter

Fixes #4895.

## Test plan

- [x] `git diff --check`
- [x] built `spark_rapids_profile_converter` from current `main`
- [x] reproduced the pre-fix SIGSEGV under GDB at the unchecked `m->domain()->str()` dereference
- [x] rebuilt with the fix and processed the affected customer profile beyond the marker that previously crashed
- [ ] complete conversion of the 15 GB decompressed profile is still running; the PR will be updated with the final result

### DIFF
--- a/src/main/cpp/profiler/spark_rapids_profile_converter.cpp
+++ b/src/main/cpp/profiler/spark_rapids_profile_converter.cpp
@@ -687,7 +687,12 @@ void convert_to_nvtxt(std::ifstream& in, std::ostream& out, program_options cons
               category = it->second->category();
             }
             marker_start ms{
-              m->timestamp(), process_id, thread_id, color, category, m->name()->str()};
+              m->timestamp(),
+              process_id,
+              thread_id,
+              color,
+              category,
+              m->name() ? m->name()->str() : ""};
             auto [ignored, inserted] = marker_start_map.insert({m->id(), ms});
             if (not inserted) {
               std::ostringstream oss;
@@ -970,8 +975,8 @@ int convert_to_nvtxw(std::ifstream& in,
                             thread_id,
                             color,
                             category,
-                            m->name()->str(),
-                            m->domain()->str()};
+                            m->name() ? m->name()->str() : "",
+                            m->domain() ? m->domain()->str() : ""};
             auto [ignored, inserted] = marker_start_map.insert({m->id(), ms});
             if (not inserted) {
               std::ostringstream oss;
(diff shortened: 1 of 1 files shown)


thirtiseven:
pre-commit.ci autofix

thirtiseven (approved):
Thank you for finding this, LGTM

nvliyuan:
build

nvliyuan:
Hi @thirtiseven could you help trigger a build? seems not authorized, thx
image

thirtiseven:
build

PR #4909, opened 2026-07-29
Deduplicate input file null message

binmahone:
### Description

Define a class-level constant for the repeated `inputFile can't be null` validation message and reuse it at all three `Objects.requireNonNull` call sites.

This addresses the duplicated string literal finding without changing validation behavior or exception text.

### Testing

- `git diff --check`
- Verified the literal has a single definition and all three validation call sites reference the constant.
- No new tests were added because this is a behavior-preserving constant extraction.


binmahone:
build

GaryShen2008 (approved):

binmahone:
build

Issue #4889, opened 2026-07-24
[BUG] `get_json_object` diverges from Spark on many input classes


ttnghia:
The API `get_json_object` diverges from Apache Spark across UTF-16 escaping, UTF-8 decoding, path evaluation, and JSON validation, in the worst case emitting structurally invalid JSON or a wrong value:

| # | Class | Input JSON | Path | Spark `GetJsonObject` | Our output (incorrect) |
|---|---|---|---|---|---|
| 1 | Decoded escape not re-escaped | `{'a':'\u0022'}` | `$` | `{"a":"\""}` | `{"a":"""}` (bare `"`, structurally invalid JSON) |
| 2 | Escaped surrogate decoded | `{'a':'\uD83D\uDE00'}` | `$` | `{"a":"\uD83D\uDE00"}` | `{"a":"😀"}` (pair decoded to raw U+1F600) |
| 3 | Lone surrogate not nulled | `{'a':'\uD800'}` | `$.a` | `null` | non-null string holding the raw, invalid UTF-8 for `0xD800` |
| 4 | Raw astral not escaped | `{'a':'😀'}` (raw U+1F600) | `$` | `{"a":"\uD83D\uDE00"}` | `{"a":"😀"}` (raw U+1F600, copied verbatim) |
| 5 | Malformed UTF-8 not replaced | `{"a":"<ED A0 80>"}` (raw bytes) | `$.a` | `<EF BF BD>` (U+FFFD) | `<ED A0 80>` copied verbatim |
| 6 | Subscript past an inner array ends the row | `[[1],[8,9]]` | `$[*][1]` | `9` | `null` |
| 7 | Nested wildcard adds a nesting level | `[[{"a":[1,2]}]]` | `$[*][*][*].a[*]` | `[[1,2]]` | `[[[1,2]]]` |
| 8 | `null` at a matched field ends the row | `[{"a":null},{"a":1}]` | `$[*].a` | `1` | `null` |
| 9 | Object scan stops at the first name match | `{"a":{"c":1},"a":{"b":2}}` | `$.a.b` | `2` | `null` |
| 10 | Field name accepts any delimiter | `{xax:1}` | `$` | `null` | `{"a":1}` |
| 11 | Root scalar with trailing content accepted | `truex` | `$` | `null` | `true` |
| 12 | Root keyword ended by a non-ASCII identifier character | `trueé` | `$` | `null` | `true` |
[...]

PR #4915, opened 2026-07-30
[SkipRecovery] Add LAST_NON_NULL field selection for json_tuple

wjxiz1992:
Contributes to [NVIDIA/cudf-spark#11361](https://github.com/NVIDIA/cudf-spark/issues/11361).

## Summary

Spark `json_tuple` needs the last non-null occurrence of each requested top-level field: `{"k":"first","k":"last"}` produces `last`, and `{"k":"value","k":null}` retains `value`.

- Add the explicit `NamedFieldMatchPolicy.LAST_NON_NULL` multi-path overload, restricted to exactly one `NAMED` instruction per path. Retain its native ID and the call shape used by NVIDIA/cudf-spark#15446.
- Scan and validate the first object, capture the final matching value, and serialize it through the existing evaluator and exact-size retry. Like Spark, stop at the first object's end; malformed content inside it returns null.
- Keep existing single-path and multi-path entry points on the ordinary evaluator. Separate CUDA kernel instantiations keep the last-match scan out of default calls.
- Preserve full JNI field-name lengths, including embedded NUL, and cover all three entry points with actual null input rows.

Default repeated-key traversal belongs to #5127; offset/JNI argument hardening belongs to #5124. This PR does not add an explicit `FIRST_NON_NULL` API. `from_json` and JSON scan recovery remain outside this change.

## Validation

Executed on the sources committed as `e4b82958563faac8a92f0ae6b9f98b82c7e5a823`:

- Spark 3.5.4 CPU oracle: 19-row duplicate/null/malformed/root matrix, embedded-NUL and null rows, expanded output with adjacent sentinel, and repeated unsorted field names.
- Full `GetJsonObjectTest` on RTX 5880 Ada: **43 tests, 0 failures, 0 errors, 0 skipped**; `mvn package`: **BUILD SUCCESS**.
- Native `GetJsonObjectTest.RejectsInvalidNamedFieldMatchContracts`: **1 test passed** (invalid policy, empty path, wildcard, and nested path).
- Native build on u13: **BUILD SUCCESS** with CUDA 12.9.1 and `80-real;89-real`; the changed CUDA object contains exactly sm_80 and sm_89 code.
- Pre-commit and `git diff --check`: passed.

## Performance

RTX 5880 Ada, driver 580.95.05; same compiler, architecture settings, and pinned cuDF `5ebcdf692e6cbf3470eb0cf0634c200611828e62` for all variants. The default-path baseline uses extractor sources from merge base `ecb9d76acc5e0ea36fd4f5c5e1955067d05ec607`.

Reused the deterministic 34-cell benchmark from #5124 (`5ab1f1a6f77bd28888896eeea7af7f5c1e46d84f`), with only a local `json_path` alias for compatibility. The benchmark harness is outside this PR. Settings: `--min-samples 10 --min-time 0.5 --max-noise 1 --timeout 10`.

All 34 default-path cells completed. GPU latency geometric mean changed **-0.57%**, with individual changes from **-2.78% to +2.44%**. This single sweep did not show a material default-path slowdown.

Additional LAST_NON_NULL workload: approximately 10 MB per cell, 1/8/32 distinct requested fields, each repeated 2/8 times with a null after every non-null value. Output values are checked before timing. All six cells completed:

Compared with the previous PR head `8f194d2357801a8ebb621d3b0638bb97ba651b02`:

| Fields | Repetitions | Old GPU ms | New GPU ms | Latency change |
|---:|---:|---:|---:|---:|
| 1 | 2 | 7.531 | 6.343 | -15.77% |
| 8 | 2 | 4.009 | 3.832 | -4.41% |
| 32 | 2 | 12.290 | 11.538 | -6.12% |
| 1 | 8 | 2.321 | 1.904 | -17.95% |
| 8 | 8 | 3.330 | 3.119 | -6.35% |
| 32 | 8 | 11.530 | 11.238 | -2.54% |

These are single-sweep point estimates. The two 8-field baseline cells had 2.3% and 4.2% sample variability.

<details>
<summary>All 34 default-path latency comparisons</summary>

| Workload | Axes | Base ms | New ms | Latency change |
|---|---|---:|---:|---:|
| get_json_object | size_bytes=1000000, max_depth=2 | 0.6519 | 0.6495 | -0.37% |
| get_json_object | size_bytes=10000000, max_depth=2 | 1.7473 | 1.7438 | -0.20% |
| get_json_object | size_bytes=100000000, max_depth=2 | 9.6034 | 9.3447 | -2.69% |
| get_json_object | size_bytes=1000000000, max_depth=2 | 82.7383 | 83.6059 | +1.05% |
| get_json_object | size_bytes=1000000, max_depth=4 | 0.6898 | 0.7067 | +2.44% |
| get_json_object | size_bytes=10000000, max_depth=4 | 1.7899 | 1.7570 | -1.84% |
| get_json_object | size_bytes=100000000, max_depth=4 | 9.7785 | 9.5068 | -2.78% |
| get_json_object | size_bytes=1000000000, max_depth=4 | 84.4633 | 84.7400 | +0.33% |
| get_json_object | size_bytes=1000000, max_depth=6 | 0.6980 | 0.6900 | -1.15% |
| get_json_object | size_bytes=10000000, max_depth=6 | 1.8033 | 1.7937 | -0.53% |
| get_json_object | size_bytes=100000000, max_depth=6 | 9.6313 | 9.8199 | +1.96% |
| get_json_object | size_bytes=1000000000, max_depth=6 | 85.7169 | 85.9225 | +0.24% |
| get_json_object | size_bytes=1000000, max_depth=8 | 0.7121 | 0.6990 | -1.83% |
| get_json_object | size_bytes=10000000, max_depth=8 | 1.8326 | 1.8355 | +0.16% |
| get_json_object | size_bytes=100000000, max_depth=8 | 9.9342 | 9.8151 | -1.20% |
| get_json_object | size_bytes=1000000000, max_depth=8 | 87.0987 | 87.3988 | +0.34% |
| get_json_object_content | content=ascii | 1.7688 | 1.7589 | -0.56% |
| get_json_object_content | content=utf8_2byte | 1.8237 | 1.8165 | -0.39% |
| get_json_object_content | content=utf8_3byte | 1.7008 | 1.7019 | +0.07% |
| get_json_object_content | content=astral | 1.8231 | 1.8087 | -0.79% |
| get_json_object_content | content=escapes | 1.7915 | 1.7852 | -0.35% |
| get_json_object_content | content=malformed_utf8 | 1.8047 | 1.7995 | -0.29% |
| get_json_object_query | query=named_leaf | 0.9636 | 0.9547 | -0.92% |
| get_json_object_query | query=named_subtree | 1.7312 | 1.7225 | -0.50% |
| get_json_object_query | query=wildcard | 1.9635 | 1.9622 | -0.07% |
| get_json_object_query | query=index | 1.1173 | 1.1121 | -0.46% |
| get_json_object_query | query=wildcard_index | 1.2418 | 1.2288 | -1.05% |
| get_json_object_query | query=index_wildcard | 1.1801 | 1.1731 | -0.60% |
| get_json_object_multi_path | num_paths=1 | 1.0222 | 1.0066 | -1.53% |
| get_json_object_multi_path | num_paths=2 | 1.3935 | 1.3757 | -1.28% |
| get_json_object_multi_path | num_paths=4 | 2.1387 | 2.1188 | -0.93% |
| get_json_object_multi_path | num_paths=8 | 3.5282 | 3.4882 | -1.13% |
| get_json_object_multi_path | num_paths=16 | 6.4491 | 6.3925 | -0.88% |
| get_json_object_multi_path | num_paths=32 | 12.3677 | 12.2038 | -1.33% |

</details>

## Integration status

#5124–#5128 remain unmerged drafts and are not incorporated into this validation. Merge remains on hold for coordination and combined validation with that series, including its UTF-8, tokenizer, and writer changes. NVIDIA/cudf-spark#15446 should remain Draft until the required changes are merged and available in an official JNI artifact.


src/test/java/com/nvidia/spark/rapids/jni/GetJsonObjectTest.java
```diff
@@ -644,6 +644,127 @@ void getJsonObjectMultiplePathsTest() {
...
+      } finally {
+        for (ColumnVector cv : output) {
+          cv.close();
+        }
+      }
+    }
+  }
+
+  @Test
+  void getJsonObjectRejectsUnsupportedMatchPoliciesTest() {
```
thirtiseven, line 808: The new validation at `JSONUtils.java:190` rejects an empty path for `FIRST_NON_NULL`, but this negative test currently covers only a wildcard path for `FIRST_NON_NULL`, nested/wildcard paths for `LAST_NON_NULL`, and a null policy. Could we add an empty-path case here to pin the documented non-empty-path contract?

src/main/cpp/src/get_json_object.cu
```diff
@@ -839,19 +959,32 @@ __launch_bounds__(block_size, min_block_per_sm) CUDF_KERNEL
   auto const str = input.element<cudf::string_view>(row_idx);
   if (str.size_bytes() > 0) {
     json_parser p{char_range{str}};
-    cuda::std::tie(is_valid, out_size) =
-      evaluate_path(p, path.path_commands, dst, max_path_depth_exceeded);
+    if (path.use_named_path_selection) {
```
thirtiseven, line 998: All all-`NAMED` legacy calls now take this selection/full-document-validation path and then replay the selected value, so the common path performs a second token walk. The existing `GET_JSON_OBJECT_BENCH` already exercises a multi-level all-`NAMED` path. Could we include a before/after result from that benchmark (or point to an agreed regression budget/follow-up) to quantify the impact? The ptxas stack/register numbers do not answer the throughput question.

src/main/cpp/src/get_json_object.cu
```diff
@@ -1093,6 +1233,11 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
 
     if (h_error_check[idx]) {
       oob_indices.emplace_back(idx);
+      auto const& path = json_paths[idx];
+      auto const use_named_path_selection =
```
thirtiseven: This predicate is duplicated from the initial-launch setup at line 1178. Because the OOB retry must use the same evaluator as the sizing pass, could we compute this once (or extract a helper) and reuse it here? That would keep future path-shape changes from making the initial and retry launches diverge.

src/test/java/com/nvidia/spark/rapids/jni/GetJsonObjectTest.java
wjxiz1992: Thanks, added in aee38506. The FIRST_NON_NULL validation test now includes an empty-path rejection case; the final GetJsonObjectTest run on b0d3c3b5 passed 44/44.

src/main/cpp/src/get_json_object.cu
wjxiz1992: Measured with the existing 100 MB depth 2/4/6/8 benchmark on an RTX 5880 Ada. Baseline was 8.250-8.698 GB/s; the replay version was 6.085-6.382 GB/s (-25.2% to -29.6%); b0d3c3b5 is 8.676-8.860 GB/s (+1.9% to +5.2%), with a second run at 8.697-9.171 GB/s. I added the command and per-depth table to the PR body.

src/main/cpp/src/get_json_object.cu
wjxiz1992: Updated in aee38506. The all-NAMED predicate is now computed once per path in named_path_selection_flags and reused by both the initial sizing pass and the OOB retry.

wjxiz1992:
build

wjxiz1992:
build

src/main/cpp/src/JSONUtilsJni.cpp
```diff
@@ -20,9 +20,127 @@
...
+
+jlongArray get_json_object_multiple_paths(JNIEnv* env,
+                                          jlong j_input,
+                                          jbyteArray j_type_nums,
+                                          jobjectArray j_names,
+                                          jintArray j_indexes,
+                                          jintArray j_path_offsets,
+                                          jlong memory_budget_bytes,
+                                          jint parallel_override,
+                                          bool use_match_policy,
```
thirtiseven: [Optional] Could this be `std::optional<named_field_match_policy>` instead of a boolean plus a policy? The legacy caller passes `false` with a dummy `FIRST_NON_NULL`; `std::nullopt` versus a policy would encode the two modes directly.

> The [cuDF Developer Guide](https://github.com/rapidsai/cudf/blob/main/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#c-guidelines) recommends following the C++ Core Guidelines; [I.4](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-typed) says, “Make interfaces precisely and strongly typed.”

src/main/cpp/src/get_json_object.cu
```diff
@@ -1032,15 +1199,21 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object_batch(
...
   out_stringviews.reserve(json_paths.size());
   h_path_data.reserve(json_paths.size());
+  named_path_selection_flags.reserve(json_paths.size());
 
   for (std::size_t idx = 0; idx < num_outputs; ++idx) {
     auto const& path = json_paths[idx];
     if (path.size() > MAX_JSON_PATH_DEPTH) {
       CUDF_FAIL("JSON Path has depth exceeds the maximum allowed value.");
     }
+    named_path_selection_flags.emplace_back(
```
thirtiseven: [Optional] Could we extract this repeated predicate as `is_all_named_path(std::span<path_spec const>)` and reuse it here and in the policy validation below? This would keep the path-shape contract in one place and use the standard host-only view.

> [cuDF Developer Guide — Spans](https://github.com/rapidsai/cudf/blob/main/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#when-to-use-host_span-vs-stdspan): “For host-side data, prefer `std::span`.”

src/main/cpp/src/get_json_object.hpp
```diff
@@ -69,4 +78,31 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object_multiple_paths(
...
+ */
+std::vector<std::unique_ptr<cudf::column>> get_json_object_multiple_paths(
+  cudf::strings_column_view const& input,
+  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>> const&
+    json_paths,
+  int64_t memory_budget_bytes,
+  int32_t parallel_override,
+  named_field_match_policy match_policy,
+  rmm::cuda_stream_view stream      = cudf::get_default_stream(),
+  rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref());
```
thirtiseven: [Optional] Since this overload uses the cuDF wrapper, could we update the two existing MR defaults above as well? The wrapper currently delegates to RMM, so this is behavior-preserving and keeps the overload family consistent.

> [cuDF Developer Guide — Current Device Memory Resource](https://github.com/rapidsai/cudf/blob/main/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#current-device-memory-resource): “All memory resource parameters should be defaulted” to `cudf::get_current_device_resource_ref()`.

wjxiz1992:
build

thirtiseven (dismissed):
LGTM

wjxiz1992:
build

wjxiz1992:
build

wjxiz1992:
build

src/main/cpp/src/JSONUtilsJni.cpp
```diff
@@ -20,10 +20,128 @@
...
+    auto const size      = type_nums.size();
+    if (static_cast<std::size_t>(names.size()) != size || indexes.size() != size) {
+      JNI_THROW_NEW(
+        env, cudf::jni::ILLEGAL_ARG_EXCEPTION_CLASS, "wrong number of entries passed in", 0);
+    }
+
+    instructions.reserve(size);
+    for (std::size_t i = 0; i < size; i++) {
+      auto const instruction_type = static_cast<path_instruction_type>(type_nums[i]);
+      char const* name_str        = names[i].get();
```
firestarman: Could we preserve the explicit byte length when marshalling this path name? Converting `native_jstring` to `char const*` makes the tuple's `std::string` use C-string length, so a Java name containing `a`, followed by U+0000, followed by `b` is truncated to `a` and can select a different JSON field. Using the length-aware conversion, such as `native_jstringArray(...).as_cpp_vector()`, and placing the resulting `std::string` directly in the tuple would preserve the full name.

src/main/cpp/src/JSONUtilsJni.cpp
```diff
@@ -20,10 +20,128 @@
...
+    for (std::size_t i = 0; i < num_paths; ++i) {
+      CUDF_EXPECTS(path_offsets.front() == 0 && path_offsets[i] >= 0 &&
+                     path_offsets[i] <= path_offsets[i + 1] && path_offsets[i + 1] <= num_entries,
+                   "Invalid path offsets.");
+      auto const path_size = path_offsets[i + 1] - path_offsets[i];
+      auto path            = path_type{};
+      path.reserve(path_size);
+      for (int j = path_offsets[i]; j < path_offsets[i + 1]; ++j) {
+        auto const instruction_type = static_cast<path_instruction_type>(type_nums[j]);
+        char const* name_str        = names[j].get();
```
firestarman: The same length loss occurs in the multi-path marshalling loop here. `names[j].get()` exposes the bytes, but the subsequent `std::string` construction stops at an embedded NUL. Could this use `as_cpp_vector()` (or construct with `size_bytes()`) so Java path names containing `U+0000` are not truncated and matched against the wrong field?

src/main/cpp/src/get_json_object.cu
```diff
@@ -839,19 +993,45 @@ __launch_bounds__(block_size, min_block_per_sm) CUDF_KERNEL
   auto const str = input.element<cudf::string_view>(row_idx);
   if (str.size_bytes() > 0) {
     json_parser p{char_range{str}};
-    cuda::std::tie(is_valid, out_size) =
-      evaluate_path(p, path.path_commands, dst, max_path_depth_exceeded);
+    if (path.use_named_path_selection) {
```
firestarman, line 998: Could the duplicate-field test matrix include an actual null input-column row? It currently covers a JSON `null` literal and malformed JSON, but not a null row entering this named-selection branch. Adding one null input and null expected result would verify null propagation through the legacy single-path and multi-path calls as well as explicit `FIRST_NON_NULL` and `LAST_NON_NULL`.

src/main/cpp/src/get_json_object.cu
```diff
@@ -1255,8 +1448,47 @@ std::vector<std::unique_ptr<cudf::column>> get_json_object_multiple_paths(
...
+  std::vector<std::vector<std::tuple<path_instruction_type, std::string, int32_t>>> const&
+    json_paths,
+  int64_t memory_budget_bytes,
+  int32_t parallel_override,
+  named_field_match_policy match_policy,
+  rmm::cuda_stream_view stream,
+  rmm::device_async_resource_ref mr)
+{
+  SRJ_FUNC_RANGE();
+  CUDF_EXPECTS(match_policy == named_field_match_policy::FIRST_NON_NULL ||
```
firestarman, line 1481: Could we add direct native contract tests for these guards: an out-of-range policy, a wildcard `FIRST_NON_NULL` path, and a multi-segment `LAST_NON_NULL` path? The current Java negative tests reject these inputs before JNI, so removing or breaking the native checks would not fail that suite even though this public C++ overload documents a `cudf::logic_error` contract.

wjxiz1992:
build

ttnghia (changes requested):
Please hold on this a little bit. I've been fixing a bunch of bugs in `get_json_objects` that are described in https://github.com/NVIDIA/cudf-spark-jni/issues/4889. My code is ready for a long time but I need to prioritize something else and didn't have time to check again before pushing, there are about 3-5 PRs in total.

I did not look at this PR carefully, thus not sure if there are overlap/conflict work. Please check the issue https://github.com/NVIDIA/cudf-spark-jni/issues/4889.

I will push my code soon, then we will verify to make sure our PRs are parallell.

In addition, from my experiments with JSON code, fixing correctness bug like this could change the performance significantly. In my fix, I had to run microbenchmark after every code change. Sometimes, speed regressed significantly so you have to be extremely careful.

src/main/cpp/src/JSONUtilsJni.cpp
wjxiz1992: Thanks, updated.

src/main/cpp/src/JSONUtilsJni.cpp
wjxiz1992: Updated.

src/main/cpp/src/get_json_object.cu
wjxiz1992: Updated.

src/main/cpp/src/get_json_object.cu
wjxiz1992: Updated.

wjxiz1992:
Thanks. I'll hold this PR and coordinate with the #4889 fixes. After those land, I'll reconcile any overlap and rerun GET_JSON_OBJECT_BENCH before requesting review again.

wjxiz1992:
hi @ttnghia I compared your json issue table and found one direct overlap: item 9, where scanning stops at the first matching field name. #4915 addresses this through FIRST_NON_NULL matching for all-NAMED paths.
#4915 also incidentally adds full path-offset validation, overlapping the unnumbered offset-validation issue. The remaining items in #4889 are outside #4915’s scope. The LAST_NON_NULL API and related json_tuple support are also separate.
Please let me know whether your PR series will cover item 9 and the offset validation, so I can coordinate or re-scope #4915 accordingly.

wjxiz1992:
build

nvauto:

wjxiz1992:
build

PR #4917, opened 2026-07-30
Fix ORC timezone conversion semantics

res-life:
Contributes to NVIDIA/cudf-spark#15449.

### Description

GPU ORC timestamp conversion needs to match Apache ORC's
`java.util.TimeZone` semantics for writer and reader timezones.

This change:

- preserves ORC's `TimeZone.hasSameRules` behavior so same-rule writer/reader
  pairs skip cross-timezone conversion;
- preserves fixed-offset and short timezone IDs by resolving
  `TimeZone` directly from `ZoneId`;
- adds an ORC-compatible `convertFromUtc` path for timestamp columns;
- exposes the reader timezone's first transition so the plugin can limit
  historical rule correction to values before that transition;
- preserves Java/ORC overflow, negative timestamp, null-mask, sliced-column,
  CUDA stream, and memory-resource behavior.

Without the direct `ZoneId` overload, values such as `+05:30` and short IDs
such as `EST` were silently interpreted as GMT when checking
`hasSameRules`. That could incorrectly skip conversion and produce a constant
timezone offset error.

### Testing

- `GpuTimeZoneDBTest`: `8 passed`, including `UTC` in both directions with
  `+05:30` and `EST`.
- `pre-commit` checks passed; no C/C++/CUDA files were changed by the follow-up
  fixed-offset commit.
- CUDA 12 container build: `BUILD SUCCESS`.
- Dependent cuDF Spark A/B:
  - baseline reproduced all 7 fixed-seed failures;
  - fixed build passed the same 7 cases.

### DIFF
--- a/src/main/cpp/src/timezones.cu
+++ b/src/main/cpp/src/timezones.cu
@@ -817,9 +817,12 @@ __device__ cudf::timestamp_us convert_orc_from_utc_value(cudf::timestamp_us valu
                                                          tz_side_info const& reader)
 {
   auto const value_us = value.time_since_epoch().count();
-  auto const value_ms = spark_rapids_jni::integer_utils::floor_div(value_us, MICROS_PER_MILLI);
-  auto const offset_lookup_ms = wrapping_subtract(value_ms, reader.raw_offset);
-  auto const offset_ms        = get_transition_index(offset_lookup_ms, reader);
+  auto offset_ms      = reader.raw_offset;
+  if (!reader.is_fixed) {
+    auto const value_ms = spark_rapids_jni::integer_utils::floor_div(value_us, MICROS_PER_MILLI);
+    auto const offset_lookup_ms = wrapping_subtract(value_ms, reader.raw_offset);
+    offset_ms                   = get_transition_index(offset_lookup_ms, reader);
+  }
   auto const result_us =
     wrapping_subtract(value_us, static_cast<int64_t>(offset_ms) * MICROS_PER_MILLI);
   return cudf::timestamp_us{cudf::duration_us{result_us}};
@@ -835,17 +838,19 @@ CUDF_KERNEL void __launch_bounds__(CONVERT_TZ_BLOCK_SIZE)
                               orc_tz_side_kernel_args reader_args)
 {
   extern __shared__ char smem[];
-  char* ptr = smem;
-  int64_t const *rt_begin, *rt_end;
-  int32_t const* ro_begin;
-  stage_side_transitions(reader_args.trans,
-                         reader_args.offsets,
-                         reader_args.trans_count,
-                         ptr,
-                         rt_begin,
-                         rt_end,
-                         ro_begin);
-  __syncthreads();
+  char* ptr               = smem;
+  int64_t const *rt_begin = nullptr, *rt_end = nullptr;
+  int32_t const* ro_begin = nullptr;
+  if (!reader_args.is_fixed) {
+    stage_side_transitions(reader_args.trans,
+                           reader_args.offsets,
+                           reader_args.trans_count,
+                           ptr,
+                           rt_begin,
+                           rt_end,
+                           ro_begin);
+    __syncthreads();
+  }
 
   cudf::size_type idx = blockIdx.x * blockDim.x + threadIdx.x;
   if (idx < num_rows) {
@@ -855,7 +860,8 @@ CUDF_KERNEL void __launch_bounds__(CONVERT_TZ_BLOCK_SIZE)
                               ro_begin,
                               reader_args.initial_offset,
                               reader_args.raw_offset,
-                              reader_args.dst};
+                              reader_args.dst,
+                              reader_args.is_fixed};
     output[idx] = convert_orc_from_utc_value(input[idx], reader);
   }
 }
@@ -879,6 +885,7 @@ std::unique_ptr<column> convert_orc_from_utc_typed(cudf::column_view const& inpu
   int32_t const* reader_offsets_ptr =
     reader.tz_info_table ? reader.tz_info_table->column(1).begin<int32_t>() : nullptr;
   int32_t reader_trans_count = reader.tz_info_table ? reader.tz_info_table->column(0).size() : 0;
+  auto const is_reader_fixed = reader_trans_count == 0 && !reader.dst.has_dst;
   size_t smem_bytes          = 0;
   if (reader_trans_count > 0 && reader_trans_count <= MAX_SMEM_TRANSITIONS) {
     smem_bytes = reader_trans_count * (sizeof(int64_t) + sizeof(int32_t));
@@ -889,7 +896,8 @@ std::unique_ptr<column> convert_orc_from_utc_typed(cudf::column_view const& inpu
                                                    reader_trans_count,
                                                    reader.initial_offset,
                                                    reader.raw_offset,
-                                                   reader.dst};
+                                                   reader.dst,
+                                                   is_reader_fixed};
   int32_t num_blocks       = cudf::util::div_rounding_up_safe(input.size(), CONVERT_TZ_BLOCK_SIZE);
   auto const launch_config = cuda::make_config(cuda::grid_dims(num_blocks),
                                                cuda::block_dims<CONVERT_TZ_BLOCK_SIZE>(),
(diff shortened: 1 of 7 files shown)


res-life:
build

res-life:
build

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -595,6 +598,18 @@ private OrcTimezoneContext(Table writerTzInfoTable, Table readerTzInfoTable,
       this.readerInitialOffset = readerTzInfo.initialOffset;
       this.readerRawOffset = readerTzInfo.rawOffset;
       this.readerDstRule = dstRuleToArray(readerTzInfo.dstRule);
+      this.readerFirstTransitionUs =
+          readerTzInfo.transitions == null || readerTzInfo.transitions.length == 0
+              ? Long.MIN_VALUE
+              : TimeUnit.MILLISECONDS.toMicros(readerTzInfo.transitions[0]);
```
thirtiseven: This value is compared directly with `orcTimestamp` in cudf-spark#15450, but it is currently a UTC instant while `convertOrcFromUtc` looks up the offset at `value_ms - reader.raw_offset`. The input-side cutoff therefore needs the raw offset; otherwise positive-offset zones can skip required historical correction (for example, `Europe/Paris` leaves a 561-second gap).

```suggestion
              : TimeUnit.MILLISECONDS.toMicros(
                  readerTzInfo.transitions[0] + readerTzInfo.rawOffset);
```

src/main/cpp/src/timezones.cu
```diff
@@ -582,11 +582,13 @@ __device__ static cudf::timestamp_us convert_timestamp_between_timezones(
...
   orc_base_offset_info writer_2015_year_base_offset,
   tz_side_info const& writer,
-  tz_side_info const& reader)
+  tz_side_info const& reader,
+  bool writer_reader_rules_differ)
 {
   int64_t const decoded_us = static_cast<int64_t>(
     cuda::std::chrono::duration_cast<cudf::duration_us>(ts.time_since_epoch()).count());
   int64_t const adjusted_us = apply_orc_base_offset(decoded_us, writer_2015_year_base_offset);
+  if (!writer_reader_rules_differ) { return cudf::timestamp_us{cudf::duration_us{adjusted_us}}; }
```
firestarman, line 591: Could the same-rule path bypass transition staging entirely? This return happens only after both transition tables have been copied to shared memory and the block has synchronized, while the launch also reserves up to 12 KiB of dynamic shared memory that this path never reads. Please gate the staging/barrier on `writer_reader_rules_differ` and request zero dynamic shared memory when the rules match.

src/main/cpp/src/timezones.cu
```diff
@@ -789,12 +793,116 @@ std::unique_ptr<column> convert_timezones(cudf::column_view const& input,
...
+__device__ static T convert_orc_from_utc_value(T value, tz_side_info const& reader);
+
+template <>
+__device__ cudf::timestamp_us convert_orc_from_utc_value(cudf::timestamp_us value,
+                                                         tz_side_info const& reader)
+{
+  auto const value_us = value.time_since_epoch().count();
+  auto const value_ms = spark_rapids_jni::integer_utils::floor_div(value_us, MICROS_PER_MILLI);
+  auto const offset_lookup_ms = wrapping_subtract(value_ms, reader.raw_offset);
+  auto const offset_ms        = get_transition_index(offset_lookup_ms, reader);
```
firestarman: `convertOrcFromUtc` always performs the floor division and transition lookup, even for fixed-offset readers (`trans_count == 0 && !dst.has_dst`), and the kernel still executes a block barrier after staging nothing. Please propagate the existing `is_fixed` flag into this path, use `raw_offset` directly for fixed zones, and skip the barrier when no table is staged.

src/main/cpp/src/timezones.hpp
```diff
@@ -168,6 +170,22 @@ struct orc_tz_side {
...
+/**
+ * @brief Apply Apache ORC SerializationUtils.convertFromUtc semantics.
+ *
+ * @param input TIMESTAMP_MICROSECONDS input column.
+ * @param reader reader timezone transition data, offsets, and DST rule.
+ * @param stream CUDA stream.
+ * @param mr Device memory resource.
+ * @return converted column with the same type as input.
+ */
+[[nodiscard]] std::unique_ptr<cudf::column> convert_orc_from_utc(
```
firestarman, line 186: `convert_orc_from_utc` rejects non-`TIMESTAMP_MICROSECONDS` input with `CUDF_EXPECTS`, but this public declaration does not document the exception. Please add an `@throw cudf::logic_error` entry so the generated API documentation exposes the precondition.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -595,6 +619,18 @@ private OrcTimezoneContext(Table writerTzInfoTable, Table readerTzInfoTable,
...
+      TimeZone writerTz = TimeZone.getTimeZone(getZoneId(writerTimezone));
+      TimeZone readerTz = TimeZone.getTimeZone(getZoneId(readerTimezone));
+      this.writerReaderRulesDiffer = !writerTz.hasSameRules(readerTz);
+    }
+
+    public long getReaderFirstTransitionUs() {
```
firestarman, line 639: Could this public accessor document its return and closed-state contracts? `Long.MIN_VALUE` is the no-transition sentinel, and `ensureOpen()` throws `IllegalStateException` after closure. Please add Javadoc with `@return` and `@throws` entries.

src/test/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDBTest.java
```diff
@@ -190,6 +254,95 @@ void testConvertOrcTimezonesCorrectsIgnoredWriterTimezoneEpochBorrow() {
...
+              convertOrcTimezonesOnCPU(microseconds, timezones[0], timezones[1]);
+          ColumnVector actual =
+              GpuTimeZoneDB.convertOrcTimezones(input, timezones[0], timezones[1])) {
+        assertColumnsAreEqual(expected, actual);
+      }
+    }
+  }
+
+  @Test
+  void testConvertOrcFromUtcAllTimezones() {
```
firestarman, line 282: Could the new overloads also cover their failure contracts? Existing calls all use an open context and a microsecond timestamp column, so removing `context.ensureOpen()` or the native type check would not fail these tests. Please add cases for a closed context and a non-`TIMESTAMP_MICROSECONDS` column.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Fixed in ab539184. The cutoff is now expressed in the local ORC timestamp frame by adding the reader raw offset. The regression test uses Europe/Paris so a positive-offset zone exercises this case. Local GpuTimeZoneDBTest: 10 passed.

src/main/cpp/src/timezones.cu
res-life: Fixed in 737102bb. Same-rule conversion now requests zero dynamic shared memory and bypasses transition staging and the block barrier. Local GpuTimeZoneDBTest passed (11/11).

src/main/cpp/src/timezones.cu
res-life: Fixed in f09d61a6. Fixed-offset readers now propagate is_fixed, use raw_offset directly, and skip floor division, transition lookup, staging, and the block barrier. Local GpuTimeZoneDBTest passed (11/11).

src/main/cpp/src/timezones.hpp
res-life: Fixed in 5e3200a3 by documenting the cudf::logic_error contract for non-TIMESTAMP_MICROSECONDS input.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
res-life: Fixed in 05eff2d6. The accessor Javadoc now documents the Long.MIN_VALUE sentinel and IllegalStateException after closure.

src/test/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDBTest.java
res-life: Fixed in b09ee774. Tests now verify both context overloads reject a closed context and non-TIMESTAMP_MICROSECONDS input. Local GpuTimeZoneDBTest passed (11/11).

res-life:
build

firestarman (approved):
lgtm

PR #4930, opened 2026-08-03
Add Spark-safe UTF-8 string reverse for truncated trailing bytes

firestarman:
## Summary
- Adds a Spark-compatible string reverse JNI (`StringUtils.reverseStrings`) that follows `UTF8String.numBytesForFirstByte` and clamps character width to bytes remaining in each row (SPARK-57507).
- Avoids libcudf `strings::reverse` over-reading into the next row when Spark `StringType` holds truncated trailing multi-byte UTF-8.
- Includes C++ (`REVERSE_STRINGS`) and Java (`StringUtilsTest`) coverage for well-formed and truncated trailing cases with neighboring sentinel rows.

This unblocks cudf-spark work for NVIDIA/cudf-spark#15383 (GPU `reverse` on malformed trailing UTF-8).

## Test plan
- [x] C++ `ctest -R REVERSE_STRINGS`
- [x] Java `StringUtilsTest` reverse cases
- [x] Pre-merge CI


src/main/cpp/src/reverse_strings.cu
```diff
@@ -0,0 +1,123 @@
...
+                                              rmm::cuda_stream_view stream,
+                                              rmm::device_async_resource_ref mr)
+{
+  if (input.is_empty()) { return cudf::make_empty_column(cudf::type_id::STRING); }
+
+  // Preserve offsets/nulls; rewrite only the character bytes.
+  auto result          = std::make_unique<cudf::column>(input.parent(), stream, mr);
+  auto const sv        = cudf::strings_column_view(result->view());
+  auto const d_offsets = cudf::detail::offsetalator_factory::make_input_iterator(sv.offsets());
+  auto* d_chars        = result->mutable_view().head<char>();
```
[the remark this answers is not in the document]
firestarman: Thanks — this matches the current cuDF STRING layout rather than the older chars-child layout.

In today's cuDF, the character bytes live in the parent column's data buffer and offsets are the only child. `strings_column_view::chars_begin()` is implemented as `head<char>()`, and libcudf's own `strings::detail::reverse` uses the same pattern:

```cpp
auto result = std::make_unique<column>(input.parent(), stream, mr);
auto d_chars = result->mutable_view().head<char>();
```

Local C++ (`REVERSE_STRINGS`) and Java (`StringUtilsTest`) reverse tests pass with this pointer.

wjxiz1992 (dismissed):
Defensive review complete: no blocking findings on 43ef77b. The SPARK-57507 byte-width/clamping semantics, current cuDF STRING layout, slice and row bounds, stream/MR handling, and JNI/CMake integration are consistent. The Greptile parent-pointer concern does not apply to the cuDF revision pinned here.

src/main/cpp/src/reverse_strings.cu
```diff
@@ -0,0 +1,123 @@
...
+#include <cudf/strings/strings_column_view.hpp>
+#include <cudf/utilities/default_stream.hpp>
+#include <cudf/utilities/memory_resource.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+#include <rmm/exec_policy.hpp>
+
+#include <cuda/iterator>
+#include <cuda/std/algorithm>
+#include <thrust/for_each.h>
```
res-life, line 33: **[SHOULD FIX] Include the standard headers used by this translation unit directly**

This file uses bare `uint8_t` and `std::make_unique`, but currently gets both declarations through transitive CUDA/cuDF includes. That makes the build sensitive to unrelated dependency-header changes. Please keep the repository's prevalent bare `uint8_t` spelling and include its C header directly; no `uint8_t` to `std::uint8_t` rename is needed.

```suggestion
#include <thrust/for_each.h>

#include <memory>
#include <stdint.h>
```

src/main/cpp/src/reverse_strings.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cudf/column/column.hpp>
+#include <cudf/strings/strings_column_view.hpp>
+#include <cudf/utilities/default_stream.hpp>
+#include <cudf/utilities/memory_resource.hpp>
```
res-life, line 21: **[SHOULD FIX] Make this public header self-contained for `std::unique_ptr`**

The declaration below exposes `std::unique_ptr<cudf::column>`, but this header does not include `<memory>` itself. A consumer should not need a particular transitive include order for this declaration to compile.

```suggestion
#include <cudf/utilities/memory_resource.hpp>

#include <memory>
```

src/main/cpp/tests/reverse_strings.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include "reverse_strings.hpp"
+
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/column_utilities.hpp>
+#include <cudf_test/column_wrapper.hpp>
+
+#include <cudf/strings/strings_column_view.hpp>
+
+#include <string>
```
res-life, line 30: **[SHOULD FIX] Include the defining headers for the test helper's types**

`bytes_to_string` uses `std::initializer_list` and bare `uint8_t`, while the test currently relies on other headers to expose both. Please add their defining headers directly and retain the repository-consistent bare `uint8_t` spelling.

```suggestion
#include <initializer_list>
#include <stdint.h>
#include <string>
```

src/main/cpp/tests/reverse_strings.cpp
```diff
@@ -0,0 +1,95 @@
...
+struct ReverseStringsTests : public cudf::test::BaseFixture {};
+
+TEST_F(ReverseStringsTests, EmptyAndNulls)
+{
+  auto const input  = cudf::test::strings_column_wrapper{};
+  auto const result = spark_rapids_jni::reverse_strings(cudf::strings_column_view{input});
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(*result, input);
+
+  auto const with_nulls =
+    cudf::test::strings_column_wrapper({"abc", "", "世界"}, {true, false, true});
```
res-life, line 50: **[SHOULD FIX] Cover a valid empty string and a nonzero-offset slice**

The only empty row here is masked null, so it exits through the null branch instead of exercising the valid-empty branch. The new implementation also combines an offset-aware input device view with normalized output offsets, but no test passes a sliced strings view. Please add both cases so an input-slice offset regression cannot pass unnoticed.

A concrete test edit is:

```diff
 #include <cudf_test/column_wrapper.hpp>
 
+#include <cudf/copying.hpp>
 #include <cudf/strings/strings_column_view.hpp>
@@
   CUDF_TEST_EXPECT_COLUMNS_EQUAL(*reversed, expected);
+
+  auto const slice_source = cudf::test::strings_column_wrapper(
+    {"prefix", "", "ignored", "世界"}, {true, true, false, true});
+  auto const slice = cudf::slice(slice_source, {1, 4}).front();
+  auto const slice_expected =
+    cudf::test::strings_column_wrapper({"", "ignored", "界世"}, {true, false, true});
+  auto const slice_reversed =
+    spark_rapids_jni::reverse_strings(cudf::strings_column_view{slice});
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(*slice_reversed, slice_expected);
 }
```

src/main/cpp/tests/reverse_strings.cpp
```diff
@@ -0,0 +1,95 @@
...
+     bytes_to_string({0xE4, 0xB8, 0x41}),
+     bytes_to_string({0x43, 0x96}),
+     bytes_to_string({0xF0, 0x90, 0x41}),
+     bytes_to_string({0x44, 0x88, 0x8D}),
+     bytes_to_string({0xCE, 0xE4, 0xB8, 0x96}),
+     bytes_to_string({0x45})});
+
+  auto const result = spark_rapids_jni::reverse_strings(cudf::strings_column_view{input});
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(*result, expected);
+}
```
res-life, line 109: **[SHOULD FIX] Exercise Spark's disallowed UTF-8 lead-byte boundaries**

The classifier has distinct width-1 behavior for `0xC0`-`0xC1` and `0xF5`-`0xFF`, but the current tests only cover continuation bytes and valid 2-, 3-, and 4-byte lead bytes. A boundary change at either branch would therefore preserve the current test results while diverging from `UTF8String.numBytesForFirstByte`. Please add a focused regression test containing both ends of both disallowed ranges.

```suggestion
}

TEST_F(ReverseStringsTests, DisallowedUtf8LeadBytesAreWidthOne)
{
  auto const input = cudf::test::strings_column_wrapper(
    {bytes_to_string({0x41, 0xC0, 0x42, 0xC1, 0x43, 0xF5, 0x44, 0xFF, 0x45})});
  auto const expected = cudf::test::strings_column_wrapper(
    {bytes_to_string({0x45, 0xFF, 0x44, 0xF5, 0x43, 0xC1, 0x42, 0xC0, 0x41})});

  auto const result = spark_rapids_jni::reverse_strings(cudf::strings_column_view{input});
  CUDF_TEST_EXPECT_COLUMNS_EQUAL(*result, expected);
}
```

src/main/cpp/src/reverse_strings.cu
firestarman: Done in f99f545e09 — added `#include <memory>` and `#include <stdint.h>` (C header for bare `uint8_t`).

src/main/cpp/src/reverse_strings.hpp
firestarman: Done in f99f545e09 — `#include <memory>` added to the public header.

src/main/cpp/tests/reverse_strings.cpp
firestarman: Done in f99f545e09 — added `#include <initializer_list>`, `#include <stdint.h>`, and `#include <cstring>`.

src/main/cpp/tests/reverse_strings.cpp
firestarman: Done in f99f545e09 — extended `EmptyAndNulls` with a valid empty string and a nonzero-offset `cudf::slice` case. Local `ctest -R REVERSE_STRINGS` passes.

src/main/cpp/tests/reverse_strings.cpp
firestarman: Done in f99f545e09 — added `DisallowedUtf8LeadBytesAreWidthOne` covering `0xC0`/`0xC1` and `0xF5`/`0xFF`.

firestarman:
build

res-life (approved):
LGTM

firestarman:
build

firestarman:
build

PR #4974, opened 2026-08-10
Clean up casts, including reinterpret_cast

res-life:
Follow-up to https://github.com/NVIDIA/cudf-spark-jni/pull/4957.

Clean up the `reinterpret_cast` expressions that have a safer direct replacement:

- use value-level `cuda::std::bit_cast` for hash object representations;
- use timestamp duration accessors instead of reading timestamp representations through integer pointers;
- remove redundant `parse_uri` casts and use `static_cast` for `void*` buffer conversions; and
- use `cuda::std::bit_cast` for pointer-address alignment calculations on supported 64-bit CUDA targets.

The remaining casts represent byte views or raw storage/layout pointers. They are intentionally unchanged because mechanically replacing them with pointer `bit_cast` would not address alignment, object lifetime, or aliasing requirements.

### DIFF
--- a/src/main/cpp/src/hash/hash.cuh
+++ b/src/main/cpp/src/hash/hash.cuh
@@ -30,6 +30,12 @@
 #include <thrust/reverse.h>
 
 namespace spark_rapids_jni {
+
+struct java_big_decimal_bytes {
+  cuda::std::array<cuda::std::byte, sizeof(__int128_t)> bytes;
+  cudf::size_type length;
+};
+
 /**
  * Normalization of floating point NaNs, passthrough for all other values.
  */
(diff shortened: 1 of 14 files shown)


src/main/cpp/src/hash/hash.cuh
```diff
@@ -99,10 +101,9 @@ __device__ __inline__ cuda::std::pair<__int128_t, cudf::size_type> to_java_bigde
...
 
   // Convert to big endian by reversing the range of nonzero bytes. Only those bytes are hashed.
-  __int128_t big_endian_value = 0;
-  auto big_endian_data        = reinterpret_cast<cuda::std::byte*>(&big_endian_value);
-  thrust::reverse_copy(thrust::seq, data, data + length, big_endian_data);
+  cuda::std::array<cuda::std::byte, key_size> big_endian_data{};
+  thrust::reverse_copy(thrust::seq, data.begin(), data.begin() + length, big_endian_data.begin());
 
-  return {big_endian_value, length};
+  return {cuda::std::bit_cast<__int128_t>(big_endian_data), length};
```
thirtiseven: Could we make this helper return the byte array directly? Both callers immediately bit-cast the returned `__int128_t` back to `cuda::std::array<cuda::std::byte, ...>` before hashing, so the integer is now only a transport for the byte representation. Returning the bytes and length (perhaps via a small named struct) would remove this round trip and make the helper contract match what its callers consume.

thirtiseven (approved):
LGTM

PR #4979, opened 2026-08-11
Refactor protobuf decoder internals

thirtiseven:
## Summary

This PR simplifies the internal protobuf decode pipeline without adding new decoding semantics. It removes decoder-specific column cleanup, clarifies temporary versus output memory ownership, and consolidates the views and contexts passed through scan and extraction paths.

The goal is to leave the existing scalar, repeated primitive/string/bytes, enum, and nested struct behavior intact while making the decoder easier to extend for repeated message fields in the next 3c change.

### Nested column and null handling

- Removes the custom recursive propagation of parent null masks through nested `STRUCT` and `LIST` descendants.
- Relies on libcudf struct construction to sanitize descendant nulls consistently with other libcudf callers.
- Avoids retaining allocated null masks when a column contains no nulls, making all-valid output columns non-nullable.

### Memory-resource ownership

- Uses the current device memory resource for temporary decode state, including field locations, lookup tables, validity buffers, error state, and CUB workspace.
- Uses the current device memory resource for transient column-construction helpers, including the string index pairs used to build all-null string columns.
- Keeps buffers owned by returned columns on the caller-provided output memory resource.
- Transfers the decoded character buffer directly into `LIST<UINT8>` output for protobuf `bytes` fields instead of allocating and copying it again.

### Decode plumbing

- Constructs named schema/runtime contexts, field decode requests, descriptor lookups, and field scan views instead of repeating nested aggregate initializers at call sites.
- Passes `nested_location_provider` directly to the grandchild-location kernel so host and device code use the same location abstraction.
- Removes the unused output-memory-resource parameter from enum lookup-table construction, the unused repeated-message child location provider, and unused error parameters from repeated string/bytes builders.

### CUDA error handling

- Checks raw CUDA kernel launches at their wrapper or call site with `CUDF_CHECK_CUDA`.
- Checks both CUB batched-copy calls with `CUDF_CUDA_TRY` so launch/setup failures are reported where they occur.

## Scope

This PR does not add repeated `MessageType` decoding or change protobuf compatibility behavior. Those remain follow-up work under NVIDIA/cudf-spark#14069.


src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -866,20 +753,8 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
-  // top-level child, then recursively through nested STRUCT/LIST children, so callers that
-  // access backing grandchildren directly still observe logically-null rows.
-  if (struct_null_count > 0) {
-    auto const* struct_mask_ptr = static_cast<cudf::bitmask_type const*>(struct_mask.data());
-    for (auto& child : top_level_children) {
-      apply_parent_mask_to_row_aligned_column(
-        *child, struct_mask_ptr, struct_null_count, num_rows, stream, mr);
-      propagate_nulls_to_descendants(*child, stream, mr);
-    }
+    if (null_count > 0) { struct_mask = std::move(mask); }
```
igorpeshansky, line 758: Would this make all-valid struct columns non-nullable? This is the most user-visible change in this PR, and belongs in the PR description.

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -255,8 +257,7 @@ struct enum_string_lookup_tables {
 enum_string_lookup_tables make_enum_string_lookup_tables(
   cudf::detail::host_vector<int32_t> const& valid_enums,
   std::vector<cudf::detail::host_vector<uint8_t>> const& enum_name_bytes,
-  rmm::cuda_stream_view stream,
-  rmm::device_async_resource_ref mr)
```
igorpeshansky, line 259: Probably worth mentioning this unused parameter in the PR description as well…

src/main/cpp/src/protobuf/protobuf_kernels.cuh
```diff
@@ -536,22 +516,23 @@ inline std::unique_ptr<cudf::column> extract_and_build_string_or_bytes_column(
...
   rmm::device_async_resource_ref mr)
 {
-  int32_t def_len = has_default ? static_cast<int32_t>(default_bytes.size()) : 0;
-  rmm::device_uvector<uint8_t> d_default(0, stream, mr);
+  int32_t def_len       = has_default ? static_cast<int32_t>(default_bytes.size()) : 0;
+  auto const scratch_mr = cudf::get_current_device_resource_ref();
+  rmm::device_uvector<uint8_t> d_default(0, stream, scratch_mr);
   if (has_default && def_len > 0) {
     d_default = cudf::detail::make_device_uvector_async(
       default_bytes, stream, cudf::get_current_device_resource_ref());
```
igorpeshansky: `scratch_mr`?

src/main/cpp/src/protobuf/protobuf_kernels.cuh
```diff
@@ -467,7 +443,9 @@ inline std::pair<rmm::device_buffer, cudf::size_type> make_null_mask_from_valid(
   auto pred  = [ptr = valid.data()] __device__(cudf::size_type i) {
     return static_cast<bool>(ptr[i]);
   };
-  return cudf::detail::valid_if(begin, end, pred, stream, mr);
+  auto [mask, null_count] = cudf::detail::valid_if(begin, end, pred, stream, mr);
+  if (null_count == 0) { mask = rmm::device_buffer{}; }
```
igorpeshansky, line 448: Worth a comment that dropping the mask would indicate to the factory that this is non-nullable…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -255,8 +257,7 @@ struct enum_string_lookup_tables {
...
+  rmm::cuda_stream_view stream)
 {
   auto d_valid_enums = cudf::detail::make_device_uvector_async(
     valid_enums, stream, cudf::get_current_device_resource_ref());
```
igorpeshansky: Worth factoring out into `scratch_mr`?

Also in `build_enum_string_values_column`…

src/main/cpp/tests/protobuf_helpers.cu
```diff
@@ -53,6 +53,26 @@ TEST_F(ProtobufHelpersTest, NullMaskFromPaddedValidUsesZeroLogicalRows)
   EXPECT_EQ(0, null_count);
 }
 
+TEST_F(ProtobufHelpersTest, NullMaskFromAllValidRowsIsEmpty)
```
igorpeshansky: Nit: this breaks apart two padded-valid tests. I'd put this at the end…

src/main/cpp/src/protobuf/protobuf_builders.cu
```diff
@@ -158,8 +158,10 @@ std::unique_ptr<cudf::column> make_null_column(cudf::data_type dtype,
     case cudf::type_id::FLOAT64:
       return cudf::make_fixed_width_column(dtype, num_rows, cudf::mask_state::ALL_NULL, stream, mr);
     case cudf::type_id::STRING: {
-      rmm::device_uvector<cudf::strings::detail::string_index_pair> pairs(num_rows, stream, mr);
-      thrust::fill(rmm::exec_policy_nosync(stream, mr),
+      auto const scratch_mr = cudf::get_current_device_resource_ref();
+      rmm::device_uvector<cudf::strings::detail::string_index_pair> pairs(
```
igorpeshansky, line 162: Technically, `pairs` isn't "temporary decode state" — it's a construction helper, so worth mentioning in the PR description…

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Yes. All-valid struct columns are now non-nullable. I called this out explicitly in the PR description.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Done. I added the removal of the unused output-memory-resource parameter to the PR description.

src/main/cpp/src/protobuf/protobuf_builders.cu
thirtiseven: Updated the PR description.

src/main/cpp/src/protobuf/protobuf_kernels.cuh
thirtiseven: Added a comment.

src/main/cpp/src/protobuf/protobuf_builders.cu

src/main/cpp/tests/protobuf_helpers.cu
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf_kernels.cuh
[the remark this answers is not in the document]
thirtiseven: Done.

src/main/cpp/src/protobuf/protobuf_kernels.cu
```diff
@@ -901,6 +903,7 @@ void maybe_check_required_fields(required_field_input_view input,
     static_cast<int>(field_indices.size()),
     !decode_ctx.row_force_null->is_empty() ? decode_ctx.row_force_null->data() : nullptr,
     decode_ctx.error->data());
+  CUDF_CHECK_CUDA(stream.value());
```
igorpeshansky, line 906: [Really optional] Since this PR already adds `scratch_mr` locals, there's one on line 896 above that can also be factored out (to complete that migration)…

igorpeshansky (approved):
LGTM :shipit: with one optional drive-by fix.

PR #4988, opened 2026-08-11
Encapsulate task priority state

res-life:
## Summary

Fix sonar alarm: Global variables should be const.

- encapsulate task-priority allocation state in `task_priority_tracker`
- replace mutable namespace-scope variables with a function-local static tracker
- keep the priority counter, attempt map, and protecting mutex in one object

## Why

The task-priority counter and attempt map require mutable process-wide state, but storing them as namespace-scope variables triggers the Sonar rule against non-const globals. Encapsulating the state makes the synchronization boundary explicit and uses thread-safe function-local static initialization while preserving the existing priority assignment behavior.

## Validation

- `pre-commit run --files src/main/cpp/src/TaskPriorityJni.cpp`
- `./build/build-in-docker clean package -DCMAKE_CUDA_ARCHITECTURES=89-real -DCPP_PARALLEL_LEVEL=32 -Dsubmodule.check.skip=true -DskipTests`

### DIFF
--- a/src/main/cpp/src/TaskPriorityJni.cpp
+++ b/src/main/cpp/src/TaskPriorityJni.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, NVIDIA CORPORATION.
+ * Copyright (c) 2025-2026, NVIDIA CORPORATION.
  *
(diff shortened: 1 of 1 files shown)


firestarman (approved):
lgtm

PR #4991, opened 2026-08-12
Follow `libcudf` semantics on temp allocation memory resources, cudaMemcpy arguments.

### LINKED ISSUE #4925: [FEA] Pass memory resource as the second argument to `rmm::exec_policy(stream)` and `rmm::exec_policy_nosync(stream)`

cudf is migrating `rmm::exec_policy(stream)` to `rmm::exec_policy(stream, temporary_mr)` (and the `_nosync` version), tracked by the issue https://github.com/rapidsai/cudf/issues/20780. Currently, there are many instances of `rmm::exec_policy(stream)` (and `_nosync`) calls across 23 files in our source code that still omit the memory resource argument, so their temporary allocations go through RMM's default current device resource instead of the cudf device resource. 

Pass `cudf::get_current_device_resource_ref()` as the second argument of  `rmm::exec_policy(stream)` (and `_nosync` version) to follow cudf convention. Then, we can add the libcudf's hook to `.pre-commit-config.yaml` scoped to this repository to enforce such convention.
```yaml
- id: exec-policy-nosync-memory-resource
  name: exec-policy-nosync-memory-resource
  description: 'Enforce that rmm::exec_policy_nosync is always passed a memory resource as its second argument'
  entry: 'rmm::exec_policy_nosync\([^,)]+\)'
  language: pygrep
  types_or: [c, c++, cuda]
  files: '^src/main/cpp/src/'
```

mythrocks:
Fixes #4925.

This commit brings the native code in `spark-rapids-jni` in line with some changed semantics in `libcudf`:

1. This change switches to passing an explicit memory-resource instance (`cudf::get_current_device_resource_ref()`) to `rmm::exec_policy()` and `rmm::exec_policy_nosync()`, to be used as the mr for the allocation of temporaries.  
   1. This is in line with the proposal in https://github.com/NVIDIA/cudf/issues/20780.
2. As an extension, for any calls in `spark-rapids-jni` to `libcudf` functions that accept a memory-resource, rather than depend on the default, we explicitly pass `cudf::get_current_device_resource_ref()`. 
3. Tangentially, `libcudf` has replaced its use of `rmm::exec_policy()` to `rmm::exec_policy_nosync()`, with the understanding that callers who need `sync` behaviour are expected to add one.  This commit makes a similar change in `spark-rapids-jni`.
4. And finally, for `cudaMemcpy*()` calls, rather than to explicitly set the direction of copy via `cudaMemcpyDeviceToHost`/`cudaMemcpyHostToDevice`, we switch to using `cudaMemcpyDefault` (allowing the runtime to choose the direction based on the memory addresses).
   1. This is in line with https://github.com/NVIDIA/cudf/pull/23605.


src/main/cpp/benchmarks/from_json.cu
```diff
@@ -314,8 +315,10 @@ std::unique_ptr<cudf::column> generate_map_of_array_input(std::size_t num_rows,
   rmm::device_uvector<cudf::size_type> sizes(num_rows, stream);
   fn.d_chars = nullptr;
   fn.d_sizes = sizes.data();
-  thrust::for_each_n(
-    rmm::exec_policy(stream), thrust::counting_iterator<cudf::size_type>{0}, row_count, fn);
+  thrust::for_each_n(rmm::exec_policy(stream, cudf::get_current_device_resource_ref()),
```
ttnghia: The changes in this PR are not enough. libcudf enforce to use `exec_policy_nosync` in all places, forbidding `exec_policy` which may have unexpected stream sync.

mythrocks: Good point.  We might as well change those also.

But I'm not convinced it would be a matter of string-replace.  I'll check that the call sites are safe to change.

ttnghia: I'm sure it is just string replacement. It's safe to replace all `exec_policy` by `exec_policy_nosync`. The policy `exec_policy_nosync` will just fallback to sync version if a stream sync is required.

mythrocks: Appreciated.  This is a case of trust-but-verify.  (Or in our case, "thrust"-but-verify.)

> The policy `exec_policy_nosync` will just fallback to sync version if a stream sync is required.

My fear is for a case where the results from a `thrust` call might be immediately consumed from the host, or some such.

I've reviewed the call sites.  I don't think I see a case where we can't get away with `_nosync`.  I'm making the switch now.

mythrocks: Done now.

I'll update the PR description to mention this change.

.pre-commit-config.yaml
```diff
@@ -45,6 +45,13 @@ repos:
         entry: '(?i)@(?:java\.lang\.)?deprecated\b'
         language: pygrep
         types: [java]
+      - id: use-cuda-memcpy-default
```
ttnghia, line 48: Can we also add hook to enforce `exec_policy_nosync`?

mythrocks: Yep, that makes sense to do.  I've updated the pre-commit yaml.

ttnghia (approved):

PR #4995, opened 2026-08-13
Detect EXCEPTION time parser policy disagreement

res-life:
Contributes to NVIDIA/cudf-spark#15596.

### Description

Spark's `EXCEPTION` time parser policy requires detecting strings that are rejected by the corrected parser but accepted by the legacy parser. The formatted timestamp kernel previously selected only one parser and therefore could not report that disagreement.

This change adds a single time parser policy parameter to `CastStrings.parseTimestampWithFormat`. The existing boolean overloads are retained for compatibility and delegate to the new policy-based API.

Under the `EXCEPTION` policy:

- Parse using the strict `CORRECTED` rules first.
- If `CORRECTED` succeeds, return its result without running `LEGACY`.
- If both `CORRECTED` and `LEGACY` fail, preserve the `CORRECTED` behavior and return null.
- If `CORRECTED` fails but `LEGACY` succeeds, treat the result as a parser-policy disagreement and throw `CastException`.

The CUDA kernel evaluates corrected parsing first, evaluates legacy parsing only for rejected rows, records the first disagreement, and reports it through `CastException`.

Regression coverage verifies corrected successes, multiple corrected/legacy disagreement forms, inputs for which both parsers fail, corrected null behavior, null and empty inputs, first-disagreement reporting in a mixed batch, and rejection of conflicting policy flags. Expected results were checked against Apache Spark 3.3, 3.4, 3.5, 4.0, and 4.1; Spark 4.2 retains the same `checkParsedDiff` control flow.

### Testing

- `./build/build-in-docker package -DCMAKE_CUDA_ARCHITECTURES=89-real -DCPP_PARALLEL_LEVEL=32 -Dsubmodule.check.skip=true -Dtest=CastStringsTest`
  - `CastStringsTest`: 48 tests, 0 failures, 0 errors
- `./build/build-in-docker install -DCMAKE_CUDA_ARCHITECTURES=89-real -DCPP_PARALLEL_LEVEL=32 -Dsubmodule.check.skip=true -DskipTests`


firestarman (commented):
LGTM for the JNI part.

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -38,27 +38,28 @@
...
+#define CATCH_CAST_EXCEPTION(env, ret_val)                                                    \
+  JNI_CATCH_BEGIN(env, ret_val)                                                               \
+  catch (spark_rapids_jni::cast_error const& e)                                               \
+  {                                                                                           \
+    if (env->ExceptionOccurred()) { return ret_val; }                                         \
+    jclass ex_class = env->FindClass(JNI_CAST_ERROR_CLASS);                                   \
+    if (ex_class != NULL) {                                                                   \
+      jmethodID ctor_id = env->GetMethodID(ex_class, "<init>", "([BI)V");                     \
+      if (ctor_id != NULL) {                                                                  \
+        std::string n_msg = e.get_string_with_error();                                        \
```
firestarman: Reconstructing `n_msg` from this `char const*` uses the null-terminated `std::string` constructor, so an embedded NUL truncates the reported value (for example, `2024-05-06\0x` reaches Java as `2024-05-06`). Please preserve the stored length—e.g. return `std::string const&` from `get_string_with_error()` and bind it here, or expose the size and use the pointer-plus-count constructor—before building the byte array.

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -278,55 +282,68 @@ struct parse_with_format_fn {
...
+  cudf::size_type* first_exception_row;
   bool* validity;
   cudf::timestamp_us* output;
 
   // Sets the row's null bit. The output value is left uninitialized — under cuDF's null mask
   // convention, masked-off positions are not read by downstream consumers.
   __device__ void set_invalid(cudf::size_type idx) const { validity[idx] = false; }
 
-  __device__ void operator()(cudf::size_type idx) const
+  __device__ bool parse(unsigned char const* p,
```
firestarman, line 302: The corrected and fallback parser programs are passed as separate pointer/count pairs, which makes it possible for an address and extent to get out of sync. Could you carry each pair as `cudf::device_span<format_token const>` through this helper and construct the spans from the corresponding `device_uvector`s at the call site? That keeps each buffer and size coupled without changing behavior.

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -278,55 +282,68 @@ struct parse_with_format_fn {
...
       set_invalid(idx);
       return;
     }
-    validity[idx] = true;
-    output[idx]   = cudf::timestamp_us{cudf::duration_us{result_us}};
+    auto const sv = d_strings.element<cudf::string_view>(idx);
+    auto const* p = reinterpret_cast<unsigned char const*>(sv.data());
+
+    cudf::timestamp_us parsed;
+    if (parse(p, sv.size_bytes(), tokens, num_tokens, legacy, parsed)) {
```
firestarman: Under EXCEPTION policy, the primary token stream still retains the special `yyyy/MM/dd` acceptance of one-digit month/day fields. As a result, `2024/5/6` succeeds here and returns before the legacy comparison, whereas Spark’s strict corrected parser rejects it and the legacy parser accepts it, producing an upgrade exception. Please compile the primary stream with strict corrected field widths when exception policy is enabled, while retaining the existing compatibility behavior outside that policy, and add a regression for this case.

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -278,55 +282,68 @@ struct parse_with_format_fn {
...
+      validity[idx] = true;
+      output[idx]   = parsed;
+      return;
+    }
+
+    set_invalid(idx);
+    if (first_exception_row != nullptr &&
+        parse(p, sv.size_bytes(), legacy_tokens, num_legacy_tokens, true, parsed)) {
+      atomicMin(first_exception_row, idx);
```
firestarman: Repository-local device atomics use `cuda::atomic_ref` so the scope and memory order are explicit. Could you wrap the shared scalar in `cuda::atomic_ref<cudf::size_type, cuda::thread_scope_device>` and use `fetch_min(idx, cuda::memory_order_relaxed)` here? No inter-thread payload is published through this reduction, so relaxed ordering is sufficient.

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -336,23 +353,44 @@ std::unique_ptr<cudf::column> parse_timestamp_strings_with_format(
...
+  rmm::device_uvector<format_token> device_tokens(host_tokens.size(), stream, temp_mr);
   CUDF_CUDA_TRY(cudaMemcpyAsync(device_tokens.data(),
                                 host_tokens.data(),
                                 sizeof(format_token) * host_tokens.size(),
                                 cudaMemcpyHostToDevice,
                                 stream.value()));
+  rmm::device_uvector<format_token> device_legacy_tokens(
+    host_legacy_tokens.size(), stream, temp_mr);
+  if (exception_policy) {
+    CUDF_CUDA_TRY(cudaMemcpyAsync(device_legacy_tokens.data(),
```
firestarman: `host_legacy_tokens` is backed by a pageable `std::vector`, so this raw H2D call cannot provide fully asynchronous transfer behavior and also bypasses libcudf’s copy-path optimizations. Please use `cudf::detail::cuda_memcpy_async` with typed device/host spans; the host vector remains alive through the later scalar read, satisfying the wrapper’s deferred-source lifetime requirement.

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -363,18 +401,30 @@ std::unique_ptr<cudf::column> parse_timestamp_strings_with_format(
...
+                         static_cast<int>(device_tokens.size()),
+                         legacy,
+                         device_legacy_tokens.data(),
+                         static_cast<int>(device_legacy_tokens.size()),
+                         first_exception_row ? first_exception_row->data() : nullptr,
+                         validity.begin(),
+                         result->mutable_view().begin<cudf::timestamp_us>()});
+
+  if (first_exception_row) {
+    auto const row = first_exception_row->value(stream);
```
firestarman, line 425: This kernel-produced scalar is copied back to the host on every non-empty EXCEPTION-policy call. Please use `cudf::detail::device_scalar<cudf::size_type>` for the pointer and allocation; it provides libcudf’s pinned bounce-buffer path for internal scalar transfers. The host-side decision still synchronizes the stream, but avoids the implicit pageable-memory copy overhead.

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -373,12 +373,31 @@ public static ColumnVector toDate(ColumnView input, boolean ansiEnabled) {
...
+   * Parse a string column using the selected time parser policy.
+   *
+   * @param input the input string column.
+   * @param format Spark format pattern (e.g. {@code "yyyy-MM-dd HH:mm:ss"}).
+   * @param legacy true for {@code LegacyTimeParserPolicy}, false for CORRECTED/EXCEPTION.
+   * @param exceptionPolicy when true, throw {@link CastException} if CORRECTED rejects a row
+   *                        that LEGACY accepts.
+   * @return a timestamp_us column where invalid rows have nulls.
+   */
+  public static ColumnVector parseTimestampWithFormat(ColumnView input, String format,
```
firestarman: This public overload has two unchecked failure modes, but the generated Javadoc will not list them in a Throws section from the parameter text alone. Please add `@throws CastException` for corrected/legacy parser disagreements and `@throws IllegalArgumentException` when both policy flags are enabled.

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -402,6 +423,7 @@ private static native long parseTimestampStringsToIntermediate(
 
   private static native long parseDateStringsToDate(long input);
 
-  private static native long parseTimestampWithFormat(long input, String format, boolean legacy);
+  private static native long parseTimestampWithFormat(
+      long input, String format, boolean legacy, boolean exceptionPolicy);
```
firestarman: NIT: We can introduce an integer to represent all the policies instead of two booleans in the future.

src/main/java/com/nvidia/spark/rapids/jni/CastException.java
```diff
@@ -15,13 +15,19 @@
...
+
 /**
  * CastException is an exception thrown by the JNI in the even of a casting error.
  */
 public class CastException extends RuntimeException {
   private final int rowWithError;
   private final String stringWithError;
-  
+
+  CastException(byte[] utf8StringWithError, int rowWithError) {
```
firestarman, line 27: I am not a fan of this constructor which requires the input bytes are ut8 encoded. We can construct a String from UTF8 byte array first, then call `CastException(String, int)` instead.

firestarman (commented):
The native part review is done by AI, and manually reviewed the Java part.

src/main/cpp/src/CastStringJni.cpp
```diff
@@ -38,27 +38,28 @@
...
+  JNI_CATCH_BEGIN(env, ret_val)                                                               \
+  catch (spark_rapids_jni::cast_error const& e)                                               \
+  {                                                                                           \
+    if (env->ExceptionOccurred()) { return ret_val; }                                         \
+    jclass ex_class = env->FindClass(JNI_CAST_ERROR_CLASS);                                   \
+    if (ex_class != NULL) {                                                                   \
+      jmethodID ctor_id = env->GetMethodID(ex_class, "<init>", "([BI)V");                     \
+      if (ctor_id != NULL) {                                                                  \
+        std::string const& n_msg = e.get_string_with_error();                                 \
+        cudf::jni::native_jbyteArray j_msg{                                                   \
```
firestarman: Could this handler avoid `native_jbyteArray` and use the non-throwing JNI calls directly? The wrapper checks for pending Java exceptions after `NewByteArray` and `SetByteArrayRegion` by throwing a C++ `jni_exception`. Because this construction happens inside the active `cast_error` catch handler, that new C++ exception is outside the protected `JNI_TRY`; an allocation failure can therefore let a C++ exception escape the JNI entry point instead of returning with the pending `OutOfMemoryError`, potentially terminating the JVM. Please allocate/copy with `NewByteArray` and `SetByteArrayRegion` directly, return when either leaves an exception pending, and explicitly raise OOM if allocation returns null without one.

src/main/cpp/src/cast_string.hpp
```diff
@@ -52,9 +52,9 @@ struct cast_error : public std::runtime_error {
   /**
    * @brief Get the string that caused a parsing error
    *
-   * @return char const* const problematic string
+   * @return std::string const& problematic string
    */
-  [[nodiscard]] char const* get_string_with_error() const { return _string_with_error.c_str(); }
+  [[nodiscard]] std::string const& get_string_with_error() const { return _string_with_error; }
```
firestarman, line 58: Could you include `<string>` directly in this header? It exposes `std::string` in the public `cast_error` API but currently obtains the declaration only through a transitive include, so an unrelated dependency cleanup could make this header stop compiling on its own.

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -337,26 +357,44 @@ std::unique_ptr<cudf::column> parse_timestamp_strings_with_format(
...
   std::string const& format,
   bool legacy,
+  bool exception_policy,
   rmm::cuda_stream_view stream,
   rmm::device_async_resource_ref mr)
 {
   SRJ_FUNC_RANGE();
-  auto const host_tokens = compile_format(format, legacy);
-  auto const num_rows    = input.size();
+  CUDF_EXPECTS(!(legacy && exception_policy),
```
firestarman, line 368: Could you add a native C++ regression for this conflicting-policy guard? The existing conflict test calls the Java overload, which rejects the flags before entering JNI, so removing or weakening this native `CUDF_EXPECTS` would leave the current test suite green. A direct `EXPECT_THROW` call to `parse_timestamp_strings_with_format(..., true, true)` would independently protect the documented C++ contract.

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -337,26 +359,65 @@ std::unique_ptr<cudf::column> parse_timestamp_strings_with_format(
...
-    host_tokens.size(), stream, cudf::get_current_device_resource_ref());
+  auto const temp_mr   = cudf::get_current_device_resource_ref();
+  auto const pinned_mr = cudf::get_pinned_memory_resource();
+  rmm::device_uvector<format_token> host_token_staging(host_tokens.size(), stream, pinned_mr);
+  rmm::device_uvector<format_token> host_legacy_token_staging(
+    host_legacy_tokens.size(), stream, pinned_mr);
+  rmm::device_uvector<cudf::size_type> host_exception_row_staging(
+    exception_policy ? 1 : 0, stream, pinned_mr);
+  // Pinned allocations are stream ordered, so wait before initializing them from the host.
+  stream.synchronize();
```
thirtiseven: Could we avoid synchronizing the entire stream here? This blocks every non-empty CORRECTED/LEGACY call before the parser kernel is enqueued. If the barrier is only needed so these stream-ordered pinned allocations can be initialized from the host, is there an immediately writable pinned staging path that would avoid draining prior GPU work?

src/main/java/com/nvidia/spark/rapids/jni/CastStrings.java
```diff
@@ -373,12 +373,33 @@ public static ColumnVector toDate(ColumnView input, boolean ansiEnabled) {
...
    * @return a timestamp_us column where invalid rows have nulls.
    */
   public static ColumnVector parseTimestampWithFormat(ColumnView input, String format,
       boolean legacy) {
-    return new ColumnVector(parseTimestampWithFormat(input.getNativeView(), format, legacy));
+    return parseTimestampWithFormat(input, format, legacy, false);
+  }
+
+  /**
+   * Parse a string column using the selected time parser policy.
```
thirtiseven: nit: The detailed Javadoc above says that a space in the pattern can also match 'T', but the native parser uses exact literal matching and the tests expect 'T' to be rejected. Could we update that sentence to say that a space matches exactly one space?

src/main/cpp/src/parse_timestamp_with_format.cu
```diff
@@ -337,26 +360,42 @@ std::unique_ptr<cudf::column> parse_timestamp_strings_with_format(
...
-  rmm::device_uvector<format_token> device_tokens(
-    host_tokens.size(), stream, cudf::get_current_device_resource_ref());
-  CUDF_CUDA_TRY(cudaMemcpyAsync(device_tokens.data(),
-                                host_tokens.data(),
-                                sizeof(format_token) * host_tokens.size(),
-                                cudaMemcpyDefault,
-                                stream.value()));
+  auto const temp_mr = cudf::get_current_device_resource_ref();
+  auto host_token_staging =
+    cudf::detail::make_pinned_vector_async<format_token>(host_tokens.size(), stream);
```
thirtiseven, line 381: nit, optional:
`make_pinned_vector_async` still synchronizes `stream` inside `rmm_host_allocator::allocate()` before returning the writable host vector, so this still drains prior work for every non-empty CORRECTED/LEGACY call; the synchronization has just moved into the helper. The previous callback-based version avoided blocking the host. Could we keep that deferred initialization approach, or use an equivalent path that does not synchronize the caller's stream here?

res-life: Thanks for pointing this out. We previously tried the deferred callback-based initialization, but that implementation did not compile because the stream-ordered host resource is not compatible with `rmm::device_uvector`. The current code follows the cuDF staging-helper convention. `make_pinned_vector_async` currently synchronizes during allocation due to the underlying `thrust::host_vector` requirement, as noted by its internal TODO. Since avoiding that would require custom allocation and lifetime management outside the supported helpers, I suggest keeping the current implementation for now.

res-life: Thanks for pointing this out. We previously tried the deferred callback-based initialization, but that implementation did not compile because the stream-ordered host resource is not compatible with rmm::device_uvector. The current code follows the cuDF staging-helper convention. make_pinned_vector_async currently synchronizes during allocation due to the underlying thrust::host_vector requirement, as noted by its internal TODO. Since avoiding that would require custom allocation and lifetime management outside the supported helpers, I suggest keeping the current implementation for now.

thirtiseven: Thanks, that makes sense.

thirtiseven (approved):
LGTM

PR #5003, opened 2026-08-14
Update HLLPP host UDF streams for cuda::stream_ref

vyasr:
Updates the HLLPP host UDF code for the cuDF host UDF stream API change by using cuda::stream_ref and stream.get() for kernel launches.

Also updates the thirdparty/cudf submodule to 5b6960d072, which includes NVIDIA/cudf#23649, which is where the above signatures changed.


igorpeshansky (approved):
LGTM :shipit:

mythrocks (approved):

PR #5009, opened 2026-08-17
Forward CCACHE_BASEDIR to local Docker builds

### LINKED ISSUE #5008: [FEA] Share ccache entries across local JNI checkouts

`build/run-in-docker` mounts a shared `CCACHE_DIR`, but it does not configure or forward `CCACHE_BASEDIR`. Compiler commands contain the absolute repository path, so equivalent builds from different local checkouts or Git worktrees produce different cache keys.

For example, builds from:

- `/home/user/work/spark-rapids-jni`
- `/home/user/work/cudf-spark-jni`

cannot effectively share compiler cache entries even when the source revision, cuDF pin, toolchain, and build flags are identical.

This is separate from #1344, which tracks sharing caches through CI or cloud storage.

Default `CCACHE_BASEDIR` to the repository root in `build/run-in-docker` and pass it into the container while preserving an explicit user override:

```bash
CCACHE_BASEDIR=${CCACHE_BASEDIR:-"$REPODIR"}
```

```bash
-e CCACHE_BASEDIR="$CCACHE_BASEDIR" \
```

This normalizes repository-local absolute paths while keeping source contents, generated headers, compiler identity, CUDA architectures, and compiler options in the cache key.

The premerge Jenkins pipeline already sets `CCACHE_BASEDIR` to its workspace, so this would align the local Docker wrapper with the existing CI setup.

- Pass `CCACHE_BASEDIR` manually through `DOCKER_RUN_EXTRA_ARGS` for every build.
- Keep all checkouts at the same absolute path.
- Share the complete `target` directory, which is unsafe because CMake and Ninja outputs contain checkout-specific absolute paths.

**Additional context**

The change should only affect local Docker builds. Different repository layouts may receive fewer cache hits, but ccache still validates source contents, included files, toolchain identity, and compiler flags before reusing an object.

thirtiseven:
## Description

`build/run-in-docker` mounts a shared ccache directory, but compiler paths from different local checkouts currently produce different cache keys.

Default `CCACHE_BASEDIR` to the selected working directory and forward it to the container. An explicitly configured `CCACHE_BASEDIR` is preserved.

Closes #5008.

## Testing

- `bash -n build/run-in-docker`
- ShellCheck error-level validation
- Simulated Docker invocations verifying the default, `WORKDIR`, and explicit `CCACHE_BASEDIR` behavior


gerashegalov (dismissed):
LGTM for normal builds 

The [manual](https://ccache.dev/manual/4.11.2.html#_compiling_in_different_directories) requires debug-path mapping such as
  -fdebug-prefix-map/-fdebug-compilation-dir, or disabling hash_dir with the risk of incorrect embedded paths. Either fix through debug path normalization  or document the limitation

gerashegalov (approved):
LGTM

PR #5030, opened 2026-08-20
Use cuDF to count the total deleted row count for cuDF based Delta reader for deletion vectors

### LINKED ISSUE #14628: [TASK] Use new cuDF API to count the total deleted row count for cuDF-based Delta reader for deletion vectors
A new cuDF-based Delta reader has been added in https://github.com/NVIDIA/spark-rapids/issues/13404 for more efficient processing of deletion vectors during the scan. The new reader uses libcudf APIs to apply deletion vectors after the parquet file scan. However, when it needs to know the total number of rows deleted, it currently loads the deletion vector into host memory, and counts deleted rows by going through the deletion vector bitmap. The cases where it needs to know the total deleted row count include:

- When 0 column is read (`SELECT count(*)` queries), the reader needs to know only the total number of rows alive after applying the deletion vector. 
- When the reader reads partition columns, it first populates the partition column values in host memory and then converts them into a column to attach to the `ColumnarBatch`. It needs to know the number of rows alive to populate the partition column values.

In both cases, the number of rows alive is computed by `total row count in row groups in files - total deleted row count in row groups in deletion vectors`. The former is read from the metadata, but the latter should be counted by counting through the deletion vector.

https://github.com/rapidsai/cudf/pull/21963 has added a new API in libcudf to run the counting on GPU. The plugin reader should use the new API instead.

sdrp713:
Fixes [#14628](https://github.com/NVIDIA/cudf-spark/issues/14628)

## Summary

Expose the new libcudf deletion vector row count API through `DeletionVectorUtils.computeNumDeletedRows`.

This allows cudf-spark to count deleted rows on the GPU instead of traversing the deletion vector bitmap in host code.

## Testing

- JNI Maven build and tests passed on CUDA 12.9 with architecture 86.
- Validated end-to-end with DBR 17.3, Spark 4.0.0, and Scala 2.13.
- Targeted Delta deletion-vector integration tests: 48 passed; remaining selected cases were skipped by existing runtime guards.

## Performance

100 million rows, 5 million deleted rows, 2 warmups, and 10 measured runs. The JNI and cudf-spark changes were tested together.

| Reader/query | Baseline median | PR median | Speedup |
|---|---:|---:|---:|
| PERFILE `count(*)` | 1.095745 s | 1.041387 s | 1.05× |
| PERFILE partition aggregate | 0.496434 s | 0.445686 s | 1.11× |
| COALESCING `count(*)` | 0.909243 s | 0.835445 s | 1.09× |
| COALESCING partition aggregate | 0.459016 s | 0.385882 s | 1.19× |
| MULTITHREADED `count(*)` | 0.841082 s | 0.783515 s | 1.07× |
| MULTITHREADED partition aggregate | 0.401422 s | 0.373242 s | 1.08× |

### DIFF
--- a/src/main/cpp/src/DeletionVectorUtilsJni.cpp
+++ b/src/main/cpp/src/DeletionVectorUtilsJni.cpp
@@ -26,8 +26,7 @@
 
 extern "C" {
 
-JNIEXPORT jlong JNICALL
-Java_com_nvidia_spark_rapids_jni_DeletionVectorUtils_computeNumDeletedRows(
+JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_DeletionVectorUtils_computeNumDeletedRows(
   JNIEnv* env,
   jclass,
   jlong serialized_bitmap_address,
--- a/src/main/cpp/src/DeletionVectorUtilsJni.cpp
+++ b/src/main/cpp/src/DeletionVectorUtilsJni.cpp
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2026, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cudf_jni_apis.hpp"
+#include "jni_utils.hpp"
+
+#include <cudf/io/experimental/deletion_vectors.hpp>
+
+#include <algorithm>
+#include <iterator>
+#include <utility>
+#include <vector>
+
+extern "C" {
+
+JNIEXPORT jlong JNICALL
+Java_com_nvidia_spark_rapids_jni_DeletionVectorUtils_computeNumDeletedRows(
+  JNIEnv* env,
+  jclass,
+  jlong serialized_bitmap_address,
+  jlong serialized_bitmap_length,
+  jint total_num_rows,
+  jlongArray row_group_offsets,
+  jintArray row_group_num_rows,
+  jboolean is_retention,
+  jint max_chunk_rows)
+{
+  JNI_NULL_CHECK(env, serialized_bitmap_address, "serialized bitmap address is null", 0);
+
+  JNI_TRY
+  {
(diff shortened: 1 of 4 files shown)


src/main/java/com/nvidia/spark/rapids/jni/DeletionVectorUtils.java
```diff
@@ -0,0 +1,65 @@
...
+    if (deletionVectorInfo == null) {
+      throw new NullPointerException("deletionVectorInfo");
+    }
+    if (maxChunkRows <= 0) {
+      throw new IllegalArgumentException("maxChunkRows must be positive");
+    }
+    return computeNumDeletedRows(
+        deletionVectorInfo.serializedBitmap.getAddress(),
+        deletionVectorInfo.serializedBitmap.getLength(),
+        deletionVectorInfo.totalNumRows,
```
nartal1: Are we guaranteed that row-group metadata is always populated here? DeletionVectorInfo uses Integer.MAX_VALUE for totalNumRows when the metadata is null or empty, so should we validate that case here or pass the actual row count explicitly?

src/main/cpp/src/DeletionVectorUtilsJni.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    auto const bitmap = cudf::host_span<cuda::std::byte const>{
+      reinterpret_cast<cuda::std::byte const*>(serialized_bitmap_address),
+      static_cast<std::size_t>(serialized_bitmap_length)};
+    auto const info = cudf::io::parquet::experimental::deletion_vector_info{
+      .serialized_roaring_bitmaps = {bitmap},
+      .deletion_vector_row_counts = {total_num_rows},
+      .row_group_offsets          = std::move(offsets),
+      .row_group_num_rows         = n_row_group_num_rows.to_vector(),
+      .are_retention_vectors      = static_cast<bool>(is_retention)};
```
jihoonson, line 63: Any reason that we have to add this API in a new file instead of the existing `DeletionVectorJni.cpp`? The existing file already has convenience util functions you can reuse to do these.

src/main/java/com/nvidia/spark/rapids/jni/DeletionVectorUtils.java
jihoonson: I think this is fine. `totalNumRows` will be initialized either `Integer.MAX_VALUE` or a sum of `rowGroupNumRows`, which is what the cudf API expects.

src/main/cpp/src/DeletionVectorUtilsJni.cpp
jihoonson: Same question for `DeletionVectorUtils.java` vs `DeletionVector.java`.

sdrp713: Yeah I guess since these changes are specific to cudf-spark I decided to keep it in our JNI repo

PR #5072, opened 2026-09-02
Fix singular protobuf message occurrence merging

### LINKED ISSUE #4983: [BUG] Singular protobuf message occurrences use last-one-wins instead of merge semantics
**Describe the bug**

When a singular embedded message field appears more than once, the protobuf decoder keeps only the last occurrence.

protobuf-java merges singular message occurrences instead:

- Singular child fields use last-one-wins across all fragments.
- Repeated child fields append values from every fragment.
- Nested singular messages are merged recursively.
- Required fields may be satisfied by different fragments.

The current behavior can therefore discard fields that were present in an earlier occurrence or incorrectly report them as missing.

Add the following test to `ProtobufTest`:

```java
@Test
void testDuplicateSingularMessageOccurrencesMerge() {
  Byte[] firstFragment =
      concat(box(tag(1, WT_VARINT)), box(encodeVarint(1)));
  Byte[] secondFragment =
      concat(box(tag(2, WT_VARINT)), box(encodeVarint(2)));
  Byte[] row = concat(
      box(tag(1, WT_LEN)), encodeMessage(firstFragment),
      box(tag(1, WT_LEN)), encodeMessage(secondFragment));

  ProtobufSchemaDescriptor schema =
          .addField(1, DType.STRUCT).down()
              .addField(1, DType.INT32)
              .addField(2, DType.INT32)
          .up()
          .build();

  try (Table input = new Table.TestBuilder()
           .column(new Byte[][]{row})
           .build();
       ColumnVector expected = ColumnVector.fromStructs(
           new StructType(
               true,
               new StructType(
                   true,
                   new BasicType(true, DType.INT32),
                   new BasicType(true, DType.INT32))),
           struct(struct(1, 2)));
       ColumnVector actual =
           Protobuf.decodeToStruct(input.getColumn(0), schema, true)) {
    AssertUtils.assertStructColumnsAreEqual(expected, actual);
  }
}
```

The assertion currently fails because the first fragment is discarded. The result contains `null` for child field 1 and `2` for child field 2.

**Expected behavior**

The decoder should match protobuf-java singular-message merge semantics:

- Merge fields from every occurrence of the singular message.
- Apply last-one-wins only to duplicate singular scalar fields within the merged message.
- Append repeated child values in wire order.
- Merge nested singular messages recursively.
- Validate required fields after all message fragments have been merged.
- Preserve the same behavior in permissive and fail-fast modes for valid input.

**Environment details (please complete the following information)**

- Environment location: Standalone Java unit test
- Spark configuration settings related to the issue: Not applicable; reproduces directly through `Protobuf.decodeToStruct`

**Additional context**

Observed on `main` at `35f19df6`.

This requires retaining and decoding multiple occurrences of a singular message instead of storing only its final location. It should be fixed separately from the wire/string, enum, and nested-validation corrections tracked by #4980, #4981, and #4982.

thirtiseven:
Fixes #4983.

## Summary

Protobuf merges repeated occurrences of a singular embedded message instead of applying last-one-wins to the message as a whole. This change brings the JNI decoder in line with protobuf-java for the singular-message paths supported on `main`.

## Changes

- Count and gather all occurrences when a singular message appears more than once in a row.
- Validate each message fragment independently before concatenating it, so malformed data cannot consume bytes from the next fragment.
- Decode the merged payload through the existing recursive message path, preserving last-one-wins for singular scalar children, append semantics for repeated children, recursive singular-message merging, and required-field validation after the merge.
- Preserve zero-length message presence, null rows, permissive/fail-fast behavior, and sliced-input offsets.
- Add focused regression coverage for top-level and recursively nested messages, repeated children, required fields split across fragments, empty fragments, sliced inputs, and malformed fragment boundaries.

## Testing

- `pre-commit` hooks for all changed files
- `mvn -q compiler:testCompile -DskipTests`
- JNI build and relevant protobuf tests passed locally

### DIFF
--- a/src/main/cpp/src/protobuf/protobuf.cu
+++ b/src/main/cpp/src/protobuf/protobuf.cu
@@ -459,11 +459,10 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
     for (auto const ni : merge_positions) {
       nested_merge_buffers[ni].emplace(std::move(*merge_bundle.fields[ni]));
     }
-    if (!merge_bundle.scan_descriptors.empty()) {
-      auto scan_bundle =
-        make_field_occurrence_scan_bundle(merge_bundle.scan_descriptors, stream, scratch_mr);
-      launch_scan_singular_message_occurrences(*d_in, scan_bundle.view(), d_error.data(), stream);
-    }
+    for_each_field_occurrence_scan_batch(
+      merge_bundle.scan_descriptors, stream, scratch_mr, [&](field_occurrence_scan_view fields) {
+        launch_scan_singular_message_occurrences(*d_in, fields, d_error.data(), stream);
+      });
   }
 
   // Store decoded columns by schema index for ordered assembly at the end.
--- a/src/main/cpp/src/protobuf/protobuf.cu
+++ b/src/main/cpp/src/protobuf/protobuf.cu
@@ -28,6 +28,7 @@
 
 #include <algorithm>
 #include <array>
+#include <optional>
 #include <ranges>
 #include <set>
 #include <string>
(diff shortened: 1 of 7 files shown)


src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -426,6 +434,38 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
...
   }
 
+  std::vector<std::optional<singular_message_merge_buffers>> nested_merge_buffers(num_nested);
+  if (num_nested > 0) {
+    auto h_multiple_nested_fields = cudf::detail::make_pinned_vector_async<int>(num_nested, stream);
+    CUDF_CUDA_TRY(cudf::detail::memcpy_async(h_multiple_nested_fields.data(),
+                                             d_multiple_nested_fields.data(),
+                                             num_nested * sizeof(int),
+                                             stream));
+    stream.sync();
```
thirtiseven, line 444: This adds a D2H copy and `stream.sync()` to every decode with a top-level message field, even when no row contains duplicates; the recursive path has the same synchronization at each level containing singular message children. The control-path overhead is real, but its end-to-end impact is not established here. Let’s defer measuring and optimizing it to the follow-up benchmark PR so this correctness fix stays scoped.

PR #5099, opened 2026-09-09
Canonicalize BOOL8 values in MurmurHash32

### LINKED ISSUE #5095: Canonicalize noncanonical BOOL8 values in MurmurHash32
Follow-up from NVIDIA/cudf#23265 and review discussion https://github.com/NVIDIA/cudf/pull/23265#discussion_r3868829030.

The Spark RAPIDS JNI MurmurHash32 implementation appears to convert BOOL8 directly to `uint32_t`. For noncanonical stored BOOL8 bytes such as 2 or 255, that can hash the raw byte instead of Spark boolean value 1, even though any nonzero BOOL8 value represents true.

Canonicalize boolean inputs (`key ? 1 : 0`) and add regression coverage constructed from raw BOOL8 bytes.

sylvesterkaczmarek:
Fixes #5095.

Canonicalize BOOL8 values before Spark-compatible MurmurHash32 hashing so every non-zero stored byte hashes as boolean true (`1`) instead of exposing its raw byte value.

The regression constructs a BOOL8 column directly from raw bytes `{0, 1, 2, 255}` because `fixed_width_column_wrapper<bool>` canonicalizes inputs during test setup. It verifies that 1, 2 and 255 hash identically while false remains distinct.

Validation:
- repository pre-commit hooks pass for both changed C++/CUDA files
- `git diff --check` passes

CUDA test execution is not available on the current macOS host; the added native regression is intended for the repository's NVIDIA/CUDA CI environment.

### DIFF
--- a/src/main/cpp/src/hash/murmur_hash.cuh
+++ b/src/main/cpp/src/hash/murmur_hash.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023-2026, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -134,7 +134,9 @@ struct MurmurHash3_32 {
 template <>
 murmur_hash_value_type __device__ inline MurmurHash3_32<bool>::operator()(bool const& key) const
 {
-  return compute<uint32_t>(key);
+  // BOOL8 stores false as zero and true as any non-zero byte. Canonicalize
+  // before hashing so noncanonical true values hash exactly like Spark true.
+  return compute<uint32_t>(key ? 1u : 0u);
 }
 
 template <>
--- a/src/main/cpp/tests/hash.cpp
+++ b/src/main/cpp/tests/hash.cpp
@@ -217,12 +217,11 @@ TEST_F(SparkMurmurHash3Test, NonCanonicalBool)
                                                   0);
 
   auto const output = spark_rapids_jni::murmur_hash3_32(cudf::table_view({col->view()}), 42);
-  auto const host   = cudf::test::to_host<int32_t>(output->view()).first;
 
-  ASSERT_EQ(raw.size(), host.size());
-  EXPECT_EQ(host[1], host[2]) << "byte 2 must hash the same as byte 1";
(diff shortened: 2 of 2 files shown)


src/main/cpp/tests/hash.cpp
```diff
@@ -198,6 +203,28 @@ TYPED_TEST(HashTestFloatTyped, TestExtremes)
...
+                                                  rmm::device_buffer{},
+                                                  0);
+
+  auto const output = spark_rapids_jni::murmur_hash3_32(cudf::table_view({col->view()}), 42);
+  auto const host   = cudf::test::to_host<int32_t>(output->view()).first;
+
+  ASSERT_EQ(raw.size(), host.size());
+  EXPECT_EQ(host[1], host[2]) << "byte 2 must hash the same as byte 1";
+  EXPECT_EQ(host[1], host[3]) << "byte 255 must hash the same as byte 1";
+  EXPECT_NE(host[0], host[1]) << "false and true must differ";
```
ttnghia: We should better use a consistent style with `EXPECT_COLUMNS_EQUAL`.

gerashegalov (dismissed):
LGTM

src/main/cpp/tests/hash.cpp
sylvesterkaczmarek: Updated to compare against the canonical BOOL8 hash column with CUDF_TEST_EXPECT_COLUMNS_EQUAL. Thanks.

ttnghia (dismissed):

src/main/cpp/src/hash/murmur_hash.cuh
```diff
@@ -134,7 +134,9 @@ struct MurmurHash3_32 {
 template <>
 murmur_hash_value_type __device__ inline MurmurHash3_32<bool>::operator()(bool const& key) const
 {
-  return compute<uint32_t>(key);
+  // BOOL8 stores false as zero and true as any non-zero byte. Canonicalize
+  // before hashing so noncanonical true values hash exactly like Spark true.
+  return compute<uint32_t>(key ? 1u : 0u);
```
mythrocks, line 139: 👏

mythrocks (approved):

mythrocks:
@sylvesterkaczmarek:  Thank you for the fix.  I've just merged this change.

sylvesterkaczmarek:
Thanks for reviewing and merging this into the 26.10 release branch. Much appreciated.

PR #5106, opened 2026-09-10
Use `cuda::device_buffer<std::byte>` for null masks

KyleFromNVIDIA:
## Description

Update usages of cudf's null masks to be
`cuda::device_buffer<std::byte>` while leaving other instances of
`rmm::device_buffer` alone.

Blocked by https://github.com/NVIDIA/cudf/pull/23989.

Issue: https://github.com/rapidsai/build-planning/issues/321

## Notes for Reviewers

Files that contain nontrivial changes that go beyond "replace X with Y" and may be of interest to reviewers:

- `src/main/cpp/tests/utilities.cpp` - changes expectations around padding

### DIFF
--- a/src/main/cpp/benchmarks/common/generate_input.cu
+++ b/src/main/cpp/benchmarks/common/generate_input.cu
@@ -547,7 +547,9 @@ std::unique_ptr<cudf::column> create_random_utf8_string_column(data_profile cons
   auto [result_bitmask, null_count] =
     profile.get_null_probability().has_value()
       ? cudf::bools_to_mask(cudf::device_span<bool const>(null_mask), stream)
-      : std::pair{std::make_unique<rmm::device_buffer>(), 0};
+      : std::pair{std::make_unique<cuda::device_buffer<std::byte>>(
+                    cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED)),
+                  0};
 
   return cudf::make_strings_column(num_rows,
                                    std::move(offsets),
@@ -652,7 +654,9 @@ std::unique_ptr<cudf::column> create_random_column(data_profile const& profile,
   auto [result_bitmask, null_count] =
     profile.get_null_probability().has_value()
       ? cudf::bools_to_mask(cudf::device_span<bool const>(null_mask))
-      : std::pair{std::make_unique<rmm::device_buffer>(), 0};
+      : std::pair{std::make_unique<cuda::device_buffer<std::byte>>(
+                    cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED)),
+                  0};
 
   return std::make_unique<cudf::column>(
     dtype, num_rows, data.release(), std::move(*result_bitmask.release()), null_count);
@@ -731,7 +735,9 @@ std::unique_ptr<cudf::column> create_random_column<cudf::struct_view>(data_profi
           return cudf::bools_to_mask(cudf::device_span<bool const>(valids),
                                      cudf::get_default_stream());
         }
-        return std::pair{std::make_unique<rmm::device_buffer>(), 0};
+        return std::pair{std::make_unique<cuda::device_buffer<std::byte>>(
+                           cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED)),
+                         0};
       }();
 
       // Adopt remaining children as evenly as possible
@@ -816,15 +822,19 @@ std::unique_ptr<cudf::column> create_random_column<cudf::list_view>(data_profile
     thrust::device_pointer_cast(offsets.end())[-1] =
       current_child_column->size();  // Always include all elements
 
-    auto offsets_column = std::make_unique<cudf::column>(cudf::data_type{cudf::type_id::INT32},
-                                                         current_num_rows + 1,
-                                                         offsets.release(),
-                                                         rmm::device_buffer{},
-                                                         0);
+    auto offsets_column =
+      std::make_unique<cudf::column>(cudf::data_type{cudf::type_id::INT32},
+                                     current_num_rows + 1,
+                                     offsets.release(),
+                                     cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED),
+                                     0);
 
-    auto [null_mask, null_count] = profile.get_null_probability().has_value()
-                                     ? cudf::bools_to_mask(cudf::device_span<bool const>(valids))
-                                     : std::pair{std::make_unique<rmm::device_buffer>(), 0};
+    auto [null_mask, null_count] =
+      profile.get_null_probability().has_value()
+        ? cudf::bools_to_mask(cudf::device_span<bool const>(valids))
+        : std::pair{std::make_unique<cuda::device_buffer<std::byte>>(
+                      cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED)),
+                    0};
 
     list_column = cudf::make_lists_column(current_num_rows,
                                           std::move(offsets_column),
@@ -900,8 +910,12 @@ std::unique_ptr<cudf::column> create_distinct_rows_column<cudf::list_view>(
   auto child_column      = cudf::sequence(num_rows, *zero);
   for (int lvl = dist_params.max_depth; lvl > 0; --lvl) {
     auto offsets_column = cudf::sequence(num_rows + 1, *zero);
-    auto list_column    = cudf::make_lists_column(
-      num_rows, std::move(offsets_column), std::move(child_column), 0, rmm::device_buffer{});
+    auto list_column =
+      cudf::make_lists_column(num_rows,
+                              std::move(offsets_column),
+                              std::move(child_column),
+                              0,
+                              cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED));
     if (auto const cv = list_column->view();
         cudf::has_nonempty_nulls(cv, cudf::get_default_stream())) {
       list_column = cudf::purge_nonempty_nulls(
@@ -931,8 +945,8 @@ std::unique_ptr<cudf::column> create_distinct_rows_column<cudf::struct_view>(
   children.push_back(cudf::sequence(num_rows, *cudf::make_fixed_width_scalar<int32_t>(0)));
   for (int lvl = dist_params.max_depth; lvl > 1; --lvl) {
     std::vector<std::unique_ptr<cudf::column>> parents;
-    parents.push_back(
-      cudf::create_structs_hierarchy(num_rows, std::move(children), 0, rmm::device_buffer{}));
+    parents.push_back(cudf::create_structs_hierarchy(
+      num_rows, std::move(children), 0, cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED)));
     std::swap(parents, children);
   }
   auto const null_count = col->null_count();
@@ -1096,10 +1110,12 @@ std::unique_ptr<cudf::column> create_string_column(cudf::size_type num_rows,
   return std::move(table->release().front());
 }
 
-std::pair<rmm::device_buffer, cudf::size_type> create_random_null_mask(
+std::pair<cuda::device_buffer<std::byte>, cudf::size_type> create_random_null_mask(
   cudf::size_type size, std::optional<double> null_probability, unsigned seed)
 {
-  if (not null_probability.has_value()) { return {rmm::device_buffer{}, 0}; }
+  if (not null_probability.has_value()) {
+    return {cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED), 0};
+  }
   CUDF_EXPECTS(*null_probability >= 0.0 and *null_probability <= 1.0,
                "Null probability must be within the range [0.0, 1.0]");
   if (*null_probability == 0.0f) {
--- a/src/main/cpp/src/cast_string.cu
+++ b/src/main/cpp/src/cast_string.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2026, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -664,20 +664,23 @@ struct string_to_integer_impl {
                                      rmm::device_async_resource_ref mr)
   {
     if (string_col.size() == 0) {
-      return std::make_unique<column>(
-        data_type{type_to_id<T>()}, 0, rmm::device_buffer{}, rmm::device_buffer{}, 0);
+      return std::make_unique<column>(data_type{type_to_id<T>()},
+                                      0,
+                                      rmm::device_buffer{},
+                                      cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED),
+                                      0);
     }
 
     rmm::device_uvector<T> data(string_col.size(), stream, mr);
-    auto const num_words = bitmask_allocation_size_bytes(string_col.size()) / sizeof(bitmask_type);
-    rmm::device_uvector<bitmask_type> null_mask(num_words, stream, mr);
+    auto null_mask =
+      cudf::create_null_mask(string_col.size(), cudf::mask_state::UNINITIALIZED, stream, mr);
 
     dim3 const blocks(util::div_rounding_up_unsafe(string_col.size(), detail::NUM_THREADS));
     dim3 const threads{detail::NUM_THREADS};
 
     detail::string_to_integer_kernel<<<blocks, threads, 0, stream.get()>>>(
       data.data(),
-      null_mask.data(),
+      reinterpret_cast<bitmask_type*>(null_mask.data()),
       string_col.chars_begin(stream),
       string_col.offsets().data<size_type>(),
       string_col.null_mask(),
@@ -685,12 +688,13 @@ struct string_to_integer_impl {
       ansi_mode,
       strip);
 
-    auto null_count = cudf::null_count(null_mask.data(), 0, string_col.size(), stream);
+    auto null_count = cudf::null_count(
+      reinterpret_cast<bitmask_type*>(null_mask.data()), 0, string_col.size(), stream);
 
     auto col = std::make_unique<column>(data_type{type_to_id<T>()},
                                         string_col.size(),
                                         data.release(),
-                                        null_mask.release(),
+                                        std::move(null_mask),
                                         null_count);
 
     if (ansi_mode) { validate_ansi_column(col->view(), string_col, stream); }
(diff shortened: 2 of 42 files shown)


thirdparty/cudf-pins/versions.json
KyleFromNVIDIA, line 52: Revert before merging:

```suggestion
      "git_tag" : "f1b809b15dfd135228ccca401cdd458650e81b8d",
```

src/main/cpp/src/utilities.cu
```diff
@@ -58,12 +62,16 @@ std::unique_ptr<rmm::device_buffer> bitmask_bitwise_or(
   auto d_input = cudf::detail::make_device_uvector_async(
     h_input, stream, rmm::mr::get_current_device_resource_ref());
 
-  std::unique_ptr<rmm::device_buffer> out =
-    std::make_unique<rmm::device_buffer>(mask_size * sizeof(cudf::bitmask_type), stream, mr);
+  std::unique_ptr<cuda::device_buffer<std::byte>> out =
+    std::make_unique<cuda::device_buffer<std::byte>>(
+      cudf::create_null_mask(mask_size * sizeof(cudf::bitmask_type) * CHAR_BIT,
```
KyleFromNVIDIA, line 67: I've been trying to use `cudf::create_null_mask()` wherever possible rather than instantiating a `cuda::device_buffer` directly. However, this presents a problem: `cudf::create_null_mask()` pads the allocation, while direct instantiation does not. This causes the test at `src/main/cpp/tests/utilities.cpp:46` to fail, since the size is different than expected.

What should we do? Is it OK if we change the test to match the new expectation, or do we really want to keep the size the same? (My vote is for the former.)

mythrocks: I ~~would have ordinarily been~~ **_am_** inclined to update the expectations in the test.

But I'm not sure of the implications of padding allocations for null-masks on other operations.  For instance, would this be to the detriment of the operations like `pack`, or `contiguous_split`?  There might be other considerations as well.

This will need some digging into, unless @abellina / @ttnghia might know, off the top of their heads.

mythrocks: Question for @KyleFromNVIDIA:

> I've been trying to use `cudf::create_null_mask()` wherever possible rather than instantiating a `cuda::device_buffer` directly...

We _could_ keep this bit similar to before (creating `cuda::device_buffer` directly), and expect no padding, no?

ttnghia: Why the size is multiplied by `CHAR_BIT`? ~~The mask size should be the same.~~

ttnghia: There should not be any error regardless of padidng the mask buffer. If test fails, I think it is legitimate to change it.

In the past we've been dealing with out-of-bound access due to padding. But anyway, if such bug shows up, it is due to the accessing kernel, not due to padding.

KyleFromNVIDIA: > Why the size is multiplied by `CHAR_BIT`?

Because the size passed to `cudf::create_null_mask()` is measured in bits, not bytes.

KyleFromNVIDIA: > We _could_ keep this bit similar to before (creating `cuda::device_buffer` directly), and expect no padding, no?

We could indeed, and that's what I was proposing with "direct instantiation". However, that would be inconsistent with the practice of using `cudf::create_null_mask()` wherever we can. I like the idea of changing the test to say that the size should be "at least" rather than "exactly".

thirdparty/cudf-pins/versions.json
mythrocks: Yes, please.  Thanks for making a note.

Why was this changed, again?

mythrocks (commented):
This change is looking good to me.  It would be good to take this change out for a spin with `cudf-spark` to be sure.

I can try, when I have a moment.

thirdparty/cudf-pins/versions.json
KyleFromNVIDIA: This PR relies on https://github.com/NVIDIA/cudf/pull/23989, so I updated the submodule in order to test it, and this file updated as well. Once the cudf PR merges, I will revert.

src/main/cpp/src/utilities.cu
igorpeshansky: Sorry, this was sitting in my drafts, just submitting it for the record.

------
`spark_rapids_jni::bitmask_bitwise_or` does not even document the return value. So let's document the potential padding (especially if we can reference `cudf::create_null_mask` for the padding spec). I'd say it's ok to check that it's at least as large as the inputs. You do have to change the subsequent copies, however, to use the _destination_ size, since the source buffer will be larger due to padding… Or lean into it and also check that the padding bits are all 0.

src/main/cpp/tests/utilities.cpp
```diff
@@ -43,12 +43,15 @@ TEST_F(UtilitiesTest, BitwiseOr)
...
-    CUDF_EXPECTS(d_result->size() == expect.size() * sizeof(cudf::bitmask_type),
+    CUDF_EXPECTS(d_result->size() >= expect.size() * sizeof(cudf::bitmask_type),
                  "Unexpected output size");
     std::vector<cudf::bitmask_type> result(expect.size());
-    cudaMemcpy(result.data(), d_result->data(), d_result->size(), cudaMemcpyDefault);
+    cudaMemcpy(result.data(),
+               d_result->data(),
+               result.size() * sizeof(cudf::bitmask_type),
+               cudaMemcpyDefault);
     CUDF_EXPECTS(std::equal(result.begin(), result.end(), expect.begin()),
```
igorpeshansky: Since `result.size()` may be larger than `expect.size()`, should this use the two-iterator form to avoid reading past the end of `expect`, i.e.:

```suggestion
    CUDF_EXPECTS(std::equal(result.begin(), result.end(), expect.begin(), expect.end()),
```

Also in line 85 below…

src/main/cpp/benchmarks/common/generate_input.cu
```diff
@@ -547,7 +547,9 @@ std::unique_ptr<cudf::column> create_random_utf8_string_column(data_profile cons
   auto [result_bitmask, null_count] =
     profile.get_null_probability().has_value()
       ? cudf::bools_to_mask(cudf::device_span<bool const>(null_mask), stream)
-      : std::pair{std::make_unique<rmm::device_buffer>(), 0};
+      : std::pair{std::make_unique<cuda::device_buffer<std::byte>>(
+                    cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED)),
```
igorpeshansky, line 551: Would it make sense to have an `empty_null_mask` helper function for this?

But see https://github.com/NVIDIA/cudf-spark-jni/pull/5106#discussion_r4007449923.

src/main/cpp/src/cast_string.cu
```diff
@@ -664,33 +664,37 @@ struct string_to_integer_impl {
...
+    auto null_mask =
+      cudf::create_null_mask(string_col.size(), cudf::mask_state::UNINITIALIZED, stream, mr);
 
     dim3 const blocks(util::div_rounding_up_unsafe(string_col.size(), detail::NUM_THREADS));
     dim3 const threads{detail::NUM_THREADS};
 
     detail::string_to_integer_kernel<<<blocks, threads, 0, stream.get()>>>(
       data.data(),
-      null_mask.data(),
+      reinterpret_cast<bitmask_type*>(null_mask.data()),
```
igorpeshansky: Could we not declare `null_mask` as `cuda::device_buffer<bitmask_type>`? Then we wouldn't need those `reinterpret_cast`s…

Also below.

src/main/cpp/src/datetime_truncate.cu
```diff
@@ -308,9 +308,10 @@ std::unique_ptr<cudf::column> truncate_datetime(cudf::column_view const& datetim
 
   auto [null_mask, null_count] =
     cudf::bools_to_mask(cudf::device_span<bool const>(validity), stream, mr);
-  output->set_null_mask(
-    null_count > 0 ? std::move(*null_mask.release()) : rmm::device_buffer{0, stream, mr},
-    null_count);
+  output->set_null_mask(null_count > 0
+                          ? std::move(*null_mask.release())
+                          : cudf::create_null_mask(0, cudf::mask_state::UNALLOCATED, stream, mr),
```
igorpeshansky, line 313: [Optional] I'm wondering if it makes sense to just have `cudf::bools_to_mask` (or a wrapper) return the correct empty `null_mask` instead of doing this conditional throughout (12 times, from my count)… The upstream change is beyond the scope of this PR, but a wrapper might not be, though it's definitely ok for a follow-up…

mythrocks (dismissed):

src/main/cpp/src/cast_string.cu
igorpeshansky, line 683: [Musing] I'm not a fan of these `reinterpret_cast`s, and it might've been better to have `cudf::create_null_mask` return a `cuda::device_buffer<bitmask_type>` (especially since it's already padded appropriately), but the column APIs expect `cuda::device_buffer<std::byte>` already, so I'm guessing that ship has sailed… In any case, beyond the scope of this PR.

[Optional] We could at least hoist those repeated `reinterpret_cast`s into a local… Also in `string_to_decimal_impl`…

thirdparty/cudf-pins/versions.json
igorpeshansky: This also applies to `thirdparty/cudf` itself, right? Not a revert, but an update to the commit on `main` after NVIDIA/cudf#23989 is merged…

src/main/cpp/src/utilities.cu
igorpeshansky: FWIW, this is also a narrowing, which may overflow (`mask_size * sizeof(cudf::bitmask_type) * CHAR_BIT` is a `size_t`, but `cudf::create_null_mask` expects a 32-bit signed `cudf::size_type`).

I also realized that my "check that the padding bits are all 0" is wrong, because the padding bits are left uninitialized.

That said, doesn't look like `spark_rapids_jni::bitmask_bitwise_or` is called from any production code, only from tests. Do we need it at all?

src/main/cpp/src/datetime_truncate.cu
```diff
@@ -308,9 +308,10 @@ std::unique_ptr<cudf::column> truncate_datetime(cudf::column_view const& datetim
 
   auto [null_mask, null_count] =
     cudf::bools_to_mask(cudf::device_span<bool const>(validity), stream, mr);
-  output->set_null_mask(
-    null_count > 0 ? std::move(*null_mask.release()) : rmm::device_buffer{0, stream, mr},
-    null_count);
+  output->set_null_mask(null_count > 0
+                          ? std::move(*null_mask.release())
```
igorpeshansky, line 312: Prior code, but doesn't the `.release()` leak the memory for the `device_buffer` object? Should this just be `std::move(*null_mask)` (like `from_json_to_raw_map.cu` does)?

Also in `number_converter.cu` (line 408), `parse_uri.cu` (line 999), and `row_conversion.cu` (line 2574). There are more occurrences in the repo, but they are definitely outside of the scope of this PR. We can defer to a follow-up and fix them in one sweep.

src/main/cpp/tests/utilities.cpp
```diff
@@ -43,12 +43,15 @@ TEST_F(UtilitiesTest, BitwiseOr)
     std::vector<cudf::bitmask_type> expect{
       0x11111001, 0xffffffff, 0xffff0000, 0xf101010f, 0xab0100ab};
     auto d_result = spark_rapids_jni::bitmask_bitwise_or({{da}, {db}}, stream);
-    CUDF_EXPECTS(d_result->size() == expect.size() * sizeof(cudf::bitmask_type),
+    CUDF_EXPECTS(d_result->size() >= expect.size() * sizeof(cudf::bitmask_type),
                  "Unexpected output size");
```
igorpeshansky, line 47: [Optional] What you have now is fine, but we could be more precise and assert the exact size of the buffer (_with_ the padding) via `cudf::bitmask_allocation_size_bytes`[^1]:

```suggestion
    CUDF_EXPECTS(d_result->size() == cudf::bitmask_allocation_size_bytes(
                                       expect.size() * sizeof(cudf::bitmask_type) * CHAR_BIT),
                 "Unexpected output size");
```

Also in lines 78-79 below…

[^1]: Technically needs an include of `<cudf/null_mask.hpp>`, but only for IWYU, because it's transitively reachable. Might need `<climits>` as well…

thirdparty/cudf-pins/versions.json
KyleFromNVIDIA: Yes

src/main/cpp/benchmarks/common/generate_input.cu
KyleFromNVIDIA: I think that's beyond the scope of both this PR and https://github.com/NVIDIA/cudf/pull/23989. These two PRs should be focused solely on converting null masks from `rm::device_buffer` to `cuda::device_buffer<std::byte>`. We can do this in a follow-up.

src/main/cpp/src/datetime_truncate.cu
KyleFromNVIDIA: We can do that in a follow-up, see my reasoning in https://github.com/NVIDIA/cudf-spark-jni/pull/5106#discussion_r4015759035.

src/main/cpp/src/cast_string.cu
KyleFromNVIDIA: I brought this up at https://github.com/NVIDIA/cudf/pull/23989#issuecomment-5638948762, feel free to weigh in over there.

src/main/cpp/src/datetime_truncate.cu
KyleFromNVIDIA: Yes, let's do that in a follow-up.

src/main/cpp/tests/utilities.cpp
KyleFromNVIDIA: We discussed this on Slack and determined that it would be fine to just check "greater than or equal".

src/main/cpp/src/utilities.cu
igorpeshansky: @KyleFromNVIDIA Ping on this… The return value should be documented, which is a good place to add the padding information, and (optionally) the `mr` parameter's doc still talks about "bloom filter"… See https://github.com/NVIDIA/cudf-spark-jni/pull/5106#discussion_r4030243876 for the concrete suggestion… Also, we didn't discuss the narrowing and whether this function is needed at all.

src/main/cpp/src/utilities.hpp
```diff
@@ -46,7 +47,7 @@ bool is_basic_spark_numeric(cudf::data_type type);
  * @param mr Device memory resource used to allocate the returned bloom filter's memory.
  *
```
igorpeshansky: Since this PR is what switched to `cudf::create_null_mask` and introduced the padding, I believe documenting the new size contract belongs here as well…

src/main/cpp/src/utilities.cu
KyleFromNVIDIA: This PR is intended only to be a mechanical translation from `rmm::device_buffer` to `cuda::device_buffer<std::byte>` (or `cuda::device_buffer<cudf::bitmask_type>` depending on what we decide at https://github.com/NVIDIA/cudf/pull/23989#issuecomment-5638948762). I'd rather not worry about getting rid of helper functions in this PR. We can do that in a follow-up.

igorpeshansky: Agreed about getting rid of helpers being out of scope, but then we _do_ need to address the narrowing and potential overflow which was introduced by this PR _as part of the translation_, as well as document the new behavior stemming from this PR. Alternatively, we could give up some consistency here and use direct instantiation instead of `cudf::create_null_mask()` (getting rid of both issues), since the cost of that switch seems to be higher than originally anticipated…

igorpeshansky (commented):
Just a response to your thread on `bitmask_bitwise_or` (https://github.com/NVIDIA/cudf-spark-jni/pull/5106#discussion_r4048336312).

src/main/cpp/src/utilities.cu
KyleFromNVIDIA: > since the cost of that switch seems to be higher than originally anticipated…

You're right. I've switched to direct instantiation and restored the tests to "exactly equal".

src/main/cpp/src/utilities.hpp
igorpeshansky: Since we're no longer padding, this can now be deferred.

For future reference, I've updated the suggestion:

```suggestion
 * @param mr Device memory resource used to allocate the returned buffer's memory.
 *
 * @return A buffer containing the bitwise-or of all the input bitmasks.
```

src/main/cpp/src/utilities.cu
igorpeshansky: You can now remove the `#include <climits>`, which was added for `CHAR_BIT`…

src/main/cpp/src/utilities.cu
```diff
@@ -27,6 +28,8 @@
 #include <cuda/functional>
 #include <cuda/stream>
 
+#include <climits>
```
igorpeshansky: Given https://github.com/NVIDIA/cudf-spark-jni/pull/5106#discussion_r4049294526, you can now remove this include, which was added for `CHAR_BIT`…

```suggestion
```

igorpeshansky:
build

KyleFromNVIDIA:
@igorpeshansky Can I get another build?

igorpeshansky:
build

igorpeshansky (dismissed):
LGTM :shipit:, though I'll probably need to re-approve once the `thirdparty/cudf` changes land…

src/main/cpp/benchmarks/common/generate_input.cu
ttnghia, line 551: Sorry if I missed the discussion about this: why not creating `cudf::device_buffer<std::byte>{}` directly?

KyleFromNVIDIA: Because we're trying to use `cudf::create_null_mask()` wherever possible for consistency. We made one exception above due to padding expectations.

KyleFromNVIDIA:
@igorpeshansky I need another build please

igorpeshansky:
build

mythrocks:
> I need another build please...

`thirdparty/cudf` has moved again.  This will need resolving, for the tests to run.

KyleFromNVIDIA:
build

KyleFromNVIDIA:
build

igorpeshansky:
build

KyleFromNVIDIA:
@igorpeshansky build please? I have an open ticket to add myself to the build list but I'm still waiting on it.

igorpeshansky:
build

igorpeshansky:
build

igorpeshansky:
build

igorpeshansky:
build

igorpeshansky:
build

PR #5120, opened 2026-09-14
Roll back profiler initialization on CUPTI setup failures

### LINKED ISSUE #5113: [BUG] Roll back partially initialized profiler state on CUPTI failures
**Describe the bug**
`Profiler.nativeInit` starts the writer thread and publishes the global `State` before several fallible CUPTI registration calls. If `cuptiSubscribe`, `cuptiEnableCallback`, or `cuptiActivityRegisterCallbacks` fails, `CATCH_STD` reports the exception to Java without rolling back the thread, global JNI reference, subscription, or native state.

Java assigns `Profiler.writer` only after `nativeInit` succeeds, so `Profiler.shutdown()` cannot clean up this partial initialization. The writer thread attaches with `AttachCurrentThread` and can remain blocked on the completed-buffer queue, preventing clean retry or JVM shutdown.

1. Enable the profiler.
2. Inject a failure from a CUPTI registration call after the writer thread is started.
3. Call `Profiler.init(...)` and observe the propagated initialization exception.
4. Retry initialization or attempt normal JVM shutdown.

A native unit test can make each CUPTI setup step fail in turn and verify that no thread, subscription, global reference, or published state remains.

**Expected behavior**
Profiler initialization should be transactional. Every failure should roll back completed setup steps, join the writer thread, release the JNI global reference, and leave the profiler ready for a clean retry.

**Environment details (please complete the following information)**
 - Environment location: Any environment with profiler support enabled
 - Spark configuration settings related to the issue: Profiler enabled; otherwise not configuration-specific

**Additional context**
Relevant code:
- https://github.com/NVIDIA/cudf-spark-jni/blob/1834dabb521815dd39eacd61bb4edd396c002feb/src/main/cpp/profiler/ProfilerJni.cpp#L437-L483
- https://github.com/NVIDIA/cudf-spark-jni/blob/1834dabb521815dd39eacd61bb4edd396c002feb/src/main/java/com/nvidia/spark/rapids/jni/Profiler.java#L36-L84

thirtiseven:
Fixes #5113.

Publish profiler state only after CUPTI initialization succeeds. On failure, unsubscribe if necessary, stop and join the writer thread, and release the JNI global reference and local state. Keep the writer bound to its session and allow buffer callbacks to handle unpublished state safely.

Check the device-reset callback registration result immediately so subsequent registrations cannot overwrite an error.

Add an independent GTest target covering 16 initialization failure scenarios, resource cleanup, and successful retry.

Validation: the author built JNI; the agent ran `PROFILER_INIT_TEST` through CTest with all 16 cases passing. Applicable pre-commit checks passed. Tests use CUPTI/JNI stubs; real CUPTI integration behavior was not tested.


gerashegalov (approved):
LGTM

PR #5171, opened 2026-09-20
Fix historical ORC timezone conversion

res-life:
Contributes to https://github.com/NVIDIA/cudf-spark/issues/15895.

Companion Spark PR: [NVIDIA/cudf-spark#16062](https://github.com/NVIDIA/cudf-spark/pull/16062). It switches the Spark ORC read path to the fused JNI APIs added here.

### Description

Fuse the complete ORC-to-Spark timezone conversion into one stream-ordered JNI/CUDA kernel and fix the remaining historical timezone mismatches between Apache ORC and Spark.

### Correctness fixes

Item 1 fixes the bug reported in [NVIDIA/cudf-spark#15895](https://github.com/NVIDIA/cudf-spark/issues/15895). Items 2–5 fix additional correctness issues found while developing this PR. Item 6 records the existing semantics preserved by the fused implementation. Item 7 addresses a further floating-point schema-evolution mismatch found during end-to-end validation.

1. Fix the physical ORC timestamp mismatch in NVIDIA/cudf-spark#15895 by reconstructing the local wall-clock fields exposed by ORC's legacy `java.util.TimeZone` conversion before applying Spark's `java.time` historical rebase. This corrects the reported 238-second GPU/CPU difference at New York's 1883 LMT-to-EST boundary.
2. Fix `BIGINT -> TIMESTAMP` schema evolution at historical overlaps by preserving ORC's selected instant before applying the Spark rebase. Ambiguous historical local times now retain the same earlier/later overlap occurrence as CPU ORC/Spark instead of defaulting to the earlier `java.time` offset.
3. Fix historical correction boundary checks by using separate exclusive UTC and local cutoffs for physical and integer-derived timestamps, so gaps and overlaps are compared in the correct time coordinate.
4. Fix historical conversion in zones such as `Pacific/Port_Moresby`, whose legacy `TimeZone` view has no recorded transitions while `java.time` still contains historical LMT transitions.
5. Fix transition selection for negative sub-second historical local values by using floor division. A value one microsecond before a gap's upper local boundary no longer rounds onto the transition and selects the post-transition offset.
6. Preserve existing ORC writer/reader semantics, including the 2015 base offset, negative timestamp borrow, `TimeZone.hasSameRules`, recurring DST fallback, null masks, sliced inputs, CUDA stream ordering, and the caller memory resource.
7. Fix the 748-second GPU/CPU difference in `DOUBLE -> TIMESTAMP` schema evolution at Vancouver's 1884 historical gap by applying ORC's legacy timezone conversion and millisecond rounding before Spark's historical rebase.

**Item 7 implementation status:** Implemented and validated locally in JNI commits `c978253f` and `034d0ce0`, with companion Spark commit `05bf62cce`; these commits are pending push and are not yet included in the remote PR heads.

<details>
<summary>Fixed bug details</summary>

The entries below correspond to the seven items in **Correctness fixes**. ORC first uses legacy `java.util.TimeZone` rules to produce a `java.sql.Timestamp`; Spark then reconstructs its local calendar fields and applies `java.time` rules. Those rule sets can disagree for historical timestamps. All values ending in `Z` below are UTC instants; “local” values are wall-clock fields without an offset.

#### 1. Reconstruct the legacy local fields for physical ORC timestamps

**Why the bug existed:** The previous correction treated the ORC-produced UTC instant itself as a local timestamp, computed a difference between legacy and modern local-to-UTC conversions, and added that difference back. Near a historical transition, that probes the rules at the wrong local time. Spark instead extracts local fields from the original instant using a legacy `Calendar`, then binds those fields with `java.time`.

**Example:** With both writer and reader in `America/New_York`, the physical regression produces the ORC instant `1883-11-18T15:48:19.923975Z` (`-2717655100076025` microseconds). Legacy `TimeZone` uses `-05:00` there, exposing local `10:48:19.923975`. At that local time, `java.time` uses New York's historical LMT (Local Mean Time) offset, `-04:56:02`:

```text
ORC instant:            15:48:19.923975Z
Legacy local fields:   10:48:19.923975
Spark result with LMT:  15:44:21.923975Z = -2717655338076025 us
```

The old calculation looked up local `15:48:19.923975`, after the LMT-to-EST overlap, where both conversions selected `-05:00`. It therefore applied no correction and returned a value 238 seconds too late.

**Fix:** Reconstruct the reader's legacy local fields from the ORC instant before rebasing, including Spark's overlap selection where applicable. Covered by `testConvertPhysicalOrcTimestampToSparkAcrossNewYorkLmtTransition` and `testConvertPhysicalOrcTimestampToSparkUsesLaterHistoricalOverlap`.

#### 2. Preserve ORC's occurrence at historical integer-to-timestamp overlaps

**Why the bug existed:** The old historical `BIGINT -> TIMESTAMP` path bound the original local value directly with `java.time`, which defaults to the earlier overlap occurrence. This discarded the instant already selected by ORC. [ORC's integer reader](https://github.com/apache/orc/blob/rel/release-1.9.6/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java#L1517-L1524) calls [`SerializationUtils.convertFromUtc`](https://github.com/apache/orc/blob/rel/release-1.9.6/java/core/src/java/org/apache/orc/impl/SerializationUtils.java#L1370-L1373), which looks up the legacy offset at `localMillis - rawOffset` and subtracts that offset. [Spark's rebase](https://github.com/apache/spark/blob/v3.5.6/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/RebaseDateTime.scala#L438-L477) then uses the original `Calendar` state and the following local day to choose the earlier or later occurrence.

**Example:** At `1883-11-18T17:00:00Z`, New York changed from LMT (`-04:56:02`) to EST (`-05:00`), repeating local `12:00:00` through `12:03:57.999999`. Integer `-2717668680` represents local `1883-11-18 12:02:00`:

| Interpretation | UTC result | Epoch microseconds |
| --- | --- | --- |
| Earlier LMT occurrence, selected by the old GPU path | `1883-11-18T16:58:02Z` | `-2717650918000000` |
| Later EST occurrence, selected by CPU ORC/Spark | `1883-11-18T17:02:00Z` | `-2717650680000000` |

ORC's raw offset is `-05:00`, so its lookup probes `17:02:00Z` and selects EST. Spark retains that later occurrence because the legacy standard/DST offsets are unchanged one local day later. The old GPU result was 238 seconds too early; the affected whole-second integer interval is `[-2717668800, -2717668563]`.

**Fix:** Perform ORC's conversion first, preserve its selected instant, and reproduce Spark's overlap decision during rebasing. Covered by `testConvertIntegerOrcTimestampToSparkUsesOrcHistoricalOverlapOffset`; the companion end-to-end `test_casting_from_bigint_to_timestamp_in_new_york_lmt_overlap` now passes without an expected-failure marker.

#### 3. Compare historical cutoffs in the correct coordinate system

**Why the bug existed:** The old code reused the reader's first legacy transition expressed in a local frame for both physical instants and integer-derived local timestamps. That value neither identifies where the two rule sets actually converge nor provides a valid UTC boundary. Gaps and overlaps also have distinct local boundaries.

**Example:** For `America/New_York`, the exclusive correction bounds are:

| Input being compared | Exclusive upper bound |
| --- | --- |
| ORC-produced physical UTC instant | `1883-11-18T17:00:00Z` |
| Integer-derived local wall-clock value | `1883-11-18 12:03:58` |

A physical instant at `16:59:00Z` is still before the UTC cutoff; comparing its numeric value with a local cutoff would incorrectly skip correction. Local `12:02:00` remains inside the overlap, so ending the local range at the post-transition clock reading `12:00:00` would miss it.

**Fix:** Compute the end of the historical rule difference separately in UTC and local coordinates, including the larger boundary offset for gaps/overlaps, and use a strict `<` comparison for each input path. Covered by `testHistoricalDifferenceCutoffs`.

#### 4. Rebase zones with no legacy transition table

**Why the bug existed:** A missing legacy `TimeZone` transition table was treated as evidence that no historical correction was necessary. A constant legacy view can still disagree with `java.time`, which retains older LMT transitions.

**Example:** For `Pacific/Port_Moresby`, legacy `TimeZone` has no recorded transitions and uses `+10:00`, while `java.time` uses `+09:48:40` in 1870. ORC instant `1870-01-01T00:00:00Z` exposes local `10:00:00`. Spark rebases those fields to `1870-01-01T00:11:20Z` (`-3155672920000000` microseconds). Skipping correction leaves `00:00:00Z`, 680 seconds too early. The integer-derived local value `1870-01-01 10:00:00` requires the same result.

**Fix:** Inspect the fixed `ZoneRules` transitions even when the legacy transition table is absent, derive the historical cutoffs, and load the modern rules when correction is needed. Covered on both paths by `testHistoricalRebaseWithoutTimeZoneTransitions`.

#### 5. Floor negative fractional seconds during historical transition lookup

**Why the bug existed:** The shared local-to-UTC lookup truncated negative microseconds toward zero when selecting a whole-second transition row. That can round a value just below a boundary up onto the boundary, selecting the next offset while leaving its fractional microseconds unchanged.

**Example:** Vancouver's 1884 change from `-08:12:28` to `-08:00` created the local gap `[00:00:00, 00:12:28)`. Local `1884-01-01 00:12:27.999999` is inside that gap, one microsecond before its upper boundary:

```text
Local epoch microseconds: -2713909652000001
Truncate to seconds:      -2713909652   (00:12:28; wrong transition row)
Floor to seconds:         -2713909653   (00:12:27; correct transition row)
```

Spark resolves this nonexistent local time by advancing it through the gap, producing `1884-01-01T08:24:55.999999Z` (`-2713880104000001` microseconds). Selecting the post-transition `-08:00` row too early instead produces `08:12:27.999999Z`, 748 seconds too early.

**Fix:** Enable floor division specifically for the reconstructed historical local value. Covered on both physical and integer paths by `testConvertOrcTimestampToSparkBeforeHistoricalGapAtSubsecondPrecision`.

#### 6. Preserve the existing ORC and column semantics during fusion

This item is a compatibility requirement, rather than a claim that each listed behavior was a separate new defect. The fused implementation must retain the established ORC behavior while adding the historical fixes.

**Why these details matter:** ORC's writer-specific 2015 epoch and negative fractional-second borrow must be applied before any writer/reader timezone adjustment. Matching writer/reader rules does not make the base adjustment optional. Rule equality must use `TimeZone.hasSameRules`, and dates beyond the recorded transition table still require the recurring DST rules.

**Example:** With both writer and reader in `Asia/Shanghai`, cuDF-decoded `21087883873` microseconds must first subtract the 2015 offset `28800000000`, giving `-7712116127`. That crosses into a negative timestamp and requires ORC's additional one-second borrow, producing `-7713116127`. Skipping conversion because the timezones match, or only subtracting the base offset, gives the wrong result.

**Fix/preservation:** Reuse the existing ORC base-offset, borrow, rule-comparison, and recurring-DST logic inside the fused path. Preserve the input null mask and slice offset, and use the supplied CUDA stream and memory resource for allocations and execution. For example, slicing `[ignored, null, timestamp]` to its last two rows must return `[null, converted_timestamp]`; checking the unsliced mask index would associate validity with the wrong row.

The regression cases are in [`GpuTimeZoneDBTest.java`](https://github.com/res-life/spark-rapids-jni/blob/82d67f0012e65f7d0e124a7c42e24a1ab411b949/src/test/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDBTest.java); the fused conversion and preserved ORC primitives are in [`timezones.cu`](https://github.com/res-life/spark-rapids-jni/blob/82d67f0012e65f7d0e124a7c42e24a1ab411b949/src/main/cpp/src/timezones.cu).

#### 7. Rebase floating-point timestamps after millisecond rounding

**Why the bug existed:** The old `DOUBLE -> TIMESTAMP` path incorporated a `java.time` offset into the floating-point conversion before its final millisecond rounding, then returned the rounded instant without Spark's historical rebase. ORC instead looks up a legacy `java.util.TimeZone` offset using `(long) (seconds * 1000) - rawOffset`, applies that offset to the original floating-point seconds, and rounds the resulting instant to milliseconds. Spark rebases the local fields reconstructed from that rounded instant. Near a historical gap, the truncated lookup and the rounded result can fall on different sides of the boundary.

**Example:** With reader timezone `America/Vancouver` and JDK 17.0.17 / tzdb 2025b, the input `-2713909652.000999` seconds reproduces the 748-second mismatch:

| Stage | Value |
| --- | --- |
| Initial local millisecond lookup after truncation | `-2713909652000` ms, representing local `1884-01-01 00:12:28.000` |
| ORC-converted instant after millisecond rounding | `1884-01-01T08:12:27.999Z` = `-2713880852001000` us |
| Local fields reconstructed with legacy `-08:00` | `1884-01-01 00:12:27.999` |
| CPU Spark result after historical rebasing | `1884-01-01T08:24:55.999Z` = `-2713880104001000` us |

The reconstructed local value is inside Vancouver's gap `[00:00:00, 00:12:28)`. Spark advances it through that gap. The old GPU path returned the unrebased `08:12:27.999Z`, which is **748 seconds too early**. Unlike item 5's negative-microsecond transition lookup, this defect concerns the ordering of legacy offset conversion, floating-point millisecond rounding, and the final Spark rebase.

**Fix:** Support millisecond input in `convertOrcFromUtc` so the companion Spark path can obtain ORC's legacy offset without an early conversion to microseconds. Apply that offset to the original seconds, preserve the existing rounding and overflow checks, and call `rebaseOrcInstantToSpark` on the final rounded microsecond instant. The new JNI input mode reuses the historical rebase without applying a writer base offset or negative-timestamp borrow again.

**Regression coverage:** Add `-2713909652.000999` to the existing end-to-end `test_non_utc_timestamp_regressions`, and extend `testConvertOrcTimestampToSparkBeforeHistoricalGapAtSubsecondPrecision` with the rounded instant. Additional JNI tests cover nullable slices, millisecond extrema, rounded historical instants, and empty/all-null columns. Local validation on the implementation commits above passed all 23 `GpuTimeZoneDBTest` tests and the focused four-timezone Spark run (19 passed, 1 expected skip). The original 1,000,000-row Vancouver reproducer also changed from 20,115 mismatches to zero with the same input seed.

</details>

### Performance numbers

Please refer to the **Performance numbers: 50 million rows** section in [NVIDIA/cudf-spark#16062](https://github.com/NVIDIA/cudf-spark/pull/16062) for the end-to-end CPU, Legacy GPU, and GPU benchmark results.

### Performance and structure

- Replace the previous multi-kernel correction sequence and temporary columns with one CUDA kernel per timestamp leaf.
- Lazily materialize the `java.time` transition table only for timezones whose legacy and modern historical rules differ.
- Share the same fused implementation between physical ORC timestamps and integer schema-evolution timestamps.

### Validation

- JNI Docker package with native test targets enabled: `BUILD SUCCESS`.
- Focused GPU checks passed for the New York historical overlap and Vancouver negative sub-second gap on both physical and integer paths.
- Historical-overlap GPU audit: 129 timezones, 655 boundary values, 0 mismatches against the CPU/Spark reference.
- Spark 3.5.6 package: `BUILD SUCCESS`.
- End-to-end `test_casting_from_bigint_to_timestamp_in_new_york_lmt_overlap`: passed without an expected-failure marker.
- Existing 10,000,000-row benchmark: CPU/GPU results matched; fused cross-timezone conversion showed 9 ms mean GPU overhead and 2.49x CPU/GPU speedup.

AI assistance: Codex was used to analyze and prepare this change.


pxLi:
JNI has branched off `release/26.10` (while plugin still uses `main`), plz try check if we need to re-target thx

src/main/cpp/src/datetime_utils.cuh
```diff
@@ -548,12 +548,17 @@ __device__ static timestamp_type convert_timestamp(
   //     that fall inside a DST gap (a local time that does not exist) must resolve to the
   //     post-gap `offset_after`, matching Spark/Java's `LocalDateTime.atZone()` convention.
   //     Truncation toward zero happens to land on the post-gap row, so we keep it.
+  //
+  //   - floor_local_time: historical ORC rebasing has already reconstructed the exact local
```
thirtiseven: Could we clarify that this flag limits the fix to ORC? Java's gap resolution is the same for identical local timestamps regardless of their source. The existing truncation path retains a known mismatch; the adjacent comment claiming Java parity should be corrected too.

src/main/cpp/src/timezones.cu
```diff
@@ -835,6 +843,125 @@ __device__ cudf::timestamp_us convert_orc_from_utc_value(cudf::timestamp_us valu
...
+ * historical transitions such as America/New_York's 1883 LMT-to-EST change.
+ */
+__device__ static cudf::timestamp_us convert_orc_instant_to_local_value(cudf::timestamp_us value,
+                                                                        tz_side_info const& reader)
+{
+  auto const value_us = value.time_since_epoch().count();
+  auto const value_ms = spark_rapids_jni::integer_utils::floor_div(value_us, MICROS_PER_MILLI);
+  auto const offset_ms =
+    reader.is_fixed ? reader.raw_offset : get_transition_index(value_ms, reader);
+  auto const result_us = wrapping_add(value_us, static_cast<int64_t>(offset_ms) * MICROS_PER_MILLI);
```
firestarman, line 848: Could this local-field reconstruction preserve the hybrid calendar semantics used by the Spark reference path? Before the 1582 cutover, `java.util.Calendar` exposes Julian fields, while adding the legacy offset here yields proleptic-Gregorian fields. That can shift historical-rebase results by the Julian/Gregorian date difference. Please apply the existing device-side calendar rebase logic before `convert_historical_local_to_spark` and add a pre-1582 regression test.

src/main/cpp/src/timezones.cu
```diff
@@ -922,6 +1049,200 @@ std::unique_ptr<column> convert_orc_from_utc_typed(cudf::column_view const& inpu
...
+{
+  auto results = cudf::make_timestamp_column(input.type(),
+                                             input.size(),
+                                             cudf::copy_bitmask(input, stream, mr),
+                                             input.null_count(),
+                                             stream,
+                                             mr);
+  if (input.size() == 0) { return results; }
+
+  auto const java_time_fixed_cdv = column_device_view::create(java_time_info.column(0), stream, mr);
```
firestarman: Could these short-lived `column_device_view` metadata allocations use `cudf::get_current_device_resource_ref()` instead of the caller’s output `mr`? The returned column and null mask should use `mr`, but routing temporary launch metadata through it can make a caller-supplied output resource account for or retain scratch allocations.

src/main/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDB.java
```diff
@@ -641,12 +658,42 @@ public long getReaderFirstTransitionUs() {
...
 
+    private void ensureJavaTimeInfo() {
+      ensureOpen();
+      if ((readerHistoricalDifferenceEndUtcUs == Long.MIN_VALUE
+          && readerHistoricalDifferenceEndLocalUs == Long.MIN_VALUE)
+          || readerJavaTimeTzInfoTable != null) {
+        return;
+      }
+
+      Table javaTimeInfo = getTimezoneInfo();
```
firestarman: Could this avoid retaining a complete copy of the Java-time database for every historical ORC timezone context? Each context already owns per-zone writer/reader tables, but `getTimezoneInfo()` copies both full global columns and keeps them until close. Please retain only the reader row needed here, or share the device table with reference counting.

src/test/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDBTest.java
```diff
@@ -344,6 +417,208 @@ void testOrcTimezoneContextConversionFailures() {
...
+    long orcInstant = -2_717_655_100_076_025L;
+    long decodedMicros = orcInstant + orc2015YearBaseOffsetUs(timezoneId);
+    long expectedMicros = convertPhysicalOrcTimestampToSparkOnCPU(
+        decodedMicros, timezoneId, timezoneId);
+    assertEquals(-2_717_655_338_076_025L, expectedMicros);
+
+    try (ColumnVector input = ColumnVector.timestampMicroSecondsFromLongs(decodedMicros);
+        ColumnVector expected = ColumnVector.timestampMicroSecondsFromLongs(expectedMicros);
+        GpuTimeZoneDB.OrcTimezoneContext context =
+            GpuTimeZoneDB.buildOrcTimezoneContext(timezoneId, timezoneId);
```
firestarman, line 575: Could we add a fused physical-conversion test with different writer and reader zones? The new kernel has a separate `writer_reader_rules_differ` path that stages writer transitions, but the fused tests currently construct contexts with identical timezone IDs. Comparing that case with `convertPhysicalOrcTimestampToSparkOnCPU` would cover the branch.

src/main/cpp/src/timezones.cu
```diff
@@ -922,6 +1049,200 @@ std::unique_ptr<column> convert_orc_from_utc_typed(cudf::column_view const& inpu
...
+  if (!reader_args.is_fixed) {
+    stage_side_transitions(reader_args.trans,
+                           reader_args.offsets,
+                           reader_args.trans_count,
+                           ptr,
+                           rt_begin,
+                           rt_end,
+                           ro_begin);
+  }
+  __syncthreads();
```
firestarman: nit: Could this barrier be conditional on actually staging a transition table? Fixed-offset/global-memory paths do not write shared memory, so they currently pay an unnecessary block-wide synchronization.

src/main/cpp/src/timezones.cu
```diff
@@ -922,6 +1049,200 @@ std::unique_ptr<column> convert_orc_from_utc_typed(cudf::column_view const& inpu
...
+                                             stream,
+                                             mr);
+  if (input.size() == 0) { return results; }
+
+  auto const java_time_fixed_cdv = column_device_view::create(java_time_info.column(0), stream, mr);
+  auto const java_time_dst_cdv   = column_device_view::create(java_time_info.column(1), stream, mr);
+  auto const java_time_fixed_transitions = lists_column_device_view{*java_time_fixed_cdv};
+  auto const java_time_dst_rules         = lists_column_device_view{*java_time_dst_cdv};
+
+  int64_t const* writer_trans_ptr =
```
firestarman: nit: Could the repeated transition-table pointer/count and fixed-zone setup be extracted into a small `make_orc_tz_side_kernel_args` helper and reused by both conversion paths? Keeping this construction in one place would prevent the two paths from drifting when the argument layout changes.

src/main/cpp/src/timezones.hpp
```diff
@@ -190,6 +190,44 @@ struct orc_tz_side {
...
+ * @param reader_historical_difference_end_local_us Exclusive local timestamp upper bound for
+ *        historical rule differences, or INT64_MIN when no historical rebase is needed.
+ * @param input_is_orc_timestamp True for physical ORC timestamps; false for integer-derived local
+ *        timestamps produced by ORC schema evolution.
+ * @param writer_reader_rules_differ Whether Apache ORC would convert between the writer and reader.
+ * @param stream CUDA stream.
+ * @param mr Device memory resource.
+ * @return Spark-compatible timestamps in microseconds.
+ */
+[[nodiscard]] std::unique_ptr<cudf::column> convert_orc_to_spark(
```
firestarman, line 224: nit: Would an `orc_to_spark_options` aggregate make this API safer to call? The current signature has several adjacent policy values with identical types, including two historical cutoffs and two booleans, so accidentally swapping them remains type-correct.

src/main/cpp/src/timezones.cu
res-life: Thanks for pointing this out. Pre-1582 hybrid-calendar semantics are outside the scope of this PR and are tracked separately in https://github.com/NVIDIA/cudf-spark/issues/131. This PR focuses on historical timezone conversion; the calendar-rebase fix and corresponding pre-1582 regression tests should be handled under that issue.

res-life:
build

res-life:
build

firestarman:
build

src/main/cpp/src/timezones.cu
firestarman: lgtm, but better have more eyes

src/test/java/com/nvidia/spark/rapids/jni/GpuTimeZoneDBTest.java
```diff
@@ -322,6 +391,86 @@ void testConvertOrcFromUtcAllTimezones() {
...
+        assertColumnsAreEqual(reference, actual);
+      }
+    }
+  }
+
+  @Test
+  void testOrcToSparkEmptyAndAllNullInputs() {
+    GpuTimeZoneDB.cacheDatabase();
+    for (String zone : new String[]{"UTC", "America/New_York", "Pacific/Port_Moresby"}) {
+      for (Long[] values : new Long[][]{new Long[0], new Long[257]}) {
```
firestarman, line 477: nit: Could this fixed-zone coverage include at least one non-null value for both fused entry points? Empty and all-null inputs do not exercise the no-historical-difference dispatch or writer base-offset arithmetic. A `+05:30` case with negative fractional, epoch, and positive values would cover the path.

src/main/cpp/src/timezones.cu
```diff
@@ -922,6 +1023,182 @@ std::unique_ptr<column> convert_orc_from_utc_typed(cudf::column_view const& inpu
...
+    java_time_info.column(0), stream, cudf::get_current_device_resource_ref());
+  auto const java_time_dst_cdv = column_device_view::create(
+    java_time_info.column(1), stream, cudf::get_current_device_resource_ref());
+  auto const java_time_fixed_transitions = lists_column_device_view{*java_time_fixed_cdv};
+  auto const java_time_dst_rules         = lists_column_device_view{*java_time_dst_cdv};
+
+  auto const writer_args = make_orc_tz_side_kernel_args(writer);
+  auto const reader_args = make_orc_tz_side_kernel_args(reader);
+
+  size_t smem_bytes = 0;
```
firestarman, line 1162: nit: Could the repeated shared-memory byte calculation be extracted into one helper? The same transition-count limit, alignment, and per-entry byte formula appear in three launch paths, so centralizing them would keep host sizing synchronized with the staging layout.

src/main/cpp/src/timezones.cu
```diff
@@ -922,6 +1023,182 @@ std::unique_ptr<column> convert_orc_from_utc_typed(cudf::column_view const& inpu
...
+  cuda::stream_ref stream,
+  rmm::device_async_resource_ref mr)
+{
+  auto results = cudf::make_timestamp_column(input.type(),
+                                             input.size(),
+                                             cudf::copy_bitmask(input, stream, mr),
+                                             input.null_count(),
+                                             stream,
+                                             mr);
+  if (input.size() == 0) { return results; }
```
firestarman: nit: Could this fast path also return when `input.null_count() == input.size()`? The result and copied all-null mask are already complete, so this would avoid creating device-view metadata, staging transitions, and launching a kernel whose threads all exit at the null check.

firestarman:
lgtm, but better have more eyes

firestarman:
build

res-life:
build

res-life:
build

src/main/cpp/src/timezones.cu
```diff
@@ -1015,9 +1287,74 @@ std::unique_ptr<cudf::column> convert_orc_from_utc(cudf::column_view const& inpu
                                                    rmm::device_async_resource_ref mr)
 {
   validate_timezone_table(reader.tz_info_table);
+  switch (input.type().id()) {
+    case cudf::type_id::TIMESTAMP_MILLISECONDS:
```
thirtiseven, line 1291: Millisecond inputs can exceed the `int32_t` epoch-day range used by `millis_to_year()`. For New York, input `189279560548800000` ms should use UTC−04:00, but narrowing the day count makes the DST path select UTC−05:00. Could we widen the date arithmetic and add this regression case?

src/main/cpp/src/GpuTimeZoneDBJni.cpp
```diff
@@ -197,4 +197,63 @@ JNIEXPORT jlong JNICALL Java_com_nvidia_spark_rapids_jni_GpuTimeZoneDB_convertOr
...
+    auto const options = spark_rapids_jni::orc_to_spark_options{
+      .reader_historical_difference_end_utc_us =
+        static_cast<int64_t>(reader_historical_difference_end_utc_us),
+      .reader_historical_difference_end_local_us =
+        static_cast<int64_t>(reader_historical_difference_end_local_us),
+      .input_kind                 = static_cast<spark_rapids_jni::orc_timestamp_kind>(input_kind),
+      .writer_reader_rules_differ = static_cast<bool>(writer_reader_rules_differ)};
+    return cudf::jni::release_as_jlong(spark_rapids_jni::convert_orc_to_spark(
+      *input,
+      static_cast<int64_t>(writer_tz_offset_at_orc_2015_base_us),
```
igorpeshansky: FYI, I've opened https://github.com/NVIDIA/cudf/issues/24259 to remove the need for this workaround.

src/main/cpp/src/timezones.cu
res-life: Spark Timestamp type is stored as us, will never exceed the int range.
This function is calculating ms, overflow will never happen in practice, because this code is only invoked by Spark.  Let's have a follow-up in the future if we encounter this problem.

thirtiseven (approved):
LGTM

res-life:
build

thirtiseven (approved):

PR #5173, opened 2026-09-21
Add endpoint-delta percentile interpolation

### LINKED ISSUE #5172: Support endpoint-delta exact-percentile interpolation
Spark exact percentile evaluation now has an endpoint-delta interpolation mode that computes `lower + fraction * (higher - lower)`. The JNI histogram percentile evaluator currently supports only weighted-endpoint interpolation.

Add a selectable endpoint-delta mode while preserving weighted-endpoint behavior for existing callers. This is the native prerequisite for NVIDIA/cudf-spark#15516.

firestarman:
Fixes #5172.

Spark's endpoint-delta exact-percentile interpolation can produce different IEEE results from weighted-endpoint interpolation, notably when a percentile crosses infinity. The JNI histogram evaluator currently exposes only the weighted-endpoint form, preventing cudf-spark from matching the newer Spark semantics.

This change adds `WEIGHTED_ENDPOINTS` and `ENDPOINT_DELTA` modes to the Java and native histogram APIs. Existing callers retain weighted-endpoint behavior through the original overload. Native dispatch selects a compile-time-specialized kernel once per call, and endpoint-delta temporaries preserve Spark's subtraction, multiplication, and addition order without FMA contraction. There is no user-facing behavior change until cudf-spark selects the new mode. This is the native prerequisite for NVIDIA/cudf-spark#15516.

`HistogramTest` now covers the existing default, both modes across negative infinity, finite/+Inf, -Inf/+Inf, NaN, equal endpoints, integral positions, Float input, a bit-exact no-FMA case, and large finite monotonicity. The focused suite passed all 8 tests using CUDA 12.9.1.

Performance testing is not required: sorting, allocation, and per-row control flow are unchanged; mode selection is a single host-side switch per column operation, and each device kernel is compile-time specialized without a per-row mode branch.

AI assistance was used to implement and test this change. The author reviewed the complete diff and this description before submission.


firestarman:
build

src/test/java/com/nvidia/spark/rapids/jni/HistogramTest.java
```diff
@@ -45,4 +86,81 @@ void testAllNulls() {
...
+        Arrays.asList(entry(Float.NEGATIVE_INFINITY, 1L), entry(10.0f, 1L));
+    try (ColumnVector histogram = histograms(FLOAT_HISTOGRAM_TYPE, row);
+         ColumnVector result = Histogram.percentileFromHistogram(
+             histogram, new double[]{0.2}, false, ENDPOINT_DELTA)) {
+      assertDoubleColumnExactly(result, Double.NaN);
+    }
+  }
+
+  @Test
+  void testEndpointDeltaIsMonotonicForLargeFiniteValues() {
```
res-life, line 175: Could we add a bit-exact case that specifically distinguishes separate multiply/add from FMA? The current large-finite case produces the same bits with both evaluation strategies, so it would still pass if compiler contraction changed the Spark-compatible rounding behavior.

For example, with `lower = -100.0`, `higher = -99.3`, and `fraction = 0.1`, Spark’s separate operations produce `-99.93` (`0xc058fb851eb851ec`), while `Math.fma` produces `-99.92999999999999` (`0xc058fb851eb851eb`). A two-entry histogram at percentile `0.1` would directly protect the no-FMA invariant documented in `histogram.cu`.

firestarman: Added the bit-exact regression case in 30205006a0. It uses the suggested two-entry histogram at percentile `0.1` and asserts the separate multiply/add result bits (`0xc058fb851eb851ec`). `HistogramTest` passes all 8 tests with CUDA 12.9.1. Thanks!

firestarman:
build

res-life (approved):
LGTM

PR #5181, opened 2026-09-22
Fix protobuf decoder out-of-bounds accesses

### LINKED ISSUE #5180: [BUG] spark-rapids-jni submodule-sync: cudaErrorLaunchFailure in ProtobufTest corrupts CUDA context causing 197 cascading test errors
**Describe the bug**
Build: spark-rapids-jni_submodule-sync-dev/6938

During submodule-sync validation (testing against new cudf commit cebee9124d68db0ac2de77787720070abd4b5444 with -DUSE_SANITIZER=ON), the native build succeeded but the Maven surefire test phase failed. A CUDA kernel launch failure (cudaErrorLaunchFailure - unspecified launch failure) occurred in ProtobufTest. This is a sticky CUDA context error, so all later GPU allocations failed with std::bad_alloc at cuda_memory_resource.cpp:26, cascading into 197 errors across ProtobufTest and RegexRewriteUtilsTest. The forked JVM finally crashed in RmmSparkTest (exit code 254). The trailing 'unzip: cannot find target/cudf-spark-jni-26.12.0-SNAPSHOT-cuda12.jar' (exit code 9) is a secondary effect because the jar was never produced.

Error logs:
```
[ERROR] testRequiredFieldPresent  Time elapsed: 0.004 s  <<< ERROR!
ai.rapids.cudf.CudaFatalException: std::bad_alloc: CUDA error (failed to allocate 9 bytes) at: _deps/rmm-src/cpp/src/mr/cuda_memory_resource.cpp:26: cudaErrorLaunchFailure unspecified launch failure
	at ai.rapids.cudf.Rmm.allocInternal(Native Method)
	at ai.rapids.cudf.Rmm.alloc(Rmm.java:618)
	at ai.rapids.cudf.DeviceMemoryBuffer.allocate(DeviceMemoryBuffer.java:136)
	at ai.rapids.cudf.HostColumnVector.copyToDevice(HostColumnVector.java:305)
	at com.nvidia.spark.rapids.jni.ProtobufTest.testRequiredFieldPresent(ProtobufTest.java:1227)
...
[ERROR] Tests run: 2201, Failures: 0, Errors: 197, Skipped: 4
[INFO] BUILD FAILURE
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.0:test (default-test) on project cudf-spark-jni: There are test failures.
[ERROR] The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[ERROR] Process Exit Code: 254
[ERROR] Crashed tests:
[ERROR] com.nvidia.spark.rapids.jni.RmmSparkTest
unzip:  cannot find or open target/cudf-spark-jni-26.12.0-SNAPSHOT-cuda12.jar ...
script returned exit code 9
```

**Environment details**
spark-rapids-jni submodule-sync CI; rockylinux8-cuda12.9.1 image; validating cudf commit cebee9124d68db0ac2de77787720070abd4b5444; built with -DBUILD_TESTS=ON -DUSE_SANITIZER=ON; single GPU node.

thirtiseven:
## Description
Fixes #5180.

This PR fixes two independent issues in protobuf error handling: a known sanitizer false positive for small-type atomics, and uninitialized occurrence records caused by inconsistent scan policies.

**1. Small-type atomics trigger memcheck errors on tightly allocated bool buffers**

For example:

```protobuf
message Outer {
  optional Inner inner = 1;
}
message Inner {
  required int32 id = 1;
  optional string note = 2;
}
```

Given `{ inner: { note: "oops" } }`, permissive mode must mark the outer row as null because `inner.id` is missing.

Previously, these flags were stored in `device_uvector<bool>`. For one input row, this allocates only one byte. However, `atomic_ref<bool>::store(true)` in the affected implementation uses a 32-bit read/CAS sequence. The logical one-byte update therefore accesses four bytes, exceeding the allocation. This explains why sanitizer reports an invalid **read**, even though the source operation is a store:

```text
Invalid __global__ read of size 4 bytes
at check_required_fields_kernel(...)
... allocation ... of size 1 bytes
```

This matches the issue described in [CCCL #6430](https://github.com/NVIDIA/cccl/issues/6430), which classifies these small-type atomic accesses as a sanitizer false positive on supported hardware. This report alone does not establish memory corruption outside sanitizer runs.

Atomic updates are necessary because multiple nested values or fragments can mark the same parent/root row. The fix keeps both affected flag buffers as bool arrays, guarantees 4-byte alignment, and rounds their allocation sizes up to a multiple of four bytes. The entire allocation, including padding, is zero-initialized. The owning buffer retains the padded size, while consumers use a `cudf::device_span<bool>` whose logical size is the row count. For example, five flags use eight allocated bytes rather than twenty bytes with `uint32_t` flags. This follows the padding workaround used for small-type columns in cuDF hash groupby.

**2. The count and occurrence-writing passes disagree on unexpected wire types**

A singular message field can appear multiple times on the wire, and its fragments must be merged. Consider:

```protobuf
message Outer {
  optional Inner inner = 1;
}
message Inner {
  optional int32 value = 1;
}
```

Now consider this wire sequence for field 1:

```text
VARINT 7              // Wrong wire type: a message requires LEN.
LEN { value: 1 }      // Valid message fragment.
LEN { value: 2 }      // Valid message fragment.
```

The decoder counts fragments, allocates occurrence records, writes their row indices/offsets/lengths, and then validates and merges them. Previously:

| Pass | Behavior |
|---|---|
| Count | Record the wire-type error, continue, and count two valid fragments. |
| Write occurrences | Stop at the wire-type error, leaving both allocated records uninitialized. |
| Validate fragments | Read those records using the count, interpreting uninitialized data as indices or offsets. |

An error between the two valid fragments similarly leaves only the second record uninitialized.

Marking the row invalid does not make this safe: downstream kernels still consume the intermediate records before the host handles the decoding error. The fix makes singular-message occurrence writing use the same skip-and-continue policy as counting. The malformed row still becomes null in permissive mode, and failfast still reports the wire-type error.

Regression coverage includes 1–5-row allocation boundaries, multiple malformed fragments targeting the same row, and unexpected wire types before, between, and after valid fragments. It also checks unaffected neighboring rows and the expected failfast error.

## Validation
- JNI rebuilt locally by the author; the test-library SHA256 matches the rebuilt native library.
- Ran `mvn -B compiler:testCompile surefire:test@default-test -DUSE_SANITIZER=ON -Dtest=ProtobufTest` with CUDA 12.8 Compute Sanitizer: 234 tests passed, no failures or skips; memcheck reported 0 errors.
- Ran `ProtobufHelpersTest.AtomicFlagBufferIsPaddedAndZeroed` under Compute Sanitizer memcheck: passed with 0 errors. It checks empty and nonempty buffers, alignment, padded allocation size, and zero-initialization.
- Pre-commit checks and `git diff --check` passed.
- Full JNI suite and Spark integration tests were not run.


src/main/cpp/src/protobuf/protobuf.cu
```diff
@@ -434,12 +434,9 @@ std::unique_ptr<cudf::column> decode_protobuf_to_struct(cudf::column_view const&
   // PERMISSIVE-mode row nulling support for malformed input, root enum mismatches, and missing
   // required fields.
   bool const track_permissive_null_rows = !fail_on_errors;
-  rmm::device_uvector<bool> d_row_force_null(
+  // Subword atomics may access a full 32-bit word, including beyond a bool allocation's end.
```
res-life: This allocates more memory; can illegal memory access be avoided by padding with 3 bytes?

res-life: This can be a follow-up, this is a trade off about memory and speed.
Using 32-bit may be more faster, but uses more memory.

res-life (dismissed):
LGTM

firestarman (commented):
lgtm, but better have more eyes

src/main/cpp/src/protobuf/protobuf.cu
thirtiseven: Thanks. Both approaches are used in the codebase, but I agree that padding is a better fit here. I've updated the PR to keep the bool flags and pad the allocations to a multiple of 4 bytes.

res-life (approved):
LGTM

thirtiseven:
build

thirtiseven:
build

